PR #3, opened 2017-08-29
Add expandable string enum to template2

olydis:
replay of https://github.com/Azure/autorest/pull/2505

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -195,8 +195,8 @@
     </pluginManagement>
   </build>
   <modules>
-    <module>src/generator/AutoRest.Java.Tests</module>
-    <module>src/generator/AutoRest.Java.Azure.Tests</module>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -195,8 +195,8 @@
     </pluginManagement>
   </build>
   <modules>
-    <module>test/vanilla</module>
```
RikkiGibson: This is a change from autorest.java that you probably want but it just got trampled in the replay.

RikkiGibson:
This is ready to work with and fix up for merging if you `npm install -g autorest@prerelease`.

daschult:
I've replaced this pull request with [this one](https://github.com/Azure/autorest.java/pull/10) so that I can do my work in autorest.java instead of autorest.

olydis:
I assume this one can be closed then? Reopen if that assumption is wrong.

PR #17, opened 2017-10-03
Simplify finding interfaces that an inner supports

daschult:
This pull request is just some simple cleanup. I intended to fix a bug, but then had to backtrack because I ended up breaking more functionality that I was fixing.


daschult:
@olydis Hey Johannes. Can you shed some light as to why the CoverageReporter is throwing a null pointer exception?

olydis:
@daschult Argh, that's whenever the testserver instance your tests connect against is not the one started by your tests, but by a previous test. The CI runner only ever executes one test at a time and it seems that some programming languages tests don't reliably clean up (= kill) the test server when there's a test failure... so every once in a while, there's an orphaned test server running on the CI machines. We're currently looking into moving to docker for that reason, sorry about the inconvenience and thanks for notifying, I'll restart the tests.

olydis:
> linux-x64 restart

olydis:
> win32-x64 restart

src/azurefluent/Model/MethodGroupJvaf.cs
```diff
@@ -104,34 +96,77 @@ private void DiscoverAllSupportedInterfaces()
...
+        private static Method FindFirstMethodByName(IEnumerable<Method> methods, String methodName)
+        {
+            return methods.FirstOrDefault(method => stringComparer.Equals(method.Name, methodName));
+        }
+
+        private static bool HasNonClientNonConstantRequiredParameters(Method method, int requiredParameterCount)
         {
             // When parameters are optional we generate more methods.
-            return method.Parameters.Count(x => !x.IsClientProperty && !x.IsConstant && x.IsRequired) == 2;
+            return method.Parameters.Count(x => !x.IsClientProperty && !x.IsConstant && x.IsRequired) == requiredParameterCount;
```
RikkiGibson: What does it mean for something to be a client property? Does that mean the parameter to the REST API method is pulled from an instance variable and not passed as a method argument?

It's weird that 2 used to be a magic number here, and I'm wondering how/whether this is affecting the code gen now.

olydis:
# publish job

PR #21, opened 2017-10-19
Get all vanilla tests to pass

daschult:


src/vanilla/Model/ResponseJv.cs
```diff
@@ -233,13 +270,23 @@ public IEnumerable<string> ImplImports
                 var imports = new List<string>(InterfaceImports);
                 imports.AddRange(BodyWireType.ImportSafe());
                 imports.AddRange(HeaderWireType.ImportSafe());
-                if (this.NeedsConversion && (Body is SequenceType || Headers is SequenceType))
+
+                if (ReturnValueWireType != null)
```
RikkiGibson: I'd encourage single accesses on computed properties that perform traversals. This property is probably not an offender, but the time needed to generate code has been increased substantially by unwitting repeated accesses to properties that perform traversals of the Swagger model.

It's got me considering taking computed properties that do nontrivial things and turning them into methods when I encounter them simply so that people don't use them as though it's just reading a value from a field.

package.json
```diff
@@ -12,7 +12,8 @@
     "publish-preview": "gulp publish-preview",
     "regenerate": "gulp regenerate",
     "clean": "gulp clean",
-    "nuke": "git clean -xdf"
+    "nuke": "git clean -xdf",
+    "start-test-server": "npm --prefix node_modules/@microsoft.azure/autorest.testserver run start"
```
RikkiGibson: Good idea.

daschult:
@RikkiGibson Can you take a look at my lazy changes and let me know if it's okay.

src/vanilla/Model/ResponseJv.cs
```diff
@@ -10,16 +10,59 @@ namespace AutoRest.Java.Model
...
     {
-        protected List<string> _interfaceImports = new List<string>();
-        protected List<string> _implImports = new List<string>();
+        protected readonly List<string> _interfaceImports = new List<string>();
+        protected readonly List<string> _implImports = new List<string>();
+        private readonly Lazy<string> returnValueWireType;
 
         public ResponseJv()
         {
+            returnValueWireType = createReturnValueWireTypeLazy();
```
RikkiGibson: Did something go awry when you tried to get rid of this constructor?

PR #22, opened 2017-10-20
Generate RestResponse overloads

RikkiGibson:
In addition to adding the extra overloads to return RestResponse, this PR begins a refactor that should minimize the copy and pasting needed to maintain the Javadoc generation. See MethodJv.cs and let me know what you think. I didn't push it all the way across the InterfaceTemplates, Retrofit templates, etc. just yet because I wanted to see how this would go over. Especially don't want to deal with paging and LRO specific templates until we have new overloads or something where it's a reasonable opportunity to refactor them.

I have been considering pushing it out a little further even to have a `class JavaClass` and `class JavaMethod` which would be much simpler than real ASTs but would simply allow us to capture the differences between the overloads in a more maintainable way.

### DIFF
--- a/src/vanilla/Model/CodeModelJv.cs
+++ b/src/vanilla/Model/CodeModelJv.cs
@@ -35,6 +35,19 @@ namespace AutoRest.Java.Model
         [JsonIgnore]
         public bool IsCustomBaseUri => Extensions.ContainsKey(SwaggerExtensions.ParameterizedHostExtension);
 
+        private bool? shouldGenerateXmlSerializationCached = null;
+        public bool ShouldGenerateXmlSerializationCached
+        {
+            get
+            {
+                if (!shouldGenerateXmlSerializationCached.HasValue)
+                {
+                    shouldGenerateXmlSerializationCached = ShouldGenerateXmlSerialization;
+                }
+                return shouldGenerateXmlSerializationCached.Value;
+            }
+        }
+
         [JsonIgnore]
         public string ServiceClientServiceType
         {
--- a/src/vanilla/Model/CodeModelJv.cs
+++ b/src/vanilla/Model/CodeModelJv.cs
@@ -73,6 +73,7 @@ namespace AutoRest.Java.Model
                 }
                 classes.AddRange(new[]{
                         "com.microsoft.rest.RestProxy",
+                        "com.microsoft.rest.RestResponse",
                         "com.microsoft.rest.ServiceClient",
                         "com.microsoft.rest.RestClient",
                         "rx.Single"
--- a/src/vanilla/Model/ResponseJv.cs
+++ b/src/vanilla/Model/ResponseJv.cs
@@ -206,9 +206,6 @@ namespace AutoRest.Java.Model
         public virtual string GenericBodyWireTypeString => BodyWireType.Name;
 
         [JsonIgnore]
-        public string GenericBodyWireTypeStringWrapped => GenericBodyWireTypeString;
-
-        [JsonIgnore]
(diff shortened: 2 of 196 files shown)


src/vanilla/Model/CodeModelJv.cs
```diff
@@ -35,6 +35,19 @@ public override string BaseUrl
         [JsonIgnore]
         public bool IsCustomBaseUri => Extensions.ContainsKey(SwaggerExtensions.ParameterizedHostExtension);
 
+        private bool? shouldGenerateXmlSerializationCached = null;
+        public bool ShouldGenerateXmlSerializationCached
```
daschult: Having a Cached and "NonCached" version feels a little funky to me. Is there a way to merge them into a single property?

src/vanilla/Model/ResponseJv.cs
```diff
@@ -184,68 +179,37 @@ public virtual string GenericBodyClientTypeString
...
 
         [JsonIgnore]
-        public string GenericBodyWireTypeStringWrapped
-        {
-            get
-            {
-                return GenericBodyWireTypeString;
-            }
-        }
+        public string GenericBodyWireTypeStringWrapped => GenericBodyWireTypeString;
```
daschult: What's the purpose of having "Wrapped" properties?

test/azure/src/main/java/fixtures/azureparametergrouping/ParameterGroupings.java
```diff
@@ -197,6 +242,14 @@
     /**
      * Post parameters with a shared parameter group object.
      *
+     * @throws IllegalArgumentException thrown if parameters fail the validation
+     * @return the {@link RestResponse<Void, Void>} object if successful.
+     */
+    RestResponse<Void, Void> postSharedParameterGroupObjectWithRestResponseAsync();
```
daschult: Are you sure this is correct? A method that has no parameters and doesn't return anything?

test/azure/src/main/java/fixtures/azureparametergrouping/implementation/ParameterGroupingsImpl.java
```diff
@@ -150,30 +165,43 @@ public void postOptional() {
...
+        Validator.validate(parameterGroupingPostOptionalParameters);
+    String customHeader = null;
+    Integer query = null;
         return service.postOptional(this.client.acceptLanguage(), customHeader, query, this.client.userAgent());
     }
 
     /**
      * Post a bunch of optional parameters grouped.
      *
+     * @throws IllegalArgumentException thrown if parameters fail the validation
```
daschult: This function has no parameters. What parameters is the comment talking about?

src/vanilla/Model/CodeModelJv.cs
RikkiGibson: It's declared in the core project and they're not comfortable with making it lazy because of mutability.

src/vanilla/Model/ResponseJv.cs
RikkiGibson: This one isn't virtual so there's no conceivable purpose. Removing it.

test/azure/src/main/java/fixtures/azureparametergrouping/implementation/ParameterGroupingsImpl.java
RikkiGibson: This means an if (!parameters.IsEmpty) check needs to be added somewhere as part of a refactor.

PR #24, opened 2017-10-26
Convert enum Java file generation to new pattern

daschult:
I'm going to try to break this work up into several smaller reviews rather than one big huge review at the end. One thing to remember is that before I do a review I'll make sure that everything builds, the regenerated sources are identical (or at least really close) to the old sources, and that all of our autorest.java tests still pass.

And don't worry too much about the "Dan" prefix. I'll rename everything before I merge into runtime-v2.


test/vanilla/src/main/java/fixtures/bodycomplex/models/CMYKColors.java
```diff
@@ -10,9 +10,9 @@
 
 package fixtures.bodycomplex.models;
 
-import java.util.Collection;
```
RikkiGibson: Is this the result of regenerating enums with the new generator?

daschult: Yes. The new generator automatically sorts all imports alphabetically.

src/vanilla/DanModel/JavaEnum.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        public JavaEnum(string headerCommentText, string package, string enumName)
+        {
+            HeaderCommentText = headerCommentText;
+            Package = package;
+            EnumName = enumName;
+            values = new List<JavaEnumValue>();
+        }
+
+        public JavaEnum AddValue(string valueName, string valueString)
```
RikkiGibson: I think these models should be fully immutable.

daschult: I think that's a reasonable request. I'll do that.

src/vanilla/DanModel/JavaBlock.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+
+namespace AutoRest.Java.DanModel
+{
+    public class JavaBlock
```
RikkiGibson: Is this like a facade to convey to readers of a large fluent chain what construct we're in right now?

src/vanilla/DanModel/JavaEnum.cs
```diff
@@ -0,0 +1,135 @@
...
+            return new JavaFile(GetFilePath())
+                .MaximumMultipleLineCommentWidth(MaximumMultipleLineCommentWidth)
+                .MultipleLineComment(HeaderComment())
+                .Line()
+                .Package(Package)
+                .Line()
+                .Import("com.fasterxml.jackson.annotation.JsonCreator",
+                        "com.fasterxml.jackson.annotation.JsonValue")
+                .MultipleLineComment(EnumTypeComment())
+                .Block($"public enum {EnumName}", (enumBlock) =>
```
RikkiGibson: I think the lambdas are the most compelling part of all this-- part of the difficulty with StringBuilders is that they stopped resembling the shape of the code being generated (all flat).

src/vanilla/DanModel/JavaBlock.cs
daschult: Yes, and it also serves as a way of restricting what you can add to the chain within a block. For example, there's no `Import()` method on JavaBlock, because you can't have an `import` statement there. That method only exists on JavaFile.

PR #30, opened 2017-11-03
Daschult/new generator impl interfaces

daschult:


src/vanilla/DanModel/JavaMethodGroupInterface.cs
```diff
@@ -0,0 +1,64 @@
...
+                })
+                .PublicInterface(TypeName, (typeBlock) =>
+                {
+                    foreach (JavaMethod method in Methods)
+                    {
+                        typeBlock.MultipleLineComment((comment) =>
+                        {
+                            comment.Line(method.Description);
+                            comment.Line();
+                            if (method.Parameters != null)
```
RikkiGibson: Would encourage requiring people to provide empty collections instead of null unless null is useful for representing a special case.

daschult: There's no special case for null here. This is just preferring to handle null instead of crashing.

PR #31, opened 2017-11-03
Fix build errors in tests

RikkiGibson:
A rename in the runtime caused some build errors to appear in autorest.java, so I'm fixing them.


RikkiGibson:
There's a runtime bug resulting in some test failures for LRO polling responses with invalid JSON bodies. Will rerun CI after that fix is merged and published.

RikkiGibson:
> win32-x64 restart
> linux-x64 restart

RikkiGibson:
> linux-x64 restart

PR #32, opened 2017-11-06
Use new runtime package in new generator

RikkiGibson:
Basically just merges runtime-v2 so we get the new package we've been publishing. This also makes CI available.


src/vanilla/CodeGeneratorJv.cs
```diff
@@ -82,6 +82,25 @@ protected Settings Settings
 
             //Enums
             await WriteEnumJavaFiles(codeModel).ConfigureAwait(false);
+            
+            //XML wrappers
+            if (codeModel.ShouldGenerateXmlSerializationCached)
```
daschult: If the XML wrappers now are created in each CodeGenerator implementation, then can you extract this to a protected method that is inherited by CodeGeneratorJva and CodeGeneratorJvaf instead of just having it be copied and pasted.

src/vanilla/ModelNameComparer.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using AutoRest.Core.Model;
+
+namespace AutoRest.Java
+{
+    public class ModelNameComparer : IEqualityComparer<ModelType>
```
daschult: Did this class accidentally slip into this pull request?

src/vanilla/CodeGeneratorJv.cs
RikkiGibson: Good idea. Yes.

src/vanilla/ModelNameComparer.cs
RikkiGibson: Nope, but when the XML serialization stuff is consolidated it might go back to being an inner class.

RikkiGibson:
> linux-x64 restart

PR #47, opened 2017-12-01
Replace RestClient with HttpPipeline

daschult:


olydis:
# 🤖 AutoRest automatic feature coverage report 🤖
*feature set version 1.9.2*

## ❌️ General: 0%

no tests were run for this category


## ❌️ Azure: 0%

no tests were run for this category

.gulp/regeneration.iced
```diff
@@ -1,21 +1,26 @@
-
 ###############################################
 # LEGACY 
 # Instead: have bunch of configuration files sitting in a well-known spot, discover them, feed them to AutoRest, done.
 
+rimraf = require "rimraf"
```
RikkiGibson: Should I just say OK, this works, I'm leaving it alone?

RikkiGibson: That's not a complaint, by the way

daschult: No, you should understand what it is. rimraf is a npm package that allows me to delete files and folders easily. I use it on line 34 to delete the contents of the folder at the baseFolderPath.

olydis:
# 🤖 AutoRest automatic feature coverage report 🤖
*feature set version 1.9.2*

## ❌️ General: 0%

no tests were run for this category


## ❌️ Azure: 0%

no tests were run for this category

src/azurefluent/CodeGeneratorJvaf.cs
```diff
@@ -58,7 +58,10 @@ public class CodeGeneratorJvaf : CodeGeneratorJva
             {
                 // Manager
                 await WriteAzureServiceManagerJavaFile(codeModel).ConfigureAwait(false);
+            }
 
+            if (true == Settings.Instance.Host?.GetValue<bool?>("regenerate-pom").Result)
```
RikkiGibson: Was there a reason it became more convenient to separate these out? They were one flag initially because generating a manager and POM were both part of scaffolding a new module. I'm not aware of cases where a project has one but not the other. At the same time it's also not a big deal to have to pass two flags--as long as I get on top of documenting them (hey, there's a good task to pick up).

daschult: I wouldn't say it was more convenient, but rather it was more clear. There was no way to know that adding the flag `regenerate-pom` was the trigger to regenerate the managers as well. I'm fine with adding a flag called `generate-new-module` or something like that that triggers both `regenerate-pom` and `regenerate-managers`, but I think it should be more clear than it was.

PR #48, opened 2017-12-04
Remove CodeModel sub-types

daschult:
This is just another refactoring step in simplifying our java generator.


src/azurefluent/CodeGeneratorJvaf.cs
```diff
@@ -19,51 +17,40 @@ public class CodeGeneratorJvaf : CodeGeneratorJva
         /// <returns></returns>
         public override async Task Generate(CodeModel cm)
```
RikkiGibson: Might as well do the same rename in here.

src/vanilla/DanModel/DanCodeGenerator.cs
```diff
@@ -2037,7 +2085,7 @@ public static JavaFile GetServiceClientJavaFile(CodeModelJv codeModel, Settings
                             }
                             else
                             {
-                                interfaceBlock.Annotation($"Headers({{ \"x-ms-logging-context: {codeModel.FullyQualifiedDomainName} {method.Name}\" }})");
+                                interfaceBlock.Annotation($"Headers({{ \"x-ms-logging-context: {GetFullyQualifiedDomainName(codeModel)} {method.Name}\" }})");
```
RikkiGibson: You may know already-- x-ms-logging-context is dead because the callerMethod property on HttpRequest accomplishes the same thing for us. It's just that nobody has gotten around to removing it yet.

daschult: Good point. I'll remove it.

src/vanilla/DanModel/DanCodeGenerator.cs
```diff
@@ -35,18 +39,64 @@ public static class DanCodeGenerator
         private const string azureProxyImport = "com.microsoft.azure.v2." + azureProxyType;
         private const string azureProxyType = "AzureProxy";
 
-        public static JavaFile GetAzureServiceManagerJavaFile(CodeModelJva codeModel, Settings settings)
+        private const string implPackage = "implementation";
+        internal const string modelsPackage = ".models";
+
+        public static readonly IDictionary<KeyValuePair<string, string>, string> pageClasses = new Dictionary<KeyValuePair<string, string>, string>();
```
RikkiGibson: I would be happier with this kind of stuff being instance variables and a bunch of these methods becoming instance methods in order to access them, but until (if ever?) AutoRest allows a generator to run as a daemon consuming multiple Swaggers in parallel, it's not essential.

PR #50, opened 2017-12-05
Don't add https if no protocol is specified, remove DEFAULT_BASE_URL

daschult:


daschult:
Tests pass locally, so I'm going to merge this.

PR #53, opened 2017-12-07
Remove EnumTypeJv

daschult:


src/vanilla/DanModel/DanCodeGenerator.cs
```diff
@@ -3540,7 +3547,120 @@ internal static IModelType GetPropertyModelType(Property property)
...
+        internal static IEnumerable<string> GetIModelTypeImports(IModelType modelType)
+            => GetIModelTypeImports(modelType, Settings.Instance);
+
+        private static IEnumerable<string> GetIModelTypeImports(IModelType modelType, Settings settings)
+        {
+            IEnumerable<string> result = Enumerable.Empty<string>();
+
+            if (modelType != null)
+            {
+                if (modelType is EnumType)
```
RikkiGibson: Note that `is` always returns false if the lhs is null. Also is there any case where something containing a null IModelType is something besides a malformed object model?

daschult: That's true about returning false for null. I'll remove that check in a later pull request.
I don't think a null IModelType is ever valid.

PR #57, opened 2017-12-12
Flatten runtime classes (RequestPolicy.Factory and HttpPipeline.Builder)

daschult:


RikkiGibson:
This was really all that had to get fixed? I guess that's a relief.

RikkiGibson:
There's just a lot of repetitive code in the tests, so I figured even a few find-and-replaces might be necessary.

PR #71, opened 2017-12-20
Cleanup DanCodeGenerator part 3

daschult:


daschult:
Good point. I was planning on moving us to just using arrays soon, which should really help with this.

PR #99, opened 2018-01-16
Fix optional body tests

daschult:


RikkiGibson:
Thank you. Failures probably related to the runtime changes. I'm taking care of it in the other PR.

PR #112, opened 2018-02-05
collectBytes() -> collectBytesInArray()

RikkiGibson:
This shouldn't cause much impact on your ongoing work and enables running tests with the changes I've been doing in the runtime.


RikkiGibson:
We'll need to publish before this becomes valid--so depending on how the timing works out, this PR might be moot :)

RikkiGibson:
Something funny's going on with AutoRest CI. autorest.java isn't able to communicate with the test server. Will follow up on it later.

RikkiGibson:
Looks like this PR is redundant now--correct me if I'm wrong

PR #149, opened 2018-02-26
Represent empty XML elements as empty instead of null objects

RikkiGibson:
This PR is about improving the usability of our deserialized models primarily by giving safe default/empty values. In a user-friendly library there is no reason that
```java
MyModel myModel = service.get();
myModel.myList().size()
```
should throw NPE.

I'm essentially achieving this for lists by assigning an empty list as default value in an "embedded" list case (e.g the SlideShow model) and adding safe getters for wrapped list properties (e.g. the AppleBarrel model).

I have gone back and forth on getting wrecked by JAXB and then getting wrecked by Jackson XML probably 2 or 3 times now and man it is dizzying. Jackson requires one pile of workarounds while JAXB requires another. I'm concluding that this year at least it will be best to just work around whatever Jackson XML quirks we encounter by any means necessary.

Consider the following model class:

```java
@XmlRootElement(name = "bean")
class Bean {
    @XmlElementWrapper(name = "myMap")
    private Map<String, String> myMap;
}
```

JAXB will serialize an instance to this:
```xml
<bean><myMap><entry><key>bar</key><value>other-value</value></entry><entry><key>foo</key><value>value</value></entry></myMap></bean>
```
Jackson XML will serialize an instance to this:
```xml
<bean><myMap><bar>other-value</bar><foo>value</foo></myMap></bean>
```

The types of data that our SDKs will represent as a map we'll want to serialize to the Jackson way, e.g. Azure Storage blob metadata from listBlobs.

Convincing JAXB to do a better thing requires implementing an XmlAdapter and decorating every Map property with it, or decorating every package-info.java that contains a type you want to serialize with a Map property with a reference to this custom map serializer/deserializer type. JAXB has no way to register an XmlAdapter on a JAXB Marshaller/Unmarshaller instance for every instance of a particular model type it encounters.
https://stackoverflow.com/questions/3941479/jaxb-how-to-marshall-map-into-keyvalue-key

Similarly XmlAdapters would have to be implemented for other common types
-	DateTime
-	ExpandableStringEnum

There are probably some others that I am not familiar with yet. We have lots of assumptions in the generator and runtime related to Jackson annotations.

We are so close to being there with Jackson XML. It's just that it gives null values when deserializing empty XML elements to string. It seems Jackson 2.9 has fixed that so I added test coverage for that specific case that we can just ignore until we decide it's best to update to 2.9. That will be its own hazard as it seems to break some of our polymorphism tests for reasons I haven't completely looked into.

This PR also contains some fixes related to XML element names of root list wrappers.


RikkiGibson:
Addresses Azure/autorest-clientruntime-for-java#381

src/JavaCodeGenerator.cs
```diff
@@ -1956,8 +1962,10 @@ public static JavaFile GetPageJavaFile(PageDetails pageClass, JavaSettings setti
 
         public static JavaFile GetXmlSequenceWrapperJavaFile(XmlSequenceWrapper xmlSequenceWrapper, JavaSettings settings)
         {
-            string xmlElementName = xmlSequenceWrapper.XmlElementName;
-            string xmlElementNameCamelCase = xmlElementName.ToCamelCase();
+            string xmlRootElementName = xmlSequenceWrapper.XmlRootElementName;
```
alvadb: Should these names be camel cased?

src/JavaCodeGenerator.cs
```diff
@@ -2515,8 +2535,12 @@ public static JavaFile GetModelJavaFile(ServiceModel model, JavaSettings setting
                         string expression = $"this.{property.Name}";
                         if (sourceTypeName == targetTypeName)
                         {
-                            if (settings.ShouldGenerateXmlSerialization && property.IsXmlWrapper)
+                            if (settings.ShouldGenerateXmlSerialization && property.IsXmlWrapper && property.WireType is ListType listType)
```
alvadb: Doesn't IsXmlWrapper imply the WireType will be a list?

RikkiGibson: If I remember right, the main reason for this is that the pattern matching `is ListType listType` is convenient here

src/JavaCodeGenerator.cs
RikkiGibson: No because these values are used as JsonProperty annotation arguments--if the swagger says the name of the root element is <foo> we have to be careful not to change it to <Foo>

RikkiGibson:
Some infrastructure issues around publishing a new runtime to fix the CI build errors, so I'm just going to merge

PR #176, opened 2018-04-06
Generate Java 7 compatible code in vanilla generator

RikkiGibson:
- Uses threetenbp
- Generates anonymous classes for vanilla (may need to do Azure and other variants down the line, but it seems like this is all that's necessary just to get started)

- CI will probably fail--will update when I have a story for publishing the Java 7 runtime


test/vanilla/AutoRestRFC1123DateTimeTestService.java
```diff
@@ -1,37 +0,0 @@
-/**
```
RikkiGibson: I have no idea how these files were around--they were never being compiled, but I just deleted them at some point and that deletion stuck around in here.

test/vanilla/src/main/java/fixtures/bodybyte/implementation/BytesImpl.java
```diff
@@ -296,6 +312,11 @@ public Completable putNonAsciiAsync(@NonNull byte[] byteBody) {
      */
     public Maybe<byte[]> getInvalidAsync() {
         return getInvalidWithRestResponseAsync()
-            .flatMapMaybe((BodyResponse<byte[]> res) -> res.body() == null ? Maybe.empty() : Maybe.just(res.body()));
+            .flatMapMaybe(new Function<BodyResponse<byte[]>, Maybe<byte[]>>() {
+                public Maybe<byte[]> apply(BodyResponse<byte[]> res) {
+                    return res.body() == null ? Maybe.<byte[]>empty() : Maybe.just(res.body());
+                }
+            }
+            );
```
daschult: Hmm... I'm not really a fan of the right parenthesis being on the next line after the right curly-bracket. Since this for a one-off thing I think it's fine, but if you have some extra time you might look into figuring out a way to fix this. This is the sort of thing that makes people shudder when they think of "generated code".

src/Model/RestAPIMethod.cs
```diff
@@ -154,6 +154,8 @@ public void AddImportsTo(ISet<string> imports, bool includeImplementationImports
 
             if (includeImplementationImports)
             {
+                imports.Add("io.reactivex.functions.Function");
```
daschult: Please add an entry in the ClassType.cs file for the Function type and then use the new entry here. That makes it much easier to determine usages of Function in the code.

src/JavaCodeGenerator.cs
```diff
@@ -4551,16 +4562,27 @@ private static void AddClientMethods(JavaType typeBlock, IEnumerable<ClientMetho
...
                                 else if (isFluentDelete)
                                 {
                                     function.Text($".flatMapMaybe(");
-                                    function.Lambda(returnValueTypeArgumentClientType.ToString(), "res", "Maybe.empty()");
+                                    function.AnonymousClass($"new Function<{returnValueTypeArgumentClientType}, {GenericType.Maybe(ClassType.Void)}>()", anonymousClassBlock =>
+                                    {
+                                        anonymousClassBlock.PublicMethod(
+                                            $"{GenericType.Maybe(ClassType.Void)} apply({returnValueTypeArgumentClientType} res)",
+                                            method => method.Return("Maybe.empty()"));
+                                    });
```
daschult: If you look at your usages of the AnonymousClass() function, you'll see that there is a pattern. They always take
```cs
$"new Function<{returnValueTypeArgumentClientType}, {return-type}>()"
```
as their first argument. I'd recommend inlining this into the body of the AnonymousClass function. so that AnonymousClass() takes the returnValueTypeArgumentClientType as its first parameter and return-type as its second argument.
You'll also notice that the interior call to PublicMethod() always looks the same too:
```cs
anonymousClassBlock.PublicMethod($"{return-type} apply({returnValueTypeARgumentClientType} res)", body-lambda);
```
You can abstract that away too by providing the `body-lambda` argument to AnonymousClass() so that your new invocation looks like this:
```cs
function.AnonymousClass(returnValueTypeArgumentClientType, return-type, body-lambda);
```
It's much smaller and much, much less prone to typos (did I remember to put the closing right arrow-bracket on the Function signature?).

src/JavaFileContents.cs
```diff
@@ -319,11 +319,22 @@ public void ReturnAnonymousClass(string anonymousClassDeclaration, Action<JavaCl
...
 
+        public void AnonymousClass(string anonymousClassDeclaration, Action<JavaClass> anonymousClassBlock)
+        {
+            Line($"{anonymousClassDeclaration} {{");
+            Indent(() =>
+            {
+                JavaClass javaClass = new JavaClass(this);
+                anonymousClassBlock(javaClass);
+            });
+            Line("}");
```
daschult: I think if you change this to `Text("}");` that will solve your right parenthesis on the next line problem that I mentioned in a different comment.

test/vanilla/src/main/java/fixtures/bodybyte/implementation/BytesImpl.java
daschult: I think I figured out an easy fix for this in a later comment.

src/JavaCodeGenerator.cs
RikkiGibson: Renaming `AnonymousClass()` to `AnonymousFunction()` so that it's more reasonable to specialize for usability's sake. 👍

src/JavaFileContents.cs
RikkiGibson: I'll go with that.

RikkiGibson:
Review feedback has been addressed (including the additional newline issue).

PR #177, opened 2018-04-11
Automatically add Content-Length parameter if needed

### LINKED ISSUE #160: Add Content-Length parameter when Stream request body is used
Right now we're having Swagger authors add a Content-Length parameter to their specs. This isn't quite right. Since it doesn't look like we're going to support chunked upload scenarios, we might want to just have the generator automatically add a Content-Length parameter when a `Flowable<ByteBuffer>` is going to be used for the request body.

RikkiGibson:
Fixes #160.

### DIFF
--- a/src/JavaCodeGenerator.cs
+++ b/src/JavaCodeGenerator.cs
@@ -1066,6 +1066,30 @@ namespace AutoRest.Java
                 pagedListTypes.Add(responseBodyWireListType);
             }
 
+            // If there is a stream body and no Content-Length header parameter, add one automatically
+            // Convert to list so we can use methods like FindIndex and Insert(int, T)
+            List<AutoRestParameter> autoRestMethodParameters = new List<AutoRestParameter>(autoRestMethod.Parameters);
+            int streamBodyParameterIndex = autoRestMethodParameters.FindIndex(p => p.Location == AutoRestParameterLocation.Body && p.ModelType is AutoRestPrimaryType mt && mt.KnownPrimaryType == AutoRestKnownPrimaryType.Stream);
+            if (streamBodyParameterIndex != -1 &&
+                !autoRestMethodParameters.Any(p =>
+                    p.Location == AutoRestParameterLocation.Header && p.SerializedName.EqualsIgnoreCase("Content-Type")))
+            {
+                AutoRestParameter contentLengthParameter = DependencyInjection.New<AutoRestParameter>();
+                contentLengthParameter.Method = autoRestMethod;
+                contentLengthParameter.IsRequired = true;
+                contentLengthParameter.Location = AutoRestParameterLocation.Header;
+                contentLengthParameter.SerializedName = "Content-Length";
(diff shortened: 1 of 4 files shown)


src/JavaCodeGenerator.cs
```diff
@@ -1066,6 +1066,30 @@ private static RestAPIMethod ParseRestAPIMethod(AutoRestMethod autoRestMethod, J
...
+                contentLengthParameter.Name = "contentLength";
+                contentLengthParameter.Documentation = "The content length";
+                contentLengthParameter.ModelType = DependencyInjection.New<AutoRestPrimaryType>(AutoRestKnownPrimaryType.Long);
+
+                // Add the Content-Length parameter before the body parameter
+                autoRestMethodParameters.Insert(streamBodyParameterIndex, contentLengthParameter);
+            }
+
+            autoRestMethod.ClearParameters();
+            autoRestMethod.AddRange(autoRestMethodParameters);
```
daschult: You only need to do these two lines if you insert the contentLengthParameter, right? In that case, this should probably be moved inside the if-statement. If you didn't insert the contentLengthParameter, then there's no need to clear the autoRestMethod's parameters.

test/vanilla/src/main/java/fixtures/bodyformdata/Formdatas.java
```diff
@@ -71,39 +71,43 @@
...
      *
+     * @param contentLength The content length.
      * @param fileContent File to upload.
      * @throws IllegalArgumentException thrown if parameters fail the validation.
      * @throws ErrorException thrown if the request is rejected by server.
      * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
      * @return the Flowable&lt;ByteBuffer&gt; object if successful.
      */
-    Flowable<ByteBuffer> uploadFileViaBody(@NonNull Flowable<ByteBuffer> fileContent);
+    Flowable<ByteBuffer> uploadFileViaBody(@NonNull long contentLength, @NonNull Flowable<ByteBuffer> fileContent);
```
daschult:
```java
@NonNull long contentLength
```
I guess this is a bug in the generator. I think it's pretty clear that long will never be null.

RikkiGibson: I guess it's not needed, but I also don't foresee it causing users any pain. Are primitives in general having this annotation applied?

RikkiGibson:
Moving to #178 so we can have CI

src/JavaCodeGenerator.cs
RikkiGibson: Sure, seems better. Will fix in the other branch.

test/vanilla/src/main/java/fixtures/bodyformdata/Formdatas.java
daschult: It looks like they are. I'm surprised, because I thought I had handled this case.

PR #180, opened 2018-04-19
Codegen that produce fluent type interfaces, impls and uses fluent patterns

anuchandy:


jianghaolu:
@RikkiGibson Okay. Thanks for the info. We will probably be working off fluent-gen-v2 branch for a while. Better yet I really want to rename it 😂

PR #220, opened 2018-05-17
Cognitive Services Java auto-gen related changes

milismsft:


RikkiGibson:
@milismsft Note that CI only works if the head branch is on the Azure repo, not on a fork

lmazuel:
Will not use this version for CS going forward

PR #256, opened 2018-06-24
Support for LROOptions extension handling, correctly import InnerSupp…

anuchandy:
…ort* based on multi-apiversion vs non-muti-apiversion (#255)

* Support for LROOptions extension handling, correctly import InnterSupport* based on multi-apiversion vs non-muti-apiversion

* Regenerating autorest tests


anuchandy:
This PR depends on pro operation options support in the runtime, introduced in the [PR](https://github.com/Azure/autorest-clientruntime-for-java/pull/442)

PR #265, opened 2018-07-09
Use xml name

### LINKED ISSUE #264: XML name should be used if present in JsonProperty annotation
Based on some troubleshooting work I did with @rickle-msft. It seems like we don't properly handle cases where the XML name is different than the property name.

Sounds like @lmazuel is planning on adding test coverage in autorest.testserver so we may want to wait for that.

RikkiGibson:
Resolves #264

Basically, I'm finding that the `XmlName` on an AutoRest `Property` is not relevant in the case of a composite type (i.e. a $ref). You have to instead dig in to the `Property.ModelType.XmlName`.


RikkiGibson:
Spun off the work for the recent testserver stuff to #266

Issue #266, opened 2018-07-10
Long running paging


RikkiGibson:
A long running paging operation was recently added to autorest.java. The v2 generator at least chokes on this and produces code that doesn't compile.

This is a real thing in ARM, so this will have to be fixed at some point.

jianghaolu:
Verify if v4 support LRO + Paging and triage as needed.

PR #269, opened 2018-07-11
Prefer the ElementType.XmlProperties.Name if present

RikkiGibson:
Similar to the case for a property's XmlName, we have to check the ElementType.XmlProperties.Name before checking the property's ElementXmlName. I've been hashing this question out for a while and basically concluded that it's not realistic to move the code for these two workarounds over to the modeler.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft.azure/autorest.java",
-  "version": "2.0.21",
+  "version": "2.0.22",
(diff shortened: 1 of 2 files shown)


RikkiGibson:
I really want to update the autorest.testserver tests to exercise this but I'm blocked on integrating an updated testserver until we solve the "long running paging" in #266

RikkiGibson:
The Windows CI failure is just due to a timeout on a name resolution failure.

PR #277, opened 2018-08-15
Fix an inner generation bug

jianghaolu:


src/azurefluent/TransformerJvaf.cs
```diff
@@ -153,9 +152,9 @@ private void AppendInnerToTopLevelType(IModelType type, CodeModel serviceClient)
             if (compositeType != null && !compositeType.IsResource)
             {
                 compositeType.IsInnerModel = true;
-                foreach (var t in serviceClient.ModelTypes)
+                foreach (var t in serviceClient.ModelTypes.Where(mt => !(mt as CompositeTypeJvaf).IsInnerModel))
```
RikkiGibson: If the cast is expected to always succeed it's good to use the cast syntax instead of `as`

Issue #278, opened 2018-08-16
Generated code is missing imports


hovsepm:
https://github.com/Azure/azure-sdk-for-java/pull/2172 

`
[[1;31mERROR[m] COMPILATION ERROR : 
[[1;34mINFO[m] -------------------------------------------------------------
[[1;31mERROR[m] /home/travis/build/Azure/azure-sdk-for-java/applicationinsights/resource-manager/v2015_05_01/src/main/java/com/microsoft/azure/management/applicationinsights/v2015_05_01/implementation/WorkItemConfigurationsInner.java:[337,50] cannot find symbol
  symbol:   method items()
  location: class com.microsoft.azure.management.applicationinsights.v2015_05_01.implementation.WorkItemConfigurationInner
[[1;31mERROR[m] /home/travis/build/Azure/azure-sdk-for-java/applicationinsights/resource-manager/v2015_05_01/src/main/java/com/microsoft/azure/management/applicationinsights/v2015_05_01/implementation/ApplicationInsightsComponentFavoriteImpl.java:[37,75] cannot find symbol
  symbol:   method id()
  location: variable inner of type com.microsoft.azure.management.applicationinsights.v2015_05_01.implementation.ApplicationInsightsComponentFavoriteInner
[[1;31mERROR[m] /home/travis/build/Azure/azure-sdk-for-java/applicationinsights/resource-manager/v2015_05_01/src/main/java/com/microsoft/azure/management/applicationinsights/v2015_05_01/implementation/ApplicationInsightsComponentFavoriteImpl.java:[38,70] cannot find symbol
  symbol:   method id()
  location: variable inner of type com.microsoft.azure.management.applicationinsights.v2015_05_01.implementation.ApplicationInsightsComponentFavoriteInner
[[1;31mERROR[m] /home/travis/build/Azure/azure-sdk-for-java/applicationinsights/resource-manager/v2015_05_01/src/main/java/com/microsoft/azure/management/applicationinsights/v2015_05_01/implementation/ApplicationInsightsComponentFavoriteImpl.java:[39,68] cannot find symbol
  symbol:   method id()
  location: variable inner of type com.microsoft.azure.management.applicationinsights.v2015_05_01.implementation.ApplicationInsightsComponentFavoriteInner
[...]

PR #283, opened 2018-09-21
Fix for issue #278 (Accessing non-existing id & items property) (#282)

anuchandy:
* Generate id parsing statements in NestedResource::ctr(inner, manager) body only if the resource is updatable|refreshable

* MethodJva.cs:: Emit body().items() accessor only if body() is of-type PageImpl

### DIFF
--- a/src/azure/Model/MethodJva.cs
+++ b/src/azure/Model/MethodJva.cs
@@ -607,23 +607,14 @@ namespace AutoRest.Java.Azure.Model
             else if (this.IsPagingNonPollingOperation)
             {
                 IndentedStringBuilder builder = new IndentedStringBuilder();
-                if (ReturnTypeJva.GenericBodyWireTypeString.Contains("PageImpl1<"))
-                {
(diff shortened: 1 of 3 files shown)


anuchandy:
@jianghaolu could you take a look at the second commit in this PR (which touches MethodJva.cs file). This fixes the compile time error in this [file](https://github.com/Azure/azure-sdk-for-java/blob/ef4691afeb0ea7e419ad8d73d673bd175907746f/applicationinsights/resource-manager/v2015_05_01/src/main/java/com/microsoft/azure/management/applicationinsights/v2015_05_01/implementation/WorkItemConfigurationsInner.java#L337) 

```
/applicationinsights/resource-manager/v2015_05_01/src/main/java/com/microsoft/azure/management/applicationinsights/v2015_05_01/implementation/WorkItemConfigurationsInner.java:[337,50] cannot find symbol symbol: method items() location: class 
```

Fix is - we need to emit body().items() accessor only if body() is of-type Page (PageImpl).

anuchandy:
Discussed with @jianghaolu on this, indeed an issue in the swagger (x-ms-pagable ext applied when it should not), opened a PR to fix swagger and opened an issue to add linter rule for this.

PR #388, opened 2019-08-23
Convert all bytebufs to bytebuffers

jianghaolu:


src/ClientModel/ClassType.cs
```diff
@@ -40,7 +40,7 @@ public class ClassType : IType
         public static readonly ClassType RestProxy = new ClassType("com.azure.core.implementation", "RestProxy");
         public static readonly ClassType Validator = new ClassType("com.azure.core.implementation", "Validator");
         public static readonly ClassType Function = new ClassType("io.reactivex.functions", "Function");
-        public static readonly ClassType ByteBuf = new ClassType("io.netty.buffer", "ByteBuf");
+        public static readonly ClassType ByteBuffer = new ClassType("io.netty.buffer", "ByteBuffer");
```
alzimmermsft: Wrong import

jianghaolu: when find replace fails

jianghaolu: fixed

src/JvModel/PrimaryTypeJv.cs
```diff
@@ -70,7 +70,7 @@ public string ModelTypeName
                         result = IsNullable ? "Long" : "long";
                         break;
                     case KnownPrimaryType.Stream:
-                        result = "Flowable<ByteBuffer>";
+                        result = "Flowable<ByteBufferfer>";
```
alzimmermsft: ByteBuffer?

jianghaolu: Hmm this is wrong in many ways. The path is somehow never hit. Fixing.

PR #394, opened 2019-10-16
Fix for service client LRO operations.

yaohaizh:
@jianghaolu This fix will unblock the SDK generation for the latest (2019-08) network RP, which contains a root method on the service side for LRO.


fearthecowboy:
# 🤖 AutoRest automatic publish job 🤖
## success (version: 2.1.95)

PR #395, opened 2019-10-16
Sakintoye/varargs support

sakintoye:
Adds variable arguments as syntactic sugar to a Model's property setter method that accepts a List of elements.

This provides a convenient way for users to provide var args, rather `Arrays.asList(....)` which can be clunky.

When the `varArg` flag is set to `true`, it transforms the setter method from:

```java
public Foo setBar(List<String> bar) {
        this.bar = Arrays.asList(bar);
        return this;
}
```
to:
```java
public Foo setBar(String... bar) {
        this.bar = Arrays.asList(bar);
        return this;
}
```

### Usage

In an Autorest configuration file, a user will include something like below

```markdown
model-properties:
    Foo.bar:
        varArg: true
```

### DIFF
--- a/src/ClientModel/ClientModelProperty.cs
+++ b/src/ClientModel/ClientModelProperty.cs
@@ -28,7 +28,9 @@ namespace AutoRest.Java.Model
         /// <param name="defaultValue">The default value expression of this property.</param>
         /// <param name="isReadOnly">Whether or not this property's value can be changed by the client library.</param>
         /// <param name="headerCollectionPrefix">The prefix of the headers that make up this property's values.</param>
-        public ClientModelProperty(string name, string description, string annotationArguments, bool isXmlAttribute, string xmlName, string serializedName, bool isXmlWrapper, string xmlListElementName, IType wireType, IType clientType, bool isConstant, string defaultValue, bool isReadOnly, bool wasFlattened, string headerCollectionPrefix)
+        /// <param name="useVarArg">If true, the property's setter method will be represented as variable argument (Varargs) in the model.</param>
+        public ClientModelProperty(string name, string description, string annotationArguments, bool isXmlAttribute, string xmlName, string serializedName, bool isXmlWrapper,
+            string xmlListElementName, IType wireType, IType clientType, bool isConstant, string defaultValue, bool isReadOnly, bool wasFlattened, string headerCollectionPrefix, bool useVarArg)
         {
             Name = name;
             Description = description;
@@ -45,12 +47,15 @@ namespace AutoRest.Java.Model
             IsReadOnly = isReadOnly;
             WasFlattened = wasFlattened;
             HeaderCollectionPrefix = headerCollectionPrefix;
+            UseVarArg = useVarArg;
         }
 
         /// <summary>
         /// Get the name of this property.
         /// </summary>
         public string Name { get; }
+
+        public bool UseVarArg { get; }
 
         public string GetterName
         {
--- a/src/ClientModel/ClientModelProperty.cs
+++ b/src/ClientModel/ClientModelProperty.cs
@@ -55,8 +55,6 @@ namespace AutoRest.Java.Model
         /// </summary>
         public string Name { get; }
 
-        public bool UseVarArg { get; }
-
         public string GetterName
         {
             get
@@ -149,6 +147,11 @@ namespace AutoRest.Java.Model
         /// The prefix of the headers that make up this property's values.
         /// </summary>
         public string HeaderCollectionPrefix { get; }
+
+        /// <summary>
+        /// If true, the setter method of this property will have a variable arguments signature
+        /// </summary>
+        public bool UseVarArg { get; }
 
         /// <summary>
         /// Add this ServiceModelProperty's imports to the provided ISet of imports.
--- a/src/Mapper/ModelPropertyMapper.cs
+++ b/src/Mapper/ModelPropertyMapper.cs
@@ -120,7 +120,24 @@ namespace AutoRest.Java
 
             bool wasFlattened = property.WasFlattened();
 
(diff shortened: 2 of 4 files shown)


src/ClientModel/ClientModelProperty.cs
```diff
@@ -45,13 +47,16 @@ public ClientModelProperty(string name, string description, string annotationArg
...
             HeaderCollectionPrefix = headerCollectionPrefix;
+            UseVarArg = useVarArg;
         }
 
         /// <summary>
         /// Get the name of this property.
         /// </summary>
         public string Name { get; }
 
+        public bool UseVarArg { get; }
```
chenmliu: nit: method comment

src/Mapper/ModelPropertyMapper.cs
```diff
@@ -120,7 +120,24 @@ public ClientModelProperty Map(PropertyJv property)
 
             bool wasFlattened = property.WasFlattened();
 
-            return new ClientModelProperty(property.Name, description, annotationArguments, isXmlAttribute, xmlName, serializedName, isXmlWrapper, xmlListElementName, propertyWireType, propertyClientType, isConstant, defaultValue, isReadOnly, wasFlattened, headerCollectionPrefix);
+            bool useVarArgs = false;
+
+            if (property.Parent is CompositeType parent)
+            {
+                var modelPropertyName= $"{parent.Name}.{property.Name}";
+                if (Settings.Instance.CustomSettings["ModelProperties"] is Dictionary<string, Dictionary<string, bool>> modelProperties)
```
chenmliu: Is the custom setting `ModelProperties` guaranteed to exist? If not, this if condition could throw

src/ClientModel/ClientModelProperty.cs
```diff
@@ -28,7 +28,9 @@ public class ClientModelProperty
         /// <param name="defaultValue">The default value expression of this property.</param>
         /// <param name="isReadOnly">Whether or not this property's value can be changed by the client library.</param>
         /// <param name="headerCollectionPrefix">The prefix of the headers that make up this property's values.</param>
-        public ClientModelProperty(string name, string description, string annotationArguments, bool isXmlAttribute, string xmlName, string serializedName, bool isXmlWrapper, string xmlListElementName, IType wireType, IType clientType, bool isConstant, string defaultValue, bool isReadOnly, bool wasFlattened, string headerCollectionPrefix)
+        /// <param name="useVarArg">If true, the property's setter method will be represented as variable argument (Varargs) in the model.</param>
```
chenmliu: nit: VarArgs

src/Mapper/ModelPropertyMapper.cs
```diff
@@ -120,7 +120,24 @@ public ClientModelProperty Map(PropertyJv property)
...
+                {
+                    if (modelProperties.ContainsKey(modelPropertyName) &&
+                        modelProperties[modelPropertyName].ContainsKey("varArg") &&
+                        modelProperties[modelPropertyName]["varArg"])
```
chenmliu: Does `modelProperties[modelPropertyName]["varArg"]` only verify not-null? Do we need not-empty too?

src/Mapper/ModelPropertyMapper.cs
sakintoye: ModelProperties is initialized, just as every other properties in `CustomSettings`

src/Mapper/ModelPropertyMapper.cs
sakintoye: We don't need to verify not-null or not-empty. `varArg` is a boolean type

src/Template/ModelTemplate.cs
```diff
@@ -228,7 +232,19 @@ public void Write(ClientModel model, JavaFile javaFile)
...
+                        if (property.UseVarArg)
+                        {
+                            if (!(property.WireType is ListType))
+                            {
+                                throw new InvalidOperationException($"Cannot create a varArgs for a non ListType '{property.WireType}'");
+                            }
+                            if (string.IsNullOrWhiteSpace(property.XmlListElementName)) {
+                                throw new InvalidOperationException($"Could not determine the varArgs element");
+                            }
+                            parameterType = $"{property.XmlListElementName}...";
```
chenmliu: just for my understanding, what are the three dots for? :)

src/Template/ModelTemplate.cs
```diff
@@ -257,7 +273,12 @@ public void Write(ClientModel model, JavaFile javaFile)
                                 }
                                 else
                                 {
-                                    methodBlock.Line($"this.{property.Name} = {expression};");
+                                    if (property.UseVarArg) {
```
chenmliu: curious if the conditional operator works in java?

`methodBlock.Line($"this.{property.Name} = property.UseVarArg ? Arrays.asList({expression}) : {expression};");`

src/Template/ModelTemplate.cs
sakintoye: That's the [varargs](https://docs.oracle.com/javase/8/docs/technotes/guides/language/varargs.html) syntax :)

src/Mapper/ModelPropertyMapper.cs
chenmliu: this if check seems redundant because ProcessInternal() in Program always sets it to a dictionary

jianghaolu:
Can we simplify the configuration to something like this?
```yaml
vararg-properties:Foo.bar,Foo.baz
```

I feel the `Model-properties` argument name is a bit too general.

sakintoye:
> Can we simplify the configuration to something like this?

We can. My initial thought was to set the stage for other ModelProperty level configuration, for example, if we needed to add static property, it'd be extendable as below

```java
model-properties:
    Foo.bar:
        varArg: true
        static: true
```

`static` may be a terrible example but you get the gist :) I don't know if there are reasons why that type of configuration is bad so definitely open to changing it.

> Also, comma separated list is also a bit easier to provide as a command line argument as --varag-properties=Foo.bar,Foo.baz when modifying a README is not an option

I like the comma separated list for command line argument!

PR #434, opened 2019-11-27
v4 include tests module

weidongxu-microsoft:


jianghaolu:
The problem with this is that when a generated file doesn't compile - the whole project doesn't compile - which makes it difficult to rebuild (`mvn package -Dlocal`) on the command line.

weidongxu-microsoft:
Yes, agree. I will keep it there (and anyone want to run the test could cherry-pick --no-commit to ones local). We might get it later when the generated code get more stable.

Issue #418, opened 2019-11-22
Generate code for header.json


jianghaolu:

mayurid:
Closed as issue is merged

Issue #419, opened 2019-11-22
Generate code for head.json & head-exceptions.json


jianghaolu:

mayurid:
Closing this issue as it is merged.

weidongxu-microsoft:
This issue is linked incorrectly. Not started yet.

jianghaolu:
fixed via #446.

PR #435, opened 2019-11-29
V4 header

weidongxu-microsoft:
Correction, it is https://github.com/Azure/autorest.java/issues/418

Code change https://github.com/Azure/autorest.java/commit/e0f534cf9f00fc93c7305be1c17a9a6c534e2422

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
@@ -1,5 +1,6 @@
 package com.azure.autorest.mapper;
 
+import com.azure.autorest.extension.base.model.codemodel.Header;
 import com.azure.autorest.extension.base.model.codemodel.Operation;
 import com.azure.autorest.extension.base.model.codemodel.Parameter;
 import com.azure.autorest.extension.base.model.codemodel.Response;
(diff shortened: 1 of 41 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -78,6 +81,15 @@ public ProxyMethod map(Operation operation) {
...
 
+        // method with schema in headers would require a ClientResponse
+        if (operation.getResponses().stream()
+                .filter(r -> r.getProtocol() != null && r.getProtocol().getHttp() != null && r.getProtocol().getHttp().getHeaders() != null)
+                .flatMap(r -> r.getProtocol().getHttp().getHeaders().stream().map(Header::getSchema))
+                .anyMatch(Objects::nonNull)) {
+            final String clientResponseName = operation.getOperationGroup().getLanguage().getJava().getName() + CodeNamer.toPascalCase(operation.getLanguage().getJava().getName()) + "Response";
+            returnType = GenericType.Mono(new ClassType(settings.getPackage(settings.getModelsSubpackage()), clientResponseName));
+        }
+
```
weidongxu-microsoft: This part is of same logic as ClientMapper.ParseResponse. If OK I would refactor to avoid duplicate.

jianghaolu: Feel free to make local refactorings.

jianghaolu:
I believe this resolved #418 instead of #419.

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
weidongxu-microsoft: Refactor out class name and package name.

Issue #416, opened 2019-11-22
Generate code for body-file.json


jianghaolu:

weidongxu-microsoft:
Reponse is actually a BinaryResponse, not SchemaResponse.

```
    - binary: true
      language:
        default:
          name: ''
          description: ''
      protocol:
        http:
          knownMediaType: binary
          mediaTypes:
          - image/png
          statusCodes:
          - '200'
```

jianghaolu:
Closed via #436.

PR #436, opened 2019-12-03
V4 bodyfile

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/416

Code change https://github.com/Azure/autorest.java/pull/436/commits/cce8f70b46e102936e241e689da0d761c5a7da02


extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Response.java
```diff
@@ -11,6 +11,7 @@
 
     private Schema schema;
 
+    private Boolean binary;
```
weidongxu-microsoft: We are mixing SchemaResponse and BinaryResponse here. Do we need to separate the two?

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -76,6 +76,9 @@ public ProxyMethod map(Operation operation) {
         } else {
             singleValueType = GenericType.BodyResponse(responseBodyType);
         }
+        if (operation.getResponses().stream().anyMatch(r -> Boolean.TRUE.equals(r.getBinary()))) {
+            singleValueType = ClassType.StreamResponse;
+        }
```
weidongxu-microsoft: Might not be a good solution. It will break we get mixed BinaryResponse and SchemaResponse. Any suggestion for improvement?

extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Response.java
jianghaolu: I cannot figure out polymorphic deserialization in snakeyaml. I'm fine with this.

PR #437, opened 2019-12-04
Polymorphism and update modelerfour to 4.1.60

jianghaolu:


jianghaolu:
Polymorphism isn't fully working due to https://github.com/Azure/autorest.modelerfour/issues/62

PR #448, opened 2019-12-06
Fix constants and parameter names

### LINKED ISSUE #407: Add constant support in `IType`
Today there's no constant support in IType - when we reference a constant type, we actually just use the backing type, e.g. String, int, etc, thus missing the constant value.

srnagar:
Fixes #407

### DIFF
--- a/tests/src/main/java/fixtures/header/models/HeadersResponseEnumHeaders.java
+++ b/tests/src/main/java/fixtures/header/models/HeadersResponseEnumHeaders.java
@@ -12,14 +12,14 @@ public final class HeadersResponseEnumHeaders {
      * The value property.
      */
     @JsonProperty(value = "value")
-    private GreyscaleColors value;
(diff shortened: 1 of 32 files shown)


tests/src/main/java/fixtures/header/models/HeadersResponseEnumHeaders.java
```diff
@@ -12,14 +12,14 @@
      * The value property.
      */
     @JsonProperty(value = "value")
-    private GreyscaleColors value;
+    private Enum0 value;
```
weidongxu-microsoft: This might be a autorest.core issue.

Yesterday I updated to latest autorest-beta and autorest.core, regenerate header.json with our latest java v4. And this enum is  `GreyscaleColors` (I think this should be the correct class, not Enum0), and `Enum0.java` is not generated.

Issue #442, opened 2019-12-05
[EPIC] Generate code for resources.json


weidongxu-microsoft:
Epic to group all issues & fixes encountered on generating code for ResourceManager.
`/specification/resources/resource-manager/Microsoft.Resources/stable/2019-08-01/resources.json`

Yaohai's prototype with manually edited code https://github.com/Azure/azure-libraries-for-java/tree/azure_core_prototype

Anu's unmerged poller PR https://github.com/Azure/azure-sdk-for-java/pull/6303

Notes https://github.com/weidongxu-microsoft/work-notes/blob/master/autorest-java-v4.md

weidongxu-microsoft:
At present a good portion of generated code is good to use.

Major problem to be handled:
1. Integration with mgmt lib for LRO
2. x-ms-client-flatten

weidongxu-microsoft:
Done in v4_fluentgen branch.

Leave the PR merge to https://github.com/Azure/autorest.java/issues/442

PR #460, opened 2019-12-13
v4 fluentgen

weidongxu-microsoft:
[Build Status](https://dev.azure.com/weidxu/public/_build/latest?definitionId=4&branchName=v4_fluentgen)

https://github.com/Azure/autorest.java/issues/442

Hello @jianghaolu 

This PR does not need to get merged.

It is just a showcase on current fluent code path (which is almost same as javagen at present). I've tried to fill api-version and subscriptionId missing from modelerfour, for now.

I will use `--use="./fluentgen"` to run this plugin.

FYI @yaohaizh


fluentgen/src/main/java/com/azure/autorest/fluent/FluentGen.java
```diff
@@ -0,0 +1,128 @@
...
+                    codeModel = jsonMapper.readValue(file, CodeModel.class);
+                }
+            } catch (Exception e) {
+                System.err.println("Got an error " + e.getMessage());
+                connection.sendError(1, 500, "Cannot parse input into code model: " + e.getMessage());
+                return false;
+            }
+
+            // Step 2: Transform
+            codeModel = new FluentTransformer().preTransform(codeModel);
```
weidongxu-microsoft: This is the only difference, for now.

fluentgen/src/main/java/com/azure/autorest/fluent/FluentGen.java
```diff
@@ -0,0 +1,131 @@
...
+
+            codeModel = new Transformer().transform(codeModel);
+
+            // Step 3: Map
+            new FluentMapper().preModelMap(codeModel);
```
weidongxu-microsoft: OK. One more.

weidongxu-microsoft:
CI for test https://dev.azure.com/weidxu/public/_build?definitionId=4&_a=summary
I cannot get it triggered on code commit. Probably because yaml is not on master.

It uses windows build env since my current script is bat. Should be easy to switch to linux build env.

Use `integration` profile for fluent-test.

Use a personal blob storage (on my account) to temporary host maven artifacts for management-core.
https://azuresdkmaven.blob.core.windows.net/snapshot/m2/

fluentgen/readme.md
```diff
@@ -0,0 +1,30 @@
...
+pipeline:
+
+  modelerfour:
+    flatten-models: true
+    flatten-payloads: true
+    naming:
+      preserve-uppercase-max-length: 4
+
+  fluentgen:
+    scope: java
```
jianghaolu: Should scope be `fluent`?

fluentgen/src/main/java/com/azure/autorest/fluent/FluentGen.java
```diff
@@ -0,0 +1,169 @@
...
+            for (ClientException exception : client.getExceptions()) {
+                javaPackage.addException(exception.getPackage(), exception.getName(), exception);
+            }
+
+            // Package-info
+            for (PackageInfo packageInfo : client.getPackageInfos()) {
+                javaPackage.addPackageInfo(packageInfo.getPackage(), "package-info", packageInfo);
+            }
+
+            // TODO: POM, Manager
```
jianghaolu: Pom could be handled by the vanilla gen - we can discuss. BTW I wrote an xml DSL util just like the Java one, which could be useful

fluentgen/readme.md
weidongxu-microsoft: Sure. I will change it if it works. My concern is I do not know how autorest works on the scope, and if I change it to fluent, whether it will load README.java.md, if it is fluent scope.

fluentgen/src/main/java/com/azure/autorest/fluent/FluentGen.java
weidongxu-microsoft: Sure. Fluent does not need pom at this stage. Fluent lite will need the POM but that will be months later.

fluentgen/readme.md
weidongxu-microsoft: Tried but I cannot change it to `fluent` without at least changing all README.md in swagger spec. 
Getting
`[Exception] No input files provided.`

These READMEs are all tagged with java
https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/resources/resource-manager/readme.java.md

weidongxu-microsoft:
I've refactored it to fluentnamer + fluentgen.

Test passes.

fluentgen/readme.md
jianghaolu: Ok we can use java for now. 
My concern is in AutoRest there's a default codegen lookup. For example, `--java` maps to a default npm package `autorest.java`. My understanding is that fluent will be shipped as a separate npm package. 
I think something like `--java --fluent` may be fitted for the fluent gen where `--fluent` can be used to distinguish the javagen from fluentgen while `--java` will enable autorest to pick up README.java.md.

weidongxu-microsoft:
Javagen test pass. Jianghao signed off.

fluentgen/readme.md
weidongxu-microsoft: Yes. This is a real problem.

I will follow with Garrett to see that, under same package and same scope=java, whether we can config in readme.md to have a switch --fluent to switch the pipeline. I.e. no --fluent -> preprocesser -> javagen, with --fluent -> fluentnamer -> fluentgen

PR #468, opened 2019-12-17
Filereturntype

### LINKED ISSUE #439: Streaming methods should not return void on sync & simple async methods
Currently we generate 

```java
    public Mono<StreamResponse> getFileWithResponseAsync() {
        return service.getFile(this.client.getHost());
    }

    public Mono<Void> getFileAsync() {
        return getFileWithResponseAsync()
            .flatMap((StreamResponse res) -> Mono.empty());
    }

    public void getFile() {
        getFileAsync().block();
    }

```

jianghaolu:
fixes #439

### DIFF
--- a/tests/src/main/java/fixtures/bodyboolean/models/Error.java
+++ b/tests/src/main/java/fixtures/bodyboolean/models/Error.java
@@ -9,19 +9,19 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 @Fluent
 public final class Error {
     /*
-     * The status property.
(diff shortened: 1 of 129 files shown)


tests/src/main/java/fixtures/bodyboolean/models/Error.java
```diff
@@ -9,19 +9,19 @@
 @Fluent
 public final class Error {
     /*
-     * The status property.
+     * MISSING·SCHEMA-DESCRIPTION-INTEGER
```
jianghaolu: Should not be a problem in real world swagger specs

PR #469, opened 2019-12-18
v4 plain object + abstract factory for Mappers and Templates

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/442

Certain ObjectSchema with no property, no parent, no child seems better to write as Object (as done in v1) instead of named class.

```
              - schema: &ref_33
                  type: object
                  apiVersions:
                  - version: '2019-08-01'
                  language:
                    default:
                      name: WhatIfChange-before
                      description: The snapshot of the resource before the deployment is executed.
                      namespace: Api20190801
                  protocol: {}
                required: false
                serializedName: before
                language:
                  default:
                    name: before
                    description: The snapshot of the resource before the deployment is executed.
                protocol: {}
```

Fluent would do something like this to populate Inner.
https://github.com/weidongxu-microsoft/autorest.java/blob/v4_fluentgen/fluentgen/src/main/java/com/azure/autorest/fluent/mapper/FluentMapper.java#L35
which calls `ObjectMapper.addInnerModels`.
At present the implemetation in ObjectMapper is kind of ad-hoc. Let me know if you have plan to do this consistently, e.g. as inherit (like old CompositeTypeJvaf.IsInnerModel), or some other way.


javagen/src/main/java/com/azure/autorest/mapper/ObjectMapper.java
```diff
@@ -42,23 +42,33 @@ public IType map(ObjectSchema compositeType) {
...
         }
 
         parsed.put(compositeType, result);
         return result;
     }
+
+    public static boolean isPlainObject(ObjectSchema compositeType) {
+        return compositeType.getProperties().isEmpty() && compositeType.getDiscriminator() == null
+                && compositeType.getParents() == null && compositeType.getChildren() == null
+                && (compositeType.getExtensions() == null || compositeType.getExtensions().getXmsEnum() == null);
```
weidongxu-microsoft: I am not sure whether this is complete or not.

No parent, no child, no property, no polymorphy, no enum (not sure this applies not not. seen type: object with x-ms-enum in swagger).

weidongxu-microsoft:
PS, not related to this PR. I tried to re-generate tests with 3.0.6149 (core 3.0.6158) but there seems to be multiple failures. It persists without code from this PR.

It seems to be problem with parameter on client. Still checking what went wrong.

https://github.com/weidongxu-microsoft/autorest.java/compare/v4..v4_fix_tests

javagen/src/main/java/com/azure/autorest/mapper/ObjectMapper.java
```diff
@@ -42,23 +47,59 @@ public IType map(ObjectSchema compositeType) {
...
+        return !(ClassType.Resource.equals(modelType)
+                || ClassType.ProxyResource.equals(modelType)
+                || ClassType.SubResource.equals(modelType));
+    }
+
+    public void addInnerModels(Collection<ObjectSchema> compositeTypes) {
+        innerModelNames.addAll(compositeTypes.stream()
+                .map(t -> t.getLanguage().getDefault().getName())
+                .collect(Collectors.toSet()));
+    }
```
weidongxu-microsoft: Fluent calls this to populate Type which need to be Inner.

javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -29,13 +30,13 @@ public static ModelMapper getInstance() {
...
-        if (result == null) {
-            String modelSubPackage = !settings.isFluent() ? settings.getModelsSubpackage() : (false /*compositeType.IsInnerModel*/ ? settings.getImplementationSubpackage() : "");
-            if (settings.isCustomType(compositeType.getLanguage().getJava().getName())) {
-                modelSubPackage = settings.getCustomTypesSubpackage();
-            }
-            String modelPackage = settings.getPackage(modelSubPackage);
+        ObjectMapper objectMapper = Mappers.getObjectMapper();
+
+        ClassType modelType = objectMapper.map(compositeType);
+        String modelName = modelType.getName();
```
weidongxu-microsoft: I am trying to make result from ObjectMapper as the ground truth for ObjectSchema, at least on class name + package name.

jianghaolu:
I'm able to generate with autorest core 3.0.6169

weidongxu-microsoft:
Oh yes, 3.0.6169 is good.

weidongxu-microsoft:
I've switched to use factory, so I can inherit and customize mapper / template in fluentgen.

weidongxu-microsoft:
Also fix ObjectSchema import for parent (when they are in different packages).

weidongxu-microsoft:
@jianghaolu 
Let me know whether you are OK with this PR.

I've clustered multiple things here.
1. use abstract factory so I can plugin different Mapper / Template from fluentgen
2. for ObjectSchema without any parent/children/properties, use `ClassType.Object` instead of an empty class. (Let me know if data-plane does not want this)
3. augment ObjectMapper a bit so that I can plugin a subclass to handle Inner for fluent
4. for fluent, do not generate Resource/SubResource class.
5. fix for import of parent type for ObjectSchema (when they are in different packages).

If you want I can move certain things out.

javagen/src/main/java/com/azure/autorest/mapper/ObjectMapper.java
weidongxu-microsoft: Swagger https://github.com/Azure/azure-rest-api-specs/blob/main/specification/resources/resource-manager/Microsoft.Resources/stable/2019-08-01/resources.json#L5352-L5355

We may able to remove this in 2024

Issue #477, opened 2020-01-09
Suppoert additionalProperties


weidongxu-microsoft:
```
    "DirectoryObject": {
      "type": "object",
      "discriminator": "objectType",
      "properties": {
        "objectId": {
          "type": "string",
          "description": "The object ID.",
          "readOnly": true
        },
        "objectType": {
          "type": "string",
          "description": "The object type.",
          "readOnly": true
        },
        "deletionTimestamp": {
          "type": "string",
          "format": "date-time",
          "description": "The time at which the directory object was deleted.",
          "readOnly": true
        }
      },
      "additionalProperties": {
        "type": "object"
      },
      "description": "Represents an Azure Active Directory object."
    },
```

In YAML, DirectoryObject would have immediate parent of schema type "dictionary".

```
        parents:
          all:
          - *ref_0
          - &ref_2
            type: dictionary
            elementType:
              type: any
              language:
                default:
                  name: any
                  description: <Any object>
              protocol: {}
            language:
              default:
                name: DirectoryObject
                description: Represents an Azure Active Directory object.
            protocol: {}
          immediate:
          - *ref_0
```

PR #478, opened 2020-01-09
support additionalProperties in ObjectSchema

weidongxu-microsoft:
FIx https://github.com/Azure/autorest.java/issues/477

Currently I did it in Mapper. Second thought that it might better be done in Transform (remove the parent and add a property with that schema). @jianghaolu Let me know which stage you prefer, and I will update the PR.


jianghaolu:
Preferably everything should be in the mapper.

Issue #483, opened 2020-01-15
Support ConstantSchema in model property


weidongxu-microsoft:

weidongxu-microsoft:
@jianghaolu 
Do we have any swagger test on constant model property?

I've attempted a fix https://github.com/Azure/autorest.java/pull/484, but it is not complete and even not correct.
v1 seems have much more complex logic for them.

PR #484, opened 2020-01-15
support string constant

weidongxu-microsoft:
Partly fix https://github.com/Azure/autorest.java/issues/483

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
@@ -1,8 +1,10 @@
 package com.azure.autorest.mapper;
 
 import com.azure.autorest.extension.base.model.codemodel.ArraySchema;
+import com.azure.autorest.extension.base.model.codemodel.ConstantSchema;
 import com.azure.autorest.extension.base.model.codemodel.Property;
 import com.azure.autorest.extension.base.model.codemodel.Schema;
+import com.azure.autorest.extension.base.model.codemodel.StringSchema;
 import com.azure.autorest.extension.base.model.codemodel.XmlSerlializationFormat;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.clientmodel.ClientModelProperty;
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft:
Sample for generated code on constant string.
https://github.com/Azure/azure-libraries-for-java/commit/d08e08ff8140844ac0ea1d2db5aa9b0e095f4d95

If a property is constant, constructor would init it as e.g. `action = "Allow";` But it's class would not be generated (say enum Action which only 1 string).

javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
```diff
@@ -87,6 +89,16 @@ public ClientModelProperty map(Property property) {
             }
         }
 
+        boolean isConstant = false;
+        String defaultValue = null;
+        if (property.getSchema() instanceof ConstantSchema) {
+            ConstantSchema constantSchema = (ConstantSchema) property.getSchema();
+            if (constantSchema.getValueType() instanceof StringSchema) {
+                isConstant = true;
```
srnagar: Shouldn't `isContant` be true even is the `valueType` is not of type `StringSchema`? `isConstant` should be set to true if `property.getSchema() instanceof ConstantSchema` is true.

weidongxu-microsoft: Yes, it should be. The defaultValue would also be different in this non-string case, and also that ConstantSchema would probably get a Class of its own.

The problem here is that currently I do not have any case that have a constant which is not a string, Hence I only do for string (mostly to unblock our team converting on management-plane SDK to v2).

That is reason I was asking 
> Do we have any swagger test on constant model property?

in https://github.com/Azure/autorest.java/issues/483

If you have any case (test, or real data-plane or management-plane swagger that have a non-string constant, please do let me know. Thanks.

srnagar: Well, I don't know of any swagger that has non-string constants but it should be fairly easy to create one. Take an existing swagger and replace one of the string constants with integer constants and update the schema for it. 

Also, you can go ahead and merge this change as this can be tested and updated in a separate PR for non-string constants.

weidongxu-microsoft: Got it. I will try to test non-string constant a bit later (after holiday), maybe try stealing some test case from Garrett's repo :-).

Currently still busying converting fluent java to azure core.

PR #487, opened 2020-01-17
Bug fixes

srnagar:
When generating code for KeyVault swagger, I found a few issues that needed fixing. Also, I was using an older version of autorest core to generate code. This PR also regenerates all types with newer version.


jianghaolu:
@srnagar Maybe document some changes here?

PR #488, opened 2020-01-21
Regenerate with model flattening

jianghaolu:


tests/src/main/java/fixtures/bodycomplex/models/GoblinSharkColor.java
```diff
@@ -26,7 +26,7 @@
     /**
      * Static value RED for GoblinSharkColor.
      */
-    //public static final GoblinSharkColor RED = fromString("RED");
+//    public static final GoblinSharkColor RED = fromString("RED");
```
srnagar: Both `RED` and `red` enums are now commented due to merging order of my [changes](https://github.com/Azure/autorest.java/commit/40cefe2675927b3191f13c131d5aa3b5a1af3156#diff-49b1b1ae6d90a899f8a4f6940b106feeL34) and @weidongxu-microsoft 's [changes](https://github.com/Azure/autorest.java/commit/0800fc5010ef1c4f03ce103864575790a00651da#diff-49b1b1ae6d90a899f8a4f6940b106feeL29). One of the enums should be uncommented.

weidongxu-microsoft:
There might be some more to do with flattening.

I think modelerfour have flattenedNames as something like ['property', 'flattenedProperty'] (may be more if there is more than one nested flatten) and we need to concat it to say `property.flattenedProperty`.

Also for payload flatten where would be some `List<ParameterMapping> parameterMappings` to populate.

weidongxu-microsoft:
MyBaseType.java likely not generated correctly. It got @JsonFlatten so modelerfour have flattenedNames, but class does not have any `properties.propB1`

jianghaolu:
payload flattening will come later.

jianghaolu:
There're some bad encodings in my code model for url.json but other languages are not seeing it so likely an issue for my machine.

javagen/src/main/java/com/azure/autorest/util/CodeNamer.java
```diff
@@ -297,6 +297,8 @@ private static String removeInvalidCharacters(String name, char... allowerCharac
         for (Character c : name.toCharArray()) {
             if (Character.isLetterOrDigit(c) || allowed.contains(c)) {
                 builder.append(c);
+            } else {
+                builder.append("_");
```
weidongxu-microsoft: This one seems causing a bit of problem.

Previously genereated enum is such
` SYSTEM_ASSIGNED_USER_ASSIGNED("SystemAssigned, UserAssigned")`

Now it is such
`SYSTEM_ASSIGNED__USER_ASSIGNED("SystemAssigned, UserAssigned")` , one more `_`
Any idea what best to circumvent this?

PR #490, opened 2020-01-27
Naive payload flattening support

jianghaolu:


jianghaolu:
Pull latest from autorest.testserver - a few changes there like date time precision is changed to 3 digits.

PR #491, opened 2020-01-28
fix multiple flattened parameter with same origin

weidongxu-microsoft:
1. multiple flattened parameter with same origin need to be put in same detail
2. naming in fluent (should be no need to do this in template)


weidongxu-microsoft:
regen code and tests is good. Though I will not update test since there seems be similar unicode issue in source code on my laptop.

PR #492, opened 2020-01-28
method group support additional interfaces

weidongxu-microsoft:
Support additional interface like InnerSupportsGet/InnerSupportsDelete/InnerSupportsListing for Fluent.

Implementation is in fluentgen.


jianghaolu:
This is a good example of an "extension point" we can support in the javagen for fluent features.

PR #494, opened 2020-01-28
Builder prototype for client models

jianghaolu:
This is a just a prototype of builders in client models with only ProxyMethod and ClientMethod converted. The benefits include:

- Allow populating properties step-by-step
- Reduce duplicate code
- Avoid breaking changes in client models
- Avoid passing in constructor parameters in the wrong order since many parameters are of type String

Improvements in ProxyMethodMapper are more obvious due to its 1-to-1 mapping, same as almost all client models. ClientMethod is the exception, where there's a 1-to-many mapping in ClientMethodMapper. Thus I'm re-using the builder here but I'm happy to discuss this design.


javagen/src/main/java/com/azure/autorest/mapper/ObjectMapper.java
```diff
@@ -29,38 +29,26 @@ public ClassType map(ObjectSchema compositeType) {
...
-                String classPackage;
-                String className = compositeType.getLanguage().getJava().getName();
-                if (settings.isCustomType(compositeType.getLanguage().getJava().getName())) {
-                    classPackage = settings.getPackage(settings.getCustomTypesSubpackage());
-                } else if (!settings.isFluent()) {
-                    classPackage = settings.getPackage(settings.getModelsSubpackage());
-                } else if (isInnerModel) {
-                    className += "Inner";
-                    classPackage = settings.getPackage(settings.getImplementationSubpackage());
-                }
```
weidongxu-microsoft: Where can I do this (Inner and package) in refactored code?

javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -39,10 +38,13 @@ public ClientModel map(ObjectSchema compositeType) {
         ClassType modelType = objectMapper.map(compositeType);
         String modelName = modelType.getName();
         ClientModel result = serviceModels.getModel(modelType.getName());
-        if (result == null && !ObjectMapper.isPlainObject(compositeType) && (!settings.isFluent() || ObjectMapper.nonResourceType(modelType))) {
```
weidongxu-microsoft: Eh, where do I do this after refactor?

javagen/src/main/java/com/azure/autorest/template/ServiceClientTemplate.java
```diff
@@ -129,53 +129,22 @@ public final void write(ServiceClient serviceClient, JavaFile javaFile) {
...
-                        } else if (constructor.getParameters().equals(Arrays.asList(serviceClient.getAzureEnvironmentParameter()))) {
-                            constructorBlock.line(String.format("this(%1$s.createDefaultPipeline(%2$s.class), %3$s);", ClassType.AzureProxy.getName(), serviceClient.getClassName(), serviceClient.getAzureEnvironmentParameter().getName()));
-                        } else if (constructor.getParameters().equals(Arrays.asList(serviceClient.getHttpPipelineParameter()))) {
-                            constructorBlock.line(String.format("this(%1$s, %2$s);", serviceClient.getHttpPipelineParameter().getName(), serviceClient.getAzureEnvironmentParameter().getDefaultValue()));
-                        } else if (constructor.getParameters().equals(Arrays.asList(serviceClient.getHttpPipelineParameter(), serviceClient.getAzureEnvironmentParameter()))) {
-                            constructorBlock.line(String.format("super(%1$s, %2$s);", serviceClient.getHttpPipelineParameter().getName(), serviceClient.getAzureEnvironmentParameter().getName()));
```
weidongxu-microsoft: I am not sure but Fluent would actually use some of them (not remembering which).

weidongxu-microsoft:
As commented, current refactor on ObjectMapper (deleting multiple methods not used in javagen but serving as extension point for fluentgen) would break compilation on fluentgen.
And other change will not break compilation but might break generated code.

Since I have some integration test in fluentgen, I should be able to recover from it and make a patch PR. But it will certainly take time.

My suggestion is after this batch, I will try go on and merge fluentgen module (https://github.com/Azure/autorest.java/pull/460), and we should try to make all available tests running in CI (or at least easily run by developer). Then we could continue further refactoring, being assured that at least there being no major break.

jianghaolu:
Sounds good. Let's park this PR until things are sorted out between javagen and fluentgen.

weidongxu-microsoft:
@jianghaolu 

Currently this is still OK, I can recover from it. Just notifying I might not able to handle further divergence.

I've added a CI for fluent integration test (install nodejs, install autorest, run with profile `integration` to get script to generate code for some selected management-plan json, verify compile pass, verify method signature in tests).
I'm not able to get it automatically triggered, probably because yaml is not on master. Manual trigger should be OK.

It is on this PR https://github.com/Azure/autorest.java/pull/460
When you have time you can take a look and let me know your opinion.

This PR got one difference affect javagen. I saw you commented out `removeFlattenedObjectSchemas` step. Let me know the reason so I can fix (is it because some flatten object is still being used by some other objects?).

PR #495, opened 2020-01-30
Use latest modeler and fix some flattening issues

jianghaolu:


extension-base/src/main/java/com/azure/autorest/extension/base/jsonrpc/Connection.java
```diff
@@ -163,7 +163,7 @@ private JsonNode readJson(int contentLength)
     {
         try {
             byte[] bytes = reader.readBytes(contentLength);
-            return mapper.readTree(bytes);
+            return mapper.readTree(new String(bytes, StandardCharsets.UTF_8));
```
jianghaolu: This should help - but doesn't fix the encoding issue completely

PR #502, opened 2020-02-07
v4 update modelerfour to 4.6.199

weidongxu-microsoft:
There is quite some naming change in 4.5.176. I am checking with Garrett to see whether they are final.

For now this PR serves as a head-up on the incompatibilities caused by naming change.


javagen/src/main/java/com/azure/autorest/mapper/ClientParameterMapper.java
```diff
@@ -45,11 +46,14 @@ public ClientMethodParameter map(Parameter parameter) {
         if (parameter.getSchema() != null && parameter.getSchema().getLanguage() != null) {
             parameterDescription = parameter.getSchema().getLanguage().getDefault().getDescription();
         }
+        String name = parameter.getOriginalParameter() != null
+                ? CodeNamer.toCamelCase(parameter.getOriginalParameter().getSchema().getLanguage().getJava().getName()) + CodeNamer.toPascalCase(parameter.getLanguage().getJava().getName())
```
weidongxu-microsoft: Modelerfour upgrade causing the flattened parameter name taking same name as the original parameter name. Had to change it a bit.

weidongxu-microsoft:
There seems another change on unnamed additional properties type. Previously it would get a name like `components·schemas·XXX·additionalproperties`, now it is merely empty string. Need to have transformer code to fill it.

weidongxu-microsoft:
Almost all good on 4.6.199. Only `JSON` is changed and this is a bit ambiguous anyway.

jianghaolu:
Merging for now and we can discuss these names more in the future.

PR #503, opened 2020-02-10
v4 generialize for constant in property

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/483


weidongxu-microsoft:
modelerfour only support wire type = string, boolean, float, integer.

PR #505, opened 2020-02-11
v4 support client method with only required parameters

weidongxu-microsoft:
Overload client method with only required parameters.

I choose not to overload XXXWithResponseAsync method. This overload is not useful for fluent. Let me know if you prefer to overload this as well.

Please take a look whether it is OK for you.

If it is OK, I will need to update test cases, since some parameter sequence got updated.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
@@ -134,14 +134,15 @@ public class ClientMethod {
      * Get the comma-separated list of parameter declarations for this ClientMethod.
      */
     public final String getParametersDeclaration() {
-        return getParameters().stream().map(ClientMethodParameter::getDeclaration).collect(Collectors.joining(", "));
+        List<ClientMethodParameter> methodParameters = onlyRequiredParameters ? getMethodRequiredParameters() : getMethodParameters();
+        return methodParameters.stream().map(ClientMethodParameter::getDeclaration).collect(Collectors.joining(", "));
     }
 
     /**
      * Get the comma-separated list of parameter names for this ClientMethod.
      */
     public final String getArgumentList() {
-        return getParameters().stream().map(ClientMethodParameter::getName).collect(Collectors.joining(", "));
+        return getMethodParameters().stream().map(ClientMethodParameter::getName).collect(Collectors.joining(", "));
     }
 
     /**
@@ -161,7 +162,7 @@ public class ClientMethod {
 
     public final List<ClientMethodParameter> getMethodParameters() {
         return getParameters().stream().filter(parameter -> parameter != null && !parameter.getFromClient() &&
-                parameter.getName() != null && parameter.getName().trim().isEmpty())
+                parameter.getName() != null && !parameter.getName().trim().isEmpty())
                 .sorted((p1, p2) -> Boolean.compare(!p1.getIsRequired(), !p2.getIsRequired()))
                 .collect(Collectors.toList());
     }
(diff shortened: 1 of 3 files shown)


weidongxu-microsoft:
Sample generated code

```
    /**
     * Lists all containers and does not support a prefix like data plane. Also SRP today does not return continuation token.
     * 
     * @param resourceGroupName 
     * @param accountName 
     * @param maxpagesize 
     * @param filter 
     * @throws CloudException thrown if the request is rejected by server.
     */
    public PagedFlux<ListContainerItemInner> listAsync(String resourceGroupName, String accountName, String maxpagesize, String filter) {
        return new PagedFlux<>(
            () -> listSinglePageAsync(resourceGroupName, accountName, maxpagesize, filter),
            nextLink -> listNextSinglePageAsync(nextLink));
    }

    /**
     * Lists all containers and does not support a prefix like data plane. Also SRP today does not return continuation token.
     * 
     * @param resourceGroupName 
     * @param accountName 
     * @throws CloudException thrown if the request is rejected by server.
     */
    public PagedFlux<ListContainerItemInner> listAsync(String resourceGroupName, String accountName) {
        final String maxpagesize = null;
        final String filter = null;
        return new PagedFlux<>(
            () -> listSinglePageAsync(resourceGroupName, accountName, maxpagesize, filter),
            nextLink -> listNextSinglePageAsync(nextLink));
    }

    /**
     * Lists all containers and does not support a prefix like data plane. Also SRP today does not return continuation token.
     * 
     * @param resourceGroupName 
     * @param accountName 
     * @param maxpagesize 
     * @param filter 
     * @throws CloudException thrown if the request is rejected by server.
     */
    public PagedIterable<ListContainerItemInner> list(String resourceGroupName, String accountName, String maxpagesize, String filter) {
        return new PagedIterable<>(listAsync(resourceGroupName, accountName, maxpagesize, filter));
    }

    /**
     * Lists all containers and does not support a prefix like data plane. Also SRP today does not return continuation token.
     * 
     * @param resourceGroupName 
     * @param accountName 
     * @throws CloudException thrown if the request is rejected by server.
     */
    public PagedIterable<ListContainerItemInner> list(String resourceGroupName, String accountName) {
        final String maxpagesize = null;
        final String filter = null;
        return new PagedIterable<>(listAsync(resourceGroupName, accountName, maxpagesize, filter));
    }
```

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -134,14 +134,15 @@ public final String getClientReference() {
...
+        return methodParameters.stream().map(ClientMethodParameter::getDeclaration).collect(Collectors.joining(", "));
     }
 
     /**
      * Get the comma-separated list of parameter names for this ClientMethod.
      */
     public final String getArgumentList() {
-        return getParameters().stream().map(ClientMethodParameter::getName).collect(Collectors.joining(", "));
+        return getMethodParameters().stream().map(ClientMethodParameter::getName).collect(Collectors.joining(", "));
     }
```
weidongxu-microsoft: I've change this code to enable ordering of parameters (required parameters before non-required parameters).

weidongxu-microsoft:
Preferably I should generate this (directly call the async/sync method with full parameters, instead of calling `listSinglePageAsync`).

I will find time later to update it (probably need 2 new type of client methods).

```
    public PagedFlux<ListContainerItemInner> listAsync(String resourceGroupName, String accountName) {
        final String maxpagesize = null;
        final String filter = null;
        return listAsync(resourceGroupName, accountName, maxpagesize, filter);
    }
```

jianghaolu:
> I choose not to overload XXXWithResponseAsync method. This overload is not useful for fluent. Let me know if you prefer to overload this as well.

Agreed.

PR #508, opened 2020-02-13
Add support for application/x-www-form-urlencoded

MariusVolkhart:
Previously, the generator was treating this MIME type the same as the multipart/form-data MIME type. In fact, these are two unique ways of submitting data to a server, and Retrofit has support for both of them.


jianghaolu:
Thanks!

PR #511, opened 2020-02-18
Add fluentnamer plugin

srnagar:
This PR includes changes to add a new plugin for fluent namer. The plugin sits between `preprocessor` and `javagen` plugin and when activated will use the code model yaml file generated from `preprocessor` to make necessary changes to the names of types and then output another code model yaml file which will then be consumed by `javagen` plugin.


weidongxu-microsoft:
BTW, @jianghaolu . Any advice on the fluentgen PR itself? https://github.com/Azure/autorest.java/pull/460

srnagar:
@weidongxu-microsoft initially, my thoughts were similar to what you mentioned about having 2 parallel pipelines:
javanamer -> javagen
fluentnamer -> fluentgen

However, after talking to @jianghaolu, it made sense to have just one plugin consume the code model from modelerfour and do the initial transformation. This parsing of input code model is common to both javanamer and fluentnamer and the transformation is going to be mostly common and only a few changes might be required in fluentnamer (I am assuming here, correct me if that's not the case). So, instead of duplicating the same logic in both plugins, this new approach was used.

jianghaolu:
The reason to put fluentnamer after javanamer is so that we can avoid duplicating the code for common renamings. And I suppose any new classes added after javanamer can just be named correctly by itself.

What are some specific scenarios where having fluentnamer after javanamer will make it more difficult for you?

weidongxu-microsoft:
@srnagar @jianghaolu 
There is not really concern on duplicate code. Basically the fluentnamer would just depends on preprocessor module and call the same `Transformer.transform` function. A few lines, done. (there might be some issue when package to single jar, about duplicate resource or such, but seems not too bad)

Well, even if preprocessor and fluentnamer is concatenated, fluentnamer would still depends on preprocessor module. Anyway I will need all the common rename methods. I will certainly not add a CodeNamer by myself.

BTW, about duplicate code, why do we have 2 CodeNamer java file (one in preprocessor and one in javagen), which is almost same?

That said, currently there should be no blocking issue for me to move transform after the naming stage (e.g. add apiVersion parameter to all operations, add some beginFoo operation for LRO operation), etc. I can manage it.

You can go ahead and merge. I will try to adapt if possible.

jianghaolu:
I'm okay with Weidong's proposal. It simplifies our flow too since javagen will no longer require output from `fluentnamer` (which was a bit confusing to me) @srnagar what do you think?

srnagar:
I am okay with the proposal too. I have updated the PR based on the discussion. So, the pipeline will be:

Data: modelerfour/identity -> preprocessor (may change name) -> javagen
Fluent:  modelerfour/identity -> fluentnamer -> fluentgen

PR #517, opened 2020-02-24
v4 fix method name escaping, need camelCase before escape

weidongxu-microsoft:
M4 -> "Import"
camelCase -> "import"
escape -> "importMethod"


weidongxu-microsoft:
Are we able to do something about the 2 CodeNamer? Maybe have a common module that everybody other module could share?

weidongxu-microsoft:
https://github.com/Azure/autorest.java/pull/517 is intended to mitigate following:

In transformer, setOperation to parameter does not work. E.g. "host" parameter is single instance, and all operation is referring to this single instance. So later setOperation will just override a former one.
https://github.com/Azure/autorest.java/blob/v4/preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java#L67

In this commit I just try to move the setOperation nearer to the location that this operation reference is used. This is not a fix on root.

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -126,6 +126,7 @@ public ProxyMethod map(Operation operation) {
         List<ProxyMethodParameter> parameters = new ArrayList<>();
         for (Parameter parameter : operation.getRequest().getParameters().stream()
                 .filter(p -> p.getProtocol() != null && p.getProtocol().getHttp() != null).collect(Collectors.toList())) {
+            parameter.setOperation(operation);
```
jianghaolu: This should be handled in the preprocessor.

jianghaolu: I read your comments. Scratch my comment above. I'm okay with this change for now but this does expose a defection in the design.

weidongxu-microsoft: Yes, it is more or less design issue.

Either it can be copied when changed (I've got a shallowCopy util in fluentgen to do this https://github.com/weidongxu-microsoft/autorest.java/blob/v4_fluentgen/fluentgen/src/main/java/com/azure/autorest/fluent/util/Utils.java#L34), or need other refactor elsewhere.

Currently I prefer not to do bigger change (this will just invite large merge conflict) until I've merged javagen and optionally fluentgen and your pending big refactor.

PR #518, opened 2020-02-27
v4 proxy method, add 'Content-Type' and 'Accept' headers

weidongxu-microsoft:
Add 'Content-Type' Headers for media type. Some service returns 415 if 'Content-Type' not provided, even that POST got empty body...

Add 'Accept' Headers for response media types. Some service returns XML if not specified.

Let me know if you do not want it in data-plane.


weidongxu-microsoft:
Not sure whether I should add 'Accept' as well. Got one endpoint return XML instead of JSON as response.

weidongxu-microsoft:
FIx https://github.com/Azure/autorest.java/issues/521

weidongxu-microsoft:
The import of "Headers" from core conflicts with Headers.java in test.

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -137,6 +138,15 @@ public ProxyMethod map(Operation operation) {
             .findFirst()
             .orElse(null);
 
+        Set<String> responseContentTypes = operation.getResponses().stream()
+                .filter(r -> r.getProtocol() != null && r.getProtocol().getHttp() != null && r.getProtocol().getHttp().getMediaTypes() != null)
+                .flatMap(r -> r.getProtocol().getHttp().getMediaTypes().stream())
+                .filter(s -> !s.isEmpty())
+                .collect(Collectors.toSet());
+        if (!responseContentTypes.contains("application/json")) {
+            responseContentTypes.add("application/json;q=0.9");
```
jianghaolu: What is q=0.9 for?

jianghaolu:
This might have some overlap with #523?

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
weidongxu-microsoft: This is for operation that does not specify "application/json" for media type in response (for protection that some swagger missing "application/json"). It add a default "application/json" with less importance, so that if e.g. service want to respond with application/xml specified in swagger, it got default q=1 so it will go, if e.g. service forget to specify "application/json" in swagger, it still will have this as an option.

weidongxu-microsoft:
@jianghaolu

Yes. I will let 523 goes in first, then add additional check whether these Headers would conflict with HeaderParam. In such case I will remove part of it.

Currently no harm is done to javagen, since the implementation is wholly in fluentgen (for now).

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
weidongxu-microsoft: If you prefer, I could move this to template part to handle.

jianghaolu: I think mapper should be the place for setting default values.

weidongxu-microsoft:
I've solved the potential conflict with HeaderParam in fluent.
https://github.com/Azure/autorest.java/pull/460/files#diff-2dea3bae3000a02cadd3acb541eb7557R40

PR #519, opened 2020-02-28
Support BinarySchema and minor bug fixes for paging and xml

srnagar:
While generating code for storage data-plane swagger, I ran into a few issues. This PR is to fix all those bugs that include adding BinarySchema support, disable paging if swagger doesn't have proper paging info and add yaml mapping for few missing node types.


weidongxu-microsoft:
Hit by the same BinarySchema yesterday. Nice job.

Issue #522, opened 2020-03-02
Proxy methods should include context param


srnagar:
Context should be the last parameter in all proxy method signatures.

jianghaolu:
when `--add-context-parameter` is provided on the commandline: https://github.com/Azure/autorest.java/blob/v4/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java#L91

PR #524, opened 2020-03-04
v4 support azure-core context

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/522


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -109,10 +109,14 @@ private static void ApplyParameterTransformations(JavaBlock function, ClientMeth
                         return parameterName + " != null";
                     }).collect(Collectors.joining(" || "));
             boolean conditionalAssignment = nullCheck != null && !nullCheck.isEmpty() && !transformation.getOutParameter().getIsRequired() && !clientMethod.getOnlyRequiredParameters();
+            // Use a mutable internal variable, leave the original name for effectively final variable
```
weidongxu-microsoft: Use a `fooInternal` to do conditional assignment, leave `foo` as effectively final, for lambda.

Hope it will not get into future naming conflict.

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -130,6 +131,27 @@ public ProxyMethod map(Operation operation) {
...
+            ProxyMethodParameter contextParameter = new ProxyMethodParameter(
+                "The context to associate with this operation.",
+                    ClassType.Context,
+                    ClassType.Context,
+                    "context",
+                    RequestParameterLocation.None,
+                    "context",
+                    true,
+                    false,
+                    true,
```
jianghaolu: Why is it required?

weidongxu-microsoft: I think I've copied it from v3.

Currently the flags is not used, since client code will use FluxUtil.withContext to let user provide context via reactor subscriberContext (though it probably not working nice for PagedFlux).

If we instead decide to let client provide 2 sets of client method (one with context, one without), then we would need isRequred=false here.

tests/src/main/java/fixtures/bodyboolean/Bools.java
```diff
@@ -90,7 +92,7 @@ public Bools(AutoRestBoolTestService client) {
         if (this.client.getHost() == null) {
             throw new IllegalArgumentException("Parameter this.client.getHost() is required and cannot be null.");
         }
-        return service.getTrue(this.client.getHost());
+        return FluxUtil.withContext(context -> service.getTrue(this.client.getHost(), context));
```
jianghaolu: The way we currently generate in v3 is by adding the context on the method signature: https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/BlobsImpl.java#L273

weidongxu-microsoft: Do you prefer we do this on v4 as well (add Context into client method)?

jianghaolu: @srnagar may know the correct design for this

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
jianghaolu: Just FYI client method parameters are ordered so that required params are asked first, and optional params follow.

weidongxu-microsoft: I see. I've changed it to isRequired=false

tests/src/main/java/fixtures/bodyboolean/Bools.java
weidongxu-microsoft: @srnagar 

Here is sample generated code for simple REST.

Here is sample generated code for paging (withContext is called inside SinglePageAsync and NextSinglePageAsync method).
https://github.com/Azure/autorest.java/blob/3639f7f3d5723e1533de2897f3a58c83a3936d72/tests/src/main/java/fixtures/paging/Pagings.java#L344-L386

Let me know if this is OK for you.

jianghaolu: I see this pattern is used across the existing SDKs. Not sure why they put it in the FluxUtil but fine.

weidongxu-microsoft:
@srnagar 
I will merge it into v4. Not sure whether this helps the context in data-plane. If you need something more, feel free to update. In ClientMethodTemplate, the handling for javagen and fluentgen is already separated.

Issue #527, opened 2020-03-06
v4 handle multiple inheritance


weidongxu-microsoft:
Keep first, flatten rest.

weidongxu-microsoft:
Currently handled by flatten. Better approach is by composition.

PR #528, opened 2020-03-06
v4 fluent flatten mutiple inheritance

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/527


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -46,31 +47,46 @@ public ClientModel map(ObjectSchema compositeType) {
...
+
+                ObjectSchema firstParentComplexSchema = null;
+                for (Schema parent : compositeType.getParents().getImmediate()) {
+                    if (parent instanceof ObjectSchema) {
+                        if (firstParentComplexSchema == null) {
+                            firstParentComplexSchema = (ObjectSchema) parent;
+                        } else {
+                            parentsNeedFlatten.add((ObjectSchema) parent);
+                        }
                     }
```
weidongxu-microsoft: Some mgmt-plane swagger has type with multiple base type. Currently I only take the first one as base, and flatten the rest. Could there be a better heuristic, like take the one that itself inherits a base class or got a discriminator?

v1 flatten it as well. But not sure how it pick the base class.

jianghaolu: I agree the right way to handle multiple inheritance in Java is through composition. In v1 flattening was done by modeler so we didn't have to implement that.
While we can reach out to Garrett for advice, I think keeping the same behavior as v1 should be okay.

weidongxu-microsoft: Yes, agree. I will try composition, see if I can pull off the serialization/de-serialization correctly.

weidongxu-microsoft: JsonUnwrapper should do (however even when I do not add this, the serialization/de-serialization of the properties under composition still works...).

However the complete task would affect quite much to ClientModel and ModelTemplate, since the currently ModelTemplate and ClientModelProperty appears not able to handle the properties from composition.
I will delay the implementation of composition to a future task.

PR #532, opened 2020-03-11
[draft] use fluent in readme to pick pipeline

weidongxu-microsoft:
@jianghaolu 

We are able to use `$(fluent)` in readme to pick different yaml and hence different pipeline.

However I am not able to start 2 java plugin at same time. (I don't think `--use:./preprocessor --use:./javagen` would be the real use case in published npm pakcage, unless we publish all these separately)

Idealy we need to start preprocessor and javagen (well, plus fluentnamer and fluentgen) in package.json start script. So pipeline can pickup whatever plugin need to finish the flow.

But due to stdin/stdout problem, I am not able to use `concurrently` to do this. Any advice?

Well, if we indeed can use for 4 `--use` to load 4 plugins, we can do same `$(fluent)` in plugin readme to control pipeline.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,7 +71,7 @@ AutoRest needs the below config to pick this up as a plug-in - see https://githu
 
 #### Javagen
 
-``` yaml
+``` yaml !$(fluent)
 use-extension:
   "@autorest/modelerfour": "4.8.221"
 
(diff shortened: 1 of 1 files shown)


jianghaolu:
Sorry just saw your question - let me play with it

PR #538, opened 2020-03-13
v4 fluent still need client builder at same package as client

weidongxu-microsoft:
Since method in client is package scope, client and builder must be in same package, else `setHost` will not resolve.

Let me know if you prefer fluent got a separate JavaPackage (maybe when we do lite we really do need to extend it).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,78 +71,10 @@ AutoRest needs the below config to pick this up as a plug-in - see https://githu
 
 #### Javagen
 
-``` yaml !$(fluent)
-use-extension:
-  "@autorest/modelerfour": "4.8.221"
-use: $(this-folder)//javagen
-
-pipeline:
-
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -71,78 +71,10 @@ AutoRest needs the below config to pick this up as a plug-in - see https://githu
...
-output-artifact: java-files
+``` yaml $(java) && !$(fluent)
+use: $(this-folder)/javagen
 ```
 
-``` yaml $(fluent)
-use-extension:
-  "@autorest/modelerfour": "4.9.236"
+``` yaml $(java) && $(fluent)
 use: $(this-folder)/fluentgen
```
weidongxu-microsoft: It seems easier to only load javagen or fluentgen here. Since javagen will then load preprocessor, 2 readme from them would complete the pipeline.

PR #545, opened 2020-03-20
Body array

jianghaolu:


generate
```diff
@@ -2,6 +2,7 @@
 
 COMMON_ARGUMENTS='--version=3.0.6257 --java --use:./ --output-folder=tests --sync-methods=all --client-side-validations=true --add-context-parameter=true'
 
+autorest-beta $COMMON_ARGUMENTS --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/body-array.json --namespace=fixtures.bodyarray
```
weidongxu-microsoft: BTW we should be able to use `autorest` now, since v3 is released.

PR #547, opened 2020-03-20
Fix typo in ci.yaml

jianghaolu:


eng/pipelines/fluent_integration.yaml
```diff
@@ -1,12 +1,10 @@
 trigger:
   branches:
     include:
-    - v4*
+    - v4
 
 pr:
-  branches:
```
weshaggard: This will still need to have the branches and includes lines.

PR #550, opened 2020-03-22
Fix typo in ci.yaml

jianghaolu:


mitchdenny:
Recommend that you merge this. It'll be reading the PR triggers from the target branch, not the PR so even though you've tweaked the triggers in this change, they aren't being evaluated.

jianghaolu:
@mitchdenny I'm merging the PR but do you see any error with the ci.yaml in the target branch though? I thought `v4*` would work too? David and I have spent hours on this and a little help here will be appreciated.

jianghaolu:
After merging this PR I created another PR at https://github.com/Azure/autorest.java/pull/551. As you can see, with this version of ci.yaml there builds are still not running.

PR #561, opened 2020-03-27
Remove imports of classes in same package

srnagar:


srnagar:
>It is same logic as before the equals change?

Yes.

PR #562, opened 2020-03-27
Generate additional properties and fix some corner cases

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -131,6 +131,16 @@ public ClientModel map(ObjectSchema compositeType) {
                 }
             }
             builder.imports(new ArrayList<>(modelImports));
+            if (hasAdditionalProperties) {
+                for (Property property : compositeTypeProperties) {
+                    if (property.getLanguage().getJava().getName().equals("additionalProperties")) {
+                        property.getLanguage().getJava().setName("AdditionalPropertiesProperty");
+                    }
+                    if (property.getSerializedName().equals("additionalProperties")) {
+                        property.setSerializedName("additionalProperties-original");
```
weidongxu-microsoft: Maybe we should not change the `serializedName`?

I am not sure how this one actually get implemented. Only assume that the serializedName should remain as it is.

jianghaolu: It has to be anything but remain as is. Because at serialization time the other additionalProperties will be temporarily serialized into “additionalProperties”, which conflicts with this normal property. See https://github.com/Azure/azure-sdk-for-java/pull/9624

jianghaolu: I’m not super happy with this either but cannot think of a better way. I’m open to suggestions

weidongxu-microsoft: I see. `@JsonProperty(value = "")` would let `valueToTree` generate the node with key = field name.

Yeah, not able to think a better approach with this in place.

weidongxu-microsoft: BTW, this one seems to work in basic case with only `ObjectMapper`. I didn't test any corner case (e.g. same key name with existing JsonProperty, flatten etc.).

```
    class TestClass {
        private Map<String, String> additionalProperties;

        @JsonAnySetter
        void setAdditionalProperties(String key, String value) {
                additionalProperties = new HashMap<>();
            }
            additionalProperties.put(key, value);
        }

        @JsonAnyGetter
        public Map<String, String> getAdditionalProperties(){
            return additionalProperties;
        }
    }
```

jianghaolu: This is gold. These 2 annotations should have been used in place for the custom serializers. But for backward compatibility I'll keep them there but update the new generated code to use this.

Issue #515, opened 2020-02-21
Fix checkstyle issues in generated code


srnagar:
- Remove unused imports
- Line length should be less than 120 chars
- Add `@ServiceClient` annotation to service client
- Remove public methods that don't have any implementation

weidongxu-microsoft:
Currently we've got these in mgmt generated code (only sample of different kinds of checkstyle error). I will start to fix some (other than 120char, which seems easier to pass through a formatter).

@jianghaolu @srnagar Let me know whether any of below could be conflicting with data-plane code. (I think we are using same checkstyle config in azure-sdk-for-java)

```
[ERROR] C:\github\azure-sdk-for-java\sdk\resources\mgmt\src\main\java\com\azure\management\resources\models\ResourceManagementClientBuilder.java:105:5: @ServiceClientBuilder class should not have a method name, 'build' starting with 'build' but not ending with 'Client'. [ServiceClientBuilder]
[ERROR] C:\github\azure-sdk-for-java\sdk\resources\mgmt\src\main\java\com\azure\management\resources\models\ResourceManagementClientImpl.java:97:5: Field "httpPipeline" is only assigned in constructor and it is not final. You should consider making the field final, or suppressing the warning. [EnforceFinalFields]
[ERROR] C:\github\azure-sdk-for-java\sdk\resources\mgmt\src\main\java\com\azure\management\resources\models\ResourceManagementClientImpl.java:153:5: Field "resources" is only assigned in constructor and it is not final. You should consider making the field final, or suppressing the warning. [EnforceFinalFields]
[ERROR] C:\github\azure-sdk-for-java\sdk\resources\mgmt\src\main\java\com\azure\management\resources\models\ResourceManagementClientImpl.java:210: Line is longer than 120 characters (found 143). [LineLength]
[ERROR] C:\github\azure-sdk-for-java\sdk\resources\mgmt\src\main\java\com\azure\management\resources\models\ResourcesInner.java:49:5: Field "service" is only assigned in constructor and it is not final. You should consider making the field final, or suppressing the warning. [EnforceFinalFields]
[ERROR] C:\github\azure-sdk-for-java\sdk\resources\mgmt\src\main\java\com\azure\management\resources\models\ResourcesInner.java:54:5: Field "client" is only assigned in constructor and it is not final. You should consider making the field final, or suppressing the warning. [EnforceFinalFields]
[...]

PR #566, opened 2020-04-02
v4 fix checkstyle

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/515


fluentgen/pom.xml
```diff
@@ -33,6 +40,12 @@
       <version>3.15.13</version>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.googlejavaformat</groupId>
+      <artifactId>google-java-format</artifactId>
+      <version>1.8_customized-SNAPSHOT</version>
```
weidongxu-microsoft, line 46: A customized google-java-format, tweaked to match our checkstyle.

fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentClientMethodTemplate.java
```diff
@@ -175,4 +180,57 @@ private static IType returnTypeWithoutMono(IType returnType) {
         }
         return returnTypeWithoutMono;
     }
+
+    @Override
+    protected String breakUrlOnLengthLimit(String line) {
```
weidongxu-microsoft, line 185: Break long URL.

It is ugly so for now I keep it in fluentgen.

javagen/src/main/java/com/azure/autorest/template/ProxyTemplate.java
```diff
@@ -118,6 +118,14 @@ public final void write(Proxy restAPI, JavaClass classBlock) {
         }
     }
 
+    private static String serviceInterfaceWithLengthLimit(String serviceInterfaceName) {
+        final int lengthLimit = 20;
```
weidongxu-microsoft, line 122: limit 20

Maybe we should put these to some configure file.

weidongxu-microsoft:
google-java-format requires java 11...

javagen/src/main/java/com/azure/autorest/util/CodeNamer.java
```diff
@@ -235,6 +235,9 @@ public static String getEnumMemberName(String name) {
                 }
             }
         }
+        if ("_".equals(result)) {   // "_" is keyword in Java 9
```
weidongxu-microsoft, line 238: `_` is keyword in Java 9.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ServiceClient.java
```diff
@@ -142,8 +142,10 @@ public final String getClientBaseName() {
      * @param includeImplementationImports Whether or not to include imports that are only necessary for method implementations.
      */
     public final void addImportsTo(Set<String> imports, boolean includeImplementationImports, boolean includeBuilderImports, JavaSettings settings) {
-        for (ClientMethod clientMethod : getClientMethods()) {
-            clientMethod.addImportsTo(imports, includeImplementationImports, settings);
+        if (!includeBuilderImports) {
```
weidongxu-microsoft, line 145: Builder class should not need anything from client method.

fluentgen/pom.xml
weidongxu-microsoft: However it requries Java 11.

weidongxu-microsoft: Use java version and reflect to call it. So it will pass compile, and disabled in runtime if version is lower than 11.

PR #571, opened 2020-04-09
v4 fluent naming config, fix javadoc in non-string enum

weidongxu-microsoft:
@yungezz @ChenTanyi @xccc-msft 

Note this will force it for all places, regardless what casing used in spec.

### DIFF
--- a/fluentnamer/readme.md
+++ b/fluentnamer/readme.md
@@ -19,7 +19,12 @@ pipeline:
     flatten-models: true
     flatten-payloads: true
     naming:
-      preserve-uppercase-max-length: 4
+      override:
+        api: Api
+        cpk: Cpk
+        sas: Sas
+        url: Url
+        xml: Xml
   
   # allow developer to do transformations on the code model.
   modelerfour/new-transform:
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft:
ManagedClusterAPIServerAccessProfile -> ManagedClusterApiServerAccessProfile

More in network since this is for all.

weidongxu-microsoft:
Sample javadoc

```
    /**
     * De-serializes the instance to int value.
     *
     * @return the int value.
     */
    @JsonValue
    public int toInt() {
        return this.value;
    }
```

fluentnamer/readme.md
```diff
@@ -19,7 +19,12 @@ pipeline:
     flatten-models: true
     flatten-payloads: true
     naming:
-      preserve-uppercase-max-length: 4
+      override:
+        api: Api
+        cpk: Cpk
+        sas: Sas
+        url: Url
+        xml: Xml
```
ChenTanyi, line 27: `SAS, XML, HTTP, JSON, API, URL, CPK`, it seems there is `Http` and `Json`

weidongxu-microsoft: I've deleted the `preserve-uppercase-max-length: 4` line. So all 4 letter will be changed.

PR #573, opened 2020-04-10
Add context to single page async client methods

srnagar:
Add context to single page async and paging async client methods.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
@@ -10,6 +10,7 @@ import com.azure.autorest.extension.base.model.codemodel.Schema;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientMethod;
+import com.azure.autorest.model.clientmodel.ClientMethod.Builder;
 import com.azure.autorest.model.clientmodel.ClientMethodParameter;
 import com.azure.autorest.model.clientmodel.ClientMethodType;
 import com.azure.autorest.model.clientmodel.ClientModel;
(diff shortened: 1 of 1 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -360,6 +348,37 @@ public static ClientMethodMapper getInstance() {
...
+        withContextParameters.addAll(parameters);
+        withContextParameters.add(contextParam);
+
+        methods.add(builder
+            .parameters(withContextParameters) // update builder parameters to include context
+            .returnValue(returnValue)
+            .name(proxyMethodName)
+            .onlyRequiredParameters(false)
+            .type(clientMethodType)
+            .isGroupedParameterRequired(false)
```
weidongxu-microsoft, line 376: Not sure whether need to set all these?

It seems to me that only thing need to be swapped is `withContextParameters`? All rest appears to be same as the `builder` from input parameter?

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -196,6 +197,12 @@ public static ClientMethodMapper getInstance() {
                         .type(ClientMethodType.PagingAsyncSinglePage)
                         .isGroupedParameterRequired(false)
                         .build());
+                if (JavaSettings.getInstance().isContextClientMethodParameter()) {
```
jianghaolu, line 200: nit: `JavaSettings.getInstance()` is assigned to a local variable `settings` on line 53

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
srnagar: Yes, you're right that the builder is already configured with these values from the previous client method but I didn't want to make an assumption that this method would be added only after the non-context method is added.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
srnagar: Updated to just use `settings`

PR #575, opened 2020-04-13
v4 fix docstring

weidongxu-microsoft:


javagen/src/main/java/com/azure/autorest/util/CodeNamer.java
```diff
@@ -266,7 +266,7 @@ public static String getEnumMemberName(String name) {
                 int preWidthWrapAt = -1;
                 int postWidthWrapAt = -1;
                 for (int i = 0; i != whitespacePositions.size() - 1; i++) {
-                    if (whitespacePositions.get(i) <= width && whitespacePositions.get(i + 1) > width) {
+                    if (whitespacePositions.get(i + 1) > width) {
```
weidongxu-microsoft, line 269: Break at first place that NEXT position over the line limit.

Since there is `break` in this `if` block, no need to guard against 2nd time this happens.

javagen/src/main/java/com/azure/autorest/template/ServiceClientTemplate.java
```diff
@@ -63,7 +63,7 @@ public final void write(ServiceClient serviceClient, JavaFile javaFile) {
             // Add proxy service member variable
             if (serviceClient.getProxy() != null) {
                 classBlock.javadocComment(String.format("The proxy service used to perform REST calls."));
-                classBlock.privateMemberVariable(serviceClient.getProxy().getName(), "service");
+                classBlock.privateFinalMemberVariable(serviceClient.getProxy().getName(), "service");
```
weidongxu-microsoft, line 66: Add `final` to RestProxy service.

PR #579, opened 2020-04-16
Support timeschema and represent it as String

srnagar:
This PR adds support for `time` type. A new TimeSchema is added which will use `String` representation. This is because Java's `OffsetTime` requires a zone offset while `LocalTime` doesn't support zone offset. As per ISO 8601, zone offset is optional. Before we make a decision on which type to use, this PR represents time as a string to unblock Form Recognizer.


weidongxu-microsoft:
Just a note.

Currently mgmt plane uses deserializer which fallback to GMT on OffsetDateTime if no zone is available, because some mgmt service doesn't send the "Z".
https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/resources/mgmt/src/main/java/com/azure/management/AzureServiceClient.java#L180

PR #583, opened 2020-04-17
keep track1 naming style when track1-naming=true

weidongxu-microsoft:


javagen/src/main/java/com/azure/autorest/template/ServiceClientTemplate.java
```diff
@@ -127,8 +127,8 @@ public final void write(ServiceClient serviceClient, JavaFile javaFile) {
                     comment.description(String.format("Gets the %1$s object to access its operations.", methodGroupClient.getVariableType()));
                     comment.methodReturns(String.format("the %1$s object.", methodGroupClient.getVariableType()));
                 });
-                classBlock.publicMethod(String.format("%1$s get%2$s()", methodGroupClient.getVariableType(),
-                    CodeNamer.toPascalCase(methodGroupClient.getVariableName())), function ->
+                classBlock.publicMethod(String.format("%1$s %2$s()", methodGroupClient.getVariableType(),
+                    CodeNamer.getModelNamer().modelPropertyGetterName(methodGroupClient.getVariableName())), function ->
```
weidongxu-microsoft, line 131: I am a bit lazy here. This is "client", not a "model" anymore. But still keep this until we got bigger divergence.

PR #586, opened 2020-04-17
Improve vanilla test coverage

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
```diff
@@ -105,8 +105,8 @@ public ClientModelProperty map(Property property) {
                 && sequence.getElementType().getSerialization().getXml() != null
                 && sequence.getElementType().getSerialization().getXml().getName() != null) {
                 builder.xmlListElementName(sequence.getElementType().getSerialization().getXml().getName());
-            } else if(sequence.getSerialization() != null && sequence.getSerialization().getXml() != null){
-                builder.xmlListElementName(sequence.getSerialization().getXml().getName());
+            } else {
+                builder.xmlListElementName(sequence.getElementType().getLanguage().getDefault().getName());
```
jianghaolu, line 109: @srnagar if serialization.xml.name isn't available - it doesn't seem right to use the array schema's xml name for the item name. For now I'm using the name of the item as a fallback and seems to generate this correctly so far.

srnagar: Sounds good.

vanilla-tests/pom.xml
```diff
@@ -18,12 +18,12 @@
...
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.2.0</version>
+      <version>1.4.0</version>
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-http-netty</artifactId>
-      <version>1.1.0</version>
+      <version>1.4.0</version>
```
srnagar, line 26: Just using azure-core-http-netty as dependency would transitively bring in azure-core as well. Also, we have 1.5.0 as the latest netty version.

weidongxu-microsoft: Ideally, netty should be in test scope.

PR #587, opened 2020-04-20
v4 NPE guard on primitive type

weidongxu-microsoft:
The case is async returns e.g. `Mono<Integer>` and sync return `int`.


vanilla-tests/src/main/java/fixtures/bodyboolean/Bools.java
```diff
@@ -141,7 +141,8 @@
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public boolean getTrue() {
-        return getTrueAsync().block();
+        Boolean value = getTrueAsync().block();
+        return value == null ? false : value;
```
weidongxu-microsoft, line 145: Currently I return default value for the primitive type, if result is `null`.

Or should we throw certain exception?

vanilla-tests/src/main/java/fixtures/bodyboolean/Bools.java
```diff
@@ -141,7 +141,12 @@
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public boolean getTrue() {
-        return getTrueAsync().block();
+        Boolean value = getTrueAsync().block();
+        if (value != null) {
+            return value;
+        } else {
+            throw new NullPointerException();
```
weidongxu-microsoft, line 148: Now it just make spotbugs happy. Side effect this should make the NPE explicit.

weidongxu-microsoft:
Based on https://github.com/Azure/autorest.java/pull/588

PR #593, opened 2020-04-29
remove hack, use "id" for resource type

weidongxu-microsoft:


weidongxu-microsoft:
Relate to https://github.com/Azure/azure-sdk-for-java/pull/9623

PR #602, opened 2020-05-15
v4 NPE guard in MapType validate

weidongxu-microsoft:
A rather edge case for mgmt compute.

`ManagedServiceIdentity` actually need add a `key -> null` in UPDATE to remove a user identity. Hence value in Map could be null.

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/modelflattening/AutoRestResourceFlatteningTestService.java
+++ b/vanilla-tests/src/main/java/fixtures/modelflattening/AutoRestResourceFlatteningTestService.java
@@ -366,7 +366,7 @@ public final class AutoRestResourceFlatteningTestService {
             return Mono.error(new IllegalArgumentException("Parameter this.getHost() is required and cannot be null."));
         }
         if (resourceDictionary != null) {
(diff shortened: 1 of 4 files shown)


vanilla-tests/src/main/java/fixtures/modelflattening/AutoRestResourceFlatteningTestService.java
```diff
@@ -366,7 +366,7 @@ public void putWrappedArray(List<WrappedProduct> resourceArray) {
             return Mono.error(new IllegalArgumentException("Parameter this.getHost() is required and cannot be null."));
         }
         if (resourceDictionary != null) {
-            resourceDictionary.values().forEach(e -> e.validate());
+            resourceDictionary.values().forEach(e -> {if (e != null) {e.validate();}});
```
jianghaolu, line 369: nit: space after { and before }

weidongxu-microsoft: Fixed. checkstyle actually allows code block in single line.

Issue #601, opened 2020-05-14
Boolean cmd arguments are not recognized


jianghaolu:
Running
```bash
autorest-beta --java --use:D:\git\conniey\autorest.java --input-file:D:\git\conniey\azure-sdk-for-java\sdk\servicebus\
azure-messaging-servicebus\swagger.json --output-folder:D:\git\conniey\azure-sdk-for-java\sdk\servicebus\azure-messaging
-servicebus\src\main\java\com\azure\messaging\servicebus\implementation --namespace:com.azure.messaging.servicebus.imple
mentation --enable-xml
```

will throw 
```
java.lang.IllegalArgumentException: Cannot deserialize instance of `java.lang.Boolean` out of START_OBJECT token
at [Source: UNKNOWN; line: -1, column: -1]
at com.fasterxml.jackson.databind.ObjectMapper._convert(ObjectMapper.java:3922)
at com.fasterxml.jackson.databind.ObjectMapper.convertValue(ObjectMapper.java:3873)
at com.azure.autorest.extension.base.jsonrpc.Connection.lambda$process$3(Connection.java:285)
at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.lang.Boolea
n` out of START_OBJECT token
at [Source: UNKNOWN; line: -1, column: -1]
```

Instead one has to type `--enable-xml=true`. The JSON RPC module should be able to handle boolean arguments without `=true`.

PR #604, opened 2020-05-18
handle boolean settings, as {}

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/601

It is "{}" in `jobject.get("result")`. Not sure if there is better solution.


jianghaolu:
Yes! Finally!

weidongxu-microsoft:
There is a new test we fails. https://github.com/Azure/autorest.testserver/pull/184

I will fix it in another PR.

PR #605, opened 2020-05-18
v4 sync client method context

weidongxu-microsoft:
As requested in guideline, sync method need to take `Context` as optional parameter.


weidongxu-microsoft:
@srnagar I think the same guideline works for data-plane generated code as well? Currently I add it Context to sync method for both data-plane and mgmt-plane.

PR #608, opened 2020-05-19
add beginFoo for LRO, returns PollerFlux

weidongxu-microsoft:
Only for mgmt.

Requested by Jonathan to have a method return `PollerFlux`.

This client method should be in Async client as well as Sync client. Currently I settle as type `LongRunningBegin`.

Sample generated code.
```
    public PollerFlux<PollResult<StorageAccountInner>, StorageAccountInner> beginCreate(
        String resourceGroupName, String accountName, StorageAccountCreateParameters parameters) {
        Mono<SimpleResponse<Flux<ByteBuffer>>> mono =
            createWithResponseAsync(resourceGroupName, accountName, parameters);
        return this
            .client
            .<StorageAccountInner, StorageAccountInner>getLroResultAsync(
                mono, this.client.getHttpPipeline(), StorageAccountInner.class, StorageAccountInner.class);
    }
```

`beginCreate.take(1).blockFirst()` (but it actually doing first GET query) should be similar to previous `beginCreate` in mgmt.

@ChenTanyi @xccc-msft @yungezz


weidongxu-microsoft:
With https://github.com/Azure/autorest.java/pull/605 merged

weidongxu-microsoft:
For now still keep `beginCreateWithoutPolling`

Issue #609, opened 2020-05-19
Only first method group is wrapped in Client/AsyncClient for "generate-sync-async-clients"


weidongxu-microsoft:
Other method group is missing.

weidongxu-microsoft:
Expect to have multiple Client/AsyncClient, each map to 1 operation group / method group.

Issue #610, opened 2020-05-20
Generated Sync and Async client don't provide default value for non-required parameters


idear1203:
I am trying to generate code with both below settings applied:
```
required-parameter-client-methods: true
generate-sync-async-clients: true
```

However, the generated clients contain some compilation errors caused by non-required parameters.
![image](https://user-images.githubusercontent.com/7024383/82401257-e6b56280-9a8b-11ea-8bb6-69a8102a471a.png)

Swagger spec link: https://github.com/idear1203/azure-rest-api-specs-pr/blob/synapse_update_data_plane_specs/specification/synapse/data-plane/Microsoft.Synapse/preview/2019-11-01-preview/sparkJob.json#L53-L61

Expectations:
- For non-required integers, default value should be null or 0. (I prefer null)
- For non-required booleans, default value should be false.

PR #611, opened 2020-05-20
v4 support multiple async clients

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/609


weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/610 as well

weidongxu-microsoft:
Remove "Async" suffix in async client.

PR #618, opened 2020-05-27
Add support for XML namespaces

conniey:
In swagger, sometimes XML can contain namespaces. Our Java generator currently does not generate JacksonXmlProperty if there is a namespace. If the XML property has a namespace, we'll generate `JacksonXmlProperty(namespace = "", localName ="")`, otherwise the normal JsonProperty annotation will be used (which works).
As an aside, Jackson supports having multiple annotations on it so if we wanted to support XML and JSON side by side, it would work.

```json
"CreateQueueBody": {
  "description": "The request body for creating a queue.",
  "type": "object",
  "xml":{
    "name": "entry",
    "namespace": "http://www.w3.org/2005/Atom"
  },
  "properties": {
    "content": {
      "type": "object",
      "xml": {
        "namespace": "http://www.w3.org/2005/Atom"
      },
      "properties": {
        "type": {
          "type": "string",
          "default": "application/xml",
          "xml": {
            "attribute": true
          }
        },
        "QueueProperties": {
          "$ref": "#/definitions/QueueProperties",
          "xml": {
            "name": "QueueDescription",
            "namespace": "http://schemas.microsoft.com/netservices/2010/10/servicebus/connect"
          }
        }
      }
    }
  }
```

Generated Java code:
```java
@JacksonXmlRootElement(localName = "entry", namespace = "http://www.w3.org/2005/Atom")
@Fluent
public final class CreateQueueBody {
    /*
     * TODO: Add description
     */
    @JacksonXmlProperty(localName = "content", namespace = "http://www.w3.org/2005/Atom")
    private CreateQueueBodyContent content;
...
}
```
and
```java
@JacksonXmlRootElement(localName = "null", namespace = "http://www.w3.org/2005/Atom")
@Fluent
public final class CreateQueueBodyContent {
    @JacksonXmlProperty(localName = "type", isAttribute = true)
    private String type;

    @JacksonXmlProperty(
            localName = "QueueDescription",
            namespace = "http://schemas.microsoft.com/netservices/2010/10/servicebus/connect")
    private QueueProperties queueProperties;
```

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
@@ -8,6 +8,7 @@ import com.azure.autorest.extension.base.model.codemodel.Languages;
 import com.azure.autorest.extension.base.model.codemodel.ObjectSchema;
 import com.azure.autorest.extension.base.model.codemodel.Property;
 import com.azure.autorest.extension.base.model.codemodel.Schema;
+import com.azure.autorest.extension.base.model.codemodel.XmlSerlializationFormat;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientModel;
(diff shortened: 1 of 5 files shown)


jianghaolu:
Awesome stuff!

PR #619, opened 2020-05-28
single place to get builder suffix

weidongxu-microsoft:
@srnagar

Current generator will make builder class as AutoRestSwaggerBATDictionaryServiceImplBuilder (the deleted java file) for `generate-sync-async-clients=true`. I assume it is not intended.

Anyway, I move the builder suffix to a single method `ClientModelUtil.getBuilderSuffix()`. So if need change, this would be more flexible.


weidongxu-microsoft:
@srnagar

I understand that you are busy :-) ... Just a ping to let me know if you want `AutoRestSwaggerBATDictionaryServiceImplBuilder` or `AutoRestSwaggerBATDictionaryServiceBuilder` for `generate-sync-async-clients=true`

jianghaolu:
@srnagar can confirm but I believe it's intended. We have users complaining the builder for `FooClientImpl` should be `FooClientImplBuilder` instead of `FooClientBuilder`.

weidongxu-microsoft:
Got it. I will wait for Srikanta's feedback.

For `generate-sync-async-clients` flag, the `FooClientImpl` is in `implementation` package and is not visible by user (created by `private buildInnerClient()` method). `FooClientImplBuilder` would actually build `FooAsyncClient` and `FooClient` (wrapper which calls `FooClientImpl`), hence my suspicion that it might meant to be `FooClientBuilder`.

srnagar:
@weidongxu-microsoft If the `--generate-sync-async-clients` is set and `FooClient` and `FooAsycClient` are generated then the builder can be `FooClientBuilder`. If the builder is only building `FooClientImpl` then the builder should be named as `FooClientImplBuilder`.

weidongxu-microsoft:
@srnagar BTW, you are free to change the Builder naming. Mgmt-plane does not depends on this. I only found the issue when trying to generate async/sync for mgmt-plane as vanilla style.

PR #623, opened 2020-06-02
v4 mgmt refactor

weidongxu-microsoft:
First stage for Jonathan's feedback.

Most changes is on fluent.


azure-tests/src/main/java/fixtures/azurereport/AutoRestReportServiceForAzure.java
```diff
@@ -45,7 +45,7 @@ public String getHost() {
      * @param host the host value.
      * @return the service client itself.
      */
-    public AutoRestReportServiceForAzure setHost(String host) {
+    AutoRestReportServiceForAzure setHost(String host) {
```
weidongxu-microsoft, line 48: Make it package private for now (if builder and client is in same package).

Later will move it to constructor, so that client will be immutable.

weidongxu-microsoft:
Try to handle https://github.com/Azure/autorest.testserver/pull/185

weidongxu-microsoft:
Surprised to see the base64 encoded bytes is a string *quoted*.

`"YSBzdHJpbmcgdGhhdCBnZXRzIGVuY29kZWQgd2l0aCBiYXNlNjQ="`

PR #624, opened 2020-06-03
v4 mgmt, make client immutable

weidongxu-microsoft:
As Jonathan's request. make client immutable.

Currently I do it for both data and mgmt. Let me know if data-plane does not need it.


azure-tests/src/main/java/fixtures/azurereport/AutoRestReportServiceForAzure.java
```diff
@@ -28,7 +28,7 @@
     private final AutoRestReportServiceForAzureService service;
 
     /** server parameter. */
-    private String host;
+    private final String host;
```
weidongxu-microsoft, line 31: I mgmt, I've renamed this to `endpoint`.

jianghaolu: `endpoint` is probably a more accurate name.

PR #634, opened 2020-06-17
Make service client as private final

idear1203:


idear1203:
Hi @srnagar , sorry I got some urgent tasks to do recently. I will check the failure soon.

PR #636, opened 2020-06-18
Make apiversion settable on the builder

srnagar:
`apiversion` has a default value in the swagger but it's a settable property for data-plane APIs. So, this PR changes the behavior of global parameters to be settable in the builder and if it's not set, the buildClient() method will set it to default value.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
@@ -1,7 +1,6 @@
 package com.azure.autorest.mapper;
 
 import com.azure.autorest.extension.base.model.codemodel.CodeModel;
-import com.azure.autorest.extension.base.model.codemodel.ConstantSchema;
 import com.azure.autorest.extension.base.model.codemodel.Operation;
 import com.azure.autorest.extension.base.model.codemodel.OperationGroup;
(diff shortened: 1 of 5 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
```diff
@@ -108,14 +106,14 @@ public ServiceClient map(CodeModel codeModel) {
                 serviceClientPropertyClientType = serviceClientPropertyClientType.asNullable();
             }
 
-            boolean serviceClientPropertyIsReadOnly = p.getSchema() instanceof ConstantSchema;
```
weidongxu-microsoft, line 111: Is it possible for you to add a `isFluent` guard for old logic? "api-version" is a Constant in m4, hence it should be not modifiable by definition.

IMHO, if one want "api-version" to be settable, it would more like an exception for one single property (instead of making every Constant property non-constant).

And Jonathan asked us to make it not settable in mgmt-plane.

weidongxu-microsoft: I know previously it is settable, but that is due to a bug:

```
boolean serviceClientPropertyIsReadOnly = serviceClientPropertyClientType instanceof ConstantSchema;
```

`serviceClientPropertyClientType` is client model type, `ConstantSchema` is code model type (m4). They will never match.

PR #640, opened 2020-06-22
v4 fix optional paramter of wiretype

weidongxu-microsoft:
FIx bug that optional parameter in client method is not inited.

Before fix,
```
    public PagedFlux<DeploymentResourceInner> listAsync(String resourceGroupName, String serviceName, String appName) {
        final Context context = null;
        return new PagedFlux<>(
            () -> listSinglePageAsync(resourceGroupName, serviceName, appName, version),
            nextLink -> listNextSinglePageAsync(nextLink));
    }
```

After fix,
```
    public PagedFlux<DeploymentResourceInner> listAsync(String resourceGroupName, String serviceName, String appName) {
        final List<String> version = null;
        final Context context = null;
        return new PagedFlux<>(
            () -> listSinglePageAsync(resourceGroupName, serviceName, appName, version),
            nextLink -> listNextSinglePageAsync(nextLink));
    }
```

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
@@ -87,7 +87,10 @@ public class ClientMethodTemplate implements IJavaTemplate<ClientMethod, JavaTyp
                     (parameterClientType instanceof ArrayType || parameterClientType instanceof ListType)) {
                 parameterWireType = ClassType.String;
             }
-            boolean alwaysNull = parameterWireType != parameterClientType && clientMethod.getOnlyRequiredParameters() && !parameter.getIsRequired();
+            // "addConstant" clause is a hack here.
+            // For SimpleAsyncRestResponse and PagingAsyncSinglePage, "alwaysNull" is set to true, to let "ConvertClientTypesToWireTypes" method set the optional parameter.
+            // For other ClientMethodType (which would have addConstant=false), optional parameter need to be initialized for calling method with full parameters.
+            boolean alwaysNull = addConstant && parameterWireType != parameterClientType && clientMethod.getOnlyRequiredParameters() && !parameter.getIsRequired();
 
             if (!parameter.getFromClient() && !alwaysNull && ((addOptional && clientMethod.getOnlyRequiredParameters() && !parameter.getIsRequired()) || (addConstant && parameter.getIsConstant()))) {
                 String defaultValue = parameterClientType.defaultValueExpression(parameter.getDefaultValue());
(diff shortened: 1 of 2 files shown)


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -87,7 +87,10 @@ protected static void AddOptionalAndConstantVariables(JavaBlock function, Client
                     (parameterClientType instanceof ArrayType || parameterClientType instanceof ListType)) {
                 parameterWireType = ClassType.String;
             }
-            boolean alwaysNull = parameterWireType != parameterClientType && clientMethod.getOnlyRequiredParameters() && !parameter.getIsRequired();
+            // "addConstant" clause is a hack here.
+            // For SimpleAsyncRestResponse and PagingAsyncSinglePage, "alwaysNull" is set to true, to let "ConvertClientTypesToWireTypes" method set the optional parameter.
+            // For other ClientMethodType (which would have addConstant=false), optional parameter need to be initialized for calling method with full parameters.
+            boolean alwaysNull = addConstant && parameterWireType != parameterClientType && clientMethod.getOnlyRequiredParameters() && !parameter.getIsRequired();
```
weidongxu-microsoft, line 93: So far, it is safe to 1. set the parameter to `null`, 2. convert the client type of `null` value to wire type.

Checked that following will all convert `null` to `null`, as intended.
```
Base64Util.encodeToString
Base64Url.encode
JacksonAdapter.createDefaultSerializerAdapter().serializeList
```

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -374,7 +377,7 @@ public final void write(ClientMethod clientMethod, JavaType typeBlock) {
                 typeBlock.publicMethod(clientMethod.getDeclaration(), function -> {
                     AddOptionalVariables(function, clientMethod, restAPIMethod.getParameters(), settings);
                     if (clientMethod.getReturnValue().getType() == ClassType.InputStream) {
-                        function.line("Iterator<ByteBufferBackedInputStream> iterator = %s(%S).map(ByteBufferBackedInputStream::new).toStream().iterator()",
+                        function.line("Iterator<ByteBufferBackedInputStream> iterator = %s(%S).map(ByteBufferBackedInputStream::new).toStream().iterator();",
```
weidongxu-microsoft, line 380: bug fix.

`Formatter.formatSourceAndFixImports` will fail on invalid java code. It then fails the generate script, but not triggering error.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
srnagar: I am not sure what `addConstant` is used for in other places but taking a dependency on this field to determine whether to init the optional param might cause issues when the semantics of `addConstant` change. Is there a more reliable way to determine whether optional param should be initialized or not?

weidongxu-microsoft: OK. Added `ignoreParameterNeedConvert` for this purpose.

PR #661, opened 2020-07-21
v4 fluentlite service client

weidongxu-microsoft:
1. add init test framework for fluent lite
2. add `defaultPollInterval` to client/builder
3. for fluent lite, generate a complete client class (no longer need base `AzureServiceClient`)
4. support try-catch-finally block

### DIFF
--- a/fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentServiceClientTemplate.java
+++ b/fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentServiceClientTemplate.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+package com.azure.autorest.fluent.template;
+
+import com.azure.autorest.template.ServiceClientTemplate;
+import com.azure.autorest.template.prototype.MethodTemplate;
+
+public class FluentServiceClientTemplate extends ServiceClientTemplate {
+
+    private static FluentServiceClientTemplate _instance = new FluentServiceClientTemplate();
+    static {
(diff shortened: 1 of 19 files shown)


javagen/src/main/java/com/azure/autorest/template/prototype/MethodTemplate.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+
+public class MethodTemplate {
```
weidongxu-microsoft, line 23: Add a `MethodTemplate` so things related to a method can be grouped together (import, javadoc, method signature and code).

fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentServiceClientTemplate.java
```diff
@@ -0,0 +1,138 @@
...
+                    .method(method -> {
+                        method.methodReturn("PollerFactory.create(serializerAdapter, httpPipeline, pollResultType, finalResultType, defaultPollInterval, activationResponse)");
+                    })
+                    .build();
+
+            MethodTemplate getLroFinalResultOrErrorMethod = MethodTemplate.builder()
+                    .imports(Arrays.asList(PollerFlux.class.getName(), PollResult.class.getName(),
+                            Mono.class.getName(), AsyncPollResponse.class.getName(),
+                            ManagementError.class.getName(), ManagementException.class.getName(),
+                            LongRunningOperationStatus.class.getName(), SerializerEncoding.class.getName(), IOException.class.getName()))
```
weidongxu-microsoft, line 89: Desirable to automatically infer all (or most) of these import classes from code itself.

PR #662, opened 2020-07-22
autorest.java for Android: Add implementation to fix imports in models

JianpingChen:


weidongxu-microsoft:
@JianpingChen 
I was not aware that there is an android-v4 branch. I assume that it will some day merged into v4 branch?

If it is the case:
In the following a month or two, mgmt-plane would be pushing lots of code for fluent lite development (like PR https://github.com/Azure/autorest.java/pull/661).
Personally I would try to do the bulk of it in fluentnamer and fluentgen. But inevitably some would happen to javagen. I will try to do this in extensible ways (like the factory pattern here) and not break things.

If you see any potential problem on conflict code, please let me know.

JianpingChen:
Thank you for the info, @weidongxu-microsoft ! We'll try to minimize changes to javagen to simplify merging.

javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -26,13 +26,17 @@
...
 
-    private ModelMapper() {
+    protected ModelMapper() {
     }
 
     public static ModelMapper getInstance() {
         return instance;
     }
 
+    protected ClientModel.Builder createClientModelBuilder() {
```
weidongxu-microsoft, line 36: Suggest to add a bit javadoc on it to clarify that it is the extension point for a different builder subclass.

JianpingChen: Added documentation. I agree with your last comment about extending Builder class. IMHO, the builder should not be embedded (and assuming there is never a need to override). We'd first identify these cases during our Android effort to see if we need to refactor the Builders. Anyway, I won't make any change to the builders unless we have to.

PR #669, opened 2020-07-27
Support Serializer Adapter for data plane

srnagar:
This PR enables customizing the `SerializerAdapter`  that will be used in azure-core for serializing and deserializing request/response from the client library.


srnagar:
Ignoring Android build failure as the token might not have been set up to publish test coverage reports to Storage. This does not impact the PR.

cc: @anuchandy can you please take a look?

PR #671, opened 2020-08-03
Add simple post processor

jianghaolu:
An example post processor can be found at https://github.com/jianghaolu/azure-sdk-for-java/blob/postprocessor-servicebus/sdk/servicebus/azure-messaging-servicebus/swagger/src/main/java/com/azure/messaging/servicebus/PostProcessor.java.

Running it under `sdk/servicebus/azure-messaging-servicebus/swagger` folder:
```bash
autorest --use=C:\code\autorest.java.v4\ --java --postprocessor-jar-path=C:\code\azure-sdk-for-java\sdk\servicebus\azure-messaging-servicebus\swagger\target\azure-messaging-servicebus-codegen-postprocessor-1.0.0.beta.1.jar --postprocessor-class=com.azure.messaging.servicebus.PostProcessor
```


javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -139,11 +148,35 @@ protected NodeTuple representJavaBeanProperty(Object javaBean, Property property
...
-                writeFile(javaFile.getFilePath(), formattedSource, null);
+            Map<String, String> formattedFiles = javaPackage.getJavaFiles().stream().collect(Collectors.toMap(JavaFile::getFilePath, f -> {
+                try {
+                    return formatter.formatSourceAndFixImports(f.getContents().toString());
+                } catch (FormatterException e) {
+                    throw new RuntimeException(e);
+                }
+            }));
+            //Step 5: post processing
+            String jarPath = JavaSettings.getInstance().getPostProcessorJarPath();
```
weidongxu-microsoft, line 161: Not sure whether it post-process better be done before or after formatting.

It seems better as "before".
1. formatting might do hard-to-predict formatting that might affect simple replacing task (e.g. `this.getFoo()` get line break at `.` due to certain line limit)
2. seems better to reformat after the post-process

weidongxu-microsoft:
I've regened tests in https://github.com/Azure/autorest.java/pull/672 (merged)

It should pass now if you refresh latest v4.

javagen/src/main/java/com/azure/autorest/Javagen.java
srnagar: Yes, agree with @weidongxu-microsoft  that we should format after post-processing. The formatter also catches some syntax errors which is useful to do after the postprocessing.

jianghaolu: But how do I know what strings to replace from the non-formatted code? The expected user scenario is they look at the formatted output code and they want to replace something. But that may not match the non-formatted code.

One way to work around it is to format the code again after post processing.

Also, we talked to C# folks and we are planning to not allow users do string replacements in the post processor - instead we will expose a few methods to do a set of "pre-approved" post processings, like "addCredentialToSyncAsyncClient", etc.

weidongxu-microsoft: OK. If the use case is user check the output then do the change, then it is reasonable to do post-process here.

jianghaolu:
close via https://github.com/Azure/autorest.java/pull/775

PR #674, opened 2020-08-04
Create options model for optional parameters in client methods

JianpingChen:
collapse optional parameters in client method into an options model
generate options model

### DIFF
--- a/androidgen/src/main/java/com/azure/autorest/android/mapper/AndroidOptionalParameterMapper.java
+++ b/androidgen/src/main/java/com/azure/autorest/android/mapper/AndroidOptionalParameterMapper.java
@@ -1,0 +1,128 @@
+package com.azure.autorest.android.mapper;
+
+import com.azure.autorest.extension.base.model.codemodel.Parameter;
+import com.azure.autorest.mapper.Mappers;
+import com.azure.autorest.model.clientmodel.ClientMethodParameter;
+import com.azure.autorest.model.clientmodel.ClientModel;
+import com.azure.autorest.model.clientmodel.ClientModelProperty;
+import com.azure.autorest.model.clientmodel.IType;
+import com.azure.autorest.util.CodeNamer;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class AndroidOptionalParameterMapper {
+    private String packageName;
+    private String methodName;
+    private List<Parameter> parameters = new ArrayList<Parameter>();
+
+    public AndroidOptionalParameterMapper packageName(String packageName) {
+        this.packageName = packageName;
+        return this;
+    }
+    public AndroidOptionalParameterMapper methodName(String methodName) {
+        this.methodName = methodName;
+        return this;
+    }
+
+    public AndroidOptionalParameterMapper parameters(List<Parameter> parameters) {
+        this.parameters = parameters;
+        return this;
+    }
+
+    public boolean hasOptionalParameters() {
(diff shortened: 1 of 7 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -42,13 +42,23 @@
...
 
-    private ClientMethodMapper() {
+    protected ClientMethodMapper() {
     }
 
     public static ClientMethodMapper getInstance() {
         return instance;
     }
 
+    public void addModelsTo(List<ClientModel> clientModels) {
```
weidongxu-microsoft, line 52: Seems better just have a `getOptionsModels` and return the List of `ClientModel`s, easier to understand what it for, IMHO. Also please add javadoc.

androidgen/src/main/java/com/azure/autorest/android/mapper/AndroidOptionalParameterMapper.java
```diff
@@ -0,0 +1,128 @@
...
+        this.parameters = parameters;
+        return this;
+    }
+
+    public boolean hasOptionalParameters() {
+        return parameters.size() > 0;
+    }
+
+    public String getModelTypeName() {
+        return CodeNamer.toPascalCase(String.format("%1$sOptions", methodName));
```
weidongxu-microsoft, line 40: Will it happen that 2 method in different operation group get same name (likely for `list` etc.), which result in same model name for Options class?

JianpingChen: Valid concern but that's kind of the assumption we made above this level already. For example, if two operations at two different paths have the same "operationId", they will cause name clash in the client (even though may have different signatures). We can work around that by modifying the swagger file so that they have different operationId. Fixing this assumption in autorest requires a lot more effort than we can afford. I'll just take a note and investigate later.

srnagar:
>The support for FooOptions for aggregated optional parameters in client methods seems to be a general feature. Is it preferable directly adding it to v4 javagen, with an option in JavaSetting to enable/disable it?
@jianghaolu @srnagar Your opinions?

I agree as long as this can be enabled/disabled through options.

weidongxu-microsoft:
@JianpingChen 
It is up to you to decide whether to do it in android-v4 branch, or directly in v4 branch with setting in `JavaSettings` to enable/disable it.

Either way, I would suggest to merge the part in javagen to v4 (in case some change in same file would cause conflict in andoird-v4).

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
JianpingChen: I'm just trying to follow addImportsTo. :-) In fact, passing what existed already may give this class a chance to check for duplicates (as you suggested in other comment). If people want to be smarter in renaming the newly generated model, this is a chance for doing so.

weidongxu-microsoft: Sure. I've no strong opinion here.

JianpingChen:
@weidongxu-microsoft cc: @jianghaolu : Given the time we have for android support, it is probably more practical for us to keep it in android-v4 for now and promote to v4 later, after a second thought. Changes we made in android-v4 may lead to more suggestions for v4 and we can review that together once we get Android work done within the timeline we are given. Thanks for the suggestion

PR #677, opened 2020-08-06
Remove javadoc for setter method when setter method is not generated

srnagar:
Javadocs for setter methods were getting generating for properties that were marked as required and option to move required fields to ctor was enabled. This PR removes the javadoc when no setter method is generated.

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/extensibleenums/models/Pet.java
+++ b/vanilla-tests/src/main/java/fixtures/extensibleenums/models/Pet.java
@@ -7,7 +7,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 @Fluent
 public final class Pet {
     /*
-     * The name property.
+     * name
      */
     @JsonProperty(value = "name")
     private String name;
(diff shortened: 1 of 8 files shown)


vanilla-tests/src/main/java/fixtures/extensibleenums/models/Pet.java
```diff
@@ -7,7 +7,7 @@
 @Fluent
 public final class Pet {
     /*
-     * The name property.
+     * name
```
jianghaolu, line 10: We probably need better docs here? But not critial

srnagar: I think these docs are from Swagger and not sure why it changed. I will go look at the swagger.

jianghaolu: I think we used to print "The {} property" when Swagger doc is missing but I don't know why this is changed either

srnagar: Yeah, if it was empty we would have generated "The {} property" but in this case swagger has non-empty doc. So, we just use that. It might be difficult to determine if the doc is insufficient from swagger.

weidongxu-microsoft: It appears that swagger of testservice changed.

Issue #607, opened 2020-05-19
Improve class names in plural form


idear1203:
Some class names in plural form are not good enough. It seems like current code generator simply add "s" behind the word.

Examples:
1. [DirectorysImpl](https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/storage/azure-storage-blob/src/main/java/com/azure/storage/blob/implementation/DirectorysImpl.java)
2. [SparkBatchsImpl](https://github.com/idear1203/azure-sdk-for-java-pr/blob/synapse_java_track_2_sdk/sdk/synapse/azure-analytics-synapse-spark/src/main/java/com/azure/analytics/synapse/spark/implementation/SparkBatchsImpl.java)

weidongxu-microsoft:
A potential lib to get proper plural form. https://github.com/atteo/evo-inflector
Not tried yet.

weidongxu-microsoft:
Another lib is
```
<dependency>
      <groupId>org.jibx</groupId>
      <artifactId>jibx-tools</artifactId>
      <version>1.3.3</version>
    </dependency>
```

PR #680, opened 2020-08-12
v4 plural

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/607


weidongxu-microsoft:
Currently using jibx-tools. License BSD. It also supports plural -> singular (Fluent likely need this). Downside the lib is large.

evo-inflector. License Apache 2.0. It does not support plural -> singular. For pluralize it has (almost) same effect as here.

@jianghaolu @srnagar Let me know your opinion.

PR #690, opened 2020-08-21
Enable Android Embedded Builder

anuchandy:


javagen/src/main/java/com/azure/autorest/template/ServiceAsyncClientTemplate.java
```diff
@@ -23,17 +23,8 @@ public static ServiceAsyncClientTemplate getInstance() {
...
-    imports.add("com.azure.core.annotation.ServiceClient");
-  }
-
-  protected void addClassAnnotation(JavaFile javaFile, ServiceClient serviceClient) {
-    javaFile.annotation(String.format("ServiceClient(builder = %s.class, isAsync = true)",
-            serviceClient.getInterfaceName() + ClientModelUtil.getBuilderSuffix()));
-  }
-
   @Override
-  public final void write(AsyncSyncClient asyncClient, JavaFile javaFile) {
```
JianpingChen, line 36: Try to avoid this, IMHO. This class is also used by other generators. We wouldn't know all the implications unless we also test all other generators.

javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientTemplate.java
```diff
@@ -23,17 +23,8 @@ public static ServiceSyncClientTemplate getInstance() {
...
-    imports.add("com.azure.core.annotation.ServiceClient");
-  }
-
-  protected void addClassAnnotation(JavaFile javaFile, ServiceClient serviceClient) {
-    javaFile.annotation(String.format("ServiceClient(builder = %s.class, isAsync = true)",
-            serviceClient.getInterfaceName() + ClientModelUtil.getBuilderSuffix()));
-  }
-
   @Override
-  public final void write(AsyncSyncClient syncClient, JavaFile javaFile) {
```
JianpingChen, line 36: Similar concern

javagen/src/main/java/com/azure/autorest/template/ServiceAsyncClientTemplate.java
anuchandy: discussed this offline, validated that we're not changing the contract here

PR #691, opened 2020-08-21
Fix enum name generation

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ChoiceMapper.java
```diff
@@ -50,15 +50,13 @@ public IType map(ChoiceSchema enumType) {
 
             List<ClientEnumValue> enumValues = new ArrayList<>();
             for (ChoiceValue enumValue : enumType.getChoices()) {
-                String enumName = enumValue.getValue();
```
weidongxu-microsoft, line 53: @jianghaolu Can we still keep the non-Fluent flag here?

The reason is that m4 sometimes do strange naming for enum value, e.g. `IPv4` -> `I_PV4`.

jianghaolu: Oh I see. Yes I'll make the change

jianghaolu: Actually then this PR is no longer needed

PR #705, opened 2020-09-16
v4 fluent, config m4 to use upper for enum value

weidongxu-microsoft:
@srnagar 

data-plane might want to do the same.

By default, m4 uses pascal for choice value, which removes "\_" or "-" etc. Then our CodeNamer will add the "\_" as seperator. For some value it produces very unpredictable result (e.g. "Standard_F72s_v2").


weidongxu-microsoft:
Still need improvement. It still do crazy thing like "IKEv2" -> "IK_EV2"

PR #717, opened 2020-09-23
Fix client model docs mapping

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -150,10 +150,12 @@ public ClientModel map(ObjectSchema compositeType) {
                 }
             }
 
-            if ((compositeType.getSummary() == null || compositeType.getSummary().isEmpty()) && (compositeType.getDescription() == null || compositeType.getDescription().isEmpty())) {
+            if (compositeType.getLanguage().getDefault() == null
+                    || compositeType.getLanguage().getDefault().getDescription() == null
+                    || compositeType.getLanguage().getDefault().getDescription().isEmpty()) {
```
weidongxu-microsoft, line 155: nit, this might be easier, `CoreUtils` got a `isNullOrEmpty`

weidongxu-microsoft (approved):

PR #718, opened 2020-09-23
Remove redundant JsonFlatten and escaped discriminators

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -194,7 +186,15 @@ public ClientModel map(ObjectSchema compositeType) {
                  builder.xmlName(compositeType.getLanguage().getDefault().getName());
             }
 
-            builder.needsFlatten(discriminatorNeedEscape || hasFlattenedProperty(compositeType, parentsNeedFlatten));
```
weidongxu-microsoft, line 197: This `needFlatten` logic on `discriminatorNeedEscape` is somehow caused by this issue https://github.com/Azure/azure-sdk-for-java/issues/10217

> Please make sure the following constraints are met when using this version:
> 
> The dot "." in the discriminator name must be escaped e.g. "swagger odata.type" to "java odata\\.type
> If a type Foo is annotated with JsonTypeInfo with discriminator containing dot then we also need to apply JsonFlatten annotation to the same type Foo.

jianghaolu: Right. But for classes without @JsonFlatten the discriminator doesn’t need to and shouldn’t be escaped.

jianghaolu: Adding @JsonFlatten and escape the discrimination results in the same behavior as without doing neither.

jianghaolu: You see the entire custom serializer is skipped if not annotated: https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/FlatteningSerializer.java#L83

weidongxu-microsoft: I think the origin issue is on the base class that has a flattened property (which not visible on this subclass).

If you think it is OK, you can merge it. I will double check it tomorrow. If it is not urgent, I can try it tomorrow and get back to you.

Java mgmt SDK is targeting for GA release this week and early next week. Hence I am a bit reluctant to take the risk.

jianghaolu: I can go through all the base & derived types to check any of those if flattened - if so the whole chain will be flattened. 

But yeah I can hold off this for now until later next week.

jianghaolu: FYI in track 1 management: https://github.com/Azure/azure-libraries-for-java/blob/a9cd2b5c6b54dc9e19621d92095694d4ac6a5d2a/azure-mgmt-monitor/src/main/java/com/microsoft/azure/management/monitor/RuleAction.java

weidongxu-microsoft: Track1 works because of the manually added `@JsonTypeResolver(OdataTypeDiscriminatorTypeResolver.class)` which handles the one-way `.` (not remembering which way).
From the code snippet in the issue, the actual problem is the class using the e.g. `RuleAction` got a `@JsonFlatten`.

I will add a test swagger to repo to reproduce the issue.

weidongxu-microsoft: Here it is https://github.com/Azure/autorest.java/pull/722

`MetricAlertResource` has flattened property `MetricAlertProperties`, and `MetricAlertProperties` has property `MetricAlertCriteria`.

It is really hard to see from `MetricAlertCriteria` whether it is used by a type that got flattened (maybe that flattened property nested more than one levels)

jianghaolu:
holding this off for now

PR #725, opened 2020-09-30
Fix paging APIs to include headers and host params in next page call

srnagar:
This PR fixes an issue with Java's generated code forpaging APIs. The generated API that retrieves the first page has params to set the host, headers and other params as specified in the swagger as shown below:

 ```java
@Get("/eventroutes")
@ExpectedResponses({200})
@UnexpectedResponseExceptionType(ErrorResponseException.class)
Mono<Response<EventRouteCollection>> list(
        @HeaderParam("x-ms-max-item-count") Integer maxItemCount,
        @QueryParam("api-version") String apiVersion,
        Context context);
```
 
The issue is with the API that retrieves the next page using the next link. The API only takes the `nextLink` and doesn't include the host param or the headers. 
 
```java
@Get("{nextLink}")
@ExpectedResponses({200})
@UnexpectedResponseExceptionType(ErrorResponseException.class)
Mono<Response<EventRouteCollection>> listNext(
        @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
```
 

For some services, the next link returned from first call does not include the hostname and relies on the header to determine the number of items to return in the next page. So, with the current generated code, the call to get the next page (after manually adding hostname to the returned next link) will include all remaining items because the header cannot be set. Looking at .NET and Python implementation, it seems like the host and header params are included for both first and next page APIs. We are planning to do the same for Java too. This will be done only if x-ms-pageable doesn't have `operationName` defined.

 
After the fix: 

```java
@Get("{nextLink}")
@ExpectedResponses({200})
@UnexpectedResponseExceptionType(ErrorResponseException.class)
Mono<Response<EventRouteCollection>> listNext(
        @HeaderParam("x-ms-max-item-count") Integer maxItemCount,
        @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
```

### DIFF
--- a/azure-tests/src/main/java/fixtures/paging/Pagings.java
+++ b/azure-tests/src/main/java/fixtures/paging/Pagings.java
@@ -202,73 +202,109 @@ public final class Pagings {
         @ExpectedResponses({200})
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<ProductResultValue>> getNoItemNamePagesNext(
-                @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
+                @PathParam(value = "nextLink", encoded = true) String nextLink,
+                @HostParam("$host") String host,
+                Context context);
 
         @Get("{nextLink}")
         @ExpectedResponses({200})
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<ProductResult>> getSinglePagesNext(
-                @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
+                @PathParam(value = "nextLink", encoded = true) String nextLink,
+                @HostParam("$host") String host,
+                Context context);
 
         @Get("{nextLink}")
         @ExpectedResponses({200})
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<ProductResult>> getMultiplePagesNext(
-                @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
+                @PathParam(value = "nextLink", encoded = true) String nextLink,
+                @HostParam("$host") String host,
+                @HeaderParam("client-request-id") String clientRequestId,
+                @HeaderParam("maxresults") Integer maxresults,
+                @HeaderParam("timeout") Integer timeout,
+                Context context);
 
         @Get("{nextLink}")
         @ExpectedResponses({200})
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<OdataProductResult>> getOdataMultiplePagesNext(
-                @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
+                @PathParam(value = "nextLink", encoded = true) String nextLink,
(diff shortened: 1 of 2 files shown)


azure-tests/src/main/java/fixtures/paging/Pagings.java
```diff
@@ -202,73 +202,109 @@
         @ExpectedResponses({200})
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<ProductResultValue>> getNoItemNamePagesNext(
-                @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
+                @PathParam(value = "nextLink", encoded = true) String nextLink,
+                @HostParam("$host") String host,
```
weidongxu-microsoft, line 206: Will azure-core work correctly, when `host` and `nextLink` both contains host part of the URL?

srnagar: @weidongxu-microsoft  yes, this is how the previous version of autorest also generated the code which worked with azure core. So, this should work.

weidongxu-microsoft: I didn't get it. Previous version of autorest generate this
```
        Mono<Response<ResourceGroupListResult>> listNext(
            @PathParam(value = "nextLink", encoded = true) String nextLink, Context context);
```
while updated code had one more `@HostParam("$host") String host`, which certainly is not same?

Well, if tests passes, I assume azure-core could handle this.

azure-tests/src/test/java/fixtures/paging/PagingTests.java
```diff
@@ -109,20 +110,22 @@ public void getMultiplePagesFailureUri() {
             client.getPagings().getMultiplePagesFailureUri().stream().collect(Collectors.toList());
             fail();
         } catch (Exception e) {
-            Assert.assertTrue(e.getCause() instanceof MalformedURLException);
+            assertTrue(e instanceof HttpResponseException);
```
weidongxu-microsoft, line 113: Any idea what caused this change? Is it due to the additional `host`?

weidongxu-microsoft (approved):

azure-tests/src/main/java/fixtures/paging/Pagings.java
weidongxu-microsoft: nit. Will it be better to put `@HostParam` before `@PathParam`? Not important since this method will not be exposed to user.

srnagar: Sorry, I meant the v3 version of autorest. Here's an example of what was generated by v3.

```java
    @Get("{nextUrl}")
    @ExpectedResponses({200})
    @ReturnValueWireType(ConfigurationSettingPage.class)
    Mono<PagedResponse<ConfigurationSetting>> listKeyValues(
        @HostParam("url") String url,
        @PathParam(value = "nextUrl", encoded = true) String nextUrl,
        Context context);
```

PR #747, opened 2020-10-23
v4, simple solution for naming conflict with azure-core classes

weidongxu-microsoft:


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -66,32 +66,47 @@ public static JavaSettings getInstance()
...
                     80,
                     host.getStringValue("serviceName"),
                     host.getStringValue("namespace", "").toLowerCase(),
                     host.getBooleanValue("enable-xml", false),
                     host.getBooleanValue("non-null-annotations", false),
-                    host.getBooleanValue("client-side-validations", false),
+                    host.getBooleanValue("client-side-validations", clientSideValidationsDefault),
                     host.getStringValue("client-type-prefix"),
-                    host.getBooleanValue("generate-client-interfaces", true),
-                    host.getBooleanValue("generate-client-as-impl", true),
```
weidongxu-microsoft, line 83: Previously the `defaultValue` for `getBooleanValue` will never work. The underlying `getBooleanValue` will never return `null`.

weidongxu-microsoft:
Test pass, but result not synced to PR, likely a devops bug.

https://dev.azure.com/azure-sdk/internal/_build?definitionId=1633&_a=summary
https://dev.azure.com/azure-sdk/public/_build?definitionId=1634&_a=summary

PR #749, opened 2020-10-25
Generate JsonPatchDocument if content type is json-patch+json

srnagar:
This PR generates APIs that will use `JsonPatchDocument` type introduced in `azure-core` version 2.10.0 if the REST operations have content-type set to `json-patch+json` and the body param is of type `List<Object>`.

### DIFF
--- a/vanilla-tests/pom.xml
+++ b/vanilla-tests/pom.xml
@@ -35,7 +35,7 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.10.0-alpha.20201022.0</version>
+      <version>1.5.0</version>
     </dependency>
     <dependency>
       <groupId>com.azure</groupId>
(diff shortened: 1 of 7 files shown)


vanilla-tests/pom.xml
```diff
@@ -14,11 +14,28 @@
...
+      </snapshots>
+    </repository>
+  </repositories>
+
   <dependencies>
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
-      <version>1.5.0</version>
+      <version>1.10.0-alpha.20201022.0</version>
```
weidongxu-microsoft, line 38: Upgrading 1.5.0 to 1.9.0 actually breaks a few tests (likely due to changed azure-core behavior on `null` or something).

weidongxu-microsoft (commented):
It would help for review if we know which part of `CustomParameterMapper` is different from the default one.

weidongxu-microsoft (approved):

vanilla-tests/pom.xml
srnagar: Yes, there are some changes to azure-core to use the UTC timezone. I will update the tests in a follow-up PR to fix this. For now, these tests are not related to the changes in the PR and have reverted the test dependency to use 1.5.0.

PR #761, opened 2020-10-29
v4 fluent override credential type to string

weidongxu-microsoft:
swagger is something like this
```
        "administratorLoginPassword": {
          "type": "string",
          "format": "password",
          "description": "The password of the administrator login."
        }
```


ChenTanyi (approved):
Maybe could also add comment about why change CREDENTIAL to STRING

weidongxu-microsoft:
Added some comment.

Issue #748, opened 2020-10-24
Add override methods to Fluent class from the extended class


mssfang:
Currently the autorest doesn't generate Override methods. For example,
```
@Fluent
public final class A extends B {   
 /**
   * With override method been generated, I can easily A a = new A().setMethod(true). 
   * Without override the setMethod(boolean enable) method, we can only do A a = new A(); a.setMethod(true);
   *    @Override
   *    public A setMethod(boolean enable) {
   *          super.setMethod(enable);
   *          return this;
   *     }
   **/

}

@Fluent
public class B{
   public B setMethod(boolean enable) {
        this.enable = enable;
        return this;
    }
}
```

Not sure if autorest will add it since there is a tradeoff that could add too many override methods, but it increases SDK developer's experience. Just want to bring the concern up for attention.

weidongxu-microsoft:
All, let me know the general opinion. I could try to add a setting parameter to do this for both data and mgmt (mgmt could use a similar solution, `location` is usually in base class `Resource`, which causes some inconvenience in use).

PR #765, opened 2020-11-02
v4 model override parent setter

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/748

add option `model-override-setter-from-superclass`


vanilla-tests/src/main/java/fixtures/bodycomplex/models/SmartSalmon.java
```diff
@@ -86,6 +87,34 @@ void setAdditionalProperties(String key, Object value) {
...
+    /** {@inheritDoc} */
+    @Override
+    public SmartSalmon setSpecies(String species) {
+        super.setSpecies(species);
+        return this;
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public SmartSalmon setSiblings(List<Fish> siblings) {
```
weidongxu-microsoft, line 113: sample of generated code

PR #782, opened 2020-11-09
[Autorest.Android]: Add Body-Integer Test with test server

JianpingChen:
Set up Android test framework
Add tests with sync clients for body-string and body-integer


android-tests/app/src/androidTest/java/com/azure/autoresttest/BodyStringSyncClientTests.java
```diff
@@ -0,0 +1,135 @@
...
+        Assert.assertNull(client.getNullWithRestResponse().getValue());
+    }
+
+    @Test
+    public void putNull() throws Exception {
+        try {
+            Response<Void> putResult = client.putNullWithRestResponse(null);
+            assertEquals(200, putResult.getStatusCode());
+        } catch (Exception ex) {
+            Assert.assertEquals(NullPointerException.class, ex.getClass());
```
anuchandy, line 40: looks like we are expecting the `putNullWithRestResponse` call to succeeded with 200, if so do we need the try block for NPE?

android-tests/app/src/androidTest/java/com/azure/autoresttest/BodyStringSyncClientTests.java
```diff
@@ -0,0 +1,135 @@
...
+    public void getNullBase64UrlEncoded() throws Exception {
+        Base64Url result = client.getNullBase64UrlEncodedWithRestResponse().getValue();
+        Assert.assertNull(result);
+    }
+
+    @Test
+    public void putBase64UrlEncoded() throws Exception {
+        try {
+            Response<Void> putResult = client.putBase64UrlEncodedWithRestResponse(new Base64Url("http://myhost.com/path?email=a.b@c.com&value=a%20B"));
+            assertEquals(200, putResult.getStatusCode());
```
anuchandy, line 129: the same question as above, we're expecting this to succeed, so try block may not be needed.

anuchandy (approved):
nice job on setting up the FX 👍.  one minor comment, lgtm.

PR #794, opened 2020-11-17
v4 fluentlite, bug fixes

weidongxu-microsoft:


extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Relations.java
```diff
@@ -3,23 +3,23 @@
 import java.util.List;
 
 public class Relations {
-    private List<ComplexSchema> all;
+    private List<Schema> all;
 
-    private List<ComplexSchema> immediate;
+    private List<Schema> immediate;
```
weidongxu-microsoft, line 8: It actually get a `AnySchema` in it, in this spec (though for me, that definition does not make much sense) https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cloudshell/resource-manager/Microsoft.Portal/stable/2018-10-01/CloudShell.json#L789-L809

weidongxu-microsoft: The change should not affect existing logic. For child class, it definitely still need to be `ObjectSchema`. For parent class, `AnySchema` will be ignored. Currently for parent class, `ObjectSchema` is good, `DictionarySchema` will be convert to additionalProperties.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -75,9 +76,14 @@ public static ClientMethodMapper getInstance() {
...
             ClientModel responseBodyModel = Mappers.getModelMapper().map((ObjectSchema) responseBodySchema);
-            IType listType = responseBodyModel.getProperties().stream()
+            Optional<ClientModelProperty> itemPropertyOpt = responseBodyModel.getProperties().stream()
                     .filter(p -> p.getSerializedName().equals(operation.getExtensions().getXmsPageable().getItemName()))
-                    .findFirst().get().getWireType();
+                    .findFirst();
+            if (!itemPropertyOpt.isPresent()) {
+                throw new IllegalArgumentException(String.format("[JavaCheck/SchemaError] item name %s not found among properties of client model %s",
+                        operation.getExtensions().getXmsPageable().getItemName(), responseBodyModel.getName()));
+            }
```
weidongxu-microsoft, line 85: This is spec error. Some pageable in spec does not have "value" property.

javagen/src/main/java/com/azure/autorest/util/SchemaUtil.java
```diff
@@ -90,13 +89,17 @@ public static String getDiscriminatorSerializedName(ObjectSchema compositeType)
...
+                if (parent instanceof ObjectSchema && ((ObjectSchema) parent).getDiscriminator() != null) {
                     discriminator = ((ObjectSchema) parent).getDiscriminator().getProperty().getSerializedName();
                     break;
                 }
             }
         }
+        if (discriminator == null) {
+            throw new IllegalStateException(String.format("discriminator not found in type %s and its parents",
+                    compositeType.getLanguage().getJava().getName()));
+        }
```
weidongxu-microsoft, line 102: This appear to be m4 problem. Spec put the discriminator property to a parent class.
https://github.com/Azure/azure-rest-api-specs/blob/master/specification/security/resource-manager/Microsoft.Security/stable/2020-01-01/externalSecuritySolutions.json#L180-L199

PR #803, opened 2020-11-19
v4 fluentlite bug fixes

weidongxu-microsoft:


javagen/src/main/java/com/azure/autorest/model/xmlmodel/XmlBlock.java
```diff
@@ -42,6 +44,6 @@ public final void block(String text, Consumer<XmlBlock> bodyAction) {
     }
 
     public final void tagWithInlineComment(String tag, String value, String comment) {
-        contents.line("<%1$s>%2$s</%1$s> <!-- %3$s -->", tag, value, comment);
+        contents.line("<%1$s>%2$s</%1$s> <!-- %3$s -->", tag, CodeNamer.escapeXmlComment(value), CodeNamer.escapeXmlComment(comment));
```
weidongxu-microsoft, line 47: I think the value need to be escaped uniformly.

javagen/src/test/java/com/azure/autorest/util/CodeNamerTests.java
```diff
@@ -29,5 +29,9 @@ public void testEnumMemberName() {
         Assert.assertEquals("ALL", CodeNamer.getEnumMemberName("all*"));
 
         Assert.assertEquals("SYSTEM_ASSIGNED_USER_ASSIGNED", CodeNamer.getEnumMemberName("SystemAssigned, UserAssigned"));
+
+        Assert.assertEquals("ONE_ZEROMINUTELY", CodeNamer.getEnumMemberName("_10minutely"));
```
weidongxu-microsoft, line 33: Not a pretty sight, but it better than previously `10MINUTELY` in value.

PR #804, opened 2020-11-19
Support customization code in README

jianghaolu:
Now you can add a section in readme to achieve customization like this:

### Customization
```java
public class BodyComplexCustomization extends Customization {
    @Override
    public void customize(LibraryCustomization customization) {
        PackageCustomization implementationModels = customization.getPackage("fixtures.bodycomplex.implementation.models");
        implementationModels.renameClass("Goblinshark", "GoblinShark");
        implementationModels.getClass("DotSalmon")
                .renameProperty("iswild", "isWild")
                .renameMethod("setIsWild", "setWild")
                .changeMethodReturnType("isWild", "boolean", "%s")
                .changeMethodReturnType("setWild", "void", null);
        implementationModels.getClass("CMYKColors")
                .renameEnumMember("CYAN", "TEAL");
        ClassCustomization readonlyObj = implementationModels.getClass("ReadonlyObj")
                .generateGetterAndSetter("id")
                .changeMethodReturnType("getId", "UUID", "UUID.fromString(%s)")
                .changeMethodModifier("setId", "");
        readonlyObj.methodJavadoc("getId").setDescription("Get the ID of the object.");
        readonlyObj.methodJavadoc("setId").setDescription("Set the ID of the object.")
                .setReturn("The current ReadonlyObj instance")
                .setParam("id", "The ID value");
        PackageCustomization root = customization.getPackage("fixtures.bodycomplex");
        ClassCustomization arrayClient = root.getClass("ArrayClient")
                .changeMethodReturnType("putValid", "ArrayClient", "this");
        arrayClient.classJavadoc().setDescription("The sync client containing Array operations.");
        arrayClient.methodJavadoc("putValid").setReturn("The ArrayClient itself");
    }
}

### CODE REVIEW

**Review by weidongxu-microsoft — APPROVED** [review]

File: customization-tests/swagger/README.md [hunk 1]
```diff
@@ -33,6 +33,40 @@ license-header: MICROSOFT_MIT_SMALL
...
 models-subpackage: implementation.models
 context-client-method-parameter: true
-customization-jar-path: target/bodycomplex-customization-1.0.0-beta.1.jar
-customization-class: fixtures.bodycomplex.customization.BodyComplexCustomization
+customization-class: BodyComplexCustomization
 ```
+
+### Customization
+```java
+public class BodyComplexCustomization extends Customization {
```
srnagar, line 41: Would the user be able to write this code in a README without the help of compilable code?

customization-tests/swagger/README.md
```diff
@@ -33,6 +33,40 @@ license-header: MICROSOFT_MIT_SMALL
...
+    @Override
+    public void customize(LibraryCustomization customization) {
+        PackageCustomization implementationModels = customization.getPackage("fixtures.bodycomplex.implementation.models");
```
srnagar, line 44: How does the user specify import statements if this code requires some additional classes that have to be imported?

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -26,16 +40,25 @@ public Postprocessor(Connection connection, String plugin, String sessionId) {
...
+      return false;
+    }
+
     try {
       //Step 1: post process
-      String jarPath = JavaSettings.getInstance().getCustomizationJarPath();
-      String className = JavaSettings.getInstance().getCustomizationClass();
-      if (jarPath != null && className != null) {
+      Class<? extends Customization> customizationClass;
+      if (jarPath != null) {
```
weidongxu-microsoft, line 61: Could there be a security risk if jarUrl can be loaded from remote?

customization-tests/swagger/README.md
jianghaolu: The users don't have to specify the imports if they are not ambiguous. Otherwise they could write `import aa.bbb.ccc;" above the class definition. The code here is simply injected into a maven project that has `customization-base` as a dependency.

customization-tests/swagger/README.md
jianghaolu: Without an IDE it's definitely harder to write this. But that's a tradeoff users will have to make. They can opt for easier customization coding but have to set up a maven project, or opt for easier setup but lose IDE features.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
jianghaolu: Can you elaborate on the risk here? It can be loaded from anywhere over the Internet, but it's passed in by the user and it's the user's responsibility to make sure they don't load random jars from untrusted sources.

weidongxu-microsoft: Actually, I do not exactly know. Theoretically someone can modify the url hiding it among other changes in a submitted PR, trick dev to test the PR and hence run the jar, and the jar would grab env and pass file from local disk, upload them to remote site. Will it happen? No idea. The risk probably is not significant at this stage. Does it worth some mitigation e.g. SecurityManager, or just disable remote jar? Not sure.

jianghaolu: I think we should phase out passing in a separate jar - and use either the readme or a central location in the repo. So I'll not do too much here

PR #808, opened 2020-11-22
Enable customizing the method body when changing return type

srnagar:
When customizing a method's return type, there are cases where we may need to replace the implementation completely and not just format the return value.  This PR enables replacing the entire method with custom logic. 

In the example below, the original generated code returns a `String` value. If the return type has to be changed to a `byte[]` with a special encoding, the `getBytes()` in `String` class throws a checked `UnsupportedEncodingException`. With just the formatter, changing the return type will not be possible as the try-catch block cannot be included in `return String.format(formattedString, returnValue);`

Before:
```java
public String getColor() {
     return this.color;
}
```

After:
```java
public byte[] getColor() {
    String returnValue = this.color;
    String colorStr = returnValue;
    try {
        return colorStr.getBytes("UTF-8");
    } catch (UnsupportedEncodingException ex) {
        return colorStr.getBytes();
    }
}
```


jianghaolu (approved):

jianghaolu:
Thanks!

PR #810, opened 2020-11-24
v4, add common class/annotation of azure-core to reserved words

weidongxu-microsoft:
It might not be all. I didn't add less used ones like `HeaderCollection` and `ResumeOperation`, or more confined ones like `ServiceClient` and `ServiceClientBuilder`.

### DIFF
--- a/preprocessor/src/main/java/com/azure/autorest/preprocessor/namer/CodeNamer.java
+++ b/preprocessor/src/main/java/com/azure/autorest/preprocessor/namer/CodeNamer.java
@@ -5,9 +5,11 @@ import org.atteo.evo.inflector.English;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft:
Another complete approach would be pre-compile the list of all azure-core classes from azure-core target/classes. But I guess we might not really need this...

preprocessor/src/main/java/com/azure/autorest/preprocessor/namer/CodeNamer.java
```diff
@@ -226,6 +240,17 @@ protected static String getEscapedReservedName(String name, String appendValue)
         return name;
     }
 
+    protected static String getEscapedReservedNameAndClasses(String name, String appendValue) {
```
weidongxu-microsoft, line 243: I put a new RESERVED_WORDS_CLASSES set for types.

There class/annotation names are fine for parameter/proerty.

PR #812, opened 2020-11-25
[Autorest.Android]: Test error response

JianpingChen:


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -193,6 +193,10 @@ protected void buildUnexpectedResponseExceptionTypes(ProxyMethod.Builder builder
         }
 
         if (errorType != null) {
+            if (errorType.equals(ClassType.String)){
+                builder.unexpectedResponseExceptionType(getExceptionTypeForString(errorType));
+                return;
+            }
```
anuchandy, line 199: What makes android different from Java which requires special handling for string type?

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -71,6 +71,7 @@ public Client map(CodeModel codeModel) {
                 .flatMap(og -> og.getOperations().stream())
                 .flatMap(o -> o.getExceptions().stream())
                 .map(Response::getSchema)
+                .filter(s -> s instanceof ObjectSchema)
```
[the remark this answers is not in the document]
JianpingChen: Because there is a down case two lines after which will fail when Response:getSchema returns String.

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
JianpingChen: On Android we'd throw this exception.

> com.azure.android.core.http.exception.HttpResponseException

For normal Java, it'll throw

> com.azure.core.exception.HttpResponseException

anuchandy (approved):

PR #819, opened 2020-11-30
Support credential-scopes

jianghaolu:


weidongxu-microsoft (approved):
Does data-plane has any scope other than `.default`?

jianghaolu:
I'm not aware of anything else at this moment. Azure's authentication is behind other orgs in Microsoft, e.g. O365, etc who are taking advantage of other scopes.

PR #820, opened 2020-12-01
Add dummy javadocs to client parameters & builder ctor

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ClientParameterMapper.java
```diff
@@ -52,15 +52,22 @@ public ClientMethodParameter map(Parameter parameter) {
...
         builder.isConstant(isConstant).defaultValue(defaultValue);
 
-        if (parameter.getSchema() != null && parameter.getSchema().getLanguage() != null) {
-            String description = parameter.getSchema().getLanguage().getDefault().getDescription();
-            if (description == null || description.isEmpty()) {
-                if (parameter.getLanguage() != null) {
-                    description = parameter.getLanguage().getDefault().getDescription();
-                }
+        String description = null;
+        // parameter description
```
jianghaolu, line 56: The original logic prioritizes the parameter schema's description over the parameter's own description. I feel like it should be the other way around

weidongxu-microsoft: Agree that the parameter's description should take precedence.

weidongxu-microsoft (approved):

PR #821, opened 2020-12-01
Address API review feedback for customization-base

jianghaolu:


jianghaolu:
Uploading the sources jar to apiview for review and will make further changes as needed

PR #833, opened 2020-12-08
Change group ID to com.azure.autorest and add missing info in pom.xml

jianghaolu:
This is group everything in com.azure.autorest for better organization.


jianghaolu:
After a brief discussion we will rename the artifact IDs instead.

PR #856, opened 2020-12-16
Rename Maven packages and move to com.azure.tools

jianghaolu:


azure-tests/pom.xml
```diff
@@ -5,13 +5,13 @@
...
-    <artifactId>autorest-java</artifactId>
+    <groupId>com.azure.tools</groupId>
+    <artifactId>azure-autorest-parent</artifactId>
     <version>1.0.0-beta.1</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
-  <groupId>com.azure</groupId>
-  <artifactId>autorest-azure-tests</artifactId>
+  <groupId>com.azure.tools</groupId>
+  <artifactId>azure-autorest-azure-tests</artifactId>
```
srnagar, line 14: It's a bit odd to have `azure` twice in the name. Are there any other names we can use here instead of azure for these tests?

srnagar (commented):
Do we plan to ship all these modules or just the `azure-autorest-customization` module to Maven?

azure-tests/pom.xml
jianghaolu: hmm I don't think it's weird but we can rename to `azure-autorest-arm-tests` since the autorest argument to generate these is `--azure-arm`

srnagar: Yeah, I like `arm-tests` better than `azure-tests` because it's clear that these are management related tests. However, we have started using `resourcemanager` as the new name for `arm` in track 2 libraries. So, another option is `azure-autorest-resourcemanager-tests`

weidongxu-microsoft: `arm` should be fine. Anyway if we do not release this package, we would have chance to change it later.

weidongxu-microsoft (approved):
LGTM

srnagar (approved):
The new names look good to me. Tagging @JonathanGiles to take a look at the updated names and approve.

PR #863, opened 2020-12-23
docs

iscai-msft:
@jianghaolu is doing the dev section firsts, will update the PR with this section later.

I couldn't figure out how to add a samples section like [python](https://github.com/Azure/autorest.python/tree/9839bf3392d85b4661c0b1249634a6b9f1f9706b/docs/samples) has, will try to do so in the new year.

Finally, left faq and the migrate section mostly blank, (here's the [python faq](https://github.com/Azure/autorest.python/blob/9839bf3392d85b4661c0b1249634a6b9f1f9706b/docs/faq.md), and the [python migrate](https://github.com/Azure/autorest.python/tree/9839bf3392d85b4661c0b1249634a6b9f1f9706b/docs/migrate)), lmk what your breaking changes / new features are from the previous c# generator, will do this in the new year too. 

also a lot these links are probably broken (all of the ones referring to the main docs)


docs/client/initializing.md
```diff
@@ -0,0 +1,81 @@
...
+}
+```
+
+You can also install your client with a credential, using the flag `--credential-types`. If you generate with `--credential-types=TokenCredential`, your
+client will take in an [Azure Active Directory (AAD) token credential][aad_authentication]. We always recommend
+using a [credential type][identity_credentials] obtained from the [`com.azure.identity`][azure_identity_library] package for AAD authentication. For this example,
+we use the most common [`DefaultAzureCredential`][default_azure_credential].
+
+As an installation note, the [`com.azure.identity`][azure_identity_library] is not listed as a dependency in the `pom` we generate
+(see `--regenerate-pom` in our [flag index][flag_index] for more information), so you would need to install this library separately.
```
weidongxu-microsoft, line 38: @jianghaolu What does `regenerate-pom` intend to do? Does it decide whether pom.xml is generated or not? If yes, I could use this in fluent to skip pom in Lite.

I think currently data-plane does not generate pom.xml.

docs/client/operations.md
```diff
@@ -0,0 +1,228 @@
...
+
+Long-running operations are operations which consist of an initial request sent to the service to start an operation, followed by polling the service at intervals to determine whether the operation has completed or failed, and if it has succeeded, to get the result.
+
+In concurrence with our [java guidelines][poller_guidelines], all of our long running operations are prefixed with `begin`, to signify the starting of the long running operation.
+
+For our example, we will use the long running operation generated from [this][example_swagger] swagger. Let's say we generated this swagger with namespace `com.azure.lro`.
+
+### Sync Long Running Operations
+
+Our sync long running operations return a [`SyncPoller`][lro_poller] polling object. Calling `.getFinalResult()` on this poller
```
weidongxu-microsoft, line 73: Right now, we probably do not yet support LRO in data-plane code generator.

docs/client/initializing.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+Currently, we only support generating credentials of type `TokenCredential` and / or `AzureKeyCredential`.
+
+<!-- LINKS -->
+[multiapi_generation]: ../generate/multiapi.md
+[azure_core_library]: https://pypi.org/project/azure-core/
+[msrest_library]: https://pypi.org/project/msrest/
+[azure_mgmt_core_library]: https://pypi.org/project/azure-mgmt-core/
+[azure_identity_library]: https://docs.microsoft.com/en-us/java/api/com.azure.identity?view=azure-java-stable
```
weidongxu-microsoft, line 76: One of the recommendation is to avoid `en-us` in msdoc link. It would improve i18n.

docs/samples/specification/azure_key_credential/readme.md
```diff
@@ -0,0 +1,16 @@
+# Sample Azure Key Credential Generation
+
+Use the flag `--credential-types` to specify you want your credential to be of type [`AzureKeyCredential`][azure_key_credential].
+
+### Settings
+
+``` yaml
+input-file: ../../../../node_modules/@microsoft.azure/autorest.testserver/swagger/head.json
```
weidongxu-microsoft, line 8: Maybe better just use github ULR for the json?

`../../../../node_modules/@microsoft.azure` would likely not be there until user do "npm install".

weidongxu-microsoft (approved):
LGTM in general.

I could fill fluent part later.

docs/client/initializing.md
```diff
@@ -0,0 +1,81 @@
+# <img align="center" src="../images/logo.png">  Initializing Your Java Client
+
+The first step to using your generated client in code is to import and initialize your client. Our SDKs are modelled such
+that the client is the main point of access to the generated code.
+
+## Importing Your Client
+
+You include the package from the namespace specified when generating (under flag `--namespace`). For the sake of this example,
```
jianghaolu, line 8: We don't actually require importing the client manually in the code so I would use this section to specify how to import the maven dependency.

jianghaolu: Something like:

You include the generated client through Maven. In your application's `pom.xml` add a dependency with `groupId`, `artifactId`, and `version` specified in the generated client's `pom.xml` file. For the sake for this example, let's say the `groupId` is the same as the package name specified when generating (under flag `--namespace`), the `artifactId` is `azure-pets`, and the version is `1.0.0-beta.1`.
```xml
<dependencies>
  <dependency>
    <groupId>com.azure.pets</groupId>
    <artifactId>azure-pets</artifactId>
  </dependency>
</dependencies>
```

docs/client/initializing.md
jianghaolu: For now we've only used this flag for fluent lite and it overwrites the current pom if there's any.

docs/client/models.md
```diff
@@ -0,0 +1,50 @@
+# <img align="center" src="../images/logo.png">  Accessing Models and Enums
+
+## General
+
+Models and enums are generated in the `models` namespace. So, say you are using package `azure.pets`. To access model `Dog`, you would use the following code
```
jianghaolu, line 5:
```suggestion
Models and enums are generated in the `models` sub-package. So, say you are using package `azure.pets`. To access model `Dog`, you would use the following code snippet
```

docs/client/models.md
```diff
@@ -0,0 +1,50 @@
...
+snippet
```
jianghaolu, line 6: Why is this on a separate line?

docs/client/models.md
```diff
@@ -0,0 +1,50 @@
...
+
+```java
+import azure.pets.models.Dog
+```
```
jianghaolu, line 10: Imports in Java IDEs are automatic - we can remove any instructions to import.

docs/client/models.md
```diff
@@ -0,0 +1,50 @@
...
+Enums are also listed in the `models` namespace, so say you have enum class `DogTypes`. To access the `DALMATION` enum, your code would look like
+
+```java
+package com.azure.pets;
+
+import azure.pets.models.DogTypes;
+
+public static void main(String args[])
+{
+    String myDogType = DogTypes.DALMATION;
```
jianghaolu, line 48: In Java the return type would be `DogTypes` even for expandable enums.

docs/client/operations.md
```diff
@@ -0,0 +1,228 @@
...
+import com.azure.pets.models.Dog;
+
+public static void main(String args[])
+{
+    PetsAsyncClient client = new PetsClientBuilder()
+        .credential(new DefaultAzureCredentialBuilder().build())
+        .buildAsyncClient();
+
+    client.getDog().subscribe(
+        d -> dog = d
```
jianghaolu, line 58: Where is `dog` defined? I guess we could do a simple `System.out.println()` on one of the `d`'s properties.

docs/client/operations.md
jianghaolu: That's correct. We do not return a poller on any of the generated clients as the `x-ms-long-running-operation` does not contain enough information how the polling should behave for a data plane API.

docs/client/operations.md
```diff
@@ -0,0 +1,228 @@
...
+    );
+}
+```
+
+## Paging Operations
+
+A paging operation pages through lists of data, returning an iterator for the items. Network calls get made when users start iterating through the output, not when the operation
+is initially called.
+
+For our example, we will use the long running operation generated from [this][example_swagger] swagger. Let's say we generated this swagger with namespace `com.azure.paging`.
```
jianghaolu, line 160:
```suggestion
For our example, we will use the pageable operation generated from [this][example_swagger] swagger. Let's say we generated this swagger with namespace `com.azure.paging`.
```

jianghaolu:
Looks good overall. Great job!

docs/client/operations.md
iscai-msft: I see, I had my. suspicions but wanted to get some LRO docs out as well. I can remove them until you guys add LRO support, also lmk if you have questions about how python does it (basically we inspect the first response to figure out where to find the status / result)

docs/samples/specification/azure_key_credential/readme.md
iscai-msft: Sounds good, haven't actually started running this code, but good point

docs/client/initializing.md
iscai-msft: thanks so much!

docs/client/models.md
iscai-msft: I think I just did it to make it cleaner looking at the raw markdown file. Doesn't render any differently in actual markdown, but I'll get rid of the separate line bc it looks a little weird

docs/client/operations.md
weidongxu-microsoft: ARM part of LRO is more consistent and can be handled by status code and headers like `Azure-AsyncOperation` and `Location` (I guess this is what python does), and hence it only exists in `fluent` flavor as it is for ARM. I guess data-plane could be in different and not very consistent format.

docs/samples/readme.md
```diff
@@ -0,0 +1,21 @@
+# <img align="center" src="../images/logo.png"> Sample Python Generation
```
weidongxu-microsoft, line 1: Would be `Sample Java Generation`. Easy to change later though.

docs/client/operations.md
jianghaolu: For now you can stash this section and put something like "The Java team is building the poller generation - will update when it's available."

jianghaolu (approved):
:shipit:

Issue #880, opened 2021-01-08
[BUG] Generator generates wrong JsonProperty annotation when xml enabled


gapra-msft:
Enable xml in swagger readme and run generator on 
https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.QueueStorage/preview/2018-03-28/queue.json

Code generates to something like this 
```
    @JsonProperty(value = "Metrics")
    private QueueMetrics hourMetrics;
```

It should be generating to the following
```
    @JsonProperty(value = "HourMetrics")
    private QueueMetrics hourMetrics;
```

PR #894, opened 2021-01-14
Fix customization check

srnagar:
This PR fixes an issue where the customization post-processor will fail if the customization class, jar and readme don't exist. The change is to check if the `customization-class` property is set when running the generator and skip customization if it's not set.

Fixes https://github.com/Azure/autorest.java/issues/880

### DIFF
--- a/fluentgen/pom.xml
+++ b/fluentgen/pom.xml
@@ -15,7 +15,7 @@
   <version>1.0.0-beta.1</version>
 
   <properties>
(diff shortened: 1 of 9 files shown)


weidongxu-microsoft (approved):

jianghaolu (approved):

jianghaolu:
Looks like you need to regenerate?

PR #900, opened 2021-01-19
v4, update m4 to 4.15.447 for javagen

weidongxu-microsoft:


customization-tests/src/main/java/fixtures/bodycomplex/implementation/ArraysImpl.java
```diff
@@ -50,33 +51,38 @@
...
+                @HeaderParam("Accept") String accept,
                 Context context);
 
         @Get("/complex/array/notprovided")
         @ExpectedResponses({200})
         @UnexpectedResponseExceptionType(ErrorException.class)
-        Mono<Response<ArrayWrapper>> getNotProvided(@HostParam("$host") String host, Context context);
+        Mono<Response<ArrayWrapper>> getNotProvided(
+                @HostParam("$host") String host, @HeaderParam("Accept") String accept, Context context);
     }
```
weidongxu-microsoft, line 86: Let me know whether this change will affect data-plane.

m4 added feature to fill-in `Accept` header, probably based on `produces` field in swagger.

PR #910, opened 2021-01-27
Make service version configurable and fix property name collision

### LINKED ISSUE #889: [Feature Request] Add ability to set version on generated impl builders
This was supported previously and allows users to set custom versions on requests. 

Currently sidestepping the issue by creating a custom version PipelinePolicy.

### LINKED ISSUE #902: [BUG] Generator is not creating a new variable resulting in uncompilable code
Using this swagger file

https://raw.githubusercontent.com/gapra-msft/azure-rest-api-specs/storage/queueGenerator/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-04-08/blob.json

And this swagger README

https://raw.githubusercontent.com/Azure/azure-sdk-for-java/c708ca0cf105043411598b20b7668511e0631e45/sdk/storage/azure-storage-blob/swagger/README.md

The generated code is not compilable. Basically the encryptionScope variable seems to be used twice in a bunch of the generated APIs

srnagar:
This PR makes a couple of changes:
- enables configuring service versions through the builder (fixes #889 )
- Fixes compilation failure of generated code when the parameter group name is the same as the proxy method arg name (fixes #902 )

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
@@ -109,14 +109,16 @@ public class ServiceClientMapper implements IMapper<CodeModel, ServiceClient> {
                 serviceClientPropertyClientType = serviceClientPropertyClientType.asNullable();
             }
 
-            boolean serviceClientPropertyIsReadOnly = p.getSchema() instanceof ConstantSchema;
+            // boolean serviceClientPropertyIsReadOnly = p.getSchema() instanceof ConstantSchema;
 
             String serviceClientPropertyDefaultValueExpression = serviceClientPropertyClientType.defaultValueExpression(p.getClientDefaultValue());
 
             if (serviceClientPropertyClientType == ClassType.TokenCredential) {
                 usesCredentials = true;
             } else {
-                ServiceClientProperty serviceClientProperty = new ServiceClientProperty(serviceClientPropertyDescription, serviceClientPropertyClientType, serviceClientPropertyName, serviceClientPropertyIsReadOnly, serviceClientPropertyDefaultValueExpression);
+                ServiceClientProperty serviceClientProperty =
+                        new ServiceClientProperty(serviceClientPropertyDescription, serviceClientPropertyClientType,
+                                serviceClientPropertyName, false, serviceClientPropertyDefaultValueExpression);
                 if (!serviceClientProperties.contains(serviceClientProperty)) {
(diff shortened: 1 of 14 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
```diff
@@ -109,14 +109,16 @@ public ServiceClient map(CodeModel codeModel) {
                 serviceClientPropertyClientType = serviceClientPropertyClientType.asNullable();
             }
 
-            boolean serviceClientPropertyIsReadOnly = p.getSchema() instanceof ConstantSchema;
+            // boolean serviceClientPropertyIsReadOnly = p.getSchema() instanceof ConstantSchema;
```
weidongxu-microsoft, line 112: Maybe you can set it to false only for `!setting.isFluent()`?

Mgmt won't want user to set the `api-verison`, they are not going to be compatible.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -552,7 +557,20 @@ protected void generatePagedAsyncSinglePage(ClientMethod clientMethod, JavaType
...
+                        .stream()
+                        .map(argVal -> {
+                            if(clientMethod.getParameters().stream().filter(param -> param.getName().equals(argVal))
+                                    .anyMatch(param -> clientMethod.getMethodTransformationDetails().stream()
+                                            .anyMatch(transformation -> param.getName().equals(transformation.getOutParameter().getName())))) {
+                                return argVal + "Local";
+                            }
+                            return argVal;
+                        })
+                        .collect(Collectors.toList());
```
weidongxu-microsoft, line 571: Is all these duplicate code? If so can it be moved to a shared method?

weidongxu-microsoft (approved):

jianghaolu (approved):
LGTM from data plane's perspective. But like Weidong mentioned, the mgmt library may not want API version to settable in fluent.

javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
srnagar: Done.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
srnagar: Moved to a method.

PR #912, opened 2021-01-27
Add more annotation customizations

jianghaolu:


customization-base/src/main/java/com/azure/autorest/customization/PropertyCustomization.java
```diff
@@ -73,6 +73,108 @@ public PropertyCustomization rename(String newName) {
...
+        return this;
+    }
+
+    /**
+     * Remove an annotation from the property.
+     *
+     * @param annotation the annotation to remove from the property. The leading @ can be omitted.
+     * @return the current property customization for chaining
+     */
+    public PropertyCustomization removeAnnotation(String annotation) {
```
srnagar, line 127: Will this remove the annotation from the constructor too if the property had an annotation that was also used in the ctor args?

```java
    @JsonProperty(value = "length", required = true)
    private float length;

    @JsonCreator
    public Fish(@JsonProperty(value = "length", required = true) float length) {
        this.length = length;
    }
```

srnagar:
For this change, do you also want to update the pom version for the customization library as this needs to be released to Maven.

customization-base/src/main/java/com/azure/autorest/customization/PropertyCustomization.java
jianghaolu: The user will have to manually remove the annotation on the cotr. There may be future convenience customizations to "make a property optional".

jianghaolu:
I have another small fix coming - I'll bump the pom version there.

PR #938, opened 2021-02-25
v4 fluent, support SystemData from azure-core-management

weidongxu-microsoft:
Need azure-core-managment 1.2.0


fluent-tests/src/test/java/com/azure/mgmttest/RuntimeTests.java
```diff
@@ -12,6 +12,7 @@
 import com.azure.core.http.rest.PagedIterable;
 import com.azure.core.management.AzureEnvironment;
 import com.azure.core.management.Region;
+import com.azure.core.management.ResourceAuthorIdentityType;
```
yungezz, line 15: Thanks @weidongxu-microsoft to support it. Actually it's more natural to align with Swagger `createdByType` on this,  concern is future extensibility and consistent. no idea whether will createByType will be identityType always etc. and same names cross Azure will make it easy for user to understand.

weidongxu-microsoft: "IdentityType" might be a bad name in future.. My concern here is that "CreateByType" is a bad name, **right now**, as it is returned by "lastModifiedBy".

weidongxu-microsoft: And "CreateByType" is not even consistent across RPs. Some calls it "IdentityType" (CDN), some calls it "LastModifiedByType" for the "lastModifiedBy" return type.

Issue #918, opened 2021-02-05
[Bug] Generator hardcodes enum values that have only one value even if they are not required.


gapra-msft:
Use this swagger file 
    https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-06-12/blob.json
and this README
https://raw.githubusercontent.com/Azure/azure-sdk-for-java/974e4bf50c973811dea77b3aa5e879aab8e64f95/sdk/storage/azure-storage-blob/swagger/README.md
optional Customization
https://github.com/Azure/azure-sdk-for-java/blob/974e4bf50c973811dea77b3aa5e879aab8e64f95/sdk/storage/azure-storage-blob/swagger/src/main/java/com/azure/storage/blob/customization/BlobStorageCustomization.java

in AppendBlobsImpl for example, EncryptionAlgorithm is hardcoded to AES256 when the swagger says its optional.

weidongxu-microsoft:
This is my thought. when spec says a parameter is optional, and the parameter is enum with one value, this actually means 2 possibilities. 1st, `null`; 2nd, that enum value.

There is service I saw, that `null` and enum value means two different things.

weidongxu-microsoft:
E.g., the behavior is different when `expand=null` and when `expand=InstanceViewTypes.INSTANCE_VIEW`

```
    Mono<VirtualMachineInner> getByResourceGroupAsync(
        String resourceGroupName, String vmName, InstanceViewTypes expand);
```

PR #940, opened 2021-03-02
Convert optional constants to enum

srnagar:
This PR fixes an issue where optional constants were always set. The change is to treat optional constants as single-valued enums.

Fixes https://github.com/Azure/autorest.java/issues/918


weidongxu-microsoft (approved):
LGTM. There is massive diff in generated code for test. Not sure if you would want to have a option for it.

srnagar:
@weidongxu-microsoft - added a new flag `optional-constant-as-enum`

PR #952, opened 2021-03-04
Added autorest help-content for java

timotheeguerin:
Added configuration to print help when using `--help`.

Copied the flags from the settings section above.

when running -`autorest ---help --java` it will display this after autorest help content

image


weidongxu-microsoft (approved):
Thanks Timothee. This is great. I was not aware about this help feature.

Issue #870, opened 2020-12-28
[Bug] Provide specialized Response class when response contains headers


weidongxu-microsoft:
```
    @ServiceMethod(returns = ReturnType.SINGLE)
    public Mono<HeadersResponseBoolResponse> responseBoolWithResponseAsync(String scenario) {
        if (this.client.getHost() == null) {
            return Mono.error(
                    new IllegalArgumentException("Parameter this.client.getHost() is required and cannot be null."));
        }
        if (scenario == null) {
            return Mono.error(new IllegalArgumentException("Parameter scenario is required and cannot be null."));
        }
        return FluxUtil.withContext(context -> service.responseBool(this.client.getHost(), scenario, context));
    }
```

The sync version should be
```
public HeadersResponseBoolResponse responseBoolWithResponse(String scenario, Context context)
```
instead of (which lose the deserialized headers type `HeadersResponseBoolHeaders` in response via `getDeserializedHeaders` method in `ResponseBase`)
```
public Response<Void> responseBoolWithResponse(String scenario, Context context)
```

PR #966, opened 2021-03-15
v4, bug fix, client method with sync response should return specialized Response with headers

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/870

Pending some fixes in fluent.

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/header/Headers.java
+++ b/vanilla-tests/src/main/java/fixtures/header/Headers.java
@@ -353,6 +353,29 @@ public final class Headers {
      * Send a post request with header value "User-Agent": "overwrite".
      *
      * @param userAgent Send a post request with header value "User-Agent": "overwrite".
+     * @param context The context to associate with this operation.
+     * @throws IllegalArgumentException thrown if parameters fail the validation.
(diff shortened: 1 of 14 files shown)


vanilla-tests/src/main/java/fixtures/header/Headers.java
```diff
@@ -416,6 +503,20 @@ public void responseExistingKey() {
...
+     * Get a response with header value "User-Agent": "overwrite".
+     *
+     * @param context The context to associate with this operation.
+     * @throws IllegalArgumentException thrown if parameters fail the validation.
+     * @throws ErrorException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return a response with header value "User-Agent": "overwrite".
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public HeadersResponseExistingKeyResponse responseExistingKeyWithResponse(Context context) {
```
weidongxu-microsoft, line 516: Method signature before the fix would be
```
    public Response<Void> responseExistingKeyWithResponse(Context context)
```

PR #969, opened 2021-03-17
[Autorest.Android]: Modify androidgen to create Chat client for Android

JianpingChen:
https://github.com/Azure/azure-sdk-for-android/issues/665

### DIFF
--- a/androidgen/src/main/java/com/azure/autorest/android/template/AndroidClientMethodTemplate.java
+++ b/androidgen/src/main/java/com/azure/autorest/android/template/AndroidClientMethodTemplate.java
@@ -3,9 +3,11 @@ package com.azure.autorest.android.template;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientMethod;
+import com.azure.autorest.model.clientmodel.ClientMethodParameter;
 import com.azure.autorest.model.clientmodel.GenericType;
 import com.azure.autorest.model.clientmodel.IType;
 import com.azure.autorest.model.clientmodel.ProxyMethod;
+import com.azure.autorest.model.clientmodel.ProxyMethodParameter;
 import com.azure.autorest.model.javamodel.JavaType;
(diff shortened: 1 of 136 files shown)


androidgen/src/main/java/com/azure/autorest/android/template/AndroidClientMethodTemplate.java
```diff
@@ -0,0 +1,209 @@
...
+            }
+        });
+    }
+
+    private String declareSinglePageCallback(ProxyMethod restAPIMethod, String completeFutureVariableName, String callbackVariableName) {
+        ProxyMethodParameter callbackParam = restAPIMethod.getParameters().stream().filter(param -> param.getName().equals("callback")).findFirst().get();
+        IType callbackDataType = ((GenericType) callbackParam.getClientType()).getTypeArguments()[0];
+
+        StringBuilder callbackBuilder = new StringBuilder();
+        callbackBuilder.append(String.format("Callback<%1$s> %2$s = new Callback<%1$s>() {\n", callbackDataType, callbackVariableName));
```
ellieMayVelasquez, line 72: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

JianpingChen: Thank you. I wasn't aware of that. I'll update this

ellieMayVelasquez: Happy to help :)

srnagar (approved):
The PR is being merged to non-v4 branch. It might be good to validate that this doesn't break any of the tests for fluent and javagen.

PR #975, opened 2021-03-22
Optionally use Iterable instead of List

srnagar:
This PR introduces a new option `use-iterable` to generate client methods and models using `Iterable` instead of `List`. This will allow the convenience layer to pass the iterable without converting the user-provided iterable into a List type. The general guideline for public API is to use Iterable and most libraries now have to map the iterable to a list to call the generated APIs.


jianghaolu (approved):

weidongxu-microsoft (approved):
LGTM. It might be help to put the option to one vanilla test to test it effect.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -124,7 +124,8 @@ public static JavaSettings getInstance()
                     host.getStringValue("customization-class"),
                     host.getBooleanValue("model-override-setter-from-superclass",
                             modelOverrideSetterFromSuperclassDefault),
-                    host.getBooleanValue("optional-constant-as-enum", false));
+                    host.getBooleanValue("optional-constant-as-enum", false),
+                    host.getBooleanValue("use-iterable", false));
```
weidongxu-microsoft, line 128: "use-iterable" maybe a bit too generic?

One question, does it affect client method parameter, or getter, or setter?

It seems we would prefer input as Iterable, but return value as Collection type.

vanilla-tests/src/test/java/fixtures/bodydatetime/DatetimeOperationsTests.java
```diff
@@ -82,14 +84,14 @@ public void putLocalPositiveOffsetMaxDateTime() {
     public void getLocalPositiveOffsetLowercaseMaxDateTime() {
         OffsetDateTime result = client.getDatetimeOperations().getLocalPositiveOffsetLowercaseMaxDateTime();
         OffsetDateTime expected = OffsetDateTime.parse("9999-12-31T09:59:59.999+00:00");
-        Assert.assertEquals(expected, result);
+        Assert.assertEquals(expected.toInstant().toEpochMilli(), result.toInstant().toEpochMilli());
```
weidongxu-microsoft, line 87: Any idea why now need to compare epoch?

weidongxu-microsoft (approved):

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
srnagar: This will affect all. Making changes only to the client method and not the models can potentially cause some compilation errors. Also, we need to identify input-only models, output-only models and input-output models in autorest java. .NET already does this and it would be a good addition to Java as well. When we have this feature, we can consider generating List/Iterable depending on how the models are used. For now, we'll change all List to Iterable when this flag is set and we can introduce more finer control later.

vanilla-tests/src/test/java/fixtures/bodydatetime/DatetimeOperationsTests.java
srnagar: This is not directly related to the Iterable change but a side-effect of upgrading azure-core and azure-core-http-netty packages. The serializer versions were upgraded too and all these tests failed. The epoc millis were the same but the zone offset was different and the equality check failed.

PR #982, opened 2021-03-23
v4 fluent, update doc and sample for mgmt

weidongxu-microsoft:
Only update doc with sample.


docs/samples/specification/management/readme.md
```diff
@@ -7,17 +7,18 @@ Use the flag `--azure-arm` to specify you want to generate [management plane][mg
 cd <this folder>
```
weidongxu-microsoft, line 7: Only changed this file. Rest files are all generated sample code.

PR #996, opened 2021-04-12
Code Customization Performance Improvement and Ability to Replace Method Parameters

alzimmermsft:
This PR makes the following changes:

**Performance Improvements**

Some code locations that were using `StringWriter` and `PrintWriter` were replaced with using `StringBuilder`. These locations used the writers in a localized and thread-safe manner which meant there was unneeded overhead as they use synchronization to support thread safety. Now, these locations are using `StringBuilder` which isn't thread safe as isn't doesn't synchronize. This results in performance improvements around synchronization and the ability for the operation to leverage better memory locality.

**Ability to Replace Method Parameters**

The ability to replace a methods parameters was added. This API only replaces the method parameters and doesn't update its `@param` Javadocs, so when it is used if the parameter name is change or parameters are being added or removed the Javadoc will need to be updated in an additional step.

**Other Changes**

Additionally, a minor bug was fixed with Javadoc parsing where `@deprecated` wasn't using the correct offset (it accidentally changed to using `@see`).

### DIFF
--- a/customization-base/src/main/java/com/azure/autorest/customization/Editor.java
+++ b/customization-base/src/main/java/com/azure/autorest/customization/Editor.java
@@ -70,9 +70,10 @@ public final class Editor {
         boolean fileCreated;
         try {
             fileCreated = newFile.createNewFile();
-            BufferedWriter writer = Files.newBufferedWriter(newFile.toPath());
-            writer.write(content);
-            writer.close();
+
+            try (BufferedWriter writer = Files.newBufferedWriter(newFile.toPath())) {
+                writer.write(content);
+            }
         } catch (IOException e) {
             throw new RuntimeException();
         }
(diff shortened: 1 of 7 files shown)


customization-base/src/main/java/com/azure/autorest/customization/Editor.java
```diff
@@ -71,9 +70,9 @@ public void addFile(String name, String content) {
         boolean fileCreated;
         try {
             fileCreated = newFile.createNewFile();
-            FileOutputStream stream = new FileOutputStream(newFile);
-            stream.write(content.getBytes(StandardCharsets.UTF_8));
-            stream.close();
+            BufferedWriter writer = Files.newBufferedWriter(newFile.toPath());
+            writer.write(content);
+            writer.close();
```
srnagar, line 75: Use try-with-resources or move this close() call to `finally`  block.

srnagar (approved):

jianghaolu (approved):
Feel free to merge once CI passes

PR #1007, opened 2021-04-22
Low level codegen

jianghaolu:


jianghaolu:
Need to support serviceVersion per comment https://github.com/Azure/azure-sdk-for-java/pull/20866#discussion_r618596076. But will do in a separate PR

weidongxu-microsoft:
Any sample code on the generated low level client? I am not familiar with the concept yet.

jianghaolu:
> Any sample code on the generated low level client? I am not familiar with the concept yet.
> 
> Ok, forget it. https://github.com/Azure/azure-sdk-for-java/pull/20866/files would be a sample.

Yes that would be a good one

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -292,6 +295,21 @@ private ClientResponse parseResponse(Operation method, JavaSettings settings) {
...
+        // WARNING: Only tested for low level clients
+        JavaSettings settings = JavaSettings.getInstance();
+        ModuleInfo moduleInfo = new ModuleInfo(settings.getPackage());
+
+        List<ModuleInfo.RequireModule> requireModules = moduleInfo.getRequireModules();
+        requireModules.add(new ModuleInfo.RequireModule("com.azure.core", true));
+        requireModules.add(new ModuleInfo.RequireModule("com.azure.core.experimental", true));
+
+        List<ModuleInfo.ExportModule> exportModules = moduleInfo.getExportModules();
+        exportModules.add(new ModuleInfo.ExportModule(settings.getPackage()));
```
srnagar, line 308: If the LLC client is provided with a package name that has `implementation` in it, we should not export it. Should we even support generating LLC if the package name has `implementation`?

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -292,6 +295,21 @@ private ClientResponse parseResponse(Operation method, JavaSettings settings) {
         return builder.build();
     }
 
+    private static ModuleInfo moduleInfo() {
```
srnagar, line 298: Do we generate `pom.xml` too for LLC?

srnagar (approved):

javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
```diff
@@ -139,10 +139,10 @@ public ServiceClient map(CodeModel codeModel) {
             }
         }
         serviceClientProperties.add(new ServiceClientProperty("The HTTP pipeline to send requests through.",
-                ClassType.HttpPipeline, "pipeline", true, null));
```
weidongxu-microsoft: If you do intend to use "pipeline", I can accommodate it in fluentgen, as it is not exposed to customer so not counted as breaking changes (though better not to change if no very good reason).

jianghaolu: it's the inconsistency between the naming in the client and the builder that drove me to make the change - but feel like it's probably less painful to just avoid this change

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
jianghaolu: no we don't

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
jianghaolu: If --low-level-client is turned on, the only 3 files we generate are the client, the builder, and the module-info. I don't quite understand why someone would put the client and builder in `implementation`?

jianghaolu:
Pending `RequestOptions` being added to `azure-core-experimental`

jianghaolu:
Work is moved to the `llc` branch.

PR #1008, opened 2021-04-22
Add removeMethod to ClassCustomization and Code Cleanup

alzimmermsft:


customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -285,14 +285,61 @@ public MethodCustomization addMethod(String method) {
         return getMethod(methodSignature);
     }
 
+    /**
+     * Removes a method from this class.
+     *
+     * @param methodNameOrSignature The name or signature of the method being removed.
+     * @return The current ClassCustomization.
+     */
+    public ClassCustomization removeMethod(String methodNameOrSignature) {
```
srnagar, line 294: Will this remove all overloads if just a method name is given? It would be good to document the behavior in the javadoc.

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -285,14 +285,61 @@ public MethodCustomization addMethod(String method) {
...
+        Position end = new Position(endLine, editor.getFileLine(fileName, endLine).length());
+
+        // Remove the method.
+        editor.replace(fileName, start, end, "");
+        FileEvent fileEvent = new FileEvent();
+        fileEvent.setUri(fileUri);
+        fileEvent.setType(FileChangeType.CHANGED);
+        languageClient.notifyWatchedFilesChanged(Collections.singletonList(fileEvent));
+
+        return this;
```
srnagar, line 332: Do we want to verify that this method is not called from anywhere before removing it? Also, what happens if this was the only method in a class? Would we end up with an empty class?

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
alzimmermsft: It will only remove the instance which matches the signature. I've actually been thinking that all methods which take a name or signature should throw if multiple instances are found as it leads to an ambiguous runtime behavior.

I'll update the docs to say this only removes a single instance.

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
alzimmermsft: For now I'm going to leave this unverified as this is already an advanced scenario.

If this is the only method it will end up with an empty class

srnagar: I am okay with this behavior but just document this in the javadoc so that anyone using this will know what to expect.

srnagar (approved):

PR #1009, opened 2021-04-23
Add support for FormParam

### LINKED ISSUE #1004: [FEATURE REQ] Support FormParam in Generation
When a parameters `in` is `formData` and `consumes` is, or maybe contains, `application/x-www-form-urlencoded` the parameter should generate with `@FormParam` annotation.

This is an example case: https://github.com/Azure/azure-sdk-for-js/blob/1998b841dcfa3fd17f0d8e0a4973ea61a25d2ecb/sdk/containerregistry/container-registry/swagger/containerregistry.json#L1096

srnagar:
This PR adds support for proxy method params that use `x-www-form-urlencoded` content type to be annotated with `@FormParam`. 

Fixes #1004

### DIFF
--- a/generate.bat
+++ b/generate.bat
@@ -35,6 +35,8 @@ call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com
 call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/non-string-enum.json --namespace=fixtures.nonstringenum
 call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/multiple-inheritance.json --namespace=fixtures.multipleinheritance
 call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/report.json --namespace=fixtures.report --payload-flattening-threshold=1
+call autorest $VANILLA_ARGUMENTS --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/body-formdata-urlencoded.json --namespace=fixtures.bodyformdataurlencoded
+
 rem local swagger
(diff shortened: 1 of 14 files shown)


generate.bat
```diff
@@ -35,6 +35,8 @@ call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com
 call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/non-string-enum.json --namespace=fixtures.nonstringenum
 call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/multiple-inheritance.json --namespace=fixtures.multipleinheritance
 call autorest %VANILLA_ARGUMENTS% --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/report.json --namespace=fixtures.report --payload-flattening-threshold=1
+call autorest $VANILLA_ARGUMENTS --input-file=https://raw.githubusercontent.com/Azure/autorest.testserver/master/swagger/body-formdata-urlencoded.json --namespace=fixtures.bodyformdataurlencoded
```
alzimmermsft, line 38: `$VANILLA_ARGUMENTS` or `%VANILLA_ARGUMENTS`? Everything else is using the latter but this is using the former.

javagen/src/main/java/com/azure/autorest/template/ProxyTemplate.java
```diff
@@ -111,12 +112,17 @@ public final void write(Proxy restAPI, JavaClass classBlock) {
                                 break;
 
                             case Body:
+                                if (restAPIMethod.getRequestContentType() != null && restAPIMethod.getRequestContentType().equals(ContentType.APPLICATION_X_WWW_FORM_URLENCODED)) {
```
alzimmermsft, line 115: Could invert this check and compact it:

```suggestion
                                if (ContentType.APPLICATION_X_WWW_FORM_URLENCODED).equals()restAPIMethod.getRequestContentType()) {
```

jianghaolu (approved):

alzimmermsft (approved):

pallavit (approved):
:shipit:

PR #1010, opened 2021-04-26
Fixed a Bug Where Adding a Constructor or Method with a Parameterized Parameter Failed

alzimmermsft:
This PR makes the following changes:

- Fixed bugs with constructor and method signature extraction when generic parameters or multiple parameters exist.
- Fixed a bug when a replacing a constructor's or method's parameters and there existed a an annotation (ex. `@JsonProperty()`) on a parameter.
- Further centralization of parsing and symbol walking and additional hardening of the logic.

### DIFF
--- a/customization-base/src/main/java/com/azure/autorest/customization/implementation/Utils.java
+++ b/customization-base/src/main/java/com/azure/autorest/customization/implementation/Utils.java
@@ -48,6 +48,28 @@ public class Utils {
      */
     public static final Pattern INDENT_DETERMINATION_PATTERN = Pattern.compile("^(\\s*)\\w.*$");
 
+    /*
+     * This pattern determines if a line is a beginning of constructor or method. The following is an explanation of
+     * the pattern:
+     *
+     * 1. Capture all leading space characters.
+     * 2. Capture all modifiers for the constructor or method.
+     * 3. If a method, capture the return type.
+     * 4. Capture the name of the constructor or method.
+     *
+     * The following are the groups return:
+     *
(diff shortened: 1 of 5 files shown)


customization-base/src/main/java/com/azure/autorest/customization/implementation/Utils.java
```diff
@@ -378,6 +400,48 @@ private static int walkFileUntilLineMatches(Editor editor, String fileName, int
...
+        // Ending character is exclusive of the character offset.
+        Position parametersEnd = new Position(line, matcher.group(1).length());
+
+        editor.replace(fileName, parametersStart, parametersEnd, newParameters);
+        FileEvent fileEvent = new FileEvent();
+        fileEvent.setUri(fileUri);
+        fileEvent.setType(FileChangeType.CHANGED);
+        languageClient.notifyWatchedFilesChanged(Collections.singletonList(fileEvent));
+
+        return refreshCustomizationSupplier.get();
```
srnagar, line 442: Not sure why we take the supplier and return the supplied value without using it here. Can we drop the supplier and let the caller call it after this method is executed?

srnagar (approved):

customization-base/src/main/java/com/azure/autorest/customization/implementation/Utils.java
alzimmermsft: Went with this pattern to allow for calls into this method, and others like it, to be a single line.

PR #1017, opened 2021-05-04
Fix for multiple query parameters with same name.

v-ddaian:
Add support for `collectionFormat multi` that is implemented with `explode` parameter. After identifying that we need `collectionFormat multi`, we store it in the builder and add flag `multipleQueryParams = true`. The multiple parameters part is implemented in [azure-sdk-for-java/pull/21203](https://github.com/Azure/azure-sdk-for-java/pull/21203).

This will fix the following:
* [Support URL with multiple query param with same same](https://github.com/Azure/azure-sdk-for-java/issues/13124)

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/url/multi/Queries.java
+++ b/vanilla-tests/src/main/java/fixtures/url/multi/Queries.java
@@ -14,10 +14,9 @@ import com.azure.core.http.rest.Response;
 import com.azure.core.http.rest.RestProxy;
 import com.azure.core.util.Context;
 import com.azure.core.util.FluxUtil;
-import com.azure.core.util.serializer.CollectionFormat;
-import com.azure.core.util.serializer.JacksonAdapter;
 import fixtures.url.multi.models.ErrorException;
 import java.util.List;
+import java.util.stream.Collectors;
 import reactor.core.publisher.Mono;
 
 /** An instance of this class provides access to all the operations defined in Queries. */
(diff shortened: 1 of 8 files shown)


v-ddaian:
This is ready for review.

vanilla-tests/src/main/java/fixtures/url/multi/Queries.java
```diff
@@ -50,7 +49,7 @@
         @UnexpectedResponseExceptionType(ErrorException.class)
         Mono<Response<Void>> arrayStringMultiNull(
                 @HostParam("$host") String host,
-                @QueryParam("arrayQuery") String arrayQuery,
+                @QueryParam(value = "arrayQuery", multipleQueryParams = true) List<String> arrayQuery,
```
weidongxu-microsoft, line 52: I didn't find `multipleQueryParams` in azure-core?

v-ddaian: It's actually tied to the other PR, here:
[azure-sdk-for-java/pull/21203](https://github.com/Azure/azure-sdk-for-java/pull/21203)

weidongxu-microsoft: I see, yes, would be a useful feature.

This PR on autorest.java would need to wait for the one on azure-core and its release.

weidongxu-microsoft: It appears that `multipleQueryParams` is released with azure-core 1.19.0. And I've updated with it in latest v4 branch. If you want to continue this PR, you can try refresh latest v4 (there might be conflict to resolve as we've added quite a lot of code in meantime), and it might solve your CI.

daniel-rocha:
@v-ddaian You can close this one. Superseded by https://github.com/Azure/autorest.java/pull/1161

PR #1035, opened 2021-06-03
Only Add JsonFlatten When the Flattening Property is on the Model

alzimmermsft:


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -195,7 +187,16 @@ public ClientModel map(ObjectSchema compositeType) {
                  builder.xmlName(compositeType.getLanguage().getDefault().getName());
             }
 
-            builder.needsFlatten(discriminatorNeedEscape || hasFlattenedProperty(compositeType, parentsNeedFlatten));
+            boolean needsFlattening = hasFlattenedProperty(compositeType, parentsNeedFlatten);
+            if (isPolymorphic) {
+                String discriminatorSerializedName = SchemaUtil.getDiscriminatorSerializedName(compositeType);
+                boolean discriminatorNeedEscape = needsFlattening && discriminatorSerializedName.contains(".");
+                builder.polymorphicDiscriminator(discriminatorNeedEscape
+                    ? discriminatorSerializedName.replace(".", "\\\\.")
```
srnagar, line 195: Do we want to wrap this behind a feature flag and make this optional?

weidongxu-microsoft (approved):

PR #1037, opened 2021-06-07
upgrade snakeyaml to 1.28

weidongxu-microsoft:
1.25 have security issue (DDOS https://snyk.io/vuln/maven:org.yaml:snakeyaml@1.25)

### DIFF
--- a/androidgen/src/main/java/com/azure/autorest/android/Androidgen.java
+++ b/androidgen/src/main/java/com/azure/autorest/android/Androidgen.java
@@ -23,7 +23,10 @@ import com.azure.autorest.template.Templates;
 import com.azure.autorest.util.ClientModelUtil;
 import com.google.googlejavaformat.java.Formatter;
 import org.slf4j.Logger;
+import org.yaml.snakeyaml.DumperOptions;
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.introspector.Property;
 import org.yaml.snakeyaml.nodes.NodeTuple;
 import org.yaml.snakeyaml.nodes.Tag;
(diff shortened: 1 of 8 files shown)


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/NewPlugin.java
```diff
@@ -192,8 +194,10 @@ public NewPlugin(Connection connection, String plugin, String sessionId)
                 .withIsGetterVisibility(JsonAutoDetect.Visibility.NONE));
         Representer representer = new Representer();
         representer.getPropertyUtils().setSkipMissingProperties(true);
-        Constructor constructor = new CodeModelCustomConstructor();
-        yamlMapper = new Yaml(constructor, representer);
+        LoaderOptions loaderOptions = new LoaderOptions();
+        loaderOptions.setMaxAliasesForCollections(Integer.MAX_VALUE);
```
weidongxu-microsoft, line 198: Since 1.26 the default max is 50 (my guess is to fight DDOS). However this small value fails for even simple m4 output.

androidgen/src/main/java/com/azure/autorest/android/Androidgen.java
```diff
@@ -71,7 +74,9 @@ protected NodeTuple representJavaBeanProperty(Object javaBean, Property property
                 }
             };
 
-            Yaml newYaml  = new Yaml(representer);
+            LoaderOptions loaderOptions = new LoaderOptions();
```
srnagar, line 77: Can this be refactored to share code with all other *gen libraries so we don't have to duplicate this in 3 places?

srnagar (approved):

androidgen/src/main/java/com/azure/autorest/android/Androidgen.java
weidongxu-microsoft: Alternative is to put it in extension-base, but seems not matter too much.

And there might arise some circumstance that one plugin really want to limit the max number (though not very likely, as this is only to consume output from m4, not some data from user).

PR #1055, opened 2021-06-11
v4 option to put JsonFlatten annotation on field

weidongxu-microsoft:


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -284,6 +289,15 @@ public final boolean isAzureOrFluent()
         return isAzure() || isFluent();
     }
 
+    public enum ClientFlattenAnnotationTarget {
+        TYPE, FIELD
+    }
```
weidongxu-microsoft, line 294: Might later add a `NONE` if we can generate code without JsonFlatten.

vanilla-tests/src/test/java/fixtures/discriminatorflattening/DiscriminatorTests.java
```diff
@@ -22,18 +27,40 @@ public void serializationOnDiscriminator() throws IOException {
...
+
+    @Test
+    public void serializationOnNestedFlatten() throws IOException {
+        SerializerAdapter adapter = JacksonAdapter.createDefaultSerializerAdapter();
+
+        VirtualMachineScaleSet vmss = new VirtualMachineScaleSet()
+                .setVirtualMachineProfile(new VirtualMachineScaleSetVMProfile()
+                        .setNetworkProfile(new VirtualMachineScaleSetNetworkProfile()
+                                .setNetworkInterfaceConfigurations(Collections.singletonList(new VirtualMachineScaleSetNetworkConfiguration()
+                                        .setName("name").setPrimary(true)))));
```
weidongxu-microsoft, line 47: @alzimmermsft 

Appears `@JsonFlatten` on field cannot handle nested properties.
Here `VirtualMachineScaleSetNetworkConfiguration` also have some fields been flattened, but it seems not serialized correctly.

`{"properties":{"virtualMachineProfile":{"networkProfile":{"networkInterfaceConfigurations":[{"name":"name","properties.primary":true}]}}}}`
where `properties.primary` part should be `primary` under `properties`.

vanilla-tests/src/main/java/fixtures/discriminatorflattening/models/VirtualMachineScaleSetNetworkConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+    @JsonProperty(value = "name")
+    private String name;
+
+    /*
+     * Specifies the primary network interface in case the virtual machine has
+     * more than 1 network interface.
+     */
+    @JsonFlatten
+    @JsonProperty(value = "properties.primary")
+    private Boolean primary;
```
weidongxu-microsoft, line 22: It is annotated correctly.

vanilla-tests/src/test/java/fixtures/discriminatorflattening/DiscriminatorTests.java
alzimmermsft: Thanks for identifying this @weidongxu-microsoft, I'll investigate and submit a fix.

Issue #1034, opened 2021-06-03
Pagination support in protocol methods


jianghaolu:
Generation pagination methods as described in https://gist.github.com/jianghaolu/77fea6fa1e5bda4b5746ab2235e1bfbe#pageable-operations

Test with this API: https://github.com/Azure/azure-sdk-for-java/blob/master/sdk/translation/azure-ai-documenttranslator/src/main/java/com/azure/ai/documenttranslator/BatchDocumentTranslationRestClient.java#L599

srnagar:
@qwordy has merged the PR for this work.

PR #1061, opened 2021-06-18
LLC Paging

qwordy:
Resolve https://github.com/Azure/autorest.java/issues/1034

Add paging support for low level client. Generated code example is in `protocol-tests.paging`. `impl` is internal implementation. Users will use `client` class to access service. We have two kinds of clients, sync and async client. Each paging method in swagger will generate 4 methods.

**TODO**
- Allow only path parameters. (To be confirmed)
- Java doc generation


javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -287,6 +287,20 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
...
                 imports.add("com.azure.core.util.Context");
             }
+
+            // Paging
+            if (getMethodPageDetails() != null) {
+                imports.add("com.azure.core.http.rest.PagedResponseBase");
+                imports.add("com.azure.core.http.rest.PagedResponse");
+                imports.add("com.azure.core.http.rest.PagedFlux");
+                imports.add("com.azure.core.http.rest.PagedIterable");
+                imports.add("com.google.gson.JsonArray");
```
jianghaolu, line 297: We would want to avoid having a specific json serializer in our dependencies - `azure-core` comes with `jackson` and if we have to `JacksonAdapter` will be the serializer to use.

qwordy: OK. I'll try to use class provided by azure-core.

qwordy: `JacksonAdaptor` forces me to give a type. But I don't want to generate model class from swagger. I just need string of each element in `value` and string of `nextLink`. Should we modify `azure-core` to add this ability?
```
{
  "value": [],
  "nextLink": "",
}
```

weidongxu-microsoft: You can use `Object.class` and then try parse the resulting `Object` var (it should be a `Map<String, Object>`, but do have protection code).

protocol-tests/src/test/java/fixtures/paging/PagingTests.java
```diff
@@ -0,0 +1,43 @@
+package fixtures.paging;
+
+import com.azure.core.http.rest.PagedIterable;
+import com.azure.core.util.BinaryData;
+import org.junit.Assert;
```
qwordy, line 5: `PagingTests.java` in `azure-tests` uses Junit 5. `protocal-tests` project defines
```
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.13.1</version>
      <scope>test</scope>
    </dependency>
```

jianghaolu: Feel free to update protocol-tests to use junit 5

qwordy:
Generated code misses parameters. Fixing. In the following code, `offset` is missing.
```
    public Mono<PagedResponse<BinaryData>> getMultiplePagesWithOffsetSinglePageAsync(RequestOptions requestOptions) {
        return FluxUtil.withContext(
                        context ->
                                service.getMultiplePagesWithOffset(
                                        this.client.getHost(), offset, accept, requestOptions, context))
```

qwordy:
Fixed it by adding
```
--pipeline.modelerfour.flatten-models=false --pipeline.modelerfour.flatten-payloads=false --pipeline.modelerfour.group-parameters=false
```

protocol-tests/src/test/java/fixtures/paging/PagingTests.java
qwordy: Done

qwordy:
Basic test can pass now👍

qwordy:
The issue is fixed👍

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -762,79 +777,53 @@ private static String parameterDescriptionOrDefault(ProxyMethodParameter paramet
...
-                            function.line("null));");
-                        }
-                    });
+                    }
+                    IType responseType = ((GenericType) clientMethod.getProxyMethod().getReturnType()).getTypeArguments()[0];
+                    if (responseType instanceof ClassType) {
+                        function.line("res.getDeserializedHeaders()));");
+                    } else {
+                        function.line("null));");
+                    }
```
weidongxu-microsoft, line 823: What is changed here? It seems removed an `if` branch.

javagen/src/main/java/com/azure/autorest/template/WrapperClientMethodTemplate.java
```diff
@@ -7,84 +7,83 @@
...
+
 import java.util.List;
 import java.util.stream.Collectors;
 
 /**
  * Template to generate client methods that are wrappers around the client methods generated by
  * {@link ClientMethodTemplate}.
- *
  */
 public class WrapperClientMethodTemplate implements IJavaTemplate<ClientMethod, JavaType> {
```
weidongxu-microsoft, line 18: What is changed here?

protocol-tests/pom.xml
```diff
@@ -37,16 +37,22 @@
       <artifactId>azure-core</artifactId>
       <version>1.18.0-beta.1</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
```
weidongxu-microsoft, line 42: Do you need gson?

protocol-tests/src/main/java/fixtures/paging/implementation/PagingsImpl.java
```diff
@@ -0,0 +1,3477 @@
...
+     *
+     * <p><strong>Response Body Schema</strong>
+     *
+     * <pre>{@code
+     * BinaryData
+     * }</pre>
+     *
+     * @return a DynamicRequest where customizations can be made before sent to the service.
+     */
+    @ServiceMethod(returns = ReturnType.COLLECTION)
```
weidongxu-microsoft, line 377: Not sure whether it should be `ReturnType.COLLECTION`

protocol-tests/src/main/java/fixtures/paging/implementation/PagingsImpl.java
```diff
@@ -0,0 +1,3477 @@
...
+                                        null));
+    }
+
+    /**
+     * A paging operation that must return result of the default 'value' node.
+     *
+     * <p><strong>Response Body Schema</strong>
+     *
+     * <pre>{@code
+     * BinaryData
```
weidongxu-microsoft, line 372: The schema is probably not correct.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
qwordy: Yes. OId `if` and `else` have huge code clone. I also add code for low level client.

protocol-tests/pom.xml
qwordy: Deleted.

protocol-tests/src/main/java/fixtures/paging/implementation/PagingsImpl.java
qwordy: Fixing.

protocol-tests/src/main/java/fixtures/paging/implementation/PagingsImpl.java
qwordy: I copy them from high level code.
`Mono` -> `@ServiceMethod(returns = ReturnType.SINGLE)`. 
`PagedFlux` -> `@ServiceMethod(returns = ReturnType.COLLECTION)`
`PagedIterable` -> `@ServiceMethod(returns = ReturnType.COLLECTION)`

javagen/src/main/java/com/azure/autorest/template/WrapperClientMethodTemplate.java
qwordy: Reverted.

jianghaolu (approved):

weidongxu-microsoft (approved):

PR #1062, opened 2021-06-18
Support body string in protocol client

jianghaolu:


weidongxu-microsoft (commented):
I seems not able to find `RequestOptions` class in core.

jianghaolu:
@weidongxu-microsoft It's added here https://github.com/Azure/azure-sdk-for-java/pull/22334

Issue #1107, opened 2021-07-13
Add `--skip-formatting` option


jianghaolu:
If the generated code has an error and cannot compile - right now the formatter rejects the code and the user has no way of knowing what's the problem. We should at least allow users to skip formatting and get the broken code and inspect it.

PR #1112, opened 2021-07-16
v4, add skip-formatting

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/1107

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,10 @@
 *                           @jianghaolu @srnagar
 
 # fluentgen
-/fluentnamer                @ChenTanyi @weidongxu-microsoft @yungezz
-/fluentgen                  @ChenTanyi @weidongxu-microsoft @yungezz
-/fluent-tests               @ChenTanyi @weidongxu-microsoft @yungezz
-/azure-tests                @ChenTanyi @weidongxu-microsoft @yungezz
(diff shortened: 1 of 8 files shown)


weidongxu-microsoft:
The newly added testservice case include naming conflict on `accept`. I will try to fix it in another PR.

Code (note the 2 `String accept = ...`.
```
    public Mono<Response<Void>> postReservedWordsWithResponseAsync(
            ParameterGroupingPostReservedWordsParameters parameterGroupingPostReservedWordsParameters) {
        if (this.client.getHost() == null) {
            return Mono.error(
                    new IllegalArgumentException("Parameter this.client.getHost() is required and cannot be null."));
        }
        if (parameterGroupingPostReservedWordsParameters != null) {
            parameterGroupingPostReservedWordsParameters.validate();
        }
        String fromInternal = null;
        if (parameterGroupingPostReservedWordsParameters != null) {
            fromInternal = parameterGroupingPostReservedWordsParameters.getFrom();
        }
        String from = fromInternal;
        String acceptInternal = null;
        if (parameterGroupingPostReservedWordsParameters != null) {
            acceptInternal = parameterGroupingPostReservedWordsParameters.getAccept();
        }
        String accept = acceptInternal;
        return FluxUtil.withContext(
                context -> service.postReservedWords(this.client.getHost(), from, accept, accept, context));
    }
```

PR #1113, opened 2021-07-16
v4, deduplicate parameter names in preprocess

weidongxu-microsoft:


azure-tests/src/main/java/fixtures/azureparametergrouping/ParameterGroupings.java
```diff
@@ -72,6 +73,16 @@
...
                 Context context);
 
+        @Post("/parameterGrouping/postReservedWords")
+        @ExpectedResponses({200})
+        @UnexpectedResponseExceptionType(ErrorException.class)
+        Mono<Response<Void>> postReservedWords(
+                @HostParam("$host") String host,
+                @QueryParam("from") String from,
+                @QueryParam("accept") String accept,
+                @HeaderParam("Accept") String acceptParam,
```
weidongxu-microsoft, line 83: Two `accept` if not deduplicated.

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
```diff
@@ -374,4 +379,51 @@ private Parameter createContentLengthParameter(Operation operation, Parameter bo
...
+        newParameter.getLanguage().getJava().setName(parameter.getLanguage().getJava().getName() + "Param");
+
+        iter.set(newParameter);
+
+        parameter = newParameter;
+      }
+
+      parameterNames.add(parameter.getLanguage().getJava().getName());
+    }
+  }
```
weidongxu-microsoft, line 405: Not sure if this is the best solution. @jianghaolu

weidongxu-microsoft: For now, reverted to modify the java name, instead of making a new `Parameter` object.

jianghaolu (approved):

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
jianghaolu: Fixing param names should be done in pre-processor - so I think this is the right place for it.

And yes I don't think making a copy is necessary.

Issue #1105, opened 2021-07-13
Support x-ms-client-flatten in codegen without JsonFlatten annotation


weidongxu-microsoft:
`--client-flattened-annotation-target=NONE` and probably `--pipeline.modelerfour.flatten-models=false`

PR #1114, opened 2021-07-16
v4 support client model flatten at codegen (avoid JsonFlatten)

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/1105

body-complex.json pass on this configure.

Currently it probably not work with `--required-fields-as-ctor-args`. Also have problem with `modelerfour.flatten-payloads`, as it will flatten the `properties` from payload if that is the only writable property.

A sample project https://github.com/Azure/azure-sdk-for-java/pull/23076
(regen, compared to previous version with `JsonFlatten`)

So far, compute service regenerated with this can pass all live tests.
https://github.com/Azure/azure-sdk-for-java/pull/23105

Finally get it reach same smoke test result https://github.com/weidongxu-microsoft/autorest.java/blob/v4_client-flatten-test/fluent-tests/report.md


vanilla-tests/src/main/java/fixtures/clientflattendiscriminatorflattening/models/BackendAddressPool.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param location the location value to set.
+     * @return the BackendAddressPool object itself.
+     */
+    public BackendAddressPool setLocation(String location) {
+        if (this.getProperties() == null) {
+            this.setProperties(new BackendAddressPoolPropertiesFormat());
+        }
+        this.getProperties().setLocation(location);
+        return this;
+    }
```
weidongxu-microsoft, line 60: This one is not thread safe. We could have race condition here.

Any suggestion? A `syncrhonized` `ensureProperties` method?

javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
```diff
@@ -111,16 +124,11 @@ public ClientModelProperty map(Property property) {
...
-
-        IType propertyClientType = Mappers.getSchemaMapper().map((property.getSchema())).getClientType();
-
-        if (!property.isRequired()) {
-            propertyClientType = propertyClientType.asNullable();
+        if (property.isNullable() || !property.isRequired()) {
             propertyWireType = propertyWireType.asNullable();
         }
+        // Invariant: clientType == wireType.getClientType()
+        IType propertyClientType = propertyWireType.getClientType();
```
weidongxu-microsoft, line 131: This part of logic is slightly different. In condition of `property.isNullable()`, previously it is only `propertyWireType` be nullable, now both is nullable (I assume when property is nullable, the client type of that property should also be nullable?).

PR #1116, opened 2021-07-19
Keep only one request per operation; MediaTypes

qwordy:
Keep only one request per operation
Generate MediaTypes, HttpInfrastructure, URL, write test
Make both List and Map BinaryData
Fix unchecked cast to pass style check
Redefine proxy method parameter list, allow query parameter api-version if it has default value
Test infrastructure for breaking change resilience
Make client type simple, original type is UnixTimeLong/OffsetDateTime, we need long
Fix javadoc warnings, add all @param and @return
Rename builder class name to pass linter
Fix "Exception is caught when Exception is not thrown"


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -124,6 +125,14 @@ public static ProxyMethodMapper getInstance() {
         }
         builder.responseContentTypes(responseContentTypes);
 
+        // Low-level client only requires one request per operation
+        if (settings.isLowLevelClient()) {
+            List<Request> list = operation.getRequests();
+            if (!list.isEmpty()) {
+                operation.setRequests(Collections.singletonList(list.get(0)));
```
weidongxu-microsoft, line 132: You should handle it when processing the `requests`, than override it in `operation`.

weidongxu-microsoft (approved):
LGTM with one comment.

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
qwordy: Fixed

javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodParameter.java
```diff
@@ -19,6 +19,10 @@
      * Get the description of this parameter.
      */
     private String description;
+    /**
+     * Get the raw type of this parameter.
```
weidongxu-microsoft, line 23: Better comment on this `rawType`.

qwordy: Done

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
```diff
@@ -179,6 +179,9 @@ public final String defaultValueExpression(String sourceExpression) {
     }
 
     public final IType getClientType() {
+        if (JavaSettings.getInstance().isLowLevelClient())  {
+            return this;
+        }
```
weidongxu-microsoft, line 184: This seems not very understandable to the design of `IType`.

Would it better to have `SchemaMapper` just output e.g. `Long` instead of `UnixTimeLong` in case of LLC?
@jianghaolu

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -3,24 +3,8 @@
 import com.azure.autorest.extension.base.model.codemodel.*;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.extension.base.plugin.JavaSettings.SyncMethodsGeneration;
-import com.azure.autorest.model.clientmodel.ClassType;
-import com.azure.autorest.model.clientmodel.ClientMethod;
+import com.azure.autorest.model.clientmodel.*;
```
jianghaolu, line 6: expand imports

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
jianghaolu: Yes I agree - better to handle this in the mappers

javagen/src/main/java/com/azure/autorest/model/clientmodel/PrimitiveType.java
```diff
@@ -95,6 +97,9 @@ public final String defaultValueExpression(String sourceExpression) {
     }
 
     public final IType getClientType() {
+        if (JavaSettings.getInstance().isLowLevelClient())  {
```
jianghaolu, line 100: Same here

javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -97,9 +97,8 @@ protected NodeTuple representJavaBeanProperty(Object javaBean, Property property
 
             // Service client builder
             String builderPackage = ClientModelUtil.getServiceClientBuilderPackageName(client.getServiceClient());
-            String builderSuffix = ClientModelUtil.getBuilderSuffix();
-            javaPackage.addServiceClientBuilder(builderPackage,
-                client.getServiceClient().getInterfaceName() + builderSuffix, client.getServiceClient());
+            String builderClassName = ClientModelUtil.getBuilderClassName(client.getServiceClient().getInterfaceName());
+            javaPackage.addServiceClientBuilder(builderPackage, builderClassName, client.getServiceClient());
```
weidongxu-microsoft, line 101: What is the difference?

qwordy: CI forces builder name must end with "ClientBuilder". Old code appends "Builder". I add a check if it does not end with "Client", then append "ClientBuilder".

weidongxu-microsoft: This might cause breaking changes to existing packages, many of which the Builder lives in `implementation` package and hence not need to have the `Client`.

Would it possible to only apply this logic to `shouldGenerateSyncAsyncClients` or just `isLowLevelClient`?

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
qwordy: fixed

javagen/src/main/java/com/azure/autorest/Javagen.java
qwordy: We can also specify `--title=PurviewScanningClient` to ignore name defined in swagger.

weidongxu-microsoft: Yeah, this might be easier.

qwordy: I'll revert this commit.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
qwordy: Fixed

javagen/src/main/java/com/azure/autorest/model/clientmodel/PrimitiveType.java
qwordy: Fixed

javagen/src/main/java/com/azure/autorest/Javagen.java
qwordy: Done

weidongxu-microsoft (approved):

jianghaolu (approved):
move primitive mapper logic to service client mapper

qwordy:
> move primitive mapper logic to service client mapper

I tried this approach. It is primitive type, so it will not be changed. This PR is very long. Merge it first.

Issue #1119, opened 2021-07-21
Javadoc for getters/setters use only description rather than title + description


dpwatrous:
The swagger for the Batch service currently splits the documentation of properties between `title` and `description`. This currently causes poor JavaDoc strings on model getters and setters, which use only `description` (see the example below).

Here is the same issue in autorest.python, which has some discussion on it already: https://github.com/Azure/autorest.python/issues/978 

The result of that discussion was that the python extension will concatenate title and description when generating doc strings if both are defined (note: it did this in a prior version already). In the interests of keeping Autorest behavior across languages consistent, I believe this behavior should be added to the Java extension as well.

Here is the current JavaDoc generated from a property with both title and description: 

Swagger
```
"version": {
  "type": "string",
  "title": "The version of the application to deploy. If omitted, the default version is deployed.",
  "description": "If this is omitted on a Pool, and no default version is specified for this application, the request fails with the error code InvalidApplicationPackageReferences and HTTP status code 409. If this is omitted on a Task, and no default version is specified for this application, the Task fails with a pre-processing error."
}
```

JavaDoc
```
/**
 * Set if this is omitted on a Pool, and no default version is specified for this application, the request fails with the error code InvalidApplicationPackageReferences and HTTP status code 409. If this is omitted on a Task, and no default version is specified for this application, the Task fails with a pre-processing error.
```

dpwatrous:
Tagging @jianghaolu on this at the suggestion of @iscai-msft

matthchr:
FWIW this is something that is impacting AKS as well

haolingdong-msft:
@dpwatrous The issue fix has been merged (https://github.com/Azure/autorest.java/pull/1123). Please see sample Swagger spec and generated JavaDoc below. Please let me know if there is any query.
Swagger:
```
"version": {
          "type": "string",
          "title": "The version of the application to deploy. If omitted, the default version is deployed.",
          "description": "If this is omitted, and no default version is specified for this application, the request fails with the error code InvalidApplicationPackageReferences. If you are calling the REST API directly, the HTTP status code is 409."
}
```
JavaDoc:
```
/**
     * Set the version property: The version of the application to deploy. If omitted, the default version is deployed.
     * If this is omitted, and no default version is specified for this application, the request fails with the error
     * code InvalidApplicationPackageReferences. If you are calling the REST API directly, the HTTP status code is 409.
     *
     * @param version the version value to set.
     * @return the ApplicationPackageReference object itself.
[...]

PR #1123, opened 2021-07-28
fix issue #1119 to add summary in schema together with description for code model property

haolingdong-msft:
…r code model property


javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
```diff
@@ -36,12 +37,34 @@ public ClientModelProperty map(Property property) {
...
+
+        if (descriptionInProperty.isEmpty() && summaryInProperty.isEmpty()) {
+            descriptionSb.append(String.format("The %s property.", property.getSerializedName()));
+        } else if(summaryInProperty.isEmpty()) {
+            descriptionSb.append(descriptionInProperty);
+        } else if(descriptionInProperty.isEmpty()) {
+            descriptionSb.append(summaryInProperty);
         } else {
-            builder.description(property.getLanguage().getJava().getDescription());
+            if(summaryInProperty.equals(descriptionInProperty)) {
```
haolingdong-msft, line 58: only add summary if summary is not equal to description

weidongxu-microsoft (approved):

Issue #1110, opened 2021-07-14
Design and support ServiceVersion


weidongxu-microsoft:
Reference to handwritten data-plane SDK `ServiceVersion`
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/appconfiguration/azure-data-appconfiguration/src/main/java/com/azure/data/appconfiguration/ConfigurationClientBuilder.java#L410-L413

**Solution**
`ClientBuilder` class will provide a method to set service version. It maps to `api-version` in swagger.
```
    public XXXClientBuilder serviceVersion(XXXServiceVersion version) {
        this.version = version;
        return this;
    }
```
`XXXServiceVersion` is an enum type. It defines service versions supported by service and latest version.
```
public enum XXXServiceVersion implements ServiceVersion {
    V2020_06_12("2020-06-12"),
    V2020_08_04("2020-08-04"),
    V2020_10_02("2020-10-02");

    private final String version;

    XXXServiceVersion(String version) {
        this.version = version;
    }

    @Override
    public String getVersion() {
        return this.version;
    }

    public static XXXVersion getLatest() {
        return V2020_10_02;
    }
}
```
Service version list is configured at config file or command parameter.
Config code is in yaml syntax. They are sorted in an ascending order. The bottom one is latest service version. Please add quotes around date.
```
service-versions:
  - "2020-06-12"
  - "2020-08-04"
  - "2020-10-02"
```
`ClientBuilder` class has
```
private XXXServiceVersion serviceVersion;

public XXXClientBuilder serviceVersion(XXXServiceVersion serviceVersion) {
    this.serviceVersion = serviceVersion;
    return this;
}

private XXXClientImpl buildInnerClient() {
    if (serviceVersion == null) {
        this.serviceVersion = XXXServiceVersion.getLatest();
    }
    ...
    XXXClientImpl client =
            new XXXClientImpl(pipeline, serializerAdapter, endpoint, serviceVersion.getVersion());
    ...
```
If user does not set service version, it will use latest service version by default.
`ClientImpl` class uses String type `apiVersion`.
```
private final String apiVersion;
```

qwordy:
Add a test case
```
    @Test
    public void putValidWithApiVersion() throws Exception {
        BinaryData binaryData = BinaryData.fromString("{\"id\": 2, \"name\": \"abc\", \"color\": \"Magenta\"}");
[...]

PR #1126, opened 2021-08-02
LLC service version

qwordy:
Resolve https://github.com/Azure/autorest.java/issues/1110
`ClientBuilder` class will provide a method to set service version. It maps to `api-version` in swagger.
```
    public XXXClientBuilder serviceVersion(XXXServiceVersion version) {
        this.version = version;
        return this;
    }
```
`XXXServiceVersion` is an enum type. It defines service versions supported by service and latest version.
```
public enum XXXServiceVersion implements ServiceVersion {
    V2020_06_12("2020-06-12"),
    V2020_08_04("2020-08-04"),
    V2020_10_02("2020-10-02");

    private final String version;

    XXXServiceVersion(String version) {
        this.version = version;
    }

    @Override
    public String getVersion() {
        return this.version;
    }

    public static XXXVersion getLatest() {
        return V2020_10_02;
    }
}
```
Service version list is configured at config file or command parameter.
Config code is in yaml syntax. They are sorted in an ascending order. The bottom one is latest service version. Please add quotes around date.
```
service-versions:
  - "2020-06-12"
  - "2020-08-04"
  - "2020-10-02"
```
`ClientBuilder` class has
```
private XXXServiceVersion serviceVersion;

public XXXClientBuilder serviceVersion(XXXServiceVersion serviceVersion) {
    this.serviceVersion = serviceVersion;
    return this;
}

private XXXClientImpl buildInnerClient() {
    if (serviceVersion == null) {
        this.serviceVersion = XXXServiceVersion.getLatest();
    }
    ...
    XXXClientImpl client =
            new XXXClientImpl(pipeline, serializerAdapter, endpoint, serviceVersion.getVersion());
    ...
```
If user does not set service version, it will use latest service version by default.
`ClientImpl` class uses String type `apiVersion`.
```
private final String apiVersion;
```


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/NewPlugin.java
```diff
@@ -80,6 +80,10 @@ public boolean getBooleanValue(String key, boolean defaultValue) {
         }
     }
 
+    public List<String> getListValue(String key) {
```
weidongxu-microsoft, line 83: Might not worth adding to `NewPlugin` if it only works for `String`. If you want to put it here, maybe make it more generic by using `List<T>`? Or just have the method as private in `JavaSettings`.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -545,6 +548,12 @@ public boolean shouldUseIterable() {
         return useIterable;
     }
 
+    private final List<String> serviceVersions;
+
+    public List<String> getServiceVersions() {
```
weidongxu-microsoft, line 553: Worth add some javadoc.

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -122,9 +122,29 @@ public final void write(ServiceClient serviceClient, JavaFile javaFile) {
                     javaBlock.line("this.pipelinePolicies = new ArrayList<>();");
                 });
             }
+
+            // ServiceVersion in builder
+            List<ServiceClientProperty> serviceClientProperties = new ArrayList<>(serviceClient.getProperties());
+            if (settings.isLowLevelClient()) {
+                for (int i = 0; i < serviceClientProperties.size(); i++) {
```
weidongxu-microsoft, line 129: nit. Might worth use `ListIterator`, so you can replace the value in-place (without the `set(i, ...)`).

javagen/src/main/java/com/azure/autorest/template/ServiceVersionTemplate.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void write(ServiceClient model, JavaFile javaFile) {
+        // imports
+        Set<String> imports = new HashSet<>();
+        imports.add("com.azure.core.util.ServiceVersion");
+        javaFile.declareImport(imports);
+
+        javaFile.javadocComment(comment -> {
+            comment.description("Service version");
```
weidongxu-microsoft, line 29: Could be improved to mention about the service?

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -122,9 +122,29 @@ public final void write(ServiceClient serviceClient, JavaFile javaFile) {
                     javaBlock.line("this.pipelinePolicies = new ArrayList<>();");
                 });
             }
+
+            // ServiceVersion in builder
+            List<ServiceClientProperty> serviceClientProperties = new ArrayList<>(serviceClient.getProperties());
```
jianghaolu, line 127: Should this logic be moved to ServiceClientMapper?

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
qwordy: Done

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/NewPlugin.java
qwordy: Change to inline code now

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
qwordy: The properties in service client are also used in `ClientImpl`. I need `apiVersion` in `ClientImpl` and `serviceVersion` in `ClientBuilder`. I create a new list here. `serviceClient` remains unmodified. We can also add `builderProperties` in `serviceClient`, then template code can be simplified.

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
qwordy: https://github.com/Azure/autorest.java/pull/1126#discussion_r681551103

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
jianghaolu: Is there a reason you need to keep it as `apiVersion` in `ClientImpl`? I know we send it as `api-version` on the wire but locally we can name everything as `serviceVersion` right?

qwordy: I changed `apiVersion` to `serviceVersion` in `ClientImpl` and `ClientBuilder`. Proxy method keeps `apiVersion`.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
qwordy: Added

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
qwordy: This code snippet is removed.

javagen/src/main/java/com/azure/autorest/template/ServiceVersionTemplate.java
qwordy: Added

PR #1140, opened 2021-08-12
Rename get to list in page method; Rename get to list in page method

qwordy:
Service version enum upper case
Rename get to list in page method

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
@@ -54,7 +54,7 @@ public class ProxyMethodMapper implements IMapper<Operation, Map<Request, ProxyM
 
         ProxyMethod.Builder builder = createProxyMethodBuilder()
                 .description(operation.getDescription())
-                .name(operation.getLanguage().getJava().getName())
+                .name(normalizeMethodName(operation))
                 .isResumable(false);
 
         List<HttpResponseStatus> expectedStatusCodes = operation.getResponses().stream()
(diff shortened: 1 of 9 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -369,4 +369,16 @@ private static ClassType getExceptionType(Response exception, JavaSettings setti
...
     }
+
+    private String normalizeMethodName(Operation operation) {
+        // https://azure.github.io/azure-sdk/java_introduction.html#service-methods
+        String name = operation.getLanguage().getJava().getName();
+        if (JavaSettings.getInstance().isLowLevelClient() &&
+                operation.getExtensions() != null &&
+                operation.getExtensions().getXmsPageable() != null &&
+                name.startsWith("get")) {
+            name = "list" + name.substring(3);
```
weidongxu-microsoft, line 380: This could be a risky processing that might result in naming conflict.

Also if you really want to do this, the better place would be in preprocess module, where it handles the naming for Java entities.

qwordy: Agree. I don't really want to do this. Swagger should guarantee high quality.

jianghaolu: We should not do this in codegen? Naming fixes should always be done outside the language. (Weidong if fluent can afford a blanket rename I'm okay with doing it only in fluent). This causes unpredictable behavior and often raises the question "Why is Java generating stuff differently?"

jianghaolu: If Swagger is fixable we should fix the Swagger. If for any reason we are not able to fix the Swagger, we should add a `transform` on a case-by-case basis.

weidongxu-microsoft: It is originated from the conversation here https://apiview.dev/Assemblies/Review/f2b2e0774afe4d15b1ed33d3c785dfd8/0fb2fb420d71488da88e72195ec0a7a7#com.azure.analytics.purview.account.CollectionsAsyncClient.public-PagedFlux%3CBinaryData%3E-getChildCollectionNames(String-collectionName,-RequestOptions-requestOptions)

If possible would avoid this. If really need to do this, prefer to use directive.

qwordy:
See https://github.com/Azure/autorest.java/pull/1143

Issue #1145, opened 2021-08-17
Redefining method interface


qwordy:
We have multiple overload for one method. Remove first one `get`. Keep `WithResponse` version.
```java
public BinaryData get(RequestOptions requestOptions) 
public Response<BinaryData> getWithResponse(RequestOptions requestOptions, Context context) 
```

PR #1151, opened 2021-08-17
Only keep WithResponse methods

qwordy:
Resolve https://github.com/Azure/autorest.java/issues/1145


weidongxu-microsoft (approved):
OK for me.

Though you can check `methodVisibility` method, which might be easier to read about which method does not need to be generated (return `NOT_GENERATE` for it).

PR #1154, opened 2021-08-18
v4 update ci for latest testserver

weidongxu-microsoft:
try to get coverage publish work for latest autorest.testserver

PR #1156, opened 2021-08-19
llc, refactor to use methodVisibility for some client methods

weidongxu-microsoft:
It still not apply to all client method. But IMO better than flags in code.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
@@ -496,7 +496,7 @@ public class ClientMethodMapper implements IMapper<Operation, List<ClientMethod>
                             null);
                 }
 
-                if (settings.getSyncMethods() != JavaSettings.SyncMethodsGeneration.NONE && !settings.isLowLevelClient()) {
+                if (settings.getSyncMethods() != JavaSettings.SyncMethodsGeneration.NONE) {
                     methods.add(builder
                             .returnValue(createSimpleAsyncReturnValue(operation, asyncReturnType, syncReturnType))
                             .name(proxyMethod.getSimpleAsyncMethodName())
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft:
test failure is what I've solved in v4 https://github.com/Azure/autorest.java/pull/1154

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -688,7 +686,13 @@ protected IType createProtocolPagedRestResponseReturnType() {
...
      */
     protected JavaVisibility methodVisibility(ClientMethodType methodType, boolean hasContextParameter) {
-        return VISIBLE;
+        if (JavaSettings.getInstance().isLowLevelClient()) {
+            return (methodType == ClientMethodType.SimpleAsync || methodType == ClientMethodType.SimpleSync)
+                    ? NOT_GENERATE
+                    : VISIBLE;
+        } else {
+            return VISIBLE;
+        }
```
qwordy, line 695: Two `VISIBLE` branches can be combined to one.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -513,48 +513,46 @@ public static ClientMethodMapper getInstance() {
...
                     if (settings.isContextClientMethodParameter()) {
-                        JavaVisibility visibility = methodVisibility(ClientMethodType.SimpleAsync, true);
-                        if (visibility != NOT_GENERATE) {
-                            addClientMethodWithContext(methods,
-                                    builder.methodVisibility(visibility),
-                                    parameters);
-                        }
+                        addClientMethodWithContext(methods,
+                                builder.methodVisibility(methodVisibility(ClientMethodType.SimpleAsync, true)),
+                                parameters);
```
qwordy, line 518: `settings.isLowLevelClient()` in outer if statement can be removed.

weidongxu-microsoft: good catch. removed.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
weidongxu-microsoft: Probably keep it, so it is easier to see what is configured for LLC and what is configure for default.

Issue #1147, opened 2021-08-17
Add client options to builder


qwordy:
Add client options to builder. `ClientOptions` class can config
```java
private Iterable<Header> headers;
private String applicationId;
```
Code in builder:
```java
String applicationId = CoreUtils.getApplicationId(buildClientOptions, buildLogOptions);
httpPipelinePolicies.add(new UserAgentPolicy(applicationId, CLIENT_NAME, CLIENT_VERSION, buildConfiguration));
...
HttpHeaders headers = new HttpHeaders();
buildClientOptions.getHeaders().forEach(header -> headers.set(header.getName(), header.getValue()));
if (headers.getSize() > 0) {
    httpPipelinePolicies.add(new AddHeadersPolicy(headers));
}

```

PR #1159, opened 2021-08-20
Add ClientOptions to builder

qwordy:
Resolve https://github.com/Azure/autorest.java/issues/1147
`ClientOptions` class can config
```java
private Iterable<Header> headers;
private String applicationId;
```
See generated code example in `protocol-tests`.


.gitignore
```diff
@@ -78,3 +78,6 @@ fluent-tests/pom_generated_*.xml
 
 # language server
 postprocessor/jdt-language-server
+
+# Module info
+protocol-tests/src/main/java/module-info.java
```
weidongxu-microsoft, line 83: nit, newline

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -340,6 +348,11 @@ protected void addCreateHttpPipelineMethod(JavaSettings settings, JavaClass clas
             function.line("policies.addAll(this.pipelinePolicies);");
             function.line("HttpPolicyProviders.addAfterRetryPolicies(policies);");
 
+            // clientOptions header
+            function.line("HttpHeaders headers = new HttpHeaders();");
+            function.line("clientOptions.getHeaders().forEach(header -> headers.set(header.getName(), header.getValue()));");
+            function.ifBlock("headers.getSize() > 0", block -> block.line("policies.add(new AddHeadersPolicy(headers));"));
```
weidongxu-microsoft, line 354: Should we add this before `RetryPolicy`? It seems the headers cannot change per request? Though in current data-plane code it seems always after `RetryPolicy`.

Also it might be need to be before the credential policy.

.gitignore
qwordy:
```suggestion
protocol-tests/src/main/java/module-info.java

```

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -305,15 +307,26 @@ protected void addCreateHttpPipelineMethod(JavaSettings settings, JavaClass clas
...
+
             function.line("List<HttpPipelinePolicy> policies = new ArrayList<>();");
 
 
             function.line("String clientName = properties.getOrDefault(SDK_NAME, \"UnknownName\");");
             function.line("String clientVersion = properties.getOrDefault(SDK_VERSION, \"UnknownVersion\");");
 
-            function.line("policies.add(new UserAgentPolicy(httpLogOptions.getApplicationId(), clientName, "
+            function.line("String applicationId = clientOptions.getApplicationId() == null ? " +
+                    "httpLogOptions.getApplicationId() : clientOptions.getApplicationId();");
```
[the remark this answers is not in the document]
qwordy: Updated

weidongxu-microsoft (approved):

PR #1161, opened 2021-08-20
Fix for multiple query string parameters with same name.

daniel-rocha:
Add support for collectionFormat multi that is implemented with explode parameter. After identifying that we need collectionFormat multi, we store it in the builder and add flag multipleQueryParams = true. The multiple parameters part  was implemented in https://github.com/Azure/azure-sdk-for-java/pull/21203 and has now been released.

This will fix the following: https://github.com/Azure/azure-sdk-for-java/issues/13124

Supersedes https://github.com/Azure/autorest.java/pull/1017


daniel-rocha:
Still has CI issues with a failing test. Please do not review it yet.

weidongxu-microsoft:
CI failure in fluent complains about issue related to SpotBugs check.

azure-sdk-for-java likely have same rule (unless you suppress it). You can try it first before fixing in codegen.

```
[ERROR] com.azure.mgmttest.afdx.implementation.LogAnalyticsClientImpl.lambda$getLogAnalyticsMetricsWithResponseAsync$10(String) invokes toString() method on a String [com.azure.mgmttest.afdx.implementation.LogAnalyticsClientImpl] At LogAnalyticsClientImpl.java:[line 379] DM_STRING_TOSTRING
```

weidongxu-microsoft:
fluent CI has an extra step to do similar check as done by azure-sdk-for-java repo.

Update the dependency to match that on azure-sdk-for-java https://github.com/Azure/autorest.java/pull/1163

weidongxu-microsoft:
Some of the service uses Enum for this, in this case the "toString" is needed. You might have to branch the write to handle both case (the type is String, and the type is not String).

https://github.com/Azure/azure-rest-api-specs/blob/master/specification/cdn/resource-manager/Microsoft.Cdn/stable/2020-09-01/afdx.json#L3396-L3416

weidongxu-microsoft:
@jianghaolu and @srnagar for review.

daniel-rocha:
@weidongxu-microsoft @srnagar @jianghaolu all checks have passed, please review.

vanilla-tests/src/test/java/fixtures/url/multi/QueriesTests.java
```diff
@@ -17,17 +18,21 @@ public static void setup() {
...
     @Test
     public void arrayStringMultiNull() {
-        client.getQueries().arrayStringMultiNull(null);
+        try {
+            client.getQueries().arrayStringMultiNull(null);
+            Assert.assertFalse(true); // should not reach here
+        }
+        catch (Exception e) {
+
+        }
```
weidongxu-microsoft, line 27: Code seems should not throw exception here.

The definition in swagger is `"x-nullable": true`, so I kind of assume method can take a `null`.
https://github.com/Azure/autorest.testserver/blob/main/swagger/url-multi-collectionFormat.json#L27-L39

weidongxu-microsoft (commented):
LGTM except the `null` case, when the parameter is not required.

vanilla-tests/src/test/java/fixtures/url/multi/QueriesTests.java
daniel-rocha: Good catch @weidongxu-microsoft . The code generated for arrayStringMultiNull() does not support a null parameter. It is trying to stream from a null array and causing an NPE. I will check where the code is generated and modify it to match the Swagger.

jianghaolu (approved):
LGTM once Weidong's comment is addressed

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -269,7 +269,7 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
         getReturnValue().addImportsTo(imports, includeImplementationImports);
 
         imports.add("com.azure.core.annotation.ServiceMethod");
-        imports.add("com.azure.core.annotation.ReturnType");
+        imports.add("com.azure.core.annotation.ReturnType");
```
weidongxu-microsoft, line 272: nit, remove the spaces after the line.

weidongxu-microsoft (approved):

daniel-rocha:
@weidongxu-microsoft @jianghaolu All comments addressed and checks passed. Last commit is running checks again but it's just removal of blank space, so they will all pass. Please merge when you see fit. Thanks for the help!

PR #1162, opened 2021-08-23
Add throws in javadoc

qwordy:
Tell users what exceptions will be thrown.


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -54,6 +54,7 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
         }
 
         clientMethod.getParameters().forEach(p -> commentBlock.param(p.getName(), methodParameterDescriptionOrDefault(p)));
+        commentBlock.methodThrows("HttpResponseException", "Thrown when status code is 400 or above if requestOptions is null or throwOnError in requestOptions is true");
```
weidongxu-microsoft, line 57:
```suggestion
        commentBlock.methodThrows("HttpResponseException", "thrown if status code is 400 or above, if throwOnError in requestOptions is not false");
```

jianghaolu (approved):

PR #1173, opened 2021-09-03
Polling Strategy Codegen

jianghaolu:


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -344,10 +347,33 @@ public static ClientMethodMapper getInstance() {
...
+                        && operation.getOperationGroup().getLanguage() != null
+                        && operation.getOperationGroup().getLanguage().getDefault() != null
+                        && operation.getOperationGroup().getLanguage().getDefault().getName() != null
+                        && !operation.getOperationGroup().getLanguage().getDefault().getName().isEmpty()) {
+                    operationId = operation.getOperationGroup().getLanguage().getDefault().getName() + "_"
+                            + operation.getLanguage().getDefault().getName();
+                } else {
+                    operationId = operation.getLanguage().getDefault().getName();
+                }
+                JavaSettings.PollingDetails pollingDetails = settings.getPollingConfig(operationId);
```
weidongxu-microsoft, line 361: In a recent change, I actually added `operationId` to `ProxyMethod`. Maybe it can be reused.
https://github.com/Azure/autorest.java/blob/v4/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java#L62-L75

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -684,6 +722,56 @@ private String getPageableItemName(Operation operation) {
...
+            } else {
+                intermediateTypeName = details.getIntermediateType();
+                intermediateTypePackage = JavaSettings.getInstance().getPackage();
+            }
+            pollResponseType = new ClassType.Builder().packageName(intermediateTypePackage).name(intermediateTypeName).build();
+        }
+        // azure-core wants poll response to be non-null
+        if (pollResponseType == ClassType.Void) {
+            pollResponseType = ClassType.BinaryData;
+        }
```
weidongxu-microsoft, line 745: Good point. Mgmt might able to adopt this as well.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -303,8 +313,27 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
...
 
         if (type == ClientMethodType.LongRunningBeginAsync) {
-            if (((GenericType) this.getReturnValue().getType().getClientType()).getTypeArguments()[0] instanceof GenericType) {
-                imports.add("com.fasterxml.jackson.core.type.TypeReference");
+            if (settings.isFluent()) {
+                if (((GenericType) this.getReturnValue().getType().getClientType()).getTypeArguments()[0] instanceof GenericType) {
+                    imports.add("com.fasterxml.jackson.core.type.TypeReference");
+                }
+            } else {
+                imports.add("com.azure.core.util.serializer.TypeReference");
```
weidongxu-microsoft, line 321: Just curious. Is there major difference between between `com.fasterxml.jackson.core.type.TypeReference` and `com.azure.core.util.serializer.TypeReference`? Maybe mgmt can switch to the core one as well?

weidongxu-microsoft (approved):

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -623,6 +635,52 @@ public boolean requireXMsFlattenedToFlatten() {
...
+        public String getFinalType() {
+            return finalType;
+        }
+
+        public int getPollInterval() {
+            return pollInterval != null
+                    ? Integer.parseInt(pollInterval)
+                    : 1;
+        }
+    }
```
weidongxu-microsoft, line 669: May signify the unit in name? 1 second, or 1 minute?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
jianghaolu: I think we probably prefer using classes from our own packages? I believe the `TypeReference` might be used before because that was the only one available.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
jianghaolu: I think in Mgmt you have your own `PollResult` type - I actually am envious of that. In data plane we do not have a common data structure all services agreed upon, so we have to use a `BinaryData` here. But if all services have a common type (like in mgmt), I think having a strong type may be more helpful.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
jianghaolu: Awesome!

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
weidongxu-microsoft: Yes, same thought. Probably when we start on mgmt, `TypeReference` in core did not exist yet. But if it works, we should switch to it.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
weidongxu-microsoft: `PollResult<BinaryData>` would be better than `PollResult<Void>`.

PR #1187, opened 2021-09-15
LLC Partial update

qwordy:
We define an annotation `com.azure.core.annotation.Manual`. It can annotate class method and field. It represents that the method or field is written manually and should be reserved after generation. I like `@Manual` rather than `@Generated` because we need to modify generator to add `@Generated` for all methods.
This PR adds logic before writing Java files. For each file,
1. Read content of the file if it exists. Extract manual code.
2. Merge manual code and new generated code.

In order to extract manual code, we need to understand the Java source code. In another word, parse it. We have 3 approaches, (1) write code to extract it on our own, e.g. pattern matching (2) use a Java parser (Eclipse, Spoon, JavaParser, etc.), create AST (3) use language server. I don't think (1) is reliable and robust. It is hard for me to cover all cases. (2) and (3) are better. This PR uses Spoon, a Java parser and code transformation tool. Experiments show that it can identify expected code.
Once we get the manual code, we can merge it with new generated code. We also need to understand and parse the new generated code. As we agreed before, we do not consider DELETE. We only care ADD and MODIFY. For each manual code snippet, do a signature matching among new generated code to identify whether the manual code snippet is added or is a modified version of existing method or field. If it is MODIFY, the generator will overwrite generated code with manual code. If it is ADD, just add it to class.

### DIFF
--- a/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
+++ b/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
@@ -139,7 +139,8 @@ public class JavaSettings
                     host.getStringValue("key-credential-header-name", ""),
                     host.getBooleanValue("disable-client-builder", false),
                     host.getBooleanValue("skip-formatting", false),
-                    host.getValue(new TypeReference<Map<String, PollingDetails>>() { }.getType(), "polling"));
+                    host.getValue(new TypeReference<Map<String, PollingDetails>>() { }.getType(), "polling"),
+                    host.getStringValue("output-folder", ""));
         }
         return _instance;
     }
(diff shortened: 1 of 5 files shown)


srnagar:
A couple of high-level comments:
- We are using Java Parser in [API view](https://github.com/Azure/azure-sdk-tools/blob/main/src/java/apiview-java-processor/pom.xml#L26) and in [build-tools](https://github.com/JonathanGiles/azsdkmvn/blob/master/build-tool/pom.xml#L81). So, we should stick to Java Parser if it can provide all the functionality.
- `@Manual` vs `@Generated`: the SDK developer has to remember to annotate every method with @Manual to ensure it doesn't get overwritten. There's a high possibility that they will miss this annotation and manual code gets overwritten. Using `@Generated` is a higher initial cost to update the generator but requires less work from the SDK developers. Most current libraries have manually written public APIs and if we have to include LLC methods to these clients in the future, it would require more effort to update all current methods with this annotation. Also, this requires adding one more annotation in azure-core.

weidongxu-microsoft:
Didn't aware that apiview uses JavaParser.

This lib in its own is quite OK. However last time I tried, the symbol resolver is not quite correct.

Issue #1185, opened 2021-09-13
`settings.getServiceName()` throws NPE


jianghaolu:
There are multiple usages of `settings.getServiceName()` in the generator, especially for building service version related components. `--service-name` is an optional argument which could lead to NullPointerException.

Examples include:
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/Javagen.java#L129
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java#L128

Alternative ways should be used from existing information in the code model.

qwordy:
@jianghaolu Thanks for finding this bug. I will fix it as soon as possible.

PR #1188, opened 2021-09-16
Fix service name is null; Throw exception if no service version

qwordy:
Resolve https://github.com/Azure/autorest.java/issues/1185

There are multiple usages of `settings.getServiceName()` in the generator, especially for building service version related components. `--service-name` is an optional argument which could lead to NullPointerException.
It will throw exception if no service version now.
Create an [issue](https://github.com/Azure/autorest.java/issues/1196), Load default service version from swagger file.


javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -142,7 +142,12 @@ protected NodeTuple representJavaBeanProperty(Object javaBean, Property property
             // Service version
             if (settings.isLowLevelClient() && settings.getServiceVersions() != null) {
                 String packageName = settings.getPackage();
-                String serviceName = settings.getServiceName().replaceAll("\\s", "");
+                String serviceName;
```
haolingdong-msft, line 145: do we need to add a test for the null case?

javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
```diff
@@ -125,7 +125,12 @@ public ServiceClient map(CodeModel codeModel) {
             String serviceClientPropertyDefaultValueExpression = serviceClientPropertyClientType.defaultValueExpression(p.getClientDefaultValue());
 
             if (settings.isLowLevelClient() && serviceClientPropertyName.equals("apiVersion")) {
```
weidongxu-microsoft, line 127: Do we need to test `settings.getServiceVersions() != null`?

javagen/src/main/java/com/azure/autorest/Javagen.java
qwordy: Yes, we do.

javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
qwordy: Yes, we do.

weidongxu-microsoft (approved):

haolingdong-msft (approved):

PR #1192, opened 2021-09-23
provide defaultValueExpression for IType

weidongxu-microsoft:
Case, some "null" as return value will be rejected by spotbugs.

```
    public byte[] data() {
        if (this.data == null) {
            return null;
        }
        return this.data.decodedBytes();
    }
```
will have
```
Error:  Low: Should com.azure.resourcemanager.keyvault.generated.models.KeyReleasePolicy.data() return a zero length array rather than null? [com.azure.resourcemanager.keyvault.generated.models.KeyReleasePolicy] At KeyReleasePolicy.java:[line 58] PZLA_PREFER_ZERO_LENGTH_ARRAYS
```

This fix would generate
```
    public byte[] data() {
        if (this.data == null) {
            return new byte[0];
        }
        return this.data.decodedBytes();
    }
```
instead.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/GenericType.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/GenericType.java
@@ -159,6 +159,11 @@ public class GenericType implements IType {
         return sourceExpression;
     }
 
+    @Override
+    public String defaultValueExpression() {
+        return "null";
+    }
+
     public final IType getClientType() {
         IType clientType = this;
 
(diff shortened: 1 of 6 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/GenericType.java
```diff
@@ -159,6 +159,11 @@ public final String defaultValueExpression(String sourceExpression) {
         return sourceExpression;
     }
 
+    @Override
+    public String defaultValueExpression() {
+        return "null";
```
weidongxu-microsoft, line 164: We might actually want to return empty collection when it is null from JSON. But avoid this unless necessary as this could be big behavior change.

qwordy (approved):

Issue #1194, opened 2021-09-24
Support LRO for LLC


qwordy:
LRO code is already merged in main. It does not support LLC yet. It should be compatible with LLC with modest effort.

PR #1195, opened 2021-09-24
Support LRO for LLC

qwordy:
Resolve https://github.com/Azure/autorest.java/issues/1194
LRO code is already merged in main. It does not support LLC yet. It should be compatible with LLC with modest effort.


weidongxu-microsoft (approved):

qwordy:
Fixed by another PR.

PR #1197, opened 2021-09-24
add default config for LLC

weidongxu-microsoft:
so dev does not need to duplicate these in README in SDK repo.

### DIFF
--- a/preprocessor/readme.md
+++ b/preprocessor/readme.md
@@ -15,9 +15,6 @@ pipeline:
   # "Shake the tree", and normalize the model
   modelerfour:
     input: openapi-document/multi-api/identity     # the plugin where we get inputs from
-    flatten-models: true
-    flatten-payloads: true
-    group-parameters: true
   
   # allow developer to do transformations on the code model.
   modelerfour/new-transform:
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -130,10 +130,13 @@ To use default settings globally, use `--polling={}`.
...
 
-```yaml $(low-level-client)
+```
 generate-client-interfaces: false
 generate-client-as-impl: true
 generate-sync-async-clients: true
+add-context-parameter: true
+context-client-method-parameter: true
+sync-methods: all
 ```
```
weidongxu-microsoft, line 140: Remove `yaml` from the block. This part is intended to be a description of possible configures, not the actual config for autorest.

preprocessor/readme.md
```diff
@@ -26,3 +23,27 @@ pipeline:
...
+add-context-parameter: true
+context-client-method-parameter: true
+sync-methods: all
+
+pipeline:
+  modelerfour:
+    flatten-models: false
+    flatten-payloads: false
+    group-parameters: false
+    lenient-model-deduplication: true
```
weidongxu-microsoft, line 48: flatten and group is not supported for LLC.

model is not used in LLC.

haolingdong-msft (approved):

qwordy (approved):

PR #1198, opened 2021-09-27
llc, remove sync API without Context

weidongxu-microsoft:
Do not generate e.g. `listAll(RequestOptions)`.

`listAll(RequestOptions, Context)` should be enough.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
@@ -351,7 +351,7 @@ public class ClientMethodMapper implements IMapper<Operation, List<ClientMethod>
                         }
 
                         if (settings.getSyncMethods() == JavaSettings.SyncMethodsGeneration.ALL) {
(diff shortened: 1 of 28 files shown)


haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -351,7 +351,7 @@ public static ClientMethodMapper getInstance() {
                         }
 
                         if (settings.getSyncMethods() == JavaSettings.SyncMethodsGeneration.ALL) {
-                            builder.methodVisibility(VISIBLE);
+                            builder.methodVisibility(methodVisibility(ClientMethodType.PagingSync, false));
```
haolingdong-msft, line 354: Do we need to update paging related tests in protocol-tests?

qwordy:
Protocol tests, `paging` should be changed. It is not in changed files.

haolingdong-msft (commented):
Do we need to update paging related tests in protocol-tests?

Issue #1200, opened 2021-09-27
Fix sample generation for optional request body


qwordy:
Generated sample does not set optional request body. Let's fix it.

PR #1214, opened 2021-10-13
refactor LLC example generator, bug fix of type expression error and missing body

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1200


javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java
```diff
@@ -103,22 +123,29 @@ public void write(ProtocolExample protocolExample, JavaFile javaFile) {
                     credentialExpr = "";
                 }
 
+                // client
                 String clientInit = "%s client = new %s()" +
                         ".%s(System.getenv(\"%s\"))" +
                         "%s" +
```
weidongxu-microsoft, line 129: Remaining issue on the filling of parameters that set as client property (like instanceId in device update).

PR #1231, opened 2021-10-25
Support HttpMethod OPTIONS

### LINKED ISSUE #453: [HttpInfrastructure] - OPTIONS not yet supported
Codegen for HttpInfrastructure does not support OPTIONS HTTP method yet because of a dependency in azure-core. `HttpMethod` enum in azure-core does not define this HTTP method in the current released version. PR https://github.com/Azure/azure-sdk-for-java/pull/6729 fixes azure-core but autorest will not be able to use it until the next version of core is released.

alzimmermsft:
Fixes #453

This PR adds support for `HttpMethod.OPTIONS` in code generation. Fixed by removing guard against the an HTTP method options as the generator has been updated to a version of `azure-core` which contains `OPTIONS` as a enum value in `HttpMethod`.


weidongxu-microsoft (approved):
I am updating core to 1.22.0 in PR https://github.com/Azure/autorest.java/pull/1247

PR #1239, opened 2021-11-02
Add @Generated annotation to all generated protocol methods #1099

haolingdong-msft:
pr for issue #1099 

code change:
add @Generated annotation to methods and fields in sync and async clients

example class:
customization-tests/src/main/java/fixtures/bodycomplex/ArrayAsyncClient.java


weidongxu-microsoft (approved):
You might had to check how customization works.

PR #1242, opened 2021-11-02
Use JacksonXmlText when Swagger Contains x-ms-text: true in XML Section

### LINKED ISSUE #1238: Incorrect Generation for Model with XML Attribute and Tag Value
Incorrect Jackson annotations are generated when an XML tag has both an attribute and the value in the same model. For example this XML:

```xml
<EnumerationResults ContainerName="mycontainer"> 
  <Blobs>   
    <Blob>   
      <Name Encoded="true">dir1%2Fdir2%2Ffile%EF%BF%BF.blob</Name>   
    </Blob>   
  </Blobs>   
</EnumerationResults>
```

Uses this Swagger definition:

```json
"BlobName": {
  "type": "object",
  "properties": {
    "Encoded": {
      "xml": {
        "attribute": true,
        "name": "Encoded"
      },
      "type": "boolean",
      "description": "Indicates if the blob name is encoded."
    },
    "content": {
      "xml": {
        "x-ms-text": true
      },
      "type": "string",
      "description": "The name of the blob."
    }
  }
}
```

And currently generates this (simplified) model:

```java
@JacksonXmlRootElement(localName = "BlobName")
@Fluent
public final class BlobName {
    /*
     * Indicates if the blob name is encoded.
     */
    @JacksonXmlProperty(localName = "Encoded", isAttribute = true)
    private Boolean encoded;

    /*
     * The name of the blob.
     */
    @JsonProperty(value = "content")
    private String content;
}
```

This fails to serialize and deserialize properly as Jackson attempts to find a sub-tag `content` within the current XML tag's scope. Instead the following should be generated:

```java
@JacksonXmlRootElement(localName = "BlobName")
@Fluent
public final class BlobName {
    /*
     * Indicates if the blob name is encoded.
     */
    @JacksonXmlProperty(localName = "Encoded", isAttribute = true)
    private Boolean encoded;

    /*
     * The name of the blob.
     */
    @JacksonXmlText
    private String content;
}
```

`JacksonXmlText` indicates to Jackson that the current XML tag's text value should be used for `String content`.

alzimmermsft:
Fixes #1238 

This PR resolves a code generation issue where Swagger `xml` configurations with `x-ms-test: true` were being improperly generated with `JsonProperty` instead of `JacksonXmlText`. This resulted in improper serialization and deserialization behaviors.

### DIFF
--- a/extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/XmlSerlializationFormat.java
+++ b/extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/XmlSerlializationFormat.java
@@ -2,23 +2,23 @@
 package com.azure.autorest.extension.base.model.codemodel;
 
 
+import java.util.Objects;
+
 public class XmlSerlializationFormat extends SerializationFormat {
 
     private String name;
     private String namespace;
     private String prefix;
     /**
-     * 
      * (Required)
-     * 
      */
     private boolean attribute;
     /**
-     * 
      * (Required)
-     * 
      */
     private boolean wrapped;
+
+    private boolean text;
 
     public String getName() {
         return name;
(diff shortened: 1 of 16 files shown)


extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/XmlSerlializationFormat.java
```diff
@@ -128,11 +131,16 @@ public boolean equals(Object other) {
         if (other == this) {
             return true;
         }
-        if ((other instanceof XmlSerlializationFormat) == false) {
+        if (!(other instanceof XmlSerlializationFormat)) {
             return false;
         }
         XmlSerlializationFormat rhs = ((XmlSerlializationFormat) other);
-        return ((((((this.name == rhs.name)||((this.name!= null)&&this.name.equals(rhs.name)))&&((this.namespace == rhs.namespace)||((this.namespace!= null)&&this.namespace.equals(rhs.namespace))))&&(this.attribute == rhs.attribute))&&(this.wrapped == rhs.wrapped))&&((this.prefix == rhs.prefix)||((this.prefix!= null)&&this.prefix.equals(rhs.prefix))));
+        return Objects.equals(this.name, rhs.name)
```
lmolkova, line 138: Alt Text

lmolkova (approved):

PR #1250, opened 2021-11-10
misc improvement

weidongxu-microsoft:
1. update to use the new POM properties
2. allow data-plane to use remote README
3. bug fix on LLC swagger README output, allow multiple JSON as input-file

### DIFF
--- a/postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
+++ b/postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
@@ -22,6 +22,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystemNotFoundException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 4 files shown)


postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -49,10 +50,16 @@ public boolean processInternal() {
...
+        readme = new String(Files.readAllBytes(Paths.get(new URI(readmePath))));
       }
-    } catch (IOException | URISyntaxException e) {
+    } catch (IOException | URISyntaxException | IllegalArgumentException e) {
+      logger.warn("Location of README is not valid", e);
       return false;
+    } catch (FileSystemNotFoundException e) {
+      // likely cause is the URI of the README is "https" scheme
+      logger.warn("File system of README not reachable, skip README", e);
+      // continue
```
weidongxu-microsoft, line 62: To me, it is not reasonable to fail because the README in not local file.

weidongxu-microsoft:
I probably will go ahead with the merge, as mgmt is starting to release the delayed SDKs for Ignite.

If there is suggestion about remote README in data-plane (though the positive part of not allow remote README, is that we lower the risk of caller loading untrusted code), I will take follow-up PRs.

PR #1260, opened 2021-11-18
Updates to Code Customization Framework

alzimmermsft:
This PR makes the following updates to the code customization framework:

- Updates the version of the Eclipse language server being used. (Provides a substantial 30-40% performance improvement when performing many customizations.)
- Incorporates an abstract syntax tree libraries to enable more complex and safer customizations.


customization-base/src/main/java/com/azure/autorest/customization/MethodCustomization.java
```diff
@@ -100,7 +104,22 @@ public MethodCustomization addAnnotation(String annotation) {
...
+            .getAnnotationByName(Utils.cleanAnnotationName(annotation));
+
+        if (annotationExpr.isPresent()) {
+            annotationExpr.get().remove();
+            editor.replaceFile(fileName, compilationUnit.toString());
+            Utils.sendFilesChangeNotification(languageClient, fileUri);
+            return refreshCustomization(methodSignature);
+        } else {
+            return this;
+        }
```
srnagar, line 122: This is very similar to the removeAnnotation method on the ConstructorCustomization class. We should use a utility method to reduce duplication.

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/models/SymbolKind.java
```diff
@@ -31,26 +34,32 @@
...
     TYPE_PARAMETER(26);
 
-    private int value;
+    private static final Map<Integer, SymbolKind> INT_TO_KIND_MAP;
+
+    static {
+        INT_TO_KIND_MAP = new HashMap<>();
+
+        for (SymbolKind kind : SymbolKind.values()) {
+            INT_TO_KIND_MAP.putIfAbsent(kind.value, kind);
```
srnagar, line 43: Why use `putIfAbsent` instead of just `put`? If the value of `SymbolKind` enum is a duplicate, should that be thrown as an error?

postprocessor/jdt-language-server/config_win/config.ini
```diff
@@ -1,10 +1,10 @@
...
 eclipse.p2.data.area=@config.dir/../p2
 eclipse.p2.profile=DefaultProfile
+osgi.bundles.defaultStartLevel=4
+eclipse.application=org.eclipse.jdt.ls.core.id1
+osgi.framework=file\:plugins/org.eclipse.osgi_3.17.100.v20211104-1730.jar
+osgi.framework.extensions=reference\:file\:org.eclipse.osgi.compatibility.state_1.2.500.v20210730-0750.jar
```
srnagar, line 10: We should move all these jars below into azsdkartifacts blob storage and convert them into regular maven dependencies (out of scope for this PR).

srnagar (approved):

customization-base/src/main/java/com/azure/autorest/customization/MethodCustomization.java
alzimmermsft: Yeah, I think we can add a utility method that takes an `Optional<AnnotationExpr>` and if present it does the ending logic.

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/models/SymbolKind.java
alzimmermsft: That's a good point, though I don't expect it to occur since the enum is fairly static

alzimmermsft:
Force merged as ancillary tests were failing due to a missing file

Issue #1050, opened 2021-06-08
Support partial updates to Java file to replace/update generated protocol methods


srnagar:
This is a follow-up task of https://github.com/Azure/autorest.java/issues/1049. Autorest should support updating a Java file that has a mixture of generated and handwritten methods based on the method annotation.

weidongxu-microsoft:
This probably is not easy for autorest. autorest.java send the content of the file to autorest for writing.

Fluent codegen did some tricky stuff to first read the file (relative to `output-folder`), then write a new one based on its previous content, and finally send it to autorest.
https://github.com/Azure/autorest.java/blob/v4/fluentgen/src/main/java/com/azure/autorest/fluent/model/projectmodel/Project.java#L290-L302
It is probably not the best solution for large numbers of java files.

.NET can use partial class, which is not available to Java.

qwordy:
Another approach to persist manual change is writing [customizations](https://github.com/Azure/autorest.java#customizations). It supports various types of customizations. However, it has limitations. It does not support adding custom code, e.g. changing the body of a method, adding a method. It is a common usage that add a wrapper on fundamental methods to provide high-level user experience.

qwordy:
Study of the issue.
https://microsoft-my.sharepoint.com/:w:/p/fey/EWSuAr4hPhxChdCtrxLOXYABIDpjWVZbKbhzP6bRTv3Z_A?e=UmmlRL

qwordy:
https://github.com/Azure/autorest.java/pull/1187

weidongxu-microsoft:
A few cases you need to cover in test:

### Manually add class variable and method
They need to be kept.

### Manually update method signature
E.g. https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/webpubsub/azure-messaging-webpubsub/src/main/java/com/azure/messaging/webpubsub/WebPubSubServiceClient.java#L496-L500
permission changed from `String` to `WebPubSubPermission`.

### Manually remove method
E.g. https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/webpubsub/azure-messaging-webpubsub/src/main/java/com/azure/messaging/webpubsub/WebPubSubServiceClient.java#L108-L111
method changed to package private to remove it from public access.

### Updated swagger adding new method
They need to be added.

### Updated swagger removing method
They need to be removed if not customized.

### Updated swagger update method signature
They need to be updated if not customized.

PR #1266, opened 2021-11-22
Add @Generated annotation to sync and async client itself

### LINKED ISSUE #1265: Add @Generated annotation for sync and async client class
issue https://github.com/Azure/autorest.java/issues/1050 only adds annotation @Generated to client's protocol methods. 
In order to make it easier to identify which class is generated, we want to add @Generated annotation to the client class itself

haolingdong-msft:
Fix #1265

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ServiceAsyncClientTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ServiceAsyncClientTemplate.java
@@ -54,6 +54,7 @@ public class ServiceAsyncClientTemplate implements IJavaTemplate<AsyncSyncClient
         comment.description(String.format("Initializes a new instance of the asynchronous %1$s type.",
             serviceClient.getInterfaceName())));
 
+    javaFile.annotation("Generated");
     javaFile.annotation(String.format("ServiceClient(builder = %s.class, isAsync = true)", builderClassName));
     javaFile.publicFinalClass(asyncClassName, classBlock ->
     {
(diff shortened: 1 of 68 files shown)


weidongxu-microsoft (approved):

haolingdong-msft:
@srnagar
I'm working on supporting partial update (https://github.com/Azure/autorest.java/issues/1050), we will first support adding new methods or modify existing generated methods in **sync and async clients**. 
Hence, I will first check if the class is auto generated and has @Generated annotation, if it has, then the class is in scope of partial update. Then, handle partial update for this class. 
I tried to add @Generated annotation to the class to make it easier to identify if the class is auto generated in this pr, example as below. But encounter the error that: @Generated annotation type is not applicable to class. It is only applicable to ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.FIELD. 
May I know do we have plan to make @Generated annotation applicable to class? If not, I can close this pr and I can iterate the methods to see if any method has @Generated annotation to identify class in scope of partial update. 
```java
@Generated
@ServiceClient(builder = LlcClientBuilder.class, isAsync = true)
public final class LlcAsyncClient {
    @Generated private final ParamsImpl serviceClient;
...
}
```

```java
@Retention(RetentionPolicy.SOURCE)
@Target({ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.FIELD})
public @interface Generated {
}
```

srnagar:
@haolingdong-msft currently, the `@Generated` annotation is not supported for classes. We want it to be granular at the method level as the new client APIs are generally convenience overloads and adding the `@Generated` annotation to the class can accidentally overwrite these manually added convenience methods when code is regenerated.

haolingdong-msft:
> @haolingdong-msft currently, the `@Generated` annotation is not supported for classes. We want it to be granular at the method level as the new client APIs are generally convenience overloads and adding the `@Generated` annotation to the class can accidentally overwrite these manually added convenience methods when code is regenerated.

It makes sense. I will close this pr and iterate the methods under the class to identify classes in scope of partial update. Thanks @srnagar

PR #1268, opened 2021-11-24
Add Support for Custom Http Exception Type Mappings

alzimmermsft:
This PR adds new configurations to AutoRest to enable supporting custom HTTP exception type mappings.

- `--default-http-exception-type` allows for the configuration of the default HTTP exception type.
- `--use-default-http-status-code-to-exception-type-mapping` adds support for a default status code to exception type mapping based on well-known status code to `azure-core` exception types.
- `--http-status-code-to-exception-type-mapping` allows for the configuration of a custom status code to exception type mapping.


weidongxu-microsoft (commented):
Good job! I actually considered to do this code to default exception mapping in mgmt, but then we settled to a single `ManagementException` because of class inherit hierarchy problem.

There appears to be multiple CI failure, maybe you want to double check before we do review?

I will

weidongxu-microsoft:
One discussion is whether we prefer new setting be grouped (like current `polling` option) by functionality, e.g. group by "exception-handling" here.

weidongxu-microsoft (approved):

Issue #1255, opened 2021-11-15
Add basic checkstyle and spotbugs


weidongxu-microsoft:
May start from a minimum check, as historically there is too many code that breaks a higher requirement.

- https://github.com/Azure/autorest.java/pull/1327
- https://github.com/Azure/autorest.java/pull/1331

PR #1271, opened 2021-11-29
Generate Strongly-Typed Headers Classes with Deferred Deserialization

alzimmermsft:
This PR contains a major update with how strongly-typed headers classes are handled during deserialization. A new AutoRest flag `defer-strongly-typed-header-deserialization` has been added that indicates to code generation to generate header classes with deferred deserialization. Before this change, ever response with strongly-typed headers would serialize the HTTP headers to JSON to deserialize the JSON into the strongly-typed headers object. Now, when this flag is enabled the strongly-typed headers will skips that serialization-deserialization and retain a pointer to the HTTP headers returned in the response. When the properties on the headers class are accessed a check will be made if the property has been deserialized in the past and if it hasn't the backing header will be converted to the property type and cached. So, for cases where the deserialized headers aren't returned to the caller or aren't regularly used by the caller there is no additional conversion overhead.

### DIFF
--- a/androidgen/src/main/java/com/azure/autorest/android/model/clientmodel/AndroidClientModel.java
+++ b/androidgen/src/main/java/com/azure/autorest/android/model/clientmodel/AndroidClientModel.java
@@ -12,7 +12,7 @@ public class AndroidClientModel extends ClientModel {
                                  boolean isPolymorphic, String polymorphicDiscriminator, String serializedName, boolean needsFlatten,
                                  String parentModelName, java.util.List<com.azure.autorest.model.clientmodel.ClientModel> derivedModels, String xmlName, String xmlNamespace,
                                  java.util.List<com.azure.autorest.model.clientmodel.ClientModelProperty> properties, java.util.List<com.azure.autorest.model.clientmodel.ClientModelPropertyReference> propertyReferences,
-                                 IType modelType) {
+                                 IType modelType, boolean stronglyTypedHeader) {
         super(package_Keyword,
                 name,
                 imports,
@@ -27,7 +27,8 @@ public class AndroidClientModel extends ClientModel {
                 xmlNamespace,
                 properties,
                 propertyReferences,
-                modelType);
+                modelType,
+                stronglyTypedHeader);
     }
 
     @Override
@@ -72,7 +73,8 @@ public class AndroidClientModel extends ClientModel {
                     xmlNamespace,
                     properties,
                     propertyReferences,
-                    modelType);
+                    modelType,
+                    stronglyTypedHeader);
         }
     }
 }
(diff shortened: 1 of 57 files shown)


weidongxu-microsoft (approved):

weidongxu-microsoft:
Getting a few string would be cheap. I assume the major goal in this PR is to delay the JSON to object deserialization in header? But what services do send a big JSON in header?

alzimmermsft:
> I assume the major goal in this PR is to delay the JSON to object deserialization in header? 

This is correct, the goal is to defer the conversion of response HTTP headers into a strongly-typed object defined in Swagger.

> But curious on what service does send a big JSON in header?

Storage leverages a lot of headers, sometimes upwards of 20-25, in responses, but this change is less on how big the HTTP headers and strongly-typed object is and more on the cost to do the conversion and how often the strongly-typed header object is. In many APIs the generated strongly-typed headers object is never returned to the user and may be used internally, so this PR is to remove those wasted CPU cycles and defer it until we know exactly when it is being used. Additionally, the previous method of conversion in azure-core was expensive requiring the building of a JSON tree to tear it down into a POJO which is non-trivial.

The table below outlines performance when accessing a single header value (in this case the Date header) from the response using a few different cases.

- Current: The current handling which converts the HTTP headers into a JSON tree then converts the JSON tree into the headers class.
- ObjectMapper Deserialization on Access: Uses JacksonAdapter to convert the header string value into the property type on access and caches that value.
- Custom Deserialization on Access: Uses the logic in this PR which converts well-known types with custom handling (ex DateTimeRfc1123 uses its constructor) on access and caches that value.

| Case | ns/op | Deviation (ns) |
| ----- | ------- | --------------- |
| Current | 9402.008 | 109.399 |
| ObjectMapper Deserialization on Access | 409.809 | 20.296 |
| Custom Deserialization on Access | 206.514 | 5.464 |

Although this example is an idealized case and accessing all properties may end up being a wash in terms of performance it is deemed unlikely that all header properties will be access eagerly. This case also assumes that downstream usage of the deferred headers class aren't accessed eagerly in our own code which isn't true but that can be changed.

alzimmermsft:
This PR will be merged once @g2vinay has real world performance numbers from Storage performance testing.

weidongxu-microsoft:
@alzimmermsft Maybe you could merge this, as the PR touches many codes.

We are still in the process of progressively enable more checkstyle for the codebase (with modification on existing code when necessary), which could cause more conflict with unmerged PR.
Issue: https://github.com/Azure/autorest.java/issues/1255

alzimmermsft:
> @alzimmermsft Maybe you could merge this, as the PR touches many codes.
> 
> We are still in the process of progressively enable more checkstyle for the codebase (with modification on existing code when necessary), which could cause more conflict with unmerged PR. Issue: #1255

Let me close this PR and reduce the scope of changes to those that are functional only

weidongxu-microsoft:
> Let me close this PR and reduce the scope of changes to those that are functional only. This change may be sitting for a while longer while more in-depth performance and stress testing is done to ensure it generates as expected

Alternative, you can go ahead with the merge (and improve later after testing data ready) as this feature would be behind an option.

alzimmermsft:
Closed for #1395

PR #1273, opened 2021-11-30
Support partial update

haolingdong-msft:
This pr fix issue https://github.com/Azure/autorest.java/issues/1050
In this pr, I supported partial update for .java files with `@Generated` annotation, currently they are async and sync clients. We already supported add `@Generated` annotation to members in async and sync clients in this pr: https://github.com/Azure/autorest.java/pull/1239.

I add a flag `partial-update` to enable partial update feature.

### Use Cases

I supported below partial update use cases in this pr:
1. Manually add class member (field / method / constructor) -> keep the added member
2. Manually update method signature, e.g. parameter change, method access level change -> keep the manual changed signature, and not generate the corresponding method with the same method name
3. Manually remove one class member -> if the member's definition is in swagger, this member will be auto generated again
4. Swagger add a new api -> add the new api to generated file
5. Swagger update an existing api -> if the api is auto generated, then the existing generated member will be replaced to the new one. If it is manual updated, we will keep the manual updated member.
6. Swagger delete an existing api -> if the existing api is auto generated, then it should be removed. If it is manual updated, we will keep the manual updated member.
	
### Code Change

The code change is mainly in Postprocessor.java and PartialUpdateHandler.java. It will handle partial update for each .java file after finishing customization. When handling partial update for each file, it will compare **existing file** which has manual update and **generated file** which is generated by autorest. We will keep existing file's format as much as possible.

Detailed logic:
1. Parse existing file content and generated file content using JavaParser
2. Get class members for existing file and generated file
3. Check if the file is in scope of partial update by iterate the members in generated file to see if there is a method has `@Generated` annotation. If it has `@Generated` annotation, then the file is in scope of partial update, otherwise return generated file content directly.
	**A potential issue here**: if existing file generated before we support adding `@Generated` annotation, then the existing file will be considered as all manual code change. So existing file members will be kept. One solution is to add a flag to determine whether we need to do partial update. Another solution is to check if existing file has `@Generated` annotation, but an edge case here is when use manually update all existing file, then existing file does not have `@Generated` annotation, but we still want to handle partial update. Currently I use the first solution to add a flag to enable partial update feature.
4. Iterate existing file members, keep manual updated members, and replace generated members with the corresponding newly generated one. Here we will not do the replace on the existing file member list,  we just create a new member list `updatedMembersList` and put in those manually update members and newly generated members.
5. Add remaining newly generated members to `updatedMembersList`
6. Update generated file members with `updatedMembersList`
7. Update generated file imports

### Unit Test

Unit test is in [PartialUpdateHandlerTest.java](https://github.com/Azure/autorest.java/blob/228cc14d3988b8979c743b23cc198fa27ea4cfa4/postprocessor/src/test/java/com/azure/autorest/postprocessor/util/PartialUpdateHandlerTest.java). it tests below cases:

1. Existing file manually add class member (field / method / constructor)
2. Existing file manually update method signature - parameter change and access level change
3. Existing file manually remove method
4. Generated file add API
5. Generated file remove API
6. Generated file update API
7. Generated file update API and the method is also manually updated in existing file -> keep manual update one
8. Generated file remove API and the method is also manually updated in existing file -> keep manual update one

### Integration Test

I also added integration tests to test the overall process for partial update in partial-update-tests. The test will generate code based on existing folder and [bodystring_updated.json](https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/swagger/specification/bodystring_updated.json). The existing files are in `existing` folder and the generated files are in `generated` folder. `StringOperationClient.java` covers manually code change on existing file, `EnumClient.java` covers swagger change.

Detailed test cases:
1. Manually add class member (field / method / constructor)
* existing code change:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java#L23-L24
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java#L36-L39
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java#L275

* Corresponding generated code: https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/generated/src/main/java/fixtures/bodystring/StringOperationClient.java

2. Manually update method signature - parameter change
* existing code change:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java#L60-L78

* Corresponding generated code: https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/generated/src/main/java/fixtures/bodystring/StringOperationClient.java

3. Manually update method signature -  method accessibility level change
* existing code change:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java#L80-L98

* Corresponding generated code: https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/generated/src/main/java/fixtures/bodystring/StringOperationClient.java

4. Swagger update api - method parameter name change 
*  swagger change:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/swagger/specification/bodystring_updated.json#L299

* Corresponding generated code:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/generated/src/main/java/fixtures/bodystring/EnumClient.java

5. Swagger add new api
* swagger change:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/swagger/specification/bodystring_updated.json#L430

* Corresponding generated code:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/generated/src/main/java/fixtures/bodystring/EnumClient.java

6. Swagger remove api
* swagger change:
Below api from bodystring.json is removed from bodystring_updated.json
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/existing/swagger/specification/bodystring.json#L382-L406

* Corresponding generated code:
https://github.com/Azure/autorest.java/blob/8d8d6d401f387ca085c0681c136da650ed653cfd/partial-update-tests/generated/src/main/java/fixtures/bodystring/EnumClient.java


postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -237,4 +248,147 @@ private String getBaseDirectory() {
             Utils.deleteDirectory(tempDirWithPrefix.toFile());
         }
     }
+
+    private void handlePartialUpdate(Map<String, String> fileContents) {
```
weidongxu-microsoft, line 252: Why not put them to a separate class? At least it will be easier for your unit test later.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -131,6 +141,7 @@ public boolean processInternal() {
     }
 
     private void writeToFiles(Map<String, String> fileContents) throws FormatterException {
+        handlePartialUpdate(fileContents);
```
weidongxu-microsoft, line 144: You might add a setting to control whether to enable the feature.

pom.xml
```diff
@@ -128,6 +128,7 @@
     <module>./fluentnamer</module>
     <module>./androidgen</module>
     <module>./protocol-resilience-test</module>
+    <module>./partial-update-tests</module>
```
weidongxu-microsoft, line 131: Add the folder to https://github.com/Azure/autorest.java/blob/main/.npmignore

weidongxu-microsoft (commented):
Could you show us where is your customization on the code?

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
haolingdong-msft: Agree, I also plan to do this refactoring, just first check in to show the logic.

haolingdong-msft:
> Could you show us where is your customization (hand-written change) on the test?

Enhanced 'Test' section, please refer to the 'Test' section for my hand-written change

partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java
```diff
@@ -0,0 +1,298 @@
...
+import fixtures.bodystring.implementation.EnumsImpl;
+import fixtures.bodystring.implementation.StringOperationsImpl;
+
+/** Initializes a new instance of the synchronous AutoRestSwaggerBatService type. */
+@ServiceClient(builder = AutoRestSwaggerBatServiceBuilder.class)
+public final class StringOperationClient {
+
+    @Generated private final StringOperationsImpl serviceClient;
+
+    // 1. manually add one field
```
haolingdong-msft, line 23: manually code change is in this file.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
haolingdong-msft: working on adding a flag to enable partial update feature

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
haolingdong-msft: Done  with refactoring

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -132,7 +132,9 @@ public static JavaSettings getInstance() {
                 getStringValue(host, "default-http-exception-type"),
                 getBooleanValue(host, "use-default-http-status-code-to-exception-type-mapping", false),
                 host.getValue(new TypeReference<Map<Integer, String>>() {}.getType(),
-                    "http-status-code-to-exception-type-mapping"));
+                    "http-status-code-to-exception-type-mapping"),
+                getBooleanValue(host, "handle-partial-update",  false)
```
weidongxu-microsoft, line 136: nit, space

generate
```diff
@@ -92,3 +92,7 @@ rm ./protocol-resilience-test/llcupdate1/src/main/java/module-info.java
 
 # customization
 autorest --version=3.6.6 --use:. customization-tests/swagger
+
+# partial update tests
+autorest --version=3.6.6 --use:. partial-update-tests/existing/swagger/README.md --handle-partial-update
```
weidongxu-microsoft, line 97: suggestion, move `--handle-partial-update` to your swagger/readme?

partial-update-tests/existing/swagger/README.md
```diff
@@ -0,0 +1,18 @@
...
+generate-client-as-impl: true
+namespace: fixtures.bodystring
+generate-client-interfaces: false
+sync-methods: all
+generate-sync-async-clients: true
+license-header: MICROSOFT_MIT_SMALL
+add-context-parameter: true
+models-subpackage: implementation.models
+context-client-method-parameter: true
+regenerate-pom: false
```
weidongxu-microsoft, line 16: these seems not needed, exception `namespace`

postprocessor/src/test/resources/partialupdate/StringOperationWithAddedMemberClient.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     *
+     * @param requestOptions The options to configure the HTTP request before HTTP client sends it.
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @return the response.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Response<Void> putNullWithResponse(RequestOptions requestOptions) {
```
weidongxu-microsoft, line 48: should this be `@Generated`?

postprocessor/src/test/resources/partialupdate/StringOperationWithUpdateMemberClient.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * 2. manually update method signature
+     *
+     * @param requestOptions The options to configure the HTTP request before HTTP client sends it.
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @return the response.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    Response<Void> putNullWithResponse(RequestOptions requestOptions, String test) {
```
weidongxu-microsoft, line 41: Is this for `StringOperationWithUpdateMemberGeneratedClient.java`? If so, should the method be package private (instead of public) in `StringOperationWithUpdateMemberGeneratedClient.java`?

weidongxu-microsoft (approved):

postprocessor/src/test/resources/partialupdate/StringOperationWithUpdateMemberClient.java
haolingdong-msft: `StringOperationWithUpdateMemberGeneratedClient.java`: it simulates swagger update api to add a parameter named `String updateParam` to `putNullWithResponse`. But does not change access level, so access level is still public.
`StringOperationWithUpdateMemberClient.java`: it simulates manual code change to add a parameter named `String test` and modify access level from `public` to `package-private`

postprocessor/src/test/resources/partialupdate/StringOperationWithAddedMemberClient.java
haolingdong-msft: nice catch, shoud be @Generated,  will update, thanks

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
haolingdong-msft: updated

generate
haolingdong-msft: updated

partial-update-tests/existing/swagger/README.md
haolingdong-msft: updated

haolingdong-msft:
@srnagar Could you please take a look at this pr and let me know if any comment?

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -132,7 +132,9 @@ public static JavaSettings getInstance() {
                 getStringValue(host, "default-http-exception-type"),
                 getBooleanValue(host, "use-default-http-status-code-to-exception-type-mapping", false),
                 host.getValue(new TypeReference<Map<Integer, String>>() {}.getType(),
-                    "http-status-code-to-exception-type-mapping"));
+                    "http-status-code-to-exception-type-mapping"),
+                getBooleanValue(host, "handle-partial-update", false)
```
srnagar, line 136: nit: I think we can drop `handle` from the name. Just `partial-update` should be sufficient.

partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java
```diff
@@ -0,0 +1,298 @@
...
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @return the response.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Response<Void> putBase64UrlEncodedWithResponse(BinaryData stringBody, RequestOptions requestOptions) {
+        return this.serviceClient.putBase64UrlEncodedWithResponse(stringBody, requestOptions);
+    }
+
+    public String manuallyAddedMethod() {
```
srnagar, line 275: Can you also add an annotation here and also include JavaDocs to ensure, it doesn't get overwritten?

partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java
```diff
@@ -0,0 +1,298 @@
...
+     *
+     * <pre>{@code
+     * Base64Url
+     * }</pre>
+     *
+     * @param requestOptions The options to configure the HTTP request before HTTP client sends it.
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @return null value that is expected to be base64url encoded.
+     */
+    @Generated
```
srnagar, line 292: Are we using `@Generated` annotation for `Client` classes only? What about builders? We may manually add some builder options that don't get generated. For e.g. some services support `connectionString` which we don't generate but could be added manually.

postprocessor/src/main/java/com/azure/autorest/postprocessor/util/PartialUpdateHandler.java
```diff
@@ -0,0 +1,135 @@
...
+import com.github.javaparser.ast.body.TypeDeclaration;
+import com.github.javaparser.ast.expr.SimpleName;
+import com.github.javaparser.ast.nodeTypes.NodeWithSimpleName;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.github.javaparser.StaticJavaParser.parse;
+
+public class PartialUpdateHandler {
```
srnagar, line 16: The documentation you have in the PR description is good. We should include that as JavaDoc for this class.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -237,4 +242,30 @@ private String getBaseDirectory() {
...
+                String projectBaseDirectoryPath = new File(getBaseDirectory()).getParent();
+                URI existingFilePath = Paths.get(projectBaseDirectoryPath, path).toUri();
+                // check if existingFile exists, if not, no need to handle partial update
+                if (Files.exists(Paths.get(existingFilePath))) {
+                    try {
+                        String existingFileContent = new String(Files.readAllBytes(Paths.get(existingFilePath)));
+                        String updatedContent = PartialUpdateHandler.handlePartialUpdateForFile(generatedFileContent, existingFileContent);
+                        return updatedContent;
+                    } catch (Exception e) {
+                        logger.error("Unable to get content from file path", e);
```
srnagar, line 262: If there was an error, we should throw the exception and not continue as this would overwrite an existing file that may have required a partial update.

postprocessor/src/main/java/com/azure/autorest/postprocessor/util/PartialUpdateHandler.java
```diff
@@ -0,0 +1,135 @@
...
+        if (!hasGeneratedAnnotations) {
+            return generatedFileContent;
+        }
+
+        NodeList<BodyDeclaration<?>> updatedMembersList = new NodeList<>();
+        // 2. iterate existingFileMembers, keep manual written members, and replace generated members with the corresponding newly generated one
+        for (BodyDeclaration<?> existingMember : existingFileMembers) {
+            boolean isGeneratedMethod = hasGeneratedAnnotation(existingMember);
+            if (!isGeneratedMethod) { // manual written member
+                updatedMembersList.add(existingMember);
```
srnagar, line 46: We should call this list `manuallyAddedMembersList` or something similar to make it explicit that these are the members that were not generated.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -237,4 +242,30 @@ private String getBaseDirectory() {
...
+        logger.info("Begin handle partial update...");
+        // handle partial update
+        // currently only support add additional interface or overload a generated method in sync and async client
+        fileContents.replaceAll((path, generatedFileContent) -> {
+            if (path.endsWith(".java")) { // only handle for .java file
+                // get existing file path
+                String projectBaseDirectoryPath = new File(getBaseDirectory()).getParent();
+                URI existingFilePath = Paths.get(projectBaseDirectoryPath, path).toUri();
+                // check if existingFile exists, if not, no need to handle partial update
+                if (Files.exists(Paths.get(existingFilePath))) {
```
srnagar, line 256: Why do we need to get the URI? Can we just do the following?
```java
if (Files.exists(Paths.get(projectBaseDirectoryPath, path))) 
```

postprocessor/src/test/resources/partialupdate/StringOperationWithRemovedMemberGeneratedClient.java
```diff
@@ -0,0 +1,32 @@
...
+import com.azure.core.exception.HttpResponseException;
+import com.azure.core.http.rest.RequestOptions;
+import com.azure.core.http.rest.Response;
+import fixtures.bodystring.implementation.StringOperationsImpl;
+
+/** Initializes a new instance of the synchronous AutoRestSwaggerBatService type. */
+@ServiceClient(builder = AutoRestSwaggerBatServiceBuilder.class)
+public final class StringOperationWithRemovedMemberGeneratedClient {
+
+    @Generated private final StringOperationsImpl serviceClient;
```
srnagar, line 20: nit: it would be better to have `@Generated` in a separate line. It's the pattern we use for `@JsonProperty` annotation too.

partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java
haolingdong-msft: Currently, we only include `@Generated` for `Client` classes. We can extend to Builders later.

srnagar: I guess the builder, sync and async types should have `@Generated` annotations right from the beginning as those are the most likely user-facing types that will be manually edited.

haolingdong-msft: OK, thanks for pointing out, in this issue https://github.com/Azure/autorest.java/issues/1099, we noted down that we will support only for clients as first step, so I did not add to builders. But I will create a issue to add `@Generated` annotation to builders and work on that as my next task. 
Do you think we should block this pr merge until I finish adding the annotation to builders, or we can merge this pr first ? ( I will modify this pr according to the comment and let you review again before merging)

haolingdong-msft: issue created: https://github.com/Azure/azure-sdk-for-java/issues/25895

weidongxu-microsoft: At present dev can use `--disable-client-builder` to avoid generate the Builder, if they've done customization. Builder itself is not commonly updated during swagger update.

Handling of Builder will be handled in a next PR. We do not want to have a super big PR for everything.

postprocessor/src/main/java/com/azure/autorest/postprocessor/util/PartialUpdateHandler.java
haolingdong-msft: Because `updatedMemberList` is the member list that we will set to `generatedClazz`, 
```java
generatedClazz.setMembers(updatedMembersList);
``` 
It not only contains manually added member, but also contains generated members, 'Code Change' section in description explains more details about `updatedMemberList`

postprocessor/src/test/resources/partialupdate/StringOperationWithRemovedMemberGeneratedClient.java
haolingdong-msft: In `ServiceSyncClientTemplate` (https://github.com/haolingdong-msft/autorest.java/blob/e4405a9c646d4a1306b9fb0eafd58049dfd42925/javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientTemplate.java), we add `@Generated` for member variables **in a seperate line**. But after formatting, `@Generated` comes to the same line, it only happens to member variable declaration. So may need more investigation on 
 the formatter if we want `@Generated` to be in a seperate line.
```java
// Add service client member variable
      classBlock.annotation("Generated"); // This hold a seperate line
      if (wrapServiceClient) {
        classBlock.privateFinalMemberVariable(serviceClient.getClassName(), "serviceClient");
      } else {
        classBlock.privateFinalMemberVariable(methodGroupClient.getClassName(), "serviceClient");
      }
```

weidongxu-microsoft: @srnagar This is due to your javaformat. If you want this format, please update it.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
haolingdong-msft: updated

postprocessor/src/main/java/com/azure/autorest/postprocessor/util/PartialUpdateHandler.java
haolingdong-msft: updated

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
haolingdong-msft: updated

partial-update-tests/existing/src/main/java/fixtures/bodystring/StringOperationClient.java
haolingdong-msft: updated

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
haolingdong-msft: updated

PR #1274, opened 2021-12-01
Use JacksonXmlElementWrapper Instead of a Static Inner Wrapper Class

### LINKED ISSUE #934: [FEATURE REQ] Support JacksonXmlElementWrapper for Wrapped Collections
Currently, when generating XML based APIs, wrapped collections (`<wrapperNode><node>..</node>...</wrapperNode>`) are generated using a wrapper class to contain the elements, such as:

```java
private static final class AuthorizationRulesWrapper {
    @JacksonXmlProperty(localName = "AuthorizationRule",
        namespace = "http://schemas.microsoft.com/netservices/2010/10/servicebus/connect")
    private final List<AuthorizationRuleImpl> items;
    @JsonCreator
    private AuthorizationRulesWrapper(
            @JacksonXmlProperty(localName = "AuthorizationRule",
                namespace = "http://schemas.microsoft.com/netservices/2010/10/servicebus/connect")
                List<AuthorizationRuleImpl> items) {
        this.items = items;
    }
}
/*
 * Authorization rules for resource.
 */
@JacksonXmlProperty(
        localName = "AuthorizationRules",
        namespace = "http://schemas.microsoft.com/netservices/2010/10/servicebus/connect")
private AuthorizationRulesWrapper authorizationRules;
```

This style unfortunately leads Jackson to handle `authorizationRules` as a POJO given that typing information doesn't clearly show the actual XML type is a collection. From here, Jackson uses a more generalized approach to deserializing `authorizationRules` which could lead it to use incorrect assumptions. For example, this caused an issue when upgrading Jackson where the value was treated as a string instead of a collection, fixed with configurations but if it was known as a collection no configuration changes would have been required.

The `JacksonXmlElementWrapper` annotation is used for cases where wrapped collections exist and provides better metadata to Jackson to allow it to hook into collection specific serialization configurations. This would be the equivalent generation:

```java
/*
 * Authorization rules for resource.
 */
@JacksonXmlElementWrapper(
        localName = "AuthorizationRules",
        namespace = "http://schemas.microsoft.com/netservices/2010/10/servicebus/connect")
private List<AuthorizationRuleImpl> authorizationRules;
```

alzimmermsft:
Fixes #934 

This PR updates XML code generation to use `JacksonXmlElementWrapper` instead of a private static inner class to maintain the wrapped element list. The change follows better Jackson XML serialization practices as it uses an annotation that Jackson leverages to provide the best serialization experience instead of the more generalized JsonProperty. It also clues Jackson in that the deserialization type is an enumeration whereas with JsonProperty this cannot be determined.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
@@ -11,6 +11,7 @@ import com.azure.autorest.model.clientmodel.ClientModelProperty;
 import com.azure.autorest.model.clientmodel.ClientModelPropertyAccess;
 import com.azure.autorest.model.clientmodel.ClientModelPropertyReference;
 import com.azure.autorest.model.clientmodel.ClientModels;
+import com.azure.autorest.model.clientmodel.GenericType;
 import com.azure.autorest.model.clientmodel.IType;
 import com.azure.autorest.model.clientmodel.ListType;
 import com.azure.autorest.model.clientmodel.MapType;
@@ -27,7 +28,6 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonGetter;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonSetter;
-import com.fasterxml.jackson.annotation.Nulls;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
 import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
 
@@ -77,11 +77,10 @@ public class ModelTemplate implements IJavaTemplate<ClientModel, JavaFile> {
         imports.add(JsonCreator.class.getName());
         imports.add(JacksonXmlElementWrapper.class.getName());
         imports.add(JacksonXmlProperty.class.getName());
-        imports.add(JsonSetter.class.getName());
-        imports.add(Nulls.class.getName());
 
         if (settings.isGettersAndSettersAnnotatedForSerialization()) {
             imports.add(JsonGetter.class.getName());
+            imports.add(JsonSetter.class.getName());
         }
 
         String lastParentName = model.getName();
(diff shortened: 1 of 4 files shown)


javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -232,27 +215,44 @@ public final void write(ClientModel model, JavaFile javaFile) {
...
-                } else if (settings.shouldGenerateXmlSerialization() && property.getWireType() instanceof ListType && !property.getIsXmlWrapper()) {
+                } else if (settings.shouldGenerateXmlSerialization() && property.getIsXmlWrapper()) {
+                    // Property is an XML wrapper, or simply put a node that contains a list of XML nodes.
+                    // Use JacksonXmlElementWrapper to indicate to Jackson that the current node contains a list
+                    // of values to be used as the property value.
+                    classBlock.annotation(String.format("JacksonXmlElementWrapper(localName = \"%1$s\")", property.getXmlName()));
+
+                    // Due to how the private inner class previously worked a JsonSetter annotation is also required
+                    // to configure Jackson to deserialize null into an empty collection.
+                    classBlock.annotation("JsonSetter(nulls = Nulls.AS_EMPTY)");
```
srnagar, line 244: Was this in the previous version of Jackson or was this how it worked with wrapper classes?

srnagar (commented):
I am not aware of any other services except Storage using XML but it would good to verify this with @weidongxu-microsoft  to see if any management libraries are impacted by this change.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
alzimmermsft: This is effectively what the wrapper class did. This annotation is telling Jackson to deserialize null or empty XML nodes as an empty collection instead of null. This effectively replicates what the getter was doing before but putting it into deserialization:

```java
public List<CorsRule> getCors() {
    if (this.cors == null) {
        this.cors = new CorsWrapper(new ArrayList<CorsRule>());
    }
    return this.cors.items;
}
```

srnagar: Yeah, I like the second option to explicitly return an empty list. Will the `JsonSetter(nulls = Nulls.AS_EMPTY)` annotation return a mutable collection like we were previously doing with `new CorsWrapper(new ArrayList<CorsRule>());` or with the 2nd option proposed above?

srnagar (approved):

weidongxu-microsoft:
Mgmt does not use XML, so you are good!

PR #1277, opened 2021-12-02
Revert "Use JacksonXmlElementWrapper Instead of a Static Inner Wrapper Class"

alzimmermsft:
Reverts Azure/autorest.java#1274

Turns out Jackson has edge cases with wrapped element lists that require the usage of a wrapper class. Until this changes the work done in the previous PR is being reverted.


srnagar:
It would be good to document the reason for why it didn't work in the PR description.

srnagar (approved):

PR #1285, opened 2021-12-10
client, limit Generated annotation to non-Android Client

weidongxu-microsoft:


androidgen/src/main/java/com/azure/autorest/android/template/AndroidServiceAsyncClientTemplate.java
```diff
@@ -1,9 +1,11 @@
 package com.azure.autorest.android.template;
 
+import com.azure.autorest.model.javamodel.JavaContext;
 import com.azure.autorest.template.ServiceAsyncClientTemplate;
 
 public class AndroidServiceAsyncClientTemplate extends ServiceAsyncClientTemplate {
     private static AndroidServiceAsyncClientTemplate _instance = new AndroidServiceAsyncClientTemplate();
+
```
haolingdong-msft, line 8: why do we want to limit @Generated annotation to non-android client? If we don't want to support partial update in Android, disabling partial-update flag can do the same thing.

weidongxu-microsoft: android has a different core lib and different http client, they might not have these.

androidgen/src/main/java/com/azure/autorest/android/template/AndroidServiceSyncClientTemplate.java
```diff
@@ -14,4 +15,8 @@ public static ServiceSyncClientTemplate getInstance() {
     protected void addServiceClientAnnotationImport(java.util.Set<String> imports) {
         imports.add("com.azure.android.core.rest.annotation.ServiceClient");
```
weidongxu-microsoft, line 16: @haolingdong-msft e.g. this is different in android.

haolingdong-msft (approved):

Issue #1256, opened 2021-11-15
Add unit test for Javagen and Preprocessor


weidongxu-microsoft:
Can use similar mock in Fluentgen.

weidongxu-microsoft:
For this task, 1 test will do.

PR #1287, opened 2021-12-10
refactor javagen code

XiaofeiCao:
related https://github.com/Azure/autorest.java/issues/1256
Current code style is difficult for unit testing. This pr is for refactoring `processInternal()` method in `Javagen` class:
1. organize `Step 1: Parse input yaml as CodeModel` into separate method `parseCodeModel()`.
2. organize `Step 3: Write to templates` into separate method `writeToTemplates()`

So that main flow in `processInternal()` method will look like this:
```java
// Step 1: Parse input yaml as CodeModel
String fileName = files.get(0);
CodeModel codeModel = parseCodeModel(fileName);

// Step 2: Map
Client client = Mappers.getClientMapper().map(codeModel);

// Step 3: Write to templates
JavaPackage javaPackage = writeToTemplates(settings, codeModel, client);

// Step 4:......
```

Will open another pr for adding unit test to Javagen based on this pr.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/Javagen.java
+++ b/javagen/src/main/java/com/azure/autorest/Javagen.java
@@ -140,7 +140,7 @@ public class Javagen extends NewPlugin {
         return codeModel;
     }
 
-    private JavaPackage writeToTemplates(JavaSettings settings, CodeModel codeModel, Client client) {
+    JavaPackage writeToTemplates(JavaSettings settings, CodeModel codeModel, Client client) {
         JavaPackage javaPackage = new JavaPackage(this);
         // Service client
         javaPackage
--- a/javagen/src/main/java/com/azure/autorest/Javagen.java
+++ b/javagen/src/main/java/com/azure/autorest/Javagen.java
@@ -75,175 +75,14 @@ public class Javagen extends NewPlugin {
 
         try {
             // Step 1: Parse input yaml as CodeModel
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -75,206 +75,217 @@ public boolean processInternal() {
...
-                                    if (!protocolExampleNameSet.contains(filename)) {
-                                        ProtocolExample protocolExample = new ProtocolExample(m, c, client.getServiceClient(), builderName, filename, example);
-                                        javaPackage.addProtocolExamples(protocolExample);
-                                        protocolExampleNameSet.add(filename);
-                                    }
-                                });
-                            }
-                        }));
-            }
+    private JavaPackage writeToTemplates(JavaSettings settings, CodeModel codeModel, Client client) {
```
haolingdong-msft, line 143: if considering unit test, making the function package-private instead of private may be better for unit test?

XiaofeiCao: Of course, will do that in unit testing.

XiaofeiCao: Changed, thanks for pointing out!

haolingdong-msft (approved):

PR #1289, opened 2021-12-10
add unit test for javagen

XiaofeiCao:
relate issue https://github.com/Azure/autorest.java/issues/1256

add unit test for javagen base on previous pr https://github.com/Azure/autorest.java/pull/1287

This pr does not contain assertions, just test the code can run and produce generated result.


javagen/src/test/java/com/azure/autorest/MockUnitJavagen.java
```diff
@@ -0,0 +1,44 @@
...
+        try {
+            InputStream fis = this.getClass().getClassLoader().getResourceAsStream(fileName);
+            BufferedReader reader = new BufferedReader(new InputStreamReader(fis));
+            StringBuilder sb = new StringBuilder();
+            String line;
+            while ((line = reader.readLine()) != null) {
+                sb.append(line);
+                sb.append("\n");
+            }
+            return sb.toString();
```
weidongxu-microsoft, line 34: Curious on what will be the difference between this and below
```
        try (InputStream inputStream = ResourceParserTests.class.getClassLoader().getResourceAsStream(filename);
             Reader in = new InputStreamReader(inputStream, StandardCharsets.UTF_8)) {
            int charsRead;
            while ((charsRead = in.read(buffer, 0, buffer.length)) > 0) {
                out.append(buffer, 0, charsRead);
            }
            return out.toString();
```

weidongxu-microsoft (approved):

javagen/src/test/java/com/azure/autorest/MockUnitJavagen.java
XiaofeiCao: Oh, I forgot to close the resource...

XiaofeiCao: Actually I'm surprised `\n` actually works in windows environment instead of `\r\n`.

weidongxu-microsoft: `\n` is probably fine, as snakeyaml likely able to handle both.

SDK usually does not know about current system, so it likely had to handle both.

PR #1290, opened 2021-12-11
Update Customization README Samples

alzimmermsft:
This PR updates the README structure by adding a README to the customization framework. It also updates the customization README samples to be pulled from code, requiring many fixes to them given the changes made to the framework since they were written.


weidongxu-microsoft (approved):
Good job!

weidongxu-microsoft:
One thing I am unsure of, is that when azure-core add new class (e.g. `RequestOption`, `Option`, `Generated` etc.) and they get generated to code via codegen, does it require releasing new version of `azure-autorest-customization` package?

It got a [pom in resources](https://github.com/Azure/autorest.java/blob/main/customization-base/src/main/resources/pom.xml), and used it for project configure. That is the reason I got this doubt.

alzimmermsft:
> One thing I am unsure of, is that when azure-core add new classes (e.g. `RequestOptions`, `PollingStrategy`, `Option`, `Generated` etc.) and they get generated to code via codegen, does it require releasing new version of `azure-autorest-customization` package?
> 
> It got a [pom in resources](https://github.com/Azure/autorest.java/blob/main/customization-base/src/main/resources/pom.xml), and used it for project configure. That is the reason I got this doubt.

Fortunately, it doesn't as it's dependency tree doesn't include any of the Azure SDKs

PR #1296, opened 2021-12-14
Upgrade Dependencies

### LINKED ISSUE #310: Review dependencies
Review dependency chain to make sure the minimal set of dependencies is being introduced.

alzimmermsft:
Fixes #310

This PR upgrades the Json RPC and logback-classic dependencies. It also remove dependencies on Netty and Apache Commons.


fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/OperationNameNormalization.java
```diff
@@ -103,7 +104,7 @@ private static void rename(OperationGroup operationGroup, Map<String, String> re
 
         for (Operation operation : operationGroup.getOperations()) {
             String path = operation.getRequests().iterator().next().getProtocol().getHttp().getPath();
-            path = StringUtils.strip(path.trim(), "/");
+            path = TRIM_LEADING_AND_TRAILING_FORWARD_SLASH.matcher(path.trim()).group(1);
```
weidongxu-microsoft, line 107: This one seems causing `java.lang.IllegalStateException: No match found`.

srnagar (approved):
LGTM

fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/OperationNameNormalization.java
alzimmermsft: I'll guard this with a matches or make the regex more relaxed

alzimmermsft: Decided to make the regex more relaxed

alzimmermsft: @weidongxu-microsoft, @srnagar would it be possible to have AutoRest fail if there are errors during generation for the fluent-tests? It would have made the CI error much clearer

weidongxu-microsoft: I will try add `if %errorlevel% neq 0 exit /b %errorlevel%` to bat.

weidongxu-microsoft: https://github.com/Azure/autorest.java/pull/1298

PR #1302, opened 2022-01-05
Remove RevApi from Fluent Gen POM

alzimmermsft:
This PR updates Fluent generation to no longer generate a RevApi configuration in the projects POM file. This is being removed as previously all Fluent generated libraries had a specialty exclusion which has been propagated into the RevApi configuration in azure-sdk-for-java, making the specialty configuration no longer necessary therefore meaning the code generation can be removed.

### DIFF
--- a/fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentPomTemplate.java
+++ b/fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentPomTemplate.java
@@ -30,35 +30,10 @@ public class FluentPomTemplate extends PomTemplate {
 
     @Override
     protected void writeBuildBlock(XmlBlock projectBlock, Pom pom) {
-        projectBlock.block("build", buildBlock -> {
-            buildBlock.block("plugins", pluginsBlock -> {
-                if (FluentStatic.getFluentJavaSettings().isSdkIntegration()) {
-                    // revapi-maven-plugin
-                    pluginsBlock.block("plugin", pluginBlock -> {
-                        pluginBlock.tag("groupId", "org.revapi");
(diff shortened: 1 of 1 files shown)


fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentPomTemplate.java
```diff
@@ -30,35 +30,10 @@ public static void setProject(FluentProject project) {
...
-                                });
-                            });
-                        });
-                    });
-                } else {
-                    writeStandAlonePlugins(pluginsBlock);
-                }
-            });
-        });
+        projectBlock.block("build", buildBlock -> buildBlock.block("plugins", pluginsBlock -> {
```
srnagar, line 33: We should check with @weidongxu-microsoft why they needed revapi to ignore breaking changes in `fluent` packages. If this is required, we should add this to the revapi exclusions file that is centrally maintained and not be generated for each library.

alzimmermsft: It will be added centrally using this configuration:

```json
{
  "regex": true,
  "code": "java\\.method\\.addedToInterface",
  "new": "method .* com\\.azure\\.resourcemanager\\..*",
  "justification": "resourcemanager interfaces are allowed to add methods."
}
```

weidongxu-microsoft: The classes under `fluent` is usually the Inner classes that we do not expect user to use directly, but user can use them if they choose.

It might not much for Lite packages, as they would be exposed in interface anyway.

You can remove this, and we can see how it works.

weidongxu-microsoft (approved):

PR #1311, opened 2022-01-17
LLC, fix example when query parameter is renamed via x-ms-client-name

weidongxu-microsoft:
When comparing name to JSON in example, we should use serialized name (the name used in JSON spec, not the client name which can be specified via `x-ms-client-name`).

Added a test case from https://github.com/Azure/autorest.testserver/blob/main/swagger/url-multi-collectionFormat.json


javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java
```diff
@@ -79,6 +79,7 @@ public void write(ProtocolExample protocolExample, JavaFile javaFile) {
             boolean matchRequiredParameter = false;
             for (int i = 0; i < numParam; i++) {
                 ClientMethodParameter p = method.getParameters().get(i);
+                // TODO: should use getRequestParameterName from proxy method parameter, instead of getName from client method parameter
                 if (p.getName().equalsIgnoreCase(parameterName)) {
```
haolingdong-msft, line 83: update for this TODO as well?

weidongxu-microsoft: Probably going to let you or Xiaofei to improve it :-)

weidongxu-microsoft: https://github.com/Azure/autorest.java/issues/1313

haolingdong-msft: sure, I can do that

haolingdong-msft (approved):

haolingdong-msft (approved):

protocol-tests/src/test/java/fixtures/url/multi/QueriesTests.java
```diff
@@ -0,0 +1,30 @@
...
+    public static void setup() {
+        client = new AutoRestUrlMutliCollectionFormatTestServiceClientBuilder().buildClient();
+    }
+
+    @Test
+    public void arrayStringMultiValid() {
+        RequestOptions requestOptions = new RequestOptions();
+        requestOptions.addQueryParam("arrayQuery", "ArrayQuery1");
+        requestOptions.addQueryParam("arrayQuery", "begin!*'();:@ &=+$,/?#[]end");
+        requestOptions.addQueryParam("arrayQuery", ""); // null does not work, use empty instead
```
weidongxu-microsoft, line 26: Current code in `RequestOptions` uses `UrlEscapers.QUERY_ESCAPER.escape(value)`, hence this gets to String `null` instead of empty if provided `null`.

PR #1317, opened 2022-01-19
code style, Java enum

weidongxu-microsoft:


extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/RequestParameterLocation.java
```diff
@@ -39,8 +38,7 @@ public int getValue()
         return this.ordinal();
     }
 
-
-    private RequestParameterLocation(String value) {
+    RequestParameterLocation(String value) {
```
XiaofeiCao, line 41: Why change it to package private?

Issue #1313, opened 2022-01-17
LLC, fix bug on example generation


weidongxu-microsoft:
Fix several `TODO`s here
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java

weidongxu-microsoft:
One possible improvement to code readability is to group client method parameter and proxy method parameter together, e.g. similar to https://github.com/Azure/autorest.java/blob/main/fluentgen/src/main/java/com/azure/autorest/fluent/model/clientmodel/MethodParameter.java

But note that proxy method got an extra `getAllParameters()` to return all proxy method parameter as in spec (even when they does not appear on method signature).

PR #1321, opened 2022-01-20
LLC, fix example generation with x-ms-client-name

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1313

case of the parameter not in method signature already fixed in https://github.com/Azure/autorest.java/pull/1311

this PR fix the case of the parameter in method signature (usually path parameter and body parameter).

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java
@@ -3,39 +3,49 @@
 
 package com.azure.autorest.template;
 
+import com.azure.autorest.Javagen;
 import com.azure.autorest.extension.base.model.codemodel.RequestParameterLocation;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
+import com.azure.autorest.extension.base.plugin.PluginLogger;
 import com.azure.autorest.model.clientmodel.AsyncSyncClient;
 import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientMethod;
 import com.azure.autorest.model.clientmodel.ClientMethodParameter;
 import com.azure.autorest.model.clientmodel.ProtocolExample;
+import com.azure.autorest.model.clientmodel.ProxyMethod;
 import com.azure.autorest.model.clientmodel.ProxyMethodExample;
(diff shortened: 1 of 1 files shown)


javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java
```diff
@@ -232,4 +249,24 @@ private static String getSerializedName(ProxyMethodParameter parameter) {
...
+            ProxyMethod proxyMethod,
+            List<ClientMethodParameter> clientMethodParameters) {
+        // the list of proxy method parameters will be 1-1 with list of client method parameters
+
+        Map<String, ProxyMethodParameter> proxyMethodParameterByClientParameterName = proxyMethod.getParameters().stream()
+                .collect(Collectors.toMap(p -> CodeNamer.getEscapedReservedClientMethodParameterName(p.getName()), Function.identity()));
+        List<ProxyMethodParameter> proxyMethodParameters = new ArrayList<>();
+        for (ClientMethodParameter clientMethodParameter : clientMethodParameters) {
+            ProxyMethodParameter proxyMethodParameter = proxyMethodParameterByClientParameterName.get(clientMethodParameter.getName());
+            proxyMethodParameters.add(proxyMethodParameter);
```
weidongxu-microsoft, line 263: Maybe we should put `requestParameterLocation` to client method parameter, instead of making the mapping.

javagen/src/main/java/com/azure/autorest/template/ProtocolSampleTemplate.java
```diff
@@ -74,32 +84,39 @@ public void write(ProtocolExample protocolExample, JavaFile javaFile) {
         List<String> clientParameterLines = new ArrayList<>();
         Set<ServiceClientProperty> processedServiceClientProperties = new HashSet<>();
 
+        List<ProxyMethodParameter> proxyMethodParameters = getProxyMethodParameters(method.getProxyMethod(), method.getParameters());
+
         example.getParameters().forEach((parameterName, parameterValue) -> {
```
haolingdong-msft, line 89: could we change the variable name 'example' to 'proxyMethodExample' to make it less confusion? we have both proxyMethodExample and protocolExample here. The logic here is a bit complex already :)

haolingdong-msft (approved):

Issue #1314, opened 2022-01-17
LLC/Client, use "security" from autorest/modelerfour


weidongxu-microsoft:
autorest doc
https://github.com/Azure/autorest/blob/main/docs/generate/authentication.md

m4 output
```yaml
security:
  authenticationRequired: true
  schemes:
    - type: AADToken
      scopes:
        - https://atlas.microsoft.com/.default
```

We should prefer security configure from autorest/m4, and avoid using our own definition of `credential-types`, `credential-scopes`, `key-credential-header-name` options (they can work as another override, but we should avoid using them whenever swagger and standard autorest option is good enough).

weidongxu-microsoft:
Plan is by default uses m4.
If `credential-types` provided to autorest.java, it override configuration from m4 (as if the "security" block is not available from m4).

This apply to all data-plane (LLC and existing clients).

PR #1329, opened 2022-01-26
Use "security" section from autorest/modelerfour 

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1314

Code Change:
1. Map m4 output to CodeModel POJO
2. Add `SecurityInfo` property on `ServiceClient`, later will use this property to write to template
3. Update `ServiceClientMapper`  to map security related information to `SecurityInfo`. Note: we will first get security info from  CodeModel POJO `Security`, and then if `JavaSettings` has defined security related configuration, we will overwrite m4 security definition
4. Update templates(`ProtocolSampleTemplate`, `ProtocolTestBlankTemplate`, `ServiceClientBuilderTemplate`) to use `ServiceClient.SecurityInfo`
5. Add one vanilla test: https://github.com/Azure/autorest.java/tree/9abba471175727732704905c1c8fc49f0b4716cb/vanilla-tests/src/main/java/fixtures/securityinfo
5. Use https://github.com/Azure/azure-rest-api-specs/blob/main/specification/maps/data-plane/Search/preview/2022-02-01-preview/geocoding.json to test locally. Generated builder looks like below. I also tested purview generation to make sure defining security info in JavaSetting works.


extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Scheme.java
```diff
@@ -0,0 +1,80 @@
+package com.azure.autorest.extension.base.model.codemodel;
```
weidongxu-microsoft, line 1: copyright

just config it in IDE

I will see if I got time to write a script to add them to existing classes and enable the checkstyle rule...

weidongxu-microsoft (commented):
overall LGTM

extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Scheme.java
weidongxu-microsoft: checkstyle rule merged

weidongxu-microsoft:
Maps seems having good config in swagger
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/maps/data-plane/Search/preview/2022-02-01-preview/geocoding.json

extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Scheme.java
haolingdong-msft: Thanks, will update

haolingdong-msft:
> Maps seems having good config in swagger https://github.com/Azure/azure-rest-api-specs/blob/main/specification/maps/data-plane/Search/preview/2022-02-01-preview/geocoding.json

Yes, I already use this json to test locally, it has both AzureKey and AzureToken. I will also add one case to our protocol test.

weidongxu-microsoft:
I checked autorest.testserver does not have a swagger with AADToken.

You may want to add a very small swagger in https://github.com/Azure/autorest.java/tree/main/vanilla-tests/swagger for codegen of AADToken and AzureKey.

weidongxu-microsoft (approved):

haolingdong-msft:
> I checked that autorest.testserver does not have a swagger with AADToken.
> 
> You may add a very small swagger in https://github.com/Azure/autorest.java/tree/main/vanilla-tests/swagger with AADToken and AzureKey for test. Enable at https://github.com/Azure/autorest.java/blob/main/generate.bat#L47-L56 Generate code and commit to repo, this would be enough.

Agree, I will add

weidongxu-microsoft (approved):

PR #1333, opened 2022-01-27
mgmt, name length limit

weidongxu-microsoft:


javagen/src/main/java/com/azure/autorest/util/ClassNameUtil.java
```diff
@@ -0,0 +1,35 @@
...
+package com.azure.autorest.util;
+
+public class ClassNameUtil {
+
+    public static String truncateClassName(
+            String namespace, String directory,
+            String packageName, String className) {
+        // see https://github.com/Azure/azure-sdk-for-java/blob/main/eng/common/pipelines/templates/steps/verify-path-length.yml
+        final int maxPathLength = 260;
+        final int basePathLength = 38;
```
weidongxu-microsoft, line 13: This is pretty tricky...

XiaofeiCao (approved):

PR #1344, opened 2022-02-08
Handle partial update for module-info.java

### LINKED ISSUE #1308: LLC, partial update for module-info.java
Merge generated module-info.java with existing module-info.java, when `partial-update` option is enabled,

Reason:

Generated module-info.java will be e.g.
```java
module com.azure.messaging.webpubsubnew {
    requires transitive com.azure.core;

    exports com.azure.messaging.webpubsubnew;
}
```

Then dev can customize it to be e.g.
```java
module com.azure.messaging.webpubsubnew {
    requires transitive com.azure.core;

    exports com.azure.messaging.webpubsubnew;
    exports com.azure.messaging.webpubsubnew.models;
}
```
commit to demonstrate https://github.com/Azure/azure-sdk-for-java/pull/26473/commits/094bcbc1092688d551309a32a68c59db48d6c11b

These customization needs to be perserved.

haolingdong-msft:
Fix #1308

For now, we just use the simplest solution to handle partial update for module-info.java. We compare the existing module-info.java with the generated one, if there is different, then we consider the file as manually modified, and we will use the existing file. Otherwise, we will use the generated file. This solution is good enough for now. I also listed other two solutions in this issue #1308.

Code Change:
1. `PartialUpdateHandler`: refine code structure and add logic for handling partial update for module-info.java
2. Add unit test for handling module-info.java
3. Regen code in `partial-update-test`, no change on the generated code.


weidongxu-microsoft (approved):

postprocessor/src/test/java/com/azure/autorest/postprocessor/util/PartialUpdateHandlerTest.java
```diff
@@ -136,4 +136,63 @@ public void testGeneratedFileRemoveAPIAndExistingFileUpdateMethod() throws URISy
...
+                "module com.azure.messaging.webpubsubnew {\n" +
+                "    requires transitive com.azure.core;\n" +
+                "\n" +
+                "    exports com.azure.messaging.webpubsubnew;\n" +
+                "    exports com.azure.messaging.webpubsubnew.models;\n" +
+                "}";
+        String generatedFileContent = "// Copyright (c) Microsoft Corporation. All rights reserved.\n" +
+                "// Licensed under the MIT License.\n" +
+                "// Code generated by Microsoft (R) AutoRest Code Generator.\n" +
+                "module com.azure.messaging.webpubsubnew {\n" +
```
XiaofeiCao, line 184: Maybe change the module name, so that the assert afterwards will verify it?

XiaofeiCao (approved):

postprocessor/src/test/java/com/azure/autorest/postprocessor/util/PartialUpdateHandlerTest.java
haolingdong-msft: This is a real scenario from Webpubsub, they add another directive, but don't change module name, so I want to simulate their use case here.

PR #1345, opened 2022-02-08
client, update core

weidongxu-microsoft:


javagen/src/main/java/com/azure/autorest/model/projectmodel/Project.java
```diff
@@ -50,11 +50,11 @@
...
-        private String azureCoreHttpNettyVersion = "1.11.6";
-        private String azureCoreTestVersion = "1.7.7";
-        private String azureIdentityVersion = "1.4.3";
+        private String azureCoreVersion = "1.25.0";
+        private String azureCoreManagementVersion = "1.5.2";
+        private String azureCoreHttpNettyVersion = "1.11.7";
+        private String azureCoreTestVersion = "1.7.8";
+        private String azureIdentityVersion = "1.4.4";
         private String junitVersion = "5.8.2";
         private String revapiMavenPlugin = "0.14.6";
```
weidongxu-microsoft, line 59: These does not have effect when integrated to sdk repo, as version is read from `eng/versioning`.

But it will take effect when generate code/project without sdk repo.

Issue #1312, opened 2022-01-17
LLC, improve Javadoc on query with array


weidongxu-microsoft:
Javadoc especially on query with array is almost useless.

Question: how to specify the query parameters in this `deleteByGuidsWithResponse`?
https://azuresdkdocs.blob.core.windows.net/$web/java/azure-analytics-purview-catalog/1.0.0-beta.2/index.html

Answer (hope it is correct):
https://github.com/Azure/azure-sdk-for-java/blob/0bc0e0fc315fa19c164d530826ffb2e367861bc1/sdk/purview/azure-analytics-purview-catalog/src/samples/java/com/azure/analytics/purview/catalog/generated/EntityDeleteByGuids.java

XiaofeiCao:
There are two solutions when query with array is encountered:

### Solution 1

  #### Add a description above the "Query Parameters"
  1. Content:
 ```java
/** 
 * Query parameters should be added through requestOptions e.g. <pre>{@code requestOptions.add("foo", bar); }</pre>, 
 * parameter names can be found below in "Query Parameters" section. 
 * /
```
		
  2. Change query parameter name in "Query Parameters" from parameter name to serialized name (ignore x-ms-client-name), so that user can set parameters correctly.
  3. Add an additional column named "Note"
  * For array with collection format "multi", add following into this column: 
    * values should be added one by one by calling RequestOptions.add for each value.
  * For array with other collection format and delimiter ",", add following: 
    * values should be joined into a single string delimited by ",", e.g. requestOptions.add("guid", "guid1,guid2");

  #### Overall comment looks like this:
```java
/**
 * Query parameters should be added through {@link RequestOptions#add} e.g. <pre>{@code requestOptions.add("foo", bar); }</pre>, 
 * parameter names can be found below in "Query Parameters" section. 
 * <p><strong>Query Parameters</strong>
 *
 * <table border="1">
 *     <caption>Query Parameters</caption>
 *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th><th>Note</th></tr>
 *     <tr><td>guid</td><td>String</td><td>Yes</td><td>An array of GUIDs of entities to list.</td><td>values should be added one by one by calling @{link RequestOptions#add} for each value</td></tr>
 *     <tr><td>minExtInfo</td><td>String</td><td>No</td><td>Whether to return minimal information for referred entities.</td><td></td></tr>
[...]

PR #1350, opened 2022-02-10
query with array javadoc improve

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/1312


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -202,7 +202,18 @@ private static String parameterDescriptionOrDefault(ProxyMethodParameter paramet
         if (CoreUtils.isNullOrEmpty(paramJavadoc)) {
             paramJavadoc = String.format("The %1$s parameter", parameter.getName());
         }
-        return CodeNamer.escapeXmlComment(paramJavadoc);
+        String description = CodeNamer.escapeXmlComment(paramJavadoc);
+        // query with array, add additional description
+        if (parameter.getRequestParameterLocation() == RequestParameterLocation.QUERY && parameter.getCollectionFormat() != null) {
+            if (parameter.getExplode()) {
+                // collectionFormat: multi
+                description += " Call {@link RequestOptions#addQueryParam} to add string to array";
```
weidongxu-microsoft, line 210:
```suggestion
                description += " Call {@link RequestOptions#addQueryParam} to add string to array.";
```

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -100,7 +100,7 @@ private static void optionalParametersJavadoc(String title, List<ProxyMethodPara
         for (ProxyMethodParameter parameter : parameters) {
             commentBlock.line(String.format(
                     "    <tr><td>%s</td><td>%s</td><td>%s</td><td>%s</td></tr>",
-                    parameter.getName(),
+                    parameter.getRequestParameterName(),
```
weidongxu-microsoft, line 103: nice catch!

javagen/src/main/java/com/azure/autorest/util/ProxyMethodExampleWrapper.java
```diff
@@ -0,0 +1,23 @@
...
+        this.xmsExample = xmsExample;
+        this.operationId = operationId;
+    }
+
+    public Object getXmsExample() {
+        return xmsExample;
+    }
+
+    public String getOperationId() {
+        return operationId;
```
weidongxu-microsoft, line 21: The operationId is already in ProxyMethod?

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodExampleMapper.java
```diff
@@ -32,7 +36,20 @@ public ProxyMethodExample map(Object xmsExample) {
...
     }
+
+    private String buildCodeSnippetIdentifier(String operationId) {
+        String[] operationGroupAndId = operationId.split("_");
+        if (operationGroupAndId.length != 2) {
+            return null;
+        }
+        String operationGroup = operationGroupAndId[0];
+        String oprId = operationGroupAndId[1];
+        return String.format("%s.%s.%s", JavaSettings.getInstance().getPackage(), operationGroup.toLowerCase(Locale.ROOT), oprId.toLowerCase(Locale.ROOT));
```
weidongxu-microsoft, line 53: It should be operation group + operation + example name

Maybe you can add operationId and example name, then exclude special char.

javagen/src/main/java/com/azure/autorest/util/ProxyMethodExampleWrapper.java
weidongxu-microsoft: Fine for this.

Though you also need to add the example name to this structure?

XiaofeiCao: Added.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
weidongxu-microsoft: Do you know whether BODY parameter would show up here?
BODY would have `parameter.getRequestParameterName() == null`. So if it could be here, you need to have logic to fallback to `parameter.getName()`

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodExampleMapper.java
```diff
@@ -32,7 +37,18 @@ public ProxyMethodExample map(Object xmsExample) {
...
         }
         return builder.build();
     }
+
+    private String buildCodeSnippetIdentifier(String operationId, String exampleName) {
+        return String.format("%s.generated.%s.%s", JavaSettings.getInstance().getPackage(), getValidName(operationId), getValidName(exampleName)).toLowerCase(Locale.ROOT);
+    }
+
+    private String getValidName(String exampleName) {
+        return CodeNamer.getValidName(exampleName).replace("_", "");
```
weidongxu-microsoft, line 52: Probably replace `-` as well? Or just use `CodeNamer.getValidName(name)` without `allowedCharacters`?

XiaofeiCao: Yes, I used `CodeNamer.getValidName(name)` without `allowedCharacters`, and invalid characters will be replaced by `_` :
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/util/CodeNamer.java#L356-L362
so I do an extra work of removing `_`.

XiaofeiCao: BODY parameter will not show up here because it is filtered out in the previous steps:
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java#L39-L51

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
XiaofeiCao: BODY parameter will not show up here because it is filtered out in the previous steps:
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java#L39-L51

protocol-tests/src/main/java/fixtures/url/QueriesAsyncClient.java
```diff
@@ -658,7 +658,7 @@
      * <table border="1">
      *     <caption>Query Parameters</caption>
      *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
-     *     <tr><td>arrayQuery</td><td>String</td><td>No</td><td>a null array of string using the csv-array format</td></tr>
+     *     <tr><td>arrayQuery</td><td>String</td><td>No</td><td>a null array of string using the csv-array format In the form of "," separated string.</td></tr>
```
weidongxu-microsoft, line 661: You might check if you can put a ending period (dot) to the original description, if it is not empty and is not end with a period.

Issue #1356, opened 2022-02-11
Add `sendRequest()` method to clients to support sending any HTTP request


srnagar:
Add an `sendRequest()` method (method name is not final) that can send any HTTP request. This should be optionally added to clients either that are newly generated or for ones that have previously existed.

### Sync client
```java
HttpRequest httpRequest = ...
Response<BinaryData> response = client.sendRequest(httpRequest, context);
```

### Async client

```java
HttpRequest httpRequest = ...
Mono<Response<BinaryData>> response = client.sendRequest(httpRequest);

weidongxu-microsoft:
- behind an option
- use pipeline.send without proxy method

PR #1363, opened 2022-02-17
dpg, support sendRequest method

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/1356

- code https://github.com/Azure/autorest.java/commit/f036597657d52ba9a7177afb409538c2f22faa11
- example https://github.com/Azure/autorest.java/pull/1363/commits/54d095e4695f2ff86bb930c56f58c9e026d2d12f
- test https://github.com/Azure/autorest.java/commit/f68d6d4fcf2adbbb00ca76eb321f2a811bc838e0

Client
```java
    /**
     * Sends the {@code httpRequest}.
     *
     * @param httpRequest The HTTP request to send.
     * @param context The context to associate with this operation.
     * @return the response body along with {@link Response}.
     */
    @Generated
    public Response<BinaryData> sendRequest(HttpRequest httpRequest, Context context) {
        return this.serviceClient.sendRequest(httpRequest, context);
    }
```

Implementation
```java
    /**
     * Sends the {@code httpRequest}.
     *
     * @param httpRequest The HTTP request to send.
     * @return the response body on successful completion of {@link Mono}.
     */
    public Mono<Response<BinaryData>> sendRequestAsync(HttpRequest httpRequest) {
        return FluxUtil.withContext(
                context ->
                        this.getHttpPipeline()
                                .send(httpRequest, context)
                                .flatMap(
                                        response ->
                                                BinaryData.fromFlux(response.getBody())
                                                        .map(
                                                                body ->
                                                                        new SimpleResponse<>(
                                                                                response.getRequest(),
                                                                                response.getStatusCode(),
                                                                                response.getHeaders(),
                                                                                body))));
    }

    /**
     * Sends the {@code httpRequest}.
     *
     * @param httpRequest The HTTP request to send.
     * @param context The context to associate with this operation.
     * @return the response body along with {@link Response}.
     */
    public Response<BinaryData> sendRequest(HttpRequest httpRequest, Context context) {
        return this.sendRequestAsync(httpRequest)
                .contextWrite(c -> c.putAll(FluxUtil.toReactorContext(context).readOnly()))
                .block();
    }
```

### DIFF
--- a/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
+++ b/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
@@ -137,6 +137,7 @@ public class JavaSettings {
                 }.getType(), "polling"),
                 getBooleanValue(host, "generate-samples", false),
                 getBooleanValue(host, "generate-tests", false),
+                getBooleanValue(host, "generate-send-request-method", false),
                 getBooleanValue(host, "pass-discriminator-to-child-deserialization", false),
                 getBooleanValue(host, "annotate-getters-and-setters-for-serialization", false),
                 getStringValue(host, "default-http-exception-type"),
@@ -231,6 +232,7 @@ public class JavaSettings {
         Map<String, PollingDetails> pollingConfig,
         boolean generateSamples,
         boolean generateTests,
+        boolean generateSendRequestMethod,
         boolean passDiscriminatorToChildDeserialization,
         boolean annotateGettersAndSettersForSerialization,
         String defaultHttpExceptionType,
@@ -311,6 +313,7 @@ public class JavaSettings {
         this.pollingConfig = pollingConfig;
         this.generateSamples = generateSamples;
         this.generateTests = generateTests;
+        this.generateSendRequestMethod = generateSendRequestMethod;
         this.passDiscriminatorToChildDeserialization = passDiscriminatorToChildDeserialization;
         this.annotateGettersAndSettersForSerialization = annotateGettersAndSettersForSerialization;
 
@@ -754,6 +757,12 @@ public class JavaSettings {
 
     public boolean isGenerateTests() {
         return generateTests;
+    }
+
+    private final boolean generateSendRequestMethod;
+
+    public boolean isGenerateSendRequestMethod() {
+        return generateSendRequestMethod;
     }
 
     private final boolean clientBuilderDisabled;
--- a/protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
+++ b/protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
@@ -18,6 +18,7 @@ import com.azure.core.annotation.ServiceMethod;
 import com.azure.core.exception.HttpResponseException;
 import com.azure.core.http.HttpPipeline;
 import com.azure.core.http.HttpPipelineBuilder;
+import com.azure.core.http.HttpRequest;
 import com.azure.core.http.policy.CookiePolicy;
 import com.azure.core.http.policy.RetryPolicy;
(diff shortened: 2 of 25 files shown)


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -137,6 +137,7 @@ public static JavaSettings getInstance() {
                 }.getType(), "polling"),
                 getBooleanValue(host, "generate-samples", false),
                 getBooleanValue(host, "generate-tests", false),
+                getBooleanValue(host, "generate-send-request-method", false),
```
srnagar, line 140: As we add more options, it's good to add a blurb on what this option is for and what gets generated when this option is enabled - both in code and in the README file.

protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
```diff
@@ -709,6 +711,51 @@ public DpgClientImpl(HttpPipeline httpPipeline, SerializerAdapter serializerAdap
...
 
+    /**
+     * Sends the {@code httpRequest}.
+     *
+     * @param httpRequest The HTTP request to send.
+     * @return the response body on successful completion of {@link Mono}.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Mono<Response<BinaryData>> sendRequestAsync(HttpRequest httpRequest) {
+        return this.getHttpPipeline()
```
srnagar, line 722: How are we passing the context coming from Reactor? This should be using `withContext(context -> ...)`

protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
```diff
@@ -709,6 +711,51 @@ public DpgClientImpl(HttpPipeline httpPipeline, SerializerAdapter serializerAdap
...
+                        response ->
+                                BinaryData.fromFlux(response.getBody())
+                                        .map(
+                                                body ->
+                                                        new SimpleResponse<>(
+                                                                response.getRequest(),
+                                                                response.getStatusCode(),
+                                                                response.getHeaders(),
+                                                                body)))
+                .block();
```
srnagar, line 756: Can we just call the async method and block on it?

protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
weidongxu-microsoft: Ah, thanks for point it out. On the calling side for too long, and naturally think if I write `contextWrite` after `Mono` it will automatically get to the core Context. But that is because we had `FluxUtil.withContext` in generated code.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
weidongxu-microsoft: Agree. DPG (low-level-client option) actually have an almost fixed set of options set by default, documented at https://github.com/Azure/autorest.java#settings-for-minimal-clients-low-level-clients. And dev should not remove any of them.
Presently a few compatible option to add is also documented here.

But for vanilla, it will be really hard to choose the options just based on descriptions in README (though usually the same set of default options for DPG works reasonably OK as a starter.

protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
weidongxu-microsoft: PR to core https://github.com/Azure/azure-sdk-for-java/pull/27197

PR #1368, opened 2022-02-18
Add Option to Generate BinaryData Based Overloads in ClientMethods

alzimmermsft:
This PR updates client method generation to enable the creation of `BinaryData`-based body parameter overloads to begin supporting `BinaryData` as the egress and ingress type when sending requests and receiving responses to and from the network.


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -603,6 +601,96 @@ public static ClientMethodMapper getInstance() {
...
+        // BinaryData is passed there isn't a requirement right away to translate their current SDK calls to the new
+        // BinaryData format. Instead, SDK authors will be able to port, test, and validate the transition to BinaryData
+        // when appropriate.
+        if (settings.isRequestBodyGeneratedAsBinaryData()) {
+            // Loop over each ClientMethod that was generated and for each method that has a BodyParam create a new
+            // instance that uses BinaryData as the value type.
+            //
+            // This is done using an iteration with index so that the new instance can be inserted into the list of
+            // ClientMethods to be generated next to its corresponding non-BinaryData overload.
+            for (int methodIndex = 0; methodIndex < methods.size(); methodIndex++) {
```
weidongxu-microsoft, line 624: nit, `ListIterator.add` might be more readable, or just a new output list.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -603,6 +601,96 @@ public static ClientMethodMapper getInstance() {
...
+                        .wireType(ClassType.BinaryData)
+                        .rawType(ClassType.BinaryData)
+                        .name(parameter.getName())
+                        .isRequired(parameter.getIsRequired())
+                        .isConstant(parameter.getIsConstant())
+                        .fromClient(parameter.getFromClient())
+                        .defaultValue(parameter.getDefaultValue())
+                        .annotations(parameter.getAnnotations())
+                        .location(parameter.getLocation())
+                        .build());
```
weidongxu-microsoft, line 652: These could be fragile if later someone add fields to ClientMethodParameter or ClientMethod. Should we add a e.g. `Builder.from(ClientMethodParameter parameter)` in this 2 classes instead (so at least the related code will be in same class)?

PR #1373, opened 2022-03-03
client, use static logger

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/pull/1373/commits/8f525c6fa90090365dd45d91637355c1c4041d9f
test as sample https://github.com/Azure/autorest.java/pull/1373/commits/a4c9fabfeedd36be3eeba5f14e9e761da9360c3f

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/bodyboolean/Bools.java
+++ b/vanilla-tests/src/main/java/fixtures/bodyboolean/Bools.java
@@ -25,8 +25,6 @@ import reactor.core.publisher.Mono;
 
 /** An instance of this class provides access to all the operations defined in Bools. */
 public final class Bools {
-    private static final ClientLogger LOGGER = new ClientLogger(Bools.class);
-
     /** The proxy service used to perform REST calls. */
     private final BoolsService service;
 
@@ -706,4 +704,6 @@ public final class Bools {
     public Response<Boolean> getInvalidWithResponse(Context context) {
         return getInvalidWithResponseAsync(context).block();
     }
+
+    private static final ClientLogger LOGGER = new ClientLogger(Bools.class);
 }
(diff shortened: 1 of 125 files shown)


vanilla-tests/src/main/java/fixtures/bodyboolean/Bools.java
```diff
@@ -181,7 +184,7 @@ public boolean getTrue() {
         if (value != null) {
             return value;
         } else {
-            throw new NullPointerException();
+            throw LOGGER.logExceptionAsError(new NullPointerException());
```
weidongxu-microsoft, line 187: log and throw in client method

vanilla-tests/src/main/java/fixtures/bodycomplex/models/Shark.java
```diff
@@ -105,7 +108,8 @@ public Shark setSiblings(List<Fish> siblings) {
     public void validate() {
         super.validate();
         if (getBirthday() == null) {
-            throw new IllegalArgumentException("Missing required property birthday in model Shark");
+            throw LOGGER.logExceptionAsError(
+                    new IllegalArgumentException("Missing required property birthday in model Shark"));
```
weidongxu-microsoft, line 112: log and throw in model

vanilla-tests/src/main/java/fixtures/bodycomplex/Dictionaries.java
```diff
@@ -19,12 +19,15 @@
...
 import com.azure.core.util.Context;
 import com.azure.core.util.FluxUtil;
+import com.azure.core.util.logging.ClientLogger;
 import fixtures.bodycomplex.models.DictionaryWrapper;
 import fixtures.bodycomplex.models.ErrorException;
 import reactor.core.publisher.Mono;
 
 /** An instance of this class provides access to all the operations defined in Dictionaries. */
 public final class Dictionaries {
+    private static final ClientLogger LOGGER = new ClientLogger(Dictionaries.class);
```
alzimmermsft, line 29: Is it possible to only generate the constant `ClientLogger` when the class will have logged statements?

weidongxu-microsoft: That would be harder as it would link the exact code/impl in client method / model validation to the class definition.

PR #1374, opened 2022-03-03
dpg, readme for common options

weidongxu-microsoft:


docs/generate/dataplane.md
```diff
@@ -0,0 +1,88 @@
+# <img align="center" src="../images/logo.png">  Generating a Data Plane Client
+
+The Java generator offers the ability to generate [data-plane][data-plane] client libraries.
+
+Data-plane client libraries look like [this][data-plane-client].
+It can be enabled with the flag `--low-level-client`.
```
weidongxu-microsoft, line 6: Maybe we should rename the flag to `data-plane-generator`?

PR #1378, opened 2022-03-07
dpg, handle exception option

weidongxu-microsoft:
Support `default-http-exception-type`, `use-default-http-status-code-to-exception-type-mapping`, `http-status-code-to-exception-type-mapping` in DPG (which already supported in HLC).


protocol-tests/src/main/java/fixtures/headexceptions/implementation/HeadExceptionsImpl.java
```diff
@@ -0,0 +1,194 @@
...
+    @Host("{$host}")
+    @ServiceInterface(name = "AutoRestHeadExceptio")
+    public interface HeadExceptionsService {
+        @Head("/http/success/200")
+        @ExpectedResponses({200})
+        @UnexpectedResponseExceptionType(
+                value = ResourceNotFoundException.class,
+                code = {404})
+        @UnexpectedResponseExceptionType(CustomizedException.class)
+        Mono<Response<Void>> head200(@HostParam("$host") String host, RequestOptions requestOptions, Context context);
```
weidongxu-microsoft, line 56: `CustomizedException` is the class that supposed to be manually create by dev, as grow-up.

PR #1387, opened 2022-03-10
dpg, add generate-models option for models in impl package

weidongxu-microsoft:
`client-side-validations` is not enabled, as I an not sure whether we'd like validation in client methods.


protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/models/LroProduct.java
```diff
@@ -2,14 +2,14 @@
...
+package fixtures.dpgcustomization.implementation.models;
 
 import com.azure.core.annotation.Fluent;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-/** The LROProduct model. */
+/** The LroProduct model. */
 @Fluent
-public final class LROProduct extends Product {
+public final class LroProduct extends Product {
```
weidongxu-microsoft, line 12: Here a some example of the output.

Previously I just use HLC flow to generate these and put to tests folder.
Now I switched to DPG with `generate-models` to generate these in main folder (but in implementation.models).

PR #1390, opened 2022-03-11
dpg, SyncClient wrap AsyncClient

weidongxu-microsoft:
code change
- https://github.com/Azure/autorest.java/pull/1390/commits/419b2a8c717d1bb40cef9726cd2dbb1935ddd94b
- https://github.com/Azure/autorest.java/pull/1390/commits/5c2af03b6ab497cc1541081591f94a132a936531

It is close one.
Currently can only do this for DPG.

HLC got a few other complications:
1. Some async -> sync involves complicated conversion like `Flux<Buffer>` to `InputStream`
2. Sync method has `Context` parameter, while public async method does not (hence the async client need to expose package private async methods with `Context` parameter)


protocol-tests/src/main/java/fixtures/dpgcustomization/DpgClient.java
```diff
@@ -18,22 +18,22 @@
 import com.azure.core.http.rest.Response;
 import com.azure.core.util.BinaryData;
 import com.azure.core.util.Context;
+import com.azure.core.util.FluxUtil;
 import com.azure.core.util.polling.SyncPoller;
-import fixtures.dpgcustomization.implementation.DpgClientImpl;
 
 /** Initializes a new instance of the synchronous DpgClient type. */
 @ServiceClient(builder = DpgClientBuilder.class)
 public final class DpgClient {
```
weidongxu-microsoft, line 26: sample file of SyncClient wrap AsyncClient

protocol-tests/src/main/java/fixtures/dpgcustomization/DpgClientBuilder.java
```diff
@@ -260,6 +260,6 @@ public DpgAsyncClient buildAsyncClient() {
      */
     @Generated
     public DpgClient buildClient() {
-        return new DpgClient(buildInnerClient());
+        return new DpgClient(new DpgAsyncClient(buildInnerClient()));
     }
```
weidongxu-microsoft, line 264: sample code on the change to Builder

PR #1392, opened 2022-03-14
dpg, customization doc for review

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1341

branch and PR for examples https://github.com/Azure/azure-sdk-for-java/pull/27647

### DIFF
--- a/docs/generate/dataplane-customization.md
+++ b/docs/generate/dataplane-customization.md
@@ -35,7 +35,7 @@ This ensures that future code generate will not override this customized method.
 
 ## Change Operation Behavior
 
-Generated client methods accept `ReqestOptions` for request parameters.
+Generated client methods accept `RequestOptions` for request parameters.
 For frequently invoked methods, one can customize them for better user-experience.
 
 Here is an [example](https://github.com/weidongxu-microsoft/azure-sdk-for-java/commit/c3e8ebc1bf13dd29d67da0b68d23a84197ce681e) of the modification of client to support options for client methods.
@@ -51,7 +51,7 @@ Note that the conversion only happen once in package private `MetricsAdvisorAsyn
 Generated client methods accept and produce `BinaryData` for request payload and response body.
 For frequently invoked methods, one can customize them for better user-experience.
 
-AutoRest Java `generate-models` option is provided to generate models from OpenAPI specs, to `implmenetation.models` package. They are provided only as reference.
+AutoRest Java `generate-models` option is provided to generate models from OpenAPI specs, to `implementation.models` package. They are provided only as reference.
 
 ### Model as Response Body
 
--- a/docs/generate/dataplane-customization.md
+++ b/docs/generate/dataplane-customization.md
@@ -1,0 +1,96 @@
+# Customization on Data-Plane Client
+
+## Generate Data-Plane Client
+
+Follow [Quickstart](https://aka.ms/azsdk/dpcodegen/java) to generate the Data-Plane Client from OpenAPI specs.
+
+## Change Client Behavior
+
+### Support Azure Active Directory Authentication
+
+See [Authentication in SDK](https://github.com/Azure/autorest/blob/main/docs/generate/authentication.md).
+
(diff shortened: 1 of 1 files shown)


docs/generate/dataplane-customization.md
```diff
@@ -0,0 +1,96 @@
...
+
+Usually `security: AzureKey` can provide implementation of API key authentication.
+
+For more complex API key authentication, one can customize the builder class.
+
+Here is an [example](https://github.com/weidongxu-microsoft/azure-sdk-for-java/commit/f5337b6fbc5937683e01299a509b505e01d8b7ce) of the modification of builder to support subscription key and API key.
+1. Create `MetricsAdvisorKeys` and `MetricsAdvisorKeyCredential` class in `models` package.
+2. Modify `MetricsAdvisorClientBuilder` to accept `MetricsAdvisorKeyCredential` from user, and then pass the keys in it to HTTP pipeline.
+
+Note that the `Generated` annotation on `createHttpPipeline()` is removed, as we customized this method.
```
haolingdong-msft, line 33:
```suggestion
Note that the `@Generated` annotation on `createHttpPipeline()` is removed, as we customized this method.
```

weidongxu-microsoft: `@` is not part of the annotation?

docs/generate/dataplane-customization.md
```diff
@@ -0,0 +1,96 @@
...
+This ensures that future code generate will not override this customized method.
+
+## Change Operation Behavior
```
haolingdong-msft, line 36: I don't quite get what 'Change Operation Behavior' means?

weidongxu-microsoft: Quote Laurent

> Required sections are (at the minimum):
> 
> Change operation behavior (new signature, or any wrapping)
> Change client behavior (default pipeline configuration, etc.)
> Add models manually written

For now, use the same phase as the issue for consistency at review. We might change the title later.

haolingdong-msft: ok

docs/generate/dataplane-customization.md
```diff
@@ -0,0 +1,96 @@
...
+
+Generated client methods accept `ReqestOptions` for request parameters.
```
lirenhe, line 38: typo?  RequestOptions

docs/generate/dataplane-customization.md
```diff
@@ -0,0 +1,96 @@
...
+3. Convert options from `ListDataFeedOptions` to `RequestOptions` in `listDataFeeds` method.
+
+Note that the conversion only happen once in package private `MetricsAdvisorAsyncClient.listDataFeeds(ListDataFeedOptions, Context)` method, which is invoked by all added `listDataFeeds` methods.
+
+## Add Model
+
+Generated client methods accept and produce `BinaryData` for request payload and response body.
+For frequently invoked methods, one can customize them for better user-experience.
+
+AutoRest Java `generate-models` option is provided to generate models from OpenAPI specs, to `implmenetation.models` package. They are provided only as reference.
```
lirenhe, line 54: Will it generate all models?  What if I only need a subset of all models?

weidongxu-microsoft: It will. But this is only as reference. If you check metricsadvisor, the `DataFeed` model written by hand is completely different to the model from swagger.

So if service want to use some of the generated models, they can use IDE to move them to the public `models` package. But the advise is to review them first before doing so.

docs/generate/dataplane-customization.md
weidongxu-microsoft: Thanks!

Re-run cspell on the doc, and fix the problem.

docs/generate/dataplane-customization.md
weidongxu-microsoft: In this section and example, I only want to demonstrate how to convert from/to model, not the design of user-facing class. Hence I only use the swagger generated `DataFeedDetail` class.

Issue #1343, opened 2022-02-06
Support for OASIS Request Repeatability Headers


mikekistler:
**Describe the issue or request**

The Azure API Guidelines say that all operations should be idempotent. This is mainly an issue for POST operations. For situations where a POST needs to be made idempotent, the guidelines recommend

☑️ YOU SHOULD support repeatable requests according as defined in [OASIS Repeatable Requests Version 1.0](https://docs.oasis-open.org/odata/repeatable-requests/v1.0/repeatable-requests-v1.0.html).

Some teams have begun following this advice and including the OASIS Request Repeatability headers in their REST API.

This raises the question of how the Azure SDKs should support OASIS Repeatability headers. 

**Describe your ideas for solutions**

 I think we should treat these headers specially -- meaning _not_ simply expose them as optional parameters that are set by the caller.  Instead, I think the SDK should set these headers within the method for any operation that supports them.

The headers involved are `Repeatability-Request-ID`, which the SDK should set to a freshly minted UUID, and `Repeatability-First-Sent`, which should be set to the current date/time.

The Java generator should check for the presence of these headers and produce the additional logic as described here.

weidongxu-microsoft:
@mikekistler

Do we know if autorest/modelerfour intend to do some processing on `Repeatability-Request-ID` and `Repeatability-First-Sent` headers?

Ask because change in moderfour would very likely break any codegen in its downstream.

PS: data-plane using the headers https://github.com/Azure/azure-rest-api-specs/blob/main/specification/communication/data-plane/Chat/preview/2020-11-01-preview3/communicationserviceschat.json

weidongxu-microsoft:
If there be no processing from modelerfour, codegen plan to do following:
1. Hide 2 headers from client method signature
2. When user calls the client method, fill a random GUID to `Repeatability-Request-ID` and RFC1123 datetime to `Repeatability-First-Sent` and call the internal method to send the REST call
3. [NO CHANGE] Default configure to azure-core pipeline would have request retried a few times, for most of 5xx, 408, 429

mikekistler:
@weidongxu-microsoft This sounds like a great plan. Thanks!

weidongxu-microsoft:
Discussion after Timothee and Laurent. The plan is to combine "ignore-headers" in m4 and a new pipeline (pre-request) in azure-core to have this exact effect. Laurent will raise it to architect for cross-language consistency.

weidongxu-microsoft:
[...]

PR #1394, opened 2022-03-17
client, upgrade autorest-core 3.8.1 and m4 4.23.1

weidongxu-microsoft:
Last piece for https://github.com/Azure/autorest.java/issues/1343

Code change https://github.com/Azure/autorest.java/pull/1394/commits/358f78321e44ca10c93eed13bb15ac6045768378 for breaks in autorest
Filter out "user_impersonation" due to m4 behavior break https://github.com/Azure/autorest.java/pull/1394/commits/c33fe839b5f2c18335e16a16d52889db7234abf5

With latest m4 and its bug fixed, now I can enable the test cases for `special-headers` for OASIS Request Repeatability in CI.


weidongxu-microsoft:
Fix is done in autorest core/m4, wait for new version.

PR #1395, opened 2022-03-17
Support for Custom Strongly-Typed Header Deserialization

alzimmermsft:
This PR contains a major update with how strongly-typed headers classes are handled during deserialization. A new AutoRest flag `custom-strongly-typed-header-deserialization` has been added that indicates to code generation to generate header classes with custom deserialization instead of using Jackson Databind's convertValue method. Before this change, every response with strongly-typed headers would serialize the HTTP headers to JSON to deserialize the JSON into the strongly-typed headers object. Now, when this flag is enabled the strongly-typed headers will skips that serialization-deserialization and use the raw `HttpHeaders` directly with predefined functions to convert to the strong property type. All properties are still eagerly deserialized to ensure any parsing exceptions are still thrown from the same point in code as before, if a breaking change was acceptable deserialization would be deferred to when the property getter was called like it was in #1271 (which is no longer relevant).

This requires the azure-core code change merge in https://github.com/Azure/azure-sdk-for-java/pull/27369


weidongxu-microsoft:
You might need to run `npm install`

Currently we fix the `@microsoft.azure/autorest.testserver` to 3.2.0 in package.json, and use swagger from "@microsoft.azure/autorest.testserver" (instead of from github) for codegen in local/CI.

vanilla-tests/src/main/java/fixtures/customheaderdeserialization/models/HeadersResponseBoolHeaders.java
```diff
@@ -0,0 +1,54 @@
...
+@Fluent
+public final class HeadersResponseBoolHeaders {
+    /*
+     * The value property.
+     */
+    @JsonProperty(value = "value")
+    private Boolean value;
+
+    // HttpHeaders containing the raw property values.
+    /** Creates an instance of HeadersResponseBoolHeaders class. */
```
weidongxu-microsoft, line 21: Does Javadoc need a `@param`?

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -554,6 +573,163 @@ private void addModelConstructor(ClientModel model, JavaSettings settings, JavaC
...
+        } else if (model.isStronglyTypedHeader() && settings.isCustomStronglyTypedHeaderDeserializationUsed()) {
+            classBlock.lineComment("HttpHeaders containing the raw property values.");
+            classBlock.javadocComment(settings.getMaximumJavadocCommentWidth(), (comment) ->
+                comment.description(String.format("Creates an instance of %1$s class.", model.getName())));
+            classBlock.publicConstructor(String.format("%s(HttpHeaders rawHeaders)", model.getName()), constructor -> {
+                for (ClientModelProperty property : model.getProperties()) {
+                    addConstructorCustomDeserialization(property, constructor);
+                }
+            });
+        }
```
weidongxu-microsoft, line 585: Is this ctor affect whether it is eager deserialization or lazy deserialization?
Is the difference of behavior in azure-core?

Ask because I didn't find other difference in generated code on `model.isStronglyTypedHeader() && settings.isCustomStronglyTypedHeaderDeserializationUsed()`

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
weidongxu-microsoft: OK, see the azure-core PR.

alzimmermsft:
> You might need to run `npm install` (there is many deleted files in PR, your local might be an earlier version of testserver)
> 
> Currently we fix the `@microsoft.azure/autorest.testserver` to 3.2.0 in package.json, and use swagger from this local "@microsoft.azure/autorest.testserver" (instead of from github) for codegen in local/CI.

Thanks! I was wondering why so many files were deleted. I'll regenerate to fix that.

weidongxu-microsoft (approved):

PR #1396, opened 2022-03-18
Flag to Generate without ResponseBase Sub-Types

alzimmermsft:
Adds a feature flag that enables operations with strongly-typed headers to return `ResponseBase` instead of the named `ResponseBase` sub-class. This change enables a code simplification in `azure-core` when constructing responses to no longer need to reflection to access the constructor of the `ResponseBase` sub-class and can instead instantiate `ResponseBase` directly. Additionally, this allows for smaller package sizes as the named type will no longer need to be generated.


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -145,7 +145,8 @@ public static JavaSettings getInstance() {
                 getBooleanValue(host, "use-default-http-status-code-to-exception-type-mapping", false),
                 host.getValue(new TypeReference<Map<Integer, String>>() {}.getType(),
                     "http-status-code-to-exception-type-mapping"),
-                getBooleanValue(host, "partial-update", false)
+                getBooleanValue(host, "partial-update", false),
+                getBooleanValue(host, "no-named-response-types", false)
```
weidongxu-microsoft, line 149: Emm, possible for a better name? `generic-response-type`?
Word `no` seems a bit strange to me.

Off-topic, we might want to maintain some guide on collection of recommended options for vanilla SDK. Even I do not know what best to choose if I want to make an vanilla SDK.

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/template/ResponseTemplate.java
```diff
@@ -17,47 +18,58 @@
...
 
         javaFile.declareImport(imports);
 
-        String classSignature = isStreamResponse ? String.format("%1$s extends %2$s implements Closeable", response.getName(), restResponseType) : String.format("%1$s extends %2$s", response.getName(), restResponseType);
+        String classSignature;
+        if (isStreamResponse) {
+            classSignature = String.format("%1$s extends %2$s implements Closeable", response.getName(),
+                restResponseType);
+        } else if (settings.isNoNamedResponseTypes()) {
+            classSignature = restResponseType.toString();
```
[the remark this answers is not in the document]
alzimmermsft: I believe that is correct, if there is no response type the `ResponseTemplate` isn't used. I had a hard time myself figuring that out as well.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
weidongxu-microsoft: Great. It is not urgent and not even need to be very detail. It would be already great if it works a reminder for us.

Just a note, I put the default flags in DPG at its readme (low-level-client).
https://github.com/Azure/autorest.java/blob/main/preprocessor/readme.md

And made a simple doc for potential service team working on it.
https://github.com/Azure/autorest.java/blob/main/docs/generate/dataplane.md

Of course vanilla would likely take a different format for all these.

Issue #1389, opened 2022-03-11
[BUG] codesnippet definition has line break between BEGIN and identifier


weidongxu-microsoft:
See https://github.com/Azure/azure-sdk-for-java/pull/26922#discussion_r823561419

Maybe we need to shorten/trim the identifier?

PS: Checking with Alan whether this is allowed (without white space)
```
// BEGIN:com.azure.iot.deviceupdate.generated.devicemanagementstopdeployment.devicemanagementcancelorretrydeployment
```

ref: https://github.com/Azure/azure-sdk-tools/blob/main/packages/java-packages/codesnippet-maven-plugin/README.md

PR #1397, opened 2022-03-21
dpg, remove space in sample identifier

weidongxu-microsoft:
As https://github.com/Azure/azure-sdk-tools/pull/2909, space is not required in `BEGIN`, `END`, `src_embed`, `end`
Remove the space to avoid line break by formatter.

Require codesnippet-maven-plugin 1.0.0-beta.7

Fix https://github.com/Azure/autorest.java/issues/1389

### DIFF
--- a/protocol-tests/src/samples/java/fixtures/bodystring/generated/EnumGetNotExpandable.java
+++ b/protocol-tests/src/samples/java/fixtures/bodystring/generated/EnumGetNotExpandable.java
@@ -11,10 +11,10 @@ import fixtures.bodystring.EnumClientBuilder;
 
 public class EnumGetNotExpandable {
     public static void main(String[] args) {
-        // BEGIN: fixtures.bodystring.generated.enumgetnotexpandable.enumgetnotexpandable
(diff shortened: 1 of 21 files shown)


protocol-tests/src/samples/java/fixtures/bodystring/generated/EnumGetNotExpandable.java
```diff
@@ -11,10 +11,10 @@
 
 public class EnumGetNotExpandable {
     public static void main(String[] args) {
-        // BEGIN: fixtures.bodystring.generated.enumgetnotexpandable.enumgetnotexpandable
         EnumClient enumClient = new EnumClientBuilder().host("http://localhost:3000").buildClient();
+        // BEGIN:fixtures.bodystring.generated.enumgetnotexpandable.enumgetnotexpandable
```
weidongxu-microsoft, line 15: Adjust the location of the sample code.

If we automatically embed the sample into javadoc of client method, the initialization of the client should be self-evident (another potential issue is that client initialization code could be incorrect).

XiaofeiCao (approved):

haolingdong-msft (approved):

PR #1402, opened 2022-03-23
mgmt, support RetryOptions

weidongxu-microsoft:
A minor difference to DPG, RetryPolicy override RetryOptions, if both configured.


haolingdong-msft (approved):

fluentgen/src/main/resources/Manager_Configurable.txt
```diff
@@ -9,6 +9,7 @@ public static final class Configurable {
     private final List<HttpPipelinePolicy> policies = new ArrayList<>();
     private final List<String> scopes = new ArrayList<>();
     private RetryPolicy retryPolicy;
+    private RetryOptions retryOptions;
```
haolingdong-msft, line 12: what is this .txt used for ? does it publish somewhere？

weidongxu-microsoft: This is the Java code in Manager. It is loaded into Template.

PR #1410, opened 2022-03-31
Add developer documentation for debugging AutoRest

srnagar:


weidongxu-microsoft (approved):
I usually use "run -> attach to process" after autorest debug started.

PR #1411, opened 2022-04-01
mgmt, remove duplicate property init code

weidongxu-microsoft:
Avoid 2 `httpPipeline` null check and init code in Builder for mgmt.

Previously the httpPipeline in trait will generate some code in `buildClient`, then the one in serviceClient will also generate code.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
@@ -105,7 +105,7 @@ public class ClientBuilderTrait {
         ServiceClientProperty pipelineProperty = new ServiceClientProperty("The HTTP pipeline to send requests " +
                 "through.", ClassType.HttpPipeline, "pipeline", false,
                 JavaSettings.getInstance().isAzureOrFluent()
-                        ? "new HttpPipelineBuilder().policies(new UserAgentPolicy(), new RetryPolicy(), new CookiePolicy()).build()"
+                        ? "new HttpPipelineBuilder().policies(new UserAgentPolicy(), new RetryPolicy()).build()"
                         : "createHttpPipeline()");
         Consumer<JavaBlock> pipelineMethodImpl = function -> {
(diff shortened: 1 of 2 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
```diff
@@ -105,7 +105,7 @@ private static ClientBuilderTrait createHttpTrait() {
         ServiceClientProperty pipelineProperty = new ServiceClientProperty("The HTTP pipeline to send requests " +
                 "through.", ClassType.HttpPipeline, "pipeline", false,
                 JavaSettings.getInstance().isAzureOrFluent()
-                        ? "new HttpPipelineBuilder().policies(new UserAgentPolicy(), new RetryPolicy(), new CookiePolicy()).build()"
+                        ? "new HttpPipelineBuilder().policies(new UserAgentPolicy(), new RetryPolicy()).build()"
```
weidongxu-microsoft, line 108: This change will not have real effect, and the dummy pipeline will not work anyway.

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -232,20 +232,23 @@ public final void write(ClientBuilder clientBuilder, JavaFile javaFile) {
...
                         .flatMap(trait -> trait.getTraitMethods().stream())
                         .map(ClientBuilderTraitMethod::getProperty)
                         .filter(Objects::nonNull)
                         .collect(Collectors.toList());
-                allProperties.addAll(clientProperties);
+                Set<String> propertyNames = allProperties.stream()
+                        .map(ServiceClientProperty::getName)
+                        .collect(Collectors.toSet());;
+                allProperties.addAll(clientProperties.stream()
+                        .filter(p -> !propertyNames.contains(p))    // filter out properties already contained in traits
```
weidongxu-microsoft, line 246: If the property is handled by traits, avoid it in serviceClient properties.

haolingdong-msft (approved):

PR #1412, opened 2022-04-02
client, generate unit test for models

weidongxu-microsoft:
For common setting, this feature is yet to take effect.

Mgmt usually does not enable `generate-tests` except for api-scenario.
Vanillar does not enable `generate-tests`.
DPG does usually does not have model hence not generate these tests for models.

Will see how this works for mgmt and start enabling it (usually mgmt has complex models).

Test on serialization will be done in another PR.


vanilla-tests/src/test/java/fixtures/bodycomplex/generated/CookiecuttersharkTests.java
```diff
@@ -0,0 +1,33 @@
...
+        Assertions.assertEquals("fplgmgsxnk", model.getSiblings().get(0).getSiblings().get(0).getSpecies());
+        Assertions.assertEquals(0.12664735f, model.getSiblings().get(0).getSiblings().get(0).getLength());
+        Assertions.assertEquals(
+                "eslpvlopwi", model.getSiblings().get(0).getSiblings().get(0).getSiblings().get(0).getSpecies());
+        Assertions.assertEquals(
+                0.37284225f, model.getSiblings().get(0).getSiblings().get(0).getSiblings().get(0).getLength());
+        Assertions.assertEquals(2142473497, model.getAge());
+        Assertions.assertEquals(OffsetDateTime.parse("2021-06-27T19:00:45Z"), model.getBirthday());
+    }
+}
```
weidongxu-microsoft, line 33: sample here

haolingdong-msft:
May I know the purpose of adding those model tests? Is it for testing deserialization?

weidongxu-microsoft:
@haolingdong-msft 
For test of serialization de-serialization of models. Currently most of it is done by Jackson so likely no major bug, but something might still wrong with JsonFlatten.
Later if we switch to other implementation, it might approve more useful as many mgmt lib has much more complicated model than that in autorest.testserver.

Though theoretically the correctness of model class is based on codegen test, not on unit test on generated package (and this PR does not solve everything, as codegen might have bug on mapping of code model to client model that not captured sicne this test is generated based on client model).
Practically it should help as another layer of validation (and bump up the test coverage a bit on generated lib).

Also the setter / getter code and random JSON data can be re-used for unit test on client method as well (through mock on HttpClient or method group). If we can do this it would bump much more on test coverage (though again ideally this would not be the job on generated package to verify codegen is correct).

PR #1416, opened 2022-04-07
client, support testSerialize in model validation

weidongxu-microsoft:


vanilla-tests/src/test/java/fixtures/bodycomplex/generated/SmartSalmonTests.java
```diff
@@ -30,4 +34,65 @@ public void testDeserialize() {
         Assertions.assertEquals(false, model.iswild());
         Assertions.assertEquals("sne", model.getCollegeDegree());
     }
+
+    @Test
+    public void testSerialize() {
```
weidongxu-microsoft, line 39: some example here

PR #1424, opened 2022-04-17
DPG, minor change of async client variable name in sync client

weidongxu-microsoft:


generate
```diff
@@ -10,6 +10,8 @@ PROTOCOL_RESILIENCE_ARGUMENTS='--version=3.8.1 --java --use=./ --data-plane'
 SWAGGER_PATH=node_modules/@microsoft.azure/autorest.testserver/swagger
 
 # Vanilla
+mv -f vanilla-tests/src/main/java/fixtures/report/CoverageReporter.java vanilla-tests/swagger/CoverageReporter.java
```
weidongxu-microsoft, line 13: Actually we seems not really need to run CoverageReport class. Just keep it there. This class is manual written.

javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientWrapAsyncClientTemplate.java
```diff
@@ -22,7 +22,7 @@ public static ServiceSyncClientTemplate getInstance() {
         return INSTANCE;
     }
 
-    private static final String ASYNC_CLIENT_VAR_NAME = "asyncClient";
+    private static final String ASYNC_CLIENT_VAR_NAME = "client";
```
weidongxu-microsoft, line 25: In existing hand-written data-plane sdk, we seems use `client` more.

PR #1427, opened 2022-04-20
client, always have javadoc description

weidongxu-microsoft:
Msdoc render on method of javadoc without description is not good.
https://docs.microsoft.com/en-us/java/api/com.azure.resourcemanager.mediaservices.mediaservicesmanager?view=azure-java-stable

Hence we should always add the "description" to javadoc on method. "@return" is not enough.

### DIFF
--- a/azure-tests/src/main/java/fixtures/azurespecials/models/HeadersCustomNamedRequestIdHeadResponse.java
+++ b/azure-tests/src/main/java/fixtures/azurespecials/models/HeadersCustomNamedRequestIdHeadResponse.java
@@ -29,7 +29,11 @@ public final class HeadersCustomNamedRequestIdHeadResponse
         super(request, statusCode, rawHeaders, value, headers);
     }
 
-    /** @return the deserialized response body. */
+    /**
+     * Gets the deserialized response body.
+     *
+     * @return the deserialized response body.
+     */
     @Override
     public Boolean getValue() {
         return super.getValue();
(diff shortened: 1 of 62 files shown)


azure-tests/src/main/java/fixtures/azurespecials/models/HeadersCustomNamedRequestIdHeadResponse.java
```diff
@@ -29,7 +29,11 @@ public HeadersCustomNamedRequestIdHeadResponse(
...
 
-    /** @return the deserialized response body. */
+    /**
+     * Gets the deserialized response body.
+     *
+     * @return the deserialized response body.
+     */
     @Override
     public Boolean getValue() {
         return super.getValue();
```
weidongxu-microsoft, line 39: Case 1 on Resonse class.

azure-tests/src/main/java/fixtures/paging/models/OperationResultStatus.java
```diff
@@ -54,7 +54,11 @@ public static OperationResultStatus fromString(String name) {
...
 
-    /** @return known OperationResultStatus values. */
+    /**
+     * Gets known OperationResultStatus values.
+     *
+     * @return known OperationResultStatus values.
+     */
     public static Collection<OperationResultStatus> values() {
         return values(OperationResultStatus.class);
     }
```
weidongxu-microsoft, line 64: Case 2 on Choice class.

weidongxu-microsoft:
Case 3 is on Manager class. E.g. https://docs.microsoft.com/en-us/java/api/com.azure.resourcemanager.mediaservices.mediaservicesmanager?view=azure-java-stable

PR #1438, opened 2022-05-10
DPG, support helper client method in Impl for LRO with model

weidongxu-microsoft:
this PR is on providing utilities for developer to add model to DPG (currently limit to LRO methods).

sample output https://github.com/Azure/autorest.java/pull/1438/commits/564502ae472142dda3dad1912e3c9233f3da7977

also add open and export to `module-info.java` when `generate-models=true` e.g.
```
module com.azure.communication.phonenumbersdemo {
    requires transitive com.azure.core;

    exports com.azure.communication.phonenumbersdemo;
    exports com.azure.communication.phonenumbersdemo.models;

    opens com.azure.communication.phonenumbersdemo.implementation.models to
            com.azure.core,
            com.fasterxml.jackson.databind;
    opens com.azure.communication.phonenumbersdemo.models to
            com.azure.core,
            com.fasterxml.jackson.databind;
}
```
at https://github.com/Azure/autorest.java/pull/1438/commits/0927491ac63c379a960c43048279ba250ace8c63


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -424,93 +424,88 @@ && shouldGeneratePagingMethods()) {
...
+                            !(ClassType.BinaryData.equals(methodPollingDetails.getIntermediateType())
+                                    && ClassType.BinaryData.equals(methodPollingDetails.getFinalType()))) {
+                        // a new method to be added as implementation only (not exposed to client) for developer
+                        dpgMethodPollingDetailsWithModel = methodPollingDetails;
+
+                        // DPG keep the method with BinaryData
+                        methodPollingDetails = new MethodPollingDetails(
+                                dpgMethodPollingDetailsWithModel.getPollingStrategy(),
+                                ClassType.BinaryData, ClassType.BinaryData,
+                                dpgMethodPollingDetailsWithModel.getPollIntervalInSeconds());
```
weidongxu-microsoft, line 445: Other change in the class is refactor.

1. If DPG, keep the origin client method of `<BinaryData, BinaryData>`, even if polling configuration  provide different intermediate type and final type.
2. If polling configuration provides different intermediate type and final type, generate additional LRO method with suffix "WithModel" in implementation only. This would enable developer to customize the LRO method in client without trying to convert say `PollerFlux<BinaryData, BinaryData>` to `PollerFlux<LroProduct, LroProduct>`

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -93,6 +93,8 @@ public class ClientMethod {
 
     private JavaVisibility methodVisibility;
 
+    private boolean implementationOnly = false;
```
weidongxu-microsoft, line 96: `implementationOnly` might not be a good name, or it might need to be an enum.

And it kind of group with `methodVisibility` above, but still not same or mergeable. (as we can have public method in implementation, or package private method in implementation).

haolingdong-msft (approved):

weidongxu-microsoft:
I've merged it. If any comments on it (e.g. the slightly uncomfortable name of `implemenationOnly`), I will improve in future PR as it is internal and can be changed without affecting codegen behavior.

PR #1443, opened 2022-05-11
Updates to Client Method Code Generation

alzimmermsft:
- Removed usage of `JacksonAdapter` when converting enums and `ExpandableStringEnum`s to a delimited string
- Removed usage of type casting when using `Mono.empty()` as it'll handle type casting.
- Replaced if/else null check with `Mono.justOrEmpty(Object)`.
- Replaced `Optional.ofNullable` and `orElseGet(Stream::empty)` with null check and empty `List` return (which is what the more complex `Optional`/`Stream` does.
- Added general documentation and questions to `ClientMethodTemplate` for follow-ups.


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -48,111 +52,192 @@ public static ClientMethodTemplate getInstance() {
...
+    /**
+     * Adds validations to the client method.
+     *
+     * @param function The client method code block.
+     * @param expressionsToCheck Expressions to validate as non-null.
+     * @param validateExpressions Expressions to validate with a custom validation (key is the expression, value is the
+     * validation).
+     * @param settings AutoRest generation settings, used to determine if validations should be added.
+     */
+    protected static void addValidations(JavaBlock function, List<String> expressionsToCheck,
```
srnagar, line 64: unrelated to this PR: Why is this a protected static method?

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -48,111 +52,192 @@ public static ClientMethodTemplate getInstance() {
...
+        Set<String> expressionsToCheckSet = new LinkedHashSet<>(expressionsToCheck);
+
+        for (String expressionToCheck : expressionsToCheckSet) {
+            // TODO (alzimmer): Determine if the assumption being made here are always true.
+            // 1. Assumes that the expression is nullable.
+            // 2. Assumes that the client method returns a reactive response.
+            // 3. Assumes that the reactive response is a Mono.
+            JavaIfBlock nullCheck = function.ifBlock(expressionToCheck + " == null", ifBlock ->
+                // TODO (alzimmer): Should this be a NullPointerException?
+                ifBlock.line("return Mono.error(new IllegalArgumentException(\"Parameter %s is required and "
```
srnagar, line 81: Yes, this should be `NullPointerException`

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -48,111 +52,192 @@ public static ClientMethodTemplate getInstance() {
...
-                    function.line("%1$s %2$s = new %1$s();",
-                            outParameter.getClientType(),
-                            outParameter.getName());
+                    function.line("%1$s %2$s = new %1$s();", outParameter.getClientType(), outParameter.getName());
                 } else {
-                    function.line("%1$s %2$s = null;",
-                            outParameter.getClientType(),
-                            outParameter.getName());
+                    // TODO (alzimmer): This is making the assumption that the client type is nullable.
+                    // Is that always the case? What if the client type is a primitive?
```
srnagar, line 214: Since the `if` condition checks that the parameter is required, it cannot be a primitive type here in the `else` block. Since all optional params should be non-primitive types. Also, when the param is required, the client type will be an instance of ClassType.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -48,111 +52,192 @@ public static ClientMethodTemplate getInstance() {
...
+                    function.line("%1$s %2$s = null;", outParameter.getClientType(), outParameter.getName());
                 }
+
+                // TODO (alzimmer): Should this break here? What if there are subsequent method transformation details?
```
srnagar, line 218: This might be a bug. We should continue with the rest of the transformations.

srnagar (approved):

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -331,18 +425,17 @@ private static boolean addSpecialHeadersToRequestOptions(JavaBlock function, Cli
     protected static void addSpecialHeadersToLocalVariables(JavaBlock function, ClientMethod clientMethod) {
         if (MethodUtil.isMethodIncludeRepeatableRequestHeaders(clientMethod.getProxyMethod())) {
             function.line(String.format("String %1$s = UUID.randomUUID().toString();", MethodUtil.REPEATABILITY_REQUEST_ID_VARIABLE_NAME));
+            // TODO (alzimmer): Can this be replaced with DateTimeRfc1123.toRfc1123String(OffsetDateTime)?
```
srnagar, line 428: Yes, it can.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
alzimmermsft: I'm going to file an issue for this

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
weidongxu-microsoft: This is THE oldest code that likely converted from C# via some tooling. Could be this reason.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -708,6 +799,24 @@ protected void generatePagedAsyncSinglePage(ClientMethod clientMethod, JavaType
...
+        // This code is based on the previous implementation that assume if the T type for Mono<T> is a class that
+        // it has deserialized headers. This won't always be the case, but ClassType also isn't able to maintain
+        // whether the class is an extension of ResponseBase.
+        if (type instanceof ClassType) {
+            return true;
+        } else if (type instanceof GenericType && "ResponseBase".equals(((GenericType) type).getName())) {
+            return true;
+        } else {
+            return false;
+        }
```
weidongxu-microsoft, line 817: Fluent does this.
https://github.com/Azure/autorest.java/blob/main/fluentgen/src/main/java/com/azure/autorest/fluent/util/FluentUtils.java#L320-L337

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -323,26 +428,24 @@ private static boolean addSpecialHeadersToRequestOptions(JavaBlock function, Cli
...
+            function.line(String.format("requestOptionsLocal.setHeader(\"%1$s\", DateTimeRfc1123.toRfc1123String(OffsetDateTime.now()));", MethodUtil.REPEATABILITY_FIRST_SENT_HEADER));
         }
         return requestOptionsLocal;
     }
 
     protected static void addSpecialHeadersToLocalVariables(JavaBlock function, ClientMethod clientMethod) {
         if (MethodUtil.isMethodIncludeRepeatableRequestHeaders(clientMethod.getProxyMethod())) {
             function.line(String.format("String %1$s = UUID.randomUUID().toString();", MethodUtil.REPEATABILITY_REQUEST_ID_VARIABLE_NAME));
-            function.line(String.format("String %1$s = DateTimeFormatter.ofPattern(\"EEE, dd MMM yyyy HH:mm:ss z\", Locale.ENGLISH).withZone(ZoneId.of(\"GMT\")).format(OffsetDateTime.now());", MethodUtil.REPEATABILITY_FIRST_SENT_VARIABLE_NAME));
+            function.line(String.format("String %1$s = DateTimeRfc1123.toRfc1123String(OffsetDateTime.now());", MethodUtil.REPEATABILITY_FIRST_SENT_VARIABLE_NAME));
```
weidongxu-microsoft, line 439: Oh, nice. Is it new in core?

Please add the imports here https://github.com/Azure/autorest.java/blob/7bdee3b94e32822e00a20262fcf3beaa2393c2b4/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java#L422-L427

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -221,98 +303,121 @@ protected static void applyParameterTransformations(JavaBlock function, ClientMe
...
-                            String expression;
-                            if (alwaysNull) {
-                                expression = "null";
-                            } else {
-                                expression = String.format("Base64Url.encode(%s)", parameterName);
-                            }
-                            function.line("%s %s = %s;", parameterWireTypeName, parameterWireName, expression);
+                            expression = String.format(
+                                "JacksonAdapter.createDefaultSerializerAdapter().serialize%s(%s, CollectionFormat.%s)",
+                                settings.shouldUseIterable() ? "Iterable" : "List", parameterName,
```
weidongxu-microsoft, line 391: Please do be careful that they are indeed equivalent. Else it could be nasty bug.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -48,111 +52,189 @@ public static ClientMethodTemplate getInstance() {
...
+        // expressionsToCheck to a set.
+        Set<String> expressionsToCheckSet = new LinkedHashSet<>(expressionsToCheck);
+
+        for (String expressionToCheck : expressionsToCheckSet) {
+            // TODO (alzimmer): Determine if the assumption being made here are always true.
+            // 1. Assumes that the expression is nullable.
+            // 2. Assumes that the client method returns a reactive response.
+            // 3. Assumes that the reactive response is a Mono.
+            JavaIfBlock nullCheck = function.ifBlock(expressionToCheck + " == null", ifBlock ->
+                ifBlock.line("return Mono.error(new NullPointerException(\"Parameter %s is required and "
```
weidongxu-microsoft, line 80: The change to `NullPointerException` would be a big behavior change. And this could be kind of inconsistency depends on when the code regened...

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
weidongxu-microsoft: Above code probably written by me, to handle the case that the payload schema is empty class (no property or no writable ones), so it actually gets flattened to no parameter in method signature.

Yes, I do not see issue if "break" changed to "continue", probably my bug anyway.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
weidongxu-microsoft: And if you really need to change this, remember to update javadoc as well (I am not sure whether currently we had any validation other than null check).
```
```

weidongxu-microsoft:
LGTM overall. Great work.

My major concern is the change from `IllegalArgumentException` to `NullPointerException`. I am not sure what Exception the majority hand-written code on data-plane uses. But I am concerned about inconsistency, as we definitely allow `null` for some parameter but not for other, so `IllegalArgumentException` more or less not being wrong.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
alzimmermsft: I'm thinking something similar, I think `ClassType` may want to introduce a new field that maintains which types it's implementing or extending. This will give us a richer experience throughout the code base instead of having to potentially do another Swagger lookup or wrangling external templates.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
alzimmermsft: I think it is best to revert this back to `IllegalArgumentException` for now given the effect this change could have, we'll want to have a more targeted discussion for this.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
alzimmermsft: Added here as well. Something we may want to look at in the future is making it easier to add imports to a class file. Right now, imports are only able to be determined by parameters and methods which is good for centralization but then we run into these giant conditional blocks as seen in `ClientMethod` to handle each scenario that would have been a check elsewhere in code. It took me a while to realize this and be able to update the method imports based on the changes I made.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
alzimmermsft: Yeah, even spending time writing the Javadocs for these methods and reading through what they do, I'm still not even sure what they do differently.

This would be a good area to begin a minor refactoring when we began to push forward on greenlighting new code generation features.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
weidongxu-microsoft: Emm, yes, current it is a bit hard to use. For now the simple solution is just to add whatever required import to `ClientMethod` without condition, and let formatter to handle the clean-up.

This is part of the reason I added [MethodTemplate](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/prototype/MethodTemplate.java) for mgmt to keep related imports and javadoc and codes near, because of the diversity of the type of methods there.
But it got its own drawback. It blurs the line of model and template. So that probably remain in mgmt exclusively.

weidongxu-microsoft:
I've regened mgmt storage/compute with the updated codegen. So far not found any problem.
But note that mgmt is rather homogeneous, so it only part of the assurance.

Issue #1441, opened 2022-05-11
fluent, Support `resourceGroup()` for resource that has it


weidongxu-microsoft:
For a resource model, we should be able to check whether resourceGroup is contained in ID/path, and extract for user.

PR #1446, opened 2022-05-12
mgmt, support resourceGroupName() method

weidongxu-microsoft:
Fix https://github.com/Azure/autorest.java/issues/1441

`resourceGroupName` is a very useful help method for most Azure resource. First add it to resource that feels "safe" to have it (i.e., those is based on "resourceGroup" in path/ID, and supports both create and update).


haolingdong-msft:
what does generated code look like?

haolingdong-msft (approved):

weidongxu-microsoft:
Basically just
```java
    public String resourceGroupName() {
        return resourceGroupName;
    }
```

PR #1448, opened 2022-05-12
Added Ability to Customize Constants and Miscellaneous Improvements

### LINKED ISSUE #1419: [FEATURE REQ] Support for customizing enums
As of right now, there's a way to rename an enum value on a given type, but that's the only customization available. We'd specifically like to be able to add an annotation to one of the values so we can deprecate it.

### LINKED ISSUE #1444: ClassCustomization.removeMethod Bug
There is a current workaround on using `removeAnnotation(@Override)` then removeMethod for this issue so it is not urgent

When there is `@Override` above the method, it doesn't remove `@Override` and causes format error (unable to remove method)

For example, shown through --skip-formatting:

The function call:
`ClassCustomization classCustomization = models.getClass("ErrorResponseException");`
`classCustomization .removeMethod("getValue");`

`@Override`
 `public ErrorResponse getValue() {`
          `    return (ErrorResponse) super.getValue();`
 `}`

Will leave this in the code instead of removing the entire method:
`@Override`

I am currently using this version of dependency in pom.xml:
`<dependency>`
  `    <groupId>com.azure.tools</groupId>`
  `    <artifactId>azure-autorest-customization</artifactId>`
 `     <version>1.0.0-beta.4</version>`
  `</dependency>`
  
  I am running it via:
  mvn install && autorest --java --use:@autorest/java@4.0.x

alzimmermsft:
Fixes #1444 
Fixes #1419 

- Added `ConstantCustomization` to enable customization of static fields, such as on `ExpandableStringEnum` types.
- Added the ability to add imports to `ClassCustomization` directly rather than requiring the use of a method that takes a list of imports to add.
- Added overloads to parameter and body replacement methods in `ConstructorCustomization` and `MethodCustomization` which enable adding imports required by the code modifications.
- Fixed minor bug in `JavadocCustomization` where deprecation tag would be dropped if it was the only javadoc tag.
- Fixed a bug where removing a method without Javadocs but with annotations would leave dangling annotations.

### DIFF
--- a/customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
+++ b/customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
@@ -3,7 +3,6 @@
 
 package com.azure.autorest.customization;
 
-import com.azure.autorest.customization.implementation.CodeCustomization;
 import com.azure.autorest.customization.implementation.Utils;
 import com.azure.autorest.customization.implementation.ls.EclipseLanguageClient;
 import com.azure.autorest.customization.implementation.ls.models.FileChangeType;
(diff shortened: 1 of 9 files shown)


customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -330,26 +367,12 @@ public MethodCustomization addMethod(String method, List<String> importsToAdd) {
...
-            start = new Position(methodSignatureLine - 1, 0);
-        } else {
-            start = new Position(methodSignatureLine, 0);
-        }
+        MethodCustomization methodCustomization = getMethod(methodNameOrSignature);
+
+        int methodSignatureLine = methodCustomization.getSymbol().getLocation().getRange().getStart().getLine();
+
+        // Begin by getting the method's Javadoc to determine where to begin removal of the method.
+        Position start = methodCustomization.getJavadoc().getJavadocRange().getStart();
```
weidongxu-microsoft, line 375: From the comment, it still `start` from javadoc. Does it include the annotations (maybe multiple lines of them)?

weidongxu-microsoft (approved):
LGTM

Cannot say I understand all of it :-)

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -73,6 +73,20 @@ public String getClassName() {
         return className;
     }
 
+    /**
+     * Adds imports to the class.
+     *
+     * @param imports Imports to add.
+     * @return A new {@link ClassCustomization} updated with the new imports for chaining.
+     */
+    public ClassCustomization addImports(String... imports) {
```
srnagar, line 82: Another API we should add for class customization is to add a static block. `static { ... }`

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -188,6 +204,27 @@ public PropertyCustomization getProperty(String propertyName) {
...
 
+    /**
+     * Gets the constant level customization for a constant in the class.
+     * <p>
+     * For instance properties use {@link #getProperty(String)}.
+     *
+     * @param constantName The constant name.
+     * @return The constant level customization.
+     */
+    public ConstantCustomization getConstant(String constantName) {
```
srnagar, line 215: Should we have a `removeConstant` method too?

customization-base/src/main/java/com/azure/autorest/customization/ConstantCustomization.java
```diff
@@ -0,0 +1,183 @@
...
+     * words, so the {@code newName} will be capitalized.
+     * <p>
+     * This is a refactor operation, all references of the constant will be renamed and the getter method(s) for this
+     * property will be renamed accordingly as well.
+     *
+     * @param newName The new name for the constant.
+     * @return A new instance of {@link ConstantCustomization} for chaining.
+     * @throws NullPointerException If {@code newName} is null.
+     */
+    public ConstantCustomization rename(String newName) {
```
srnagar, line 103: I don't see an API to change the value of the constant. Should that be included too?

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
alzimmermsft: It can be added in the future when it's needed

customization-base/src/main/java/com/azure/autorest/customization/ConstantCustomization.java
alzimmermsft: Also can be added in the future

Issue #1336, opened 2022-02-02
For output-only models, do not generate setter methods


srnagar:
For models that are returned as part of response only and not used in any input requests, we should not generate setter methods and make the type immutable.

sharathmalladi:
Upvoting this issue :).

weidongxu-microsoft:
I should be able to get some time on it after Sept. 12 (after the Cadl Aug. work).

There was an early impl https://github.com/Azure/autorest.java/pull/1490 but that was closed.

weidongxu-microsoft:
An additional consideration on output-only model is that, we do not want to have ctor with parameters (aka, only the default ctor), in case of "required-fields-as-ctor-args" option.

As adding required parameter to output-only model is not breaking change in REST, we do not want it to be breaking change in Java model.

weidongxu-microsoft:
Will be included in 4.1.5 (not released yet).

Note the side effect of `--output-model-immutable` is that these model will have private ctor (Jackson can still de-serialize it) to avoid breaking changes when ctor changes.

PR #1452, opened 2022-05-17
dpg, improve javadoc

weidongxu-microsoft:
At present, `RequestOptions` cannot update a query parameter.

Hence for query parameter that already either set by client, or by constant, we do not put it in Javadoc. (as user cannot do anything about it)

PS: also add `usage` from m4, which would help https://github.com/Azure/autorest.java/issues/1336 (though we likely having a flag for that feature, as remove unused setter could be breaking changes)
https://github.com/Azure/autorest.java/commit/9c6c281546a0322cba16456befebd2cac9a45a11

### DIFF
--- a/androidgen/src/main/java/com/azure/autorest/android/model/clientmodel/AndroidProxyMethodParameter.java
+++ b/androidgen/src/main/java/com/azure/autorest/android/model/clientmodel/AndroidProxyMethodParameter.java
@@ -6,6 +6,7 @@ package com.azure.autorest.android.model.clientmodel;
 import com.azure.autorest.extension.base.model.codemodel.RequestParameterLocation;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.clientmodel.IType;
+import com.azure.autorest.model.clientmodel.ParameterSynthesizedOrigin;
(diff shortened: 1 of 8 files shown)


protocol-tests/src/main/java/fixtures/constants/AutoRestSwaggerConstantServiceClient.java
```diff
@@ -0,0 +1,448 @@
...
+     *     <tr><td>query-constant</td><td>int</td><td>Yes</td><td>Constant query property on the client that is a required parameter for operation 'constants_putClientConstants'</td></tr>
+     * </table>
+     *
+     * <p><strong>Header Parameters</strong>
+     *
+     * <table border="1">
+     *     <caption>Header Parameters</caption>
+     *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
+     *     <tr><td>header-constant</td><td>boolean</td><td>Yes</td><td>Constant header property on the client that is a required parameter for operation 'constants_putClientConstants'</td></tr>
+     * </table>
```
weidongxu-microsoft, line 434: Currently, DPG seems not set the query/header if they are from client or is constant.

We should either document the constant value in Javadoc, or set it in code.

PR #1469, opened 2022-05-20
mgmt, choose parent with polymorphic

weidongxu-microsoft:
Case: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/machinelearningservices/resource-manager/Microsoft.MachineLearningServices/preview/2022-02-01-preview/mfe.json#L8604-L8624
1st allOf is not polymorphic.
2nd allOf `AutoMLVertical` is polymorphic.

Existing code takes first parent to keep, and flatten the rest.
It would fail the polymorphic hierarchy if that polymorphic chain is not via first parent.

Fix: for polymorphic, takes the one in hierarchy as parent to keep.


weidongxu-microsoft:
looped in Alan since I am changing ModelMapper.

PR #1470, opened 2022-05-23
dpg, generate swagger/readme.md

weidongxu-microsoft:
add back the feature of generating swagger/readme.md, as the decision is not to use the configure on spec repo.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
@@ -320,11 +320,16 @@ public class ClientMethod {
         if (settings.isDataPlaneClient()) {
             imports.add("com.azure.core.http.HttpMethod");
             imports.add("com.azure.core.http.rest.Response");
-            imports.add("com.azure.core.http.rest.RequestOptions");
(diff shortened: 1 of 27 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -320,11 +320,16 @@ public void addImportsTo(Set<String> imports, boolean includeImplementationImpor
...
+
+            for (ClientMethodParameter parameter : getParameters()) {
+                parameter.addImportsTo(imports, includeImplementationImports);
+            }
 
             if (includeImplementationImports) {
+                imports.add(ClassType.Context.getFullName());
                 imports.add("com.azure.core.http.rest.SimpleResponse");
                 imports.add("com.azure.core.http.HttpRequest");
                 if (settings.getAddContextParameter() || settings.isContextClientMethodParameter()) {
```
weidongxu-microsoft, line 335: We should refactor this code.

There is little reason to completely branch on dpg or not.

weidongxu-microsoft:
@haolingdong-msft 

I've been refactoring the imports in `ClientMethod` in this PR as well. It might help on you query/header parameters PR (as these parameter of different class would need these import).

PR #1477, opened 2022-05-24
ClassCustomization method to add a static code block to class

srnagar:
This PR adds new methods to `ClassCustomization` to enable adding static code blocks to a class.

### DIFF
--- a/customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
+++ b/customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
@@ -48,6 +48,10 @@ public final class ClassCustomization extends CodeCustomization {
     private static final Pattern CONSTRUCTOR_SIGNATURE_PATTERN =
         Pattern.compile("^\\s*([^/*][\\w\\s]+\\([\\w\\s<>,\\.]*\\))\\s*\\{?$", Pattern.MULTILINE);
 
+
+    private static final Pattern CLASS_DECLARATION_PATTERN =
+            Pattern.compile("^\\s*[\\w*\\s*]*class\\s+[\\w*\\s*]*\\{?$", Pattern.MULTILINE);
+
     private static final Pattern BLOCK_OPEN = Pattern.compile("\\) *\\{");
     private static final Pattern PUBLIC_MODIFIER = Pattern.compile(" *public ");
     private static final Pattern PRIVATE_MODIFIER = Pattern.compile(" *private ");
@@ -84,6 +88,51 @@ public final class ClassCustomization extends CodeCustomization {
             return Utils.addImports(Arrays.asList(imports), this, this::refreshSymbol);
         }
 
+        return this;
+    }
+
+    /**
+     * Adds a static block to the class. The {@code staticCodeBlock} should include the static keyword followed by
+     * the static code.
+     * @param staticCodeBlock The static code block including the static keyword.
+     * @return The updated {@link ClassCustomization}.
+     */
+    public ClassCustomization addStaticBlock(String staticCodeBlock) {
+        return addStaticBlock(staticCodeBlock, null);
+    }
+
+    /**
+     * Adds a static block to the class. The {@code staticCodeBlock} should include the static keyword followed by
+     * the static code.
+     * @param staticCodeBlock The static code block including the static keyword.
+     * @param importsToAdd The list of imports to add to the class.
+     * @return The updated {@link ClassCustomization}.
+     */
+    public ClassCustomization addStaticBlock(String staticCodeBlock, List<String> importsToAdd) {
+        int staticBlockStartLine;
+        // Find all constructor and field symbols.
+        List<SymbolInformation> staticBlockLocationFinder = languageClient.listDocumentSymbols(fileUri).stream()
+                .filter(symbol -> symbol.getKind() == SymbolKind.FIELD)
+                .collect(Collectors.toList());
+
+        if (Utils.isNullOrEmpty(staticBlockLocationFinder)) {
+            staticBlockStartLine = symbol.getLocation().getRange().getStart().getLine();
+        } else {
+            SymbolInformation symbol = staticBlockLocationFinder.get(staticBlockLocationFinder.size() - 1);
+            staticBlockStartLine = symbol.getLocation().getRange().getStart().getLine();
+        }
+        int indentAmount = Utils.getIndent(editor.getFileLine(fileName, staticBlockStartLine)).length();
+
+        editor.insertBlankLine(fileName, ++staticBlockStartLine, false);
+        Position staticBlockPosition = editor.insertBlankLineWithIndent(fileName, ++staticBlockStartLine, indentAmount);
+
+        // replace
+        editor.replaceWithIndentedContent(fileName, staticBlockPosition, staticBlockPosition, staticCodeBlock,
+                staticBlockPosition.getCharacter());
+        
+        if (importsToAdd != null) {
+            return Utils.addImports(importsToAdd, this, this::refreshSymbol);
+        }
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -87,6 +87,51 @@ public ClassCustomization addImports(String... imports) {
...
+    /**
+     * Adds a static block to the class. The {@code staticCodeBlock} should include the static keyword followed by
+     * the static code.
+     * @param staticCodeBlock The static code block including the static keyword.
+     * @param importsToAdd The list of imports to add to the class.
+     * @return The updated {@link ClassCustomization}.
+     */
+    public ClassCustomization addStaticBlock(String staticCodeBlock, List<String> importsToAdd) {
+        int staticBlockStartLine;
+        // Find all constructor and field symbols.
```
alzimmermsft, line 109: This says it finds all constructor and field symbols but it only looks to be finding field symbols

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -87,6 +87,51 @@ public ClassCustomization addImports(String... imports) {
...
+        List<SymbolInformation> staticBlockLocationFinder = languageClient.listDocumentSymbols(fileUri).stream()
+                .filter(symbol -> symbol.getKind() == SymbolKind.FIELD)
+                .collect(Collectors.toList());
```
alzimmermsft, line 112: We could change this to only getting the last found field instead of collecting them by using the logic outlined here: https://www.geeksforgeeks.org/find-the-last-element-of-a-stream-in-java/

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -87,6 +87,51 @@ public ClassCustomization addImports(String... imports) {
...
+
+        if (Utils.isNullOrEmpty(staticBlockLocationFinder)) {
+            staticBlockStartLine = symbol.getLocation().getRange().getStart().getLine();
+        } else {
+            SymbolInformation symbol = staticBlockLocationFinder.get(staticBlockLocationFinder.size() - 1);
+            staticBlockStartLine = symbol.getLocation().getRange().getStart().getLine();
+        }
+        int indentAmount = Utils.getIndent(editor.getFileLine(fileName, staticBlockStartLine)).length();
+
+        editor.insertBlankLine(fileName, ++staticBlockStartLine, false);
```
alzimmermsft, line 122: I'd rather track the symbol separately to determine the indent amount rather than using static block start line initially as the line where the last found field was and use it to first find indentation then pre-incrementing. `++variable` and `variable++` is generally a source of confusion or bugs and I'd rather keep this more explicit.

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -87,6 +87,51 @@ public ClassCustomization addImports(String... imports) {
...
+                .filter(symbol -> symbol.getKind() == SymbolKind.FIELD)
+                .collect(Collectors.toList());
+
+        if (Utils.isNullOrEmpty(staticBlockLocationFinder)) {
+            staticBlockStartLine = symbol.getLocation().getRange().getStart().getLine();
+        } else {
+            SymbolInformation symbol = staticBlockLocationFinder.get(staticBlockLocationFinder.size() - 1);
+            staticBlockStartLine = symbol.getLocation().getRange().getStart().getLine();
+        }
+        int indentAmount = Utils.getIndent(editor.getFileLine(fileName, staticBlockStartLine)).length();
```
alzimmermsft, line 120: Couldn't this produce the wrong indentation level if a field isn't found and the class symbol is used to determine the static block start line?

customization-base/src/test/java/com/azure/autorest/customization/ClassCustomizationTests.java
```diff
@@ -0,0 +1,101 @@
...
+                "    static { String a = \"foo\"; }",
+                "}");
+
+        Customization customization = new Customization() {
+            @Override
+            public void customize(LibraryCustomization libraryCustomization, Logger logger) {
+                ClassCustomization classCustomization = libraryCustomization.getPackage("foo")
+                        .getClass("Foo");
+
+                classCustomization.addStaticBlock("static { String a = \"foo\"; }");
```
alzimmermsft, line 43: Do we need to have the static block include `static` and `{}`? Could we handle that in the API?

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -87,6 +87,51 @@ public ClassCustomization addImports(String... imports) {
...
+    }
+
+    /**
+     * Adds a static block to the class. The {@code staticCodeBlock} should include the static keyword followed by
+     * the static code.
+     * @param staticCodeBlock The static code block including the static keyword.
+     * @param importsToAdd The list of imports to add to the class.
+     * @return The updated {@link ClassCustomization}.
+     */
+    public ClassCustomization addStaticBlock(String staticCodeBlock, List<String> importsToAdd) {
```
alzimmermsft, line 107: Do we want to make static block its own customization type? That way we can support mutating an existing static block.

srnagar: Since there could be multiple static blocks in a class, mutating an existing static block would be equivalent to returning the raw editor. Also, AutoRest doesn't generate any static blocks, so, don't think it's worth creating a new type.

customization-base/src/test/java/com/azure/autorest/customization/ClassCustomizationTests.java
srnagar: Yeah, we could handle it in the API itself. I'll update the API.

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
srnagar: Customizations shouldn't really include indentation logic ideally. But this pattern is used in all customizations. The formatter at the end of the code generation should ensure everything is indented correctly.

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
alzimmermsft: Good point and we can always change this in the future

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -32,6 +32,8 @@
  * The class level customization for an AutoRest generated class.
  */
 public final class ClassCustomization extends CodeCustomization {
+    private static final int INDENT_LENGTH = 4;
```
alzimmermsft, line 35: nit: this isn't used

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -87,6 +89,62 @@ public ClassCustomization addImports(String... imports) {
...
+                .filter(symbol -> symbol.getKind() == SymbolKind.FIELD)
+                .reduce((first, second) -> second);
+
+        int indentAmount = 4;
+        if (lastSymbol.isPresent()) {
+            // the line number of the last field declaration
+            lastSymbolLine = lastSymbol.get().getLocation().getRange().getStart().getLine();
+            indentAmount = Utils.getIndent(editor.getFileLine(fileName, lastSymbolLine)).length();
+        }
+        System.out.println("indent amount " + indentAmount);
```
alzimmermsft, line 130: nit: assume this was for testing purposes?

alzimmermsft (approved):

PR #1480, opened 2022-05-25
dpg, add examples to test swagger

weidongxu-microsoft:


protocol-tests/src/main/java/fixtures/requiredheaderquery/DpgRequiredHeaderQueryClient.java
```diff
@@ -42,6 +42,8 @@ public final class DpgRequiredHeaderQueryClient {
      * <table border="1">
      *     <caption>Query Parameters</caption>
      *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
+     *     <tr><td>parameter_optional_csv_int_array</td><td>List&lt;Integer&gt;</td><td>No</td><td>I am a required csv int array parameter. In the form of "," separated string.</td></tr>
+     *     <tr><td>parameter_optional_multi_int_array</td><td>List&lt;Integer&gt;</td><td>No</td><td>The array of integer collect by multi. Call {@link RequestOptions#addQueryParam} to add string to array.</td></tr>
      * </table>
```
weidongxu-microsoft, line 47: @haolingdong-msft 

You had a bug in javadoc.
See previous code, "Query Parameters" and the table is added, while there is no optional parameter in that table.

Please fix.

PR #1481, opened 2022-05-25
Update POMs to Simplify Release and Standardize Build

alzimmermsft:
Updates the parent POM and children POM configurations to simplify engineering for release and to standardize build further.


srnagar (approved):
We should also consider bumping up the Java version to 11

PR #1483, opened 2022-05-26
unit test, use mock connection

weidongxu-microsoft:


customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -127,7 +127,7 @@ public ClassCustomization addStaticBlock(String staticCodeBlock, List<String> im
             lastSymbolLine = lastSymbol.get().getLocation().getRange().getStart().getLine();
             indentAmount = Utils.getIndent(editor.getFileLine(fileName, lastSymbolLine)).length();
         }
-        System.out.println("indent amount " + indentAmount);
+//        System.out.println("indent amount " + indentAmount);
```
weidongxu-microsoft, line 130: I commented out this line. I assume we should not use stdout for logging.

javagen/src/test/java/com/azure/autorest/mapper/JavagenTests.java
```diff
@@ -1,18 +0,0 @@
...
-package com.azure.autorest.mapper;
-
-import com.azure.autorest.Javagen;
-import com.azure.autorest.extension.base.jsonrpc.Connection;
-import org.junit.Ignore;
-import org.junit.Test;
-
-public class JavagenTests {
-    @Test
-    @Ignore
```
weidongxu-microsoft, line 13: deleted, as it is Ignored anyway.

PR #1484, opened 2022-05-26
remove parameter table when all params are required

haolingdong-msft:


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -118,6 +121,16 @@ private static void optionalParametersJavadoc(String title, List<ProxyMethodPara
...
     }
 
+    private static boolean hasOptionalParameters(List<ProxyMethodParameter> parameters) {
+        boolean res = false;
+        for (ProxyMethodParameter parameter : parameters) {
+            if (!parameter.getIsRequired()) {
+                res = true;
+            }
+        }
+        return res;
```
weidongxu-microsoft, line 131: try use the stream

protocol-tests/src/main/java/fixtures/constants/AutoRestSwaggerConstantServiceAsyncClient.java
```diff
@@ -157,14 +149,6 @@ public Mono<Response<Void>> putNoModelAsStringRequiredTwoValueNoDefaultWithRespo
     /**
      * Puts constants to the testserver.
      *
-     * <p><strong>Query Parameters</strong>
-     *
-     * <table border="1">
-     *     <caption>Query Parameters</caption>
-     *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
-     *     <tr><td>input</td><td>String</td><td>Yes</td><td>The input parameter</td></tr>
```
weidongxu-microsoft, line 165: your base branch is wrong? "Yes" should not in their in the first place.

PR #1490, opened 2022-05-31
client, support immutable output

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1336

Please take a quick look.

I might need to add some swagger for test, for cases e.g. superclass is input/output, subclass is output; and vise versa.
A more complex case is the subclass as output, but inherits multiple superclass, and some of them get flattened.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
@@ -10,6 +10,7 @@ import com.azure.autorest.extension.base.model.codemodel.Languages;
 import com.azure.autorest.extension.base.model.codemodel.ObjectSchema;
 import com.azure.autorest.extension.base.model.codemodel.Property;
 import com.azure.autorest.extension.base.model.codemodel.Schema;
+import com.azure.autorest.extension.base.model.codemodel.SchemaContext;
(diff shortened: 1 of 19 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -244,6 +246,9 @@ public ClientModel map(ObjectSchema compositeType) {
             List<ClientModelPropertyReference> propertyReferences = new ArrayList<>();
             for (Property property : compositeTypeProperties) {
                 ClientModelProperty modelProperty = Mappers.getModelPropertyMapper().map(property);
+                if (settings.isOutputImmutable() && !compositeType.getUsage().contains(SchemaContext.INPUT)) {
+                    modelProperty = modelProperty.newBuilder().isReadOnly(true).build();
+                }
```
weidongxu-microsoft, line 251: I cannot think of a better solution.

We want to set all Properties to read-only, if Model is not used as input. But the `ModelPropertyMapper` only sees the Property, not the Model.

Doing it in template is too hard (tried and aborted).

weidongxu-microsoft (commented):
@srnagar Please take a quick look, see whether the idea is good for you. I probably need to find some time to e.g. add some swagger as test.

PR #1497, opened 2022-06-06
client, improve the check of whether type is client model

weidongxu-microsoft:
There is certain code that implicitly assume certain pattern for namespace, e.g. not start with `java.lang`.
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java#L158

Ideally, we should throw when namespace is empty or invalid.

But autorest seems having problem when configure is `--java.namespace=<>`, preprocess cannot read it, but javagen can.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
+++ b/javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
@@ -296,14 +296,34 @@ public class ClientModelUtil {
 
     private static Function<String, ClientModel> getClientModelFunction = name -> ClientModels.getInstance().getModel(name);
 
+    /**
+     * Replace the default function of getting ClientModel by name.
+     * <p>
+     * Used in Fluent for providing additional ClientModel that exists in azure-core-management,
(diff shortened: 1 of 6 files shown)


javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
```diff
@@ -302,6 +304,16 @@ public static ClientModel getClientModel(String name) {
...
 
+    public static boolean isClientModel(IType type) {
+        if (type instanceof ClassType) {
+            ClassType classType = (ClassType) type;
+            return classType.getPackage().startsWith(JavaSettings.getInstance().getPackage())
+                    && getClientModel(classType.getName()) != null;
+        } else {
+            return false;
+        }
+    }
```
weidongxu-microsoft, line 315: Use this to make a more solid inference on whether the `type` is a client model,
than previous
```java
type instanceof ClassType && ((ClassType) type).getPackage().startsWith(JavaSettings.getInstance().getPackage())
```

PR #1500, opened 2022-06-07
dpg, use require first, in swagger/readme.md

weidongxu-microsoft:
sample output (Wes/Laurent ask to use "require" instead of "input-file"):

## Generate autorest code

```yaml
require: c:/github/azure-rest-api-specs/specification/purview/data-plane/Azure.Analytics.Purview.Account/readme.md
output-folder: ../
java: true
regenerate-pom: false
title: PurviewAccountClient
data-plane: true
generate-tests: true
generate-samples: true
namespace: com.azure.analytics.purview.account
service-versions:
- 2019-11-01-preview
```

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/SwaggerReadmeTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/SwaggerReadmeTemplate.java
@@ -3,6 +3,7 @@
 
 package com.azure.autorest.template;
 
+import com.azure.autorest.extension.base.plugin.AutorestSettings;
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.projectmodel.Project;
 import com.azure.autorest.util.TemplateUtil;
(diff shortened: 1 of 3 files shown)


javagen/src/main/java/com/azure/autorest/template/SwaggerReadmeTemplate.java
```diff
@@ -78,6 +79,25 @@ public String write(Project project) {
...
     }
 
+    private static void addRequireOrInputFile(Map<String, Object> objectNode, AutorestSettings autorestSettings) {
+        // try use "require"
+        boolean useRequire = false;
+        List<String> requireList = autorestSettings.getRequire();
+        if (!CoreUtils.isNullOrEmpty(requireList)) {
+            String require = requireList.iterator().next();
+
+            if (require.contains("data-plane")) {
```
weidongxu-microsoft, line 89: this is a hack for the usual location of data-plane readme e.g. https://github.com/Azure/azure-rest-api-specs/blob/main/specification/purview/data-plane/readme.md

Issue #1453, opened 2022-05-17
DPG, method parameter from client should not require setting in every method call


weidongxu-microsoft:
Example

This 2 parameters had to be provided for a sucess API call.
https://github.com/Azure/autorest.java/blob/4e1769dc343fa85cea0153c41f3220447aa7c1c4/protocol-tests/src/test/java/fixtures/constants/ConstantsTests.java#L23-L25

However, the 2 parameters is already configured in ClientBuilder and hence already set by user.
https://github.com/Azure/autorest.java/blob/4e1769dc343fa85cea0153c41f3220447aa7c1c4/protocol-tests/src/main/java/fixtures/constants/AutoRestSwaggerConstantServiceClientBuilder.java#L147-L183

We should not require user to set them again, for every API call under that client.

Test swagger
https://github.com/Azure/autorest.testserver/blob/main/swagger/constants.json

if it's a constant (not from client), we should just set in in client method impl - don't even need to specify in the javadoc (as it would be considered an implementation detail that the user has no flexibility to change). 

However, if the user sets it somehow, then we allow it and let the service handle the error

weidongxu-microsoft:
vanilla code
https://github.com/Azure/autorest.java/blob/08ba0456418cb47be168156e59bf24329cc6d676/vanilla-tests/src/main/java/fixtures/constants/Contants.java#L1108-L1122

weidongxu-microsoft:
This would apply to constant as well.

PR #1502, opened 2022-06-08
dpg, support method parameter from client should not require setting in every method call

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1453


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -219,7 +219,9 @@ public Map<Request, List<ProxyMethod>> map(Operation operation) {
                     final boolean parameterIsRequired = parameter.isRequired();
                     final boolean parameterIsClientOrApiVersion = ClientModelUtil.getClientDefaultValueOrConstantValue(parameter) != null
                             && parameter.getLanguage().getJava().getName().equalsIgnoreCase("apiversion");
-                    if (parameterIsRequired || parameterIsClientOrApiVersion) {
+                    if (parameterIsRequired
+                            || !parameterIsRequired && (proxyMethodParameter.getIsConstant() || proxyMethodParameter.getFromClient())
```
weidongxu-microsoft, line 223: Use similar patten as e.g. `parameterIsRequired`. use a local variable with clear name for this condition of parameter (and remove `!parameterIsRequired`, this is not a condition for your case).

protocol-tests/src/main/java/fixtures/constantandclientparam/ConstantAndClientParameterServiceClientBuilder.java
```diff
@@ -0,0 +1,401 @@
...
+     * Sets Constant query property that is a non-required parameter.
+     *
+     * @param queryNonRequiredConstant the queryNonRequiredConstant value.
+     * @return the ConstantAndClientParameterServiceClientBuilder.
+     */
+    @Generated
+    public ConstantAndClientParameterServiceClientBuilder queryNonRequiredConstant(int queryNonRequiredConstant) {
+        this.queryNonRequiredConstant = queryNonRequiredConstant;
+        return this;
+    }
```
weidongxu-microsoft, line 271: Why does it get to the client/builder? Could you double check your swagger and m4 output?

It is constant, not "from client".

protocol-tests/src/main/java/fixtures/constantandclientparam/ConstantAndClientParameterServiceAsyncClient.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Pass constants from the client to this function. Will pass in constant path, query, and header parameters.
+     *
+     * <p><strong>Query Parameters</strong>
+     *
+     * <table border="1">
+     *     <caption>Query Parameters</caption>
+     *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
+     *     <tr><td>query-non-required-client-constant</td><td>Integer</td><td>No</td><td>Constant query property on the client that is a non-required parameter</td></tr>
+     *     <tr><td>query-non-required-constant</td><td>Integer</td><td>No</td><td>Constant query property that is a non-required parameter</td></tr>
```
weidongxu-microsoft, line 44: If you've set them from client, they do not need to be in javadoc. Property from builder and client should have already covered them.

But non-required-constant should not even be a constant, preprocess will transfer it to sealed choice.

protocol-tests/swagger/constant-and-from-client-parameters.json
```diff
@@ -0,0 +1,142 @@
...
+          },
+          {
+            "$ref": "#/parameters/QueryNonRequiredClientConstant"
+          },
+          {
+            "$ref": "#/parameters/QueryRequiredConstant"
+          },
+          {
+            "$ref": "#/parameters/QueryNonRequiredConstant"
+          }
```
weidongxu-microsoft, line 44: separate the case of "constant" and "from client" (don't specify both in same paramter).

1. required client
2. non-required client
3. require constant (but not client)
4. non-required constant (which likely not a constant)

protocol-tests/swagger/constant-and-from-client-parameters.json
```diff
@@ -0,0 +1,142 @@
...
+      "enum": [
+        100
+      ],
+      "x-ms-enum": {
+        "modelAsString": false
+      }
+    },
+    "QueryRequiredConstant": {
+      "name": "query-required-constant",
+      "x-ms-parameter-location": "client",
```
weidongxu-microsoft, line 116: why "client"?

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -111,7 +111,7 @@ private static void optionalParametersJavadoc(String title, List<ProxyMethodPara
         commentBlock.line(String.format("    <caption>%s</caption>", title));
         commentBlock.line("    <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>");
         for (ProxyMethodParameter parameter : parameters) {
-            if (!parameter.getIsRequired()) {
+            if (!parameter.getIsRequired() && !parameter.getIsConstant()) {
```
weidongxu-microsoft, line 114: exclude fromClient as well?

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -111,7 +111,8 @@ private static void optionalParametersJavadoc(String title, List<ProxyMethodPara
         commentBlock.line(String.format("    <caption>%s</caption>", title));
         commentBlock.line("    <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>");
         for (ProxyMethodParameter parameter : parameters) {
-            if (!parameter.getIsRequired()) {
+            boolean parameterIsConstantOrFromClient = parameter.getIsConstant() || parameter.getFromClient();
+            if (!parameter.getIsRequired() && !parameterIsConstantOrFromClient) {
```
weidongxu-microsoft, line 115: do you need to modify `hasOptionalParameters` as well? And maybe rename it (as it no longer for optional parameter only.

protocol-tests/src/main/java/fixtures/constantandclientparam/ConstantAndClientParameterServiceAsyncClient.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Pass constants from the client to this function. Will pass in constant path, query, and header parameters.
+     *
+     * <p><strong>Query Parameters</strong>
+     *
+     * <table border="1">
+     *     <caption>Query Parameters</caption>
+     *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
```
weidongxu-microsoft, line 42: Relate to the comment above. You can see this is not correct.

weidongxu-microsoft (approved):

PR #1518, opened 2022-06-21
Remove JNA and Fix Potential Race Condition

alzimmermsft:
Removes JNA as a dependency for code customizations and removes a timed `Thread.join` with an untimed `Thread.join` to prevent a race condition with the server starting up.


weidongxu-microsoft (approved):
OK for me if there is no issue with CI.

Issue #1464, opened 2022-05-19
Improve Javadoc to show optional/required properties and include sample codesnippets


srnagar:

mikekistler:
Just to make sure this detail isn't lost: 

@tg-msft made the specific suggestion to add a sentence just above the query parameters table along the lines of: "You can add these to a request with `request.queryParams.add( ...`" (or whatever the actual syntax is).

Also include the description and optional/required from the API definition for each property in the Response Body Schema section (and the Request Body Schema if it has a similar section).

PR #1519, opened 2022-06-22
Improve Javadoc to show optional/required properties and include sample codesnippets

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1464
Required/Optional flag:
image

Query table description:
image

Header table description:
image


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -55,16 +56,21 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
                     .collect(Collectors.toList());
             if (!headerParameters.isEmpty()) {
                 optionalParametersJavadoc("Header Parameters", headerParameters, commentBlock);
+                commentBlock.line("You can add these to a request with {@code new RequestOptions.addHeader()}");
```
weidongxu-microsoft, line 59: Should we use `setHeader()`?

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -247,4 +253,8 @@ private static String methodParameterDescriptionOrDefault(ClientMethodParameter
         }
         return doc;
     }
+
+    private static String appendOptionalOrRequiredAttribute(boolean isRequired, boolean isRootSchema) {
+        return isRootSchema ? "" : isRequired ? " (Required)" : " (Optional)";
```
weidongxu-microsoft, line 258: nit, not sure upper-case or lower-case be better, just for discussion.

protocol-tests/swagger/required-optional-body.json
```diff
@@ -0,0 +1,134 @@
+{
```
weidongxu-microsoft, line 1: Does this test provide more scenario than other existing tests (testserver and others), or can it be combined with existing?

We might not want to add too many test swagger, as it have some cost on dev time and CI time.

protocol-resilience-test/llcupdate1/src/main/java/fixtures/llcresi/DpgClient.java
```diff
@@ -125,7 +131,7 @@ public Response<BinaryData> putRequiredOptionalWithResponse(String requiredParam
     }
 
     /**
-     * POST a JSON or a JPEG.
+     * POST a JSON or a JPEG. You can add these to a request with {@code new RequestOptions.addHeader()}
```
weidongxu-microsoft, line 134: Why is this here?

Please be careful on reviewing the output of the swaggers, they serve this exact purpose.

weidongxu-microsoft (commented):
Overall LGTM.

Still some problems commented.

protocol-tests/swagger/required-optional-body.json
haolingdong-msft: It just contains different cases in one swagger json so that it's easier to validate the result. But autorest.testserver covers those cases as well in different swagger json. We can remove this json.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -43,8 +43,9 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
                     // ignore if synthesized by modelerfour, i.e. api-version
                     .filter(p -> p.getOrigin() == ParameterSynthesizedOrigin.NONE)
                     .collect(Collectors.toList());
-            if (!queryParameters.isEmpty()) {
+            if (!queryParameters.isEmpty() && hasParametersToPrintInJavadoc(queryParameters)) {
                 optionalParametersJavadoc("Query Parameters", queryParameters, commentBlock);
+                commentBlock.line("You can add these to a request with {@code new RequestOptions.addQueryParam()}");
```
weidongxu-microsoft, line 48: Recommend to use `{@link RequestOptions#addQueryParam}`, as not every parameter require a `new RequestOptions`

haolingdong-msft:
@srnagar Could you please take a look at this pr?

weidongxu-microsoft (approved):

PR #1538, opened 2022-06-27
generate UT on API

weidongxu-microsoft:
Sample https://github.com/Azure/azure-sdk-for-java/pull/29660

(tests on model exists before this PR)

---

Skip some LRO tests, which is hard to write UT properly.

---

Coverage (overall about 27% line)

image


weidongxu-microsoft:
Better when switched to fluent method (30%)

image

Issue #1494, opened 2022-06-01
DPG, wrong request schema for multiple media types


haolingdong-msft:
https://github.com/Azure/autorest.java/pull/1489

`Flux<ByteBuffer>` does not make sense.

```
 /**
     * json and binary types with json type first.
     *
     * <p><strong>Request Body Schema</strong>
     *
     * <pre>{@code
     * Flux<ByteBuffer>
     * }</pre>
     *
     * <p><strong>Response Body Schema</strong>
     *
     * <pre>{@code
     * String
     * }</pre>
     *
     * @param contentType Upload file type.
     * @param contentLength The contentLength parameter.
     * @param input Input parameter.
     * @param requestOptions The options to configure the HTTP request before HTTP client sends it.
     * @throws HttpResponseException thrown if the request is rejected by server.
     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
     * @return the response body along with {@link Response}.
     */
```

PR #1541, opened 2022-06-28
DPG, fix wrong requestbodyschema javadoc for binary schema

XiaofeiCao:
Fix https://github.com/Azure/autorest.java/issues/1494

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
@@ -145,7 +145,7 @@ public abstract class ClientMethodTemplateBase implements IJavaTemplate<ClientMe
         }
         commentBlock.line("<p><strong>Request Body Schema</strong></p>");
         commentBlock.line("<pre>{@code");
-        bodySchemaJavadoc(requestBodyType, commentBlock, "", null, typesInJavadoc, isBodyParamRequired, true, true);
+        bodySchemaJavadoc(requestBodyType, commentBlock, "", null, typesInJavadoc, isBodyParamRequired, true);
         commentBlock.line("}</pre>");
     }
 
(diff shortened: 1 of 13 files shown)


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -205,17 +206,34 @@ private static void bodySchemaJavadoc(IType type, JavaJavadocComment commentBloc
             } else {
                 commentBlock.line(indent + appendOptionalOrRequiredAttribute(isRequired, isRootSchema) + "{");
             }
-            bodySchemaJavadoc(((MapType) type).getValueType(), commentBlock, nextIndent, "String", typesInJavadoc, isRequired, false);
+            bodySchemaJavadoc(((MapType) type).getValueType(), commentBlock, nextIndent, "String", typesInJavadoc, isRequired, false, true);
             commentBlock.line(indent + "}");
         } else {
+            String javadoc;
+            if (isInRequestBody) {
```
XiaofeiCao, line 213: In DPG, when response has binary schema, response type for the generated proxy method is `Response<Void>`. I believe there's no `Flux<ByteBuffer>` for response body type (which shouldn't be converted anyway), but for clarity, I restrict the conversion only in request body schema.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -205,17 +206,34 @@ private static void bodySchemaJavadoc(IType type, JavaJavadocComment commentBloc
...
+                javadoc = convertToRequestBodySchemaJavadoc(type);
+            } else {
+                javadoc = type.toString();
+            }
```
weidongxu-microsoft, line 217: I think we don't need to distinguish whether the Flux in body or not.

I kind of not able to think a way that it is not in body actually.

XiaofeiCao: ok

weidongxu-microsoft (commented):
@haolingdong-msft  to also review.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
weidongxu-microsoft: Basically, for DPG, if response is not Void, it would be `BinaryData`. So it is fine to convert `Flux<ByteBuffer>` to `BinaryData`  in response (if it happens).

weidongxu-microsoft: However, above should actually have `BinaryData` as response, not `Void`.

Do you have the source?

@haolingdong-msft for possible investigation (this would be another issue not related to this PR)

weidongxu-microsoft: At least make these be good
https://github.com/Azure/autorest.testserver/blob/main/swagger/body-file.json

XiaofeiCao: I manually added the above swagger content to test the result. For fluent, response type will be `StreamResponse` and `Flux<ByteBuffer>`, for DPG, it will be `Response<Void>`.

XiaofeiCao: I can help with the investigation if haoling has other priorities.

weidongxu-microsoft: > I manually added the above swagger content to test the result. For fluent, response type will be `StreamResponse` and `Flux<ByteBuffer>`, for DPG, it will be `Response<Void>`.

The DPG likely be wrong. (what if 200 is before 204 in swagger?)

XiaofeiCao: Same `Response<Void>` if 200 is before 204.

weidongxu-microsoft (approved):

haolingdong-msft (approved):

XiaofeiCao:
issue created for wrong response type of binary response body:
https://github.com/Azure/autorest.java/issues/1542

Issue #1534, opened 2022-06-23
DPG, add content-type in header param table section when there is single content-type and body is optional


haolingdong-msft:
For multiple content type case, we will add content-type in @param or param table section depends on the body is required or not.
For single content type, if body is required, code supply the content type. If body is optional, we need to add content-type in header param table section, and the description should mention the value required.

PR #1543, opened 2022-06-29
dpg, add content type as header parameter when body is optional and content type is single

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1534

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
@@ -205,13 +205,16 @@ public class ProxyMethodMapper implements IMapper<Operation, Map<Request, List<P
             List<ProxyMethodParameter> allParameters = new ArrayList<>();
             List<ProxyMethod> proxyMethods = new ArrayList<>();
             // add content-type parameter to allParameters when body is optional and there is single content type
-            boolean isBodyParamRequired = request.getParameters()
-                    .stream().filter(p -> p.getProtocol() != null && p.getProtocol().getHttp() != null ? p.getProtocol().getHttp().getIn() == RequestParameterLocation.BODY : false)
-                    .map(Parameter::isRequired).findFirst().orElse(false);
-            if (settings.isDataPlaneClient() && MethodUtil.getContentTypeCount(operation.getRequests()) == 1 && !isBodyParamRequired) {
-                Parameter contentTypeParameter = MethodUtil.createContentTypeParameter(request, operation);
-                allParameters.add(Mappers.getProxyParameterMapper().map(contentTypeParameter));
+            if (settings.isDataPlaneClient()) {
+                boolean isBodyParamRequired = request.getParameters()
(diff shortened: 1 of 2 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -204,6 +204,14 @@ public Map<Request, List<ProxyMethod>> map(Operation operation) {
             List<ProxyMethodParameter> parameters = new ArrayList<>();
             List<ProxyMethodParameter> allParameters = new ArrayList<>();
             List<ProxyMethod> proxyMethods = new ArrayList<>();
+            // add content-type parameter to allParameters when body is optional and there is single content type
+            boolean isBodyParamRequired = request.getParameters()
+                    .stream().filter(p -> p.getProtocol() != null && p.getProtocol().getHttp() != null ? p.getProtocol().getHttp().getIn() == RequestParameterLocation.BODY : false)
+                    .map(Parameter::isRequired).findFirst().orElse(false);
```
weidongxu-microsoft, line 210: nit, move this (and the `MethodUtil.getContentTypeCount(operation.getRequests()) == 1` check) after the if block of `settings.isDataPlaneClient()`.

reason, if some variable only useful for DPG, avoid it exposed to code not really need to see that var).

javagen/src/main/java/com/azure/autorest/util/MethodUtil.java
```diff
@@ -103,6 +102,64 @@ public static Request tryMergeBinaryRequests(List<Request> requests, Operation o
         return selectedRequest;
     }
 
+    /**
+     * @param request   the request to put contentType on
+     * @param operation
+     * @return the created content type parameter
+     */
+    public static Parameter createContentTypeParameter(Request request, Operation operation) {
```
weidongxu-microsoft, line 110: Is there any change from the one that delete?

javagen/src/main/java/com/azure/autorest/util/MethodUtil.java
```diff
@@ -242,13 +237,16 @@ private static int findIndexForContentTypeParam(List<Parameter> parameters) {
...
         for (Request request : requests) {
-            for (String mediaType : request.getProtocol().getHttp().getMediaTypes()) {
-                ChoiceValue choiceValue = new ChoiceValue();
-                choiceValue.setValue(mediaType);
-                Language language = new Language();
-                language.setName(mediaType.toUpperCase(Locale.ROOT));
-                language.setDescription("Content Type " + mediaType);
-                choiceValues.add(choiceValue);
+            if (request.getProtocol().getHttp().getMediaTypes() != null
+                    && !request.getProtocol().getHttp().getMediaTypes().isEmpty()) {
```
weidongxu-microsoft, line 241: nit, `CoreUtil.isNullOrEmpty`

javagen/src/main/java/com/azure/autorest/util/MethodUtil.java
haolingdong-msft: no change

weidongxu-microsoft (approved):

PR #1565, opened 2022-07-21
Update builder to make them reusable

srnagar:
Currently, the `buildAsyncClient()` method updates the instance properties if the values are `null` when the first client is created. This causes the builder to set state and unless the state is explicitly reset, the subsequent clients will use the state that was previously set. This is an issue specifically for the pipeline.

### Issue:
```java
ClientBuilder builder = new ClientBuilder()
                           .endpoint(endpoint);

// create the first client
Client client1 = builder.buildClient();

// reuse the builder to set a custom HTTP client
builder.httpClient(myCustomHttpClient);

// create another client from the same builder
// this will not use the custom HTTP client, because the pipeline is already created with a default HTTP client
Client client2 = builder.buildClient(); 
```


javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -342,6 +345,14 @@ public final void write(ClientBuilder clientBuilder, JavaFile javaFile) {
         });
     }
 
+    private String getClientConstructorArgName(ServiceClientProperty property) {
+        if (property.getDefaultValueExpression() != null
```
weidongxu-microsoft, line 349: Is there reason to exclude variables with default value?

The error in fluent CI seems complaining that `buildDefaultPollInterval` is created but not used in client ctor. (which have default value as `Duration.ofSeconds(30)`)

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -274,17 +277,17 @@ public final void write(ClientBuilder clientBuilder, JavaFile javaFile) {
...
-                    function.line(String.format("%1$s client = new %2$s(pipeline, %3$s, defaultPollInterval, environment%4$s);",
+                    function.line(String.format("%1$s client = new %2$s(buildPipeline, %3$s, defaultPollInterval, environment%4$s);",
                             serviceClient.getClassName(),
                             serviceClient.getClassName(),
                             serializerExpression,
                             constructorArgs));
                 } else {
-                    function.line(String.format("%1$s client = new %2$s(pipeline, %3$s%4$s);",
+                    function.line(String.format("%1$s client = new %2$s(buildPipeline, %3$s%4$s);",
                             serviceClient.getClassName(), serviceClient.getClassName(), serializerExpression, constructorArgs));
```
weidongxu-microsoft, line 291: There is many `build` hardcoded, some is directly in string, some add to string.

Would it be better to use a method e.g. `getLocalBuildVariableName()`, and hide what the variable called within the method?

nit, personally I think `localPipeline` might be easier to understand? Or `myPipeline`, or `clientPipeline`.

weidongxu-microsoft (approved):
Overall LGTM.

package.json
```diff
@@ -48,6 +48,6 @@
   ],
   "devDependencies": {
     "autorest": "^3.0.0",
-    "@microsoft.azure/autorest.testserver": "3.3.31"
+    "@microsoft.azure/autorest.testserver": "3.3.34"
```
weidongxu-microsoft, line 51: appears you need to regen tests after upgrade

javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -274,17 +279,20 @@ public final void write(ClientBuilder clientBuilder, JavaFile javaFile) {
...
                             serviceClient.getClassName(),
                             serviceClient.getClassName(),
+                            getLocalBuildVariableName("pipeline"),
                             serializerExpression,
+                            getLocalBuildVariableName("defaultPollInterval"),
+                            getLocalBuildVariableName("environment"),
                             constructorArgs));
                 } else {
-                    function.line(String.format("%1$s client = new %2$s(pipeline, %3$s%4$s);",
+                    function.line(String.format("%1$s client = new %2$s(buildPipeline, %3$s%4$s);",
```
weidongxu-microsoft, line 295: there is still one "build"Pipeline

Issue #1531, opened 2022-06-22
Java code model emitter from CADL


srnagar:
Design and implement the Java emitter to generate code model from CADL. Currently, the code model is a yaml and Java's yaml support is not very flexible/efficient. Consider using json if this emitter is emitting Java specific code model for easy consumption by the AutoRest generator.

weidongxu-microsoft:
https://github.com/Azure/autorest.java/pull/1567

PR #1567, opened 2022-07-26
CADL extension

weidongxu-microsoft:
https://github.com/Azure/autorest.java/wiki/CADL-to-Java#detail

1. add emitter for CADL in `cadl-extension`, it outputs m4 with extended data (e.g. namespace, usage for Choice, usage=convenienceMethod in operation and models)
2. add simple wrapper for preprocessor and javagen in `cadl-extension/src`
3. update related code-model in Java
4. add `ImplementationDetails` class in client-model, for `ClientMethod`, `ClientModel`, and `EnumType`
5. add `mapPredefinedModel` for model that to be provided by other libs (for now, just for azure-core)
6. add one test in `cadl-tests` of https://github.com/Azure/autorest.testserver/blob/dpg_cadl/cadl/dpg-update1.cadl
7. add CI for JAR, NodeJS, and the testserver test above https://dev.azure.com/azure-sdk/internal/_build/results?buildId=1737838&view=results
8. add local cadl for tests at https://github.com/Azure/autorest.java/tree/cadl-code-model/cadl-tests/cadl

There is still some other cadl files in `cadl-tests`. They might still be valuable before we have comprehensive tests, hence still keep them.

Fix https://github.com/Azure/autorest.java/issues/1531


cadl-extension/src/code-model-builder.ts
```diff
@@ -0,0 +1,1103 @@
...
+      throw Error("Can not emit yaml for a namespace that doesn't exist.");
+    }
+
+    // java namespace
+    const namespace = getServiceNamespaceString(this.program)|| "Azure.Client";
+    const javaNamespace = getJavaNamespace(namespace);
+
+    // service version
+    this.version = getServiceVersion(this.program);
+    if (this.version === "0000-00-00") {
```
haolingdong-msft, line 119: When will gerServiceVersion() return "0000-00-00"?

weidongxu-microsoft: when no @ServiceVersion

weidongxu-microsoft:
@alzimmermsft I will wait for your PR.

cadl-extension/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.36</version>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.29.1</version>
```
alzimmermsft, line 63: nit: Should be 1.30.0

I also wonder if this dependency should be moved into the parent POM so that it doesn't need to be updated in multiple locations when a new version of azure-core is released. This will also still allow for the child projects to override this to an unreleased beta version if needed.

cadl-extension/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+      <groupId>com.google.googlejavaformat</groupId>
+      <artifactId>google-java-format</artifactId>
+      <version>1.0.0-beta.1</version>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.13.2</version>
+      <scope>test</scope>
+    </dependency>
```
alzimmermsft, line 75: We should start aiming to use JUnit 5 for all testing. This doesn't need to be changed

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
```diff
@@ -0,0 +1,165 @@
...
+        public MockConnection() {
+            super(null, null);
+        }
+    }
+
+    public CadlPlugin(String namespace) {
+        super(new MockConnection(), "dummy", "dummy");
+        SETTINGS_MAP.put("namespace", namespace);
+        JavaSettingsAccessor.setHost(this);
+        JavaSettings.getInstance();
```
alzimmermsft, line 130: Is this line doing anything? It's a getter with the return being used

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
```diff
@@ -0,0 +1,165 @@
...
+
+        // Response
+        client.getResponseModels().stream()
+                .filter(CadlPlugin::isGeneratingModel)
+                .forEach(model -> javaPackage.addClientResponse(model.getPackage(), model.getName(), model));
+    }
+
+    @Override
+    public void writeFile(String fileName, String content, List<Object> sourceMap) {
+        new File(fileName).getParentFile().mkdirs();
```
alzimmermsft, line 59: Should this first check if the directory exists before attempting to create the directories

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
```diff
@@ -0,0 +1,165 @@
...
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(fileName))) {
+            writer.write(content);
```
alzimmermsft, line 61: Since this is writing the file in on go I don't think a BufferedWriter is necessary

cadl-extension/src/main/java/com/azure/cadl/Main.java
```diff
@@ -0,0 +1,140 @@
...
+        for (JavaFile javaFile : javaPackage.getJavaFiles()) {
+            String content = javaFile.getContents().toString();
+            try {
+                content = formatter.formatSourceAndFixImports(content);
+            } catch (Exception e) {
+                LOGGER.error("Failed to format file: {}", outputFolder + javaFile.getFilePath(), e);
+//                continue;
+            }
+            cadlPlugin.writeFile(outputFolder + javaFile.getFilePath(), content, null);
+        }
```
alzimmermsft, line 94: I'd recommend copying the logic used in Javagen or Postprocessor that speeds up formatting and writing the generated files

cadl-extension/src/main/java/com/azure/cadl/Main.java
```diff
@@ -0,0 +1,140 @@
...
+            try (InputStreamReader reader = new InputStreamReader(fis, StandardCharsets.UTF_8)) {
+                int charsRead;
+                while ((charsRead = reader.read(buffer, 0, buffer.length)) > 0) {
+                    sb.append(buffer, 0, charsRead);
+                }
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            return sb.toString();
+        }
```
alzimmermsft, line 138: I'd recommend transitioning this over to `new String(Files.readAllBytes(Paths.get(path)))`;

javagen/src/main/java/com/azure/autorest/model/clientmodel/ImplementationDetails.java
```diff
@@ -0,0 +1,102 @@
...
+// Licensed under the MIT License.
+
+package com.azure.autorest.model.clientmodel;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class ImplementationDetails {
```
alzimmermsft, line 11: Mind adding javadocs to this class explaining what functions it serves

weidongxu-microsoft: Sure! thanks for remind.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
weidongxu-microsoft: I will try the change according to suggestions on read/write.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
weidongxu-microsoft: Likely no. Will try remove it.

cadl-extension/pom.xml
weidongxu-microsoft: Currently we rely on `eng/versioning/sync_versions.py` script to sync version with SDK repo.

cadl-extension/pom.xml
weidongxu-microsoft: Sure to change to 5. Copied from javagen, but javagen should also use 5 actually (just a few UT so it should be easy to upgrade to 5). Will do it in a different PR.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
weidongxu-microsoft: Agree.

cadl-extension/src/main/java/com/azure/cadl/Main.java
weidongxu-microsoft: Agree. Got it in your PR but copied this from old Javagen.

PR #1571, opened 2022-07-27
Remove Unused Templates and Always Add Getter/Setter Overrides

alzimmermsft:
Removes unused `Template` classes `ManagerTemplate` and `PageTemplate` and removes the configuration setting for having subclasses override their getter and setter methods to always be overridden.


weidongxu-microsoft (approved):
LGTM

Issue #1525, opened 2022-06-22
Generate models containing properties with built-in types


srnagar:
- [x] boolean
- [x] string
- [x] byte
- [x] int (should this always be generated as long to be resilient?)
- [x] long
- [x] double
- [x] duration
- [x] datetime
- [x] array (mainly just for byte array)
- [x] list
- [x] map

weidongxu-microsoft:
all types of int is generated as long.
all types of float is generated as double.
these is not format of bytes as Base64URL, so it is generated as ByteArray.

weidongxu-microsoft:
as change to the ModelSpecs, we might want int instead (.NET distinguish int and long)

weidongxu-microsoft:
CADL do not have a char, I assume we will use int for int8 or int 16.

int32 to int
int64 to long

uint treated same as int.

weidongxu-microsoft:
float32 for now is still double.

PR #1573, opened 2022-07-28
CADL test on builtin

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/1525


cadl-tests/src/main/java/com/cadl/builtin/BuiltinAsyncClient.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * The read operation.
+     *
+     * <p><strong>Response Body Schema</strong>
+     *
+     * <pre>{@code
+     * {
+     *     booleanProperty: boolean (Required)
```
weidongxu-microsoft, line 43: bug https://github.com/Azure/autorest.java/issues/1574

PR #1575, opened 2022-07-28
add patch template

haolingdong-msft:
Add patch template. It does not have getters and constructors. It does not support flatten client.

example output patch model:
```java
// Copyright (c) Microsoft Corporation. All rights reserved.
// Code generated by Microsoft (R) AutoRest Code Generator.

package fixtures.mergepatchjson.models;

import com.azure.core.implementation.Option;
import com.fasterxml.jackson.annotation.JsonProperty;
import java.time.OffsetDateTime;
import java.util.List;
import java.util.Map;

/** Deployment metadata. */
@Fluent
public final class DeploymentPatch {
    /*
     * The deployment identifier.
     */
    @JsonProperty(value = "deploymentId", required = true)
    private Option<String> deploymentId;

    /*
     * The deployment start datetime.
     */
    @JsonProperty(value = "startDateTime", required = true)
    private Option<OffsetDateTime> startDateTime;

    /*
     * The group identity
     */
    @JsonProperty(value = "groupId", required = true)
    private Option<String> groupId;

    /*
     * The device class subgroups for the deployment.
     */
    @JsonProperty(value = "deviceClassSubgroups")
    private Option<List<String>> deviceClassSubgroups;

    /*
     * Boolean flag indicating whether the deployment was canceled.
     */
    @JsonProperty(value = "isCanceled")
    private Option<Boolean> isCanceled;

    /*
     * Boolean flag indicating whether the deployment has been retried.
     */
    @JsonProperty(value = "isRetried")
    private Option<Boolean> isRetried;

    /*
     * Operation status filter.
     */
    @JsonProperty(value = "OperationFilterStatus", required = true)
    private Option<OperationFilterStatus> operationFilterStatus;

    /*
     * Dictionary of <string>
     */
    @JsonProperty(value = "tags")
    private Option<Map<String, String>> tags;

    /*
     * The foo property.
     */
    @JsonProperty(value = "foo")
    private Option<FooPatch> foo;

    /**
     * Set the deploymentId property: The deployment identifier.
     *
     * @param deploymentId the deploymentId value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setDeploymentId(Option<String> deploymentId) {
        this.deploymentId = deploymentId;
        return this;
    }

    /**
     * Set the startDateTime property: The deployment start datetime.
     *
     * @param startDateTime the startDateTime value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setStartDateTime(Option<OffsetDateTime> startDateTime) {
        this.startDateTime = startDateTime;
        return this;
    }

    /**
     * Set the groupId property: The group identity.
     *
     * @param groupId the groupId value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setGroupId(Option<String> groupId) {
        this.groupId = groupId;
        return this;
    }

    /**
     * Set the deviceClassSubgroups property: The device class subgroups for the deployment.
     *
     * @param deviceClassSubgroups the deviceClassSubgroups value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setDeviceClassSubgroups(Option<List<String>> deviceClassSubgroups) {
        this.deviceClassSubgroups = deviceClassSubgroups;
        return this;
    }

    /**
     * Set the isCanceled property: Boolean flag indicating whether the deployment was canceled.
     *
     * @param isCanceled the isCanceled value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setIsCanceled(Option<Boolean> isCanceled) {
        this.isCanceled = isCanceled;
        return this;
    }

    /**
     * Set the isRetried property: Boolean flag indicating whether the deployment has been retried.
     *
     * @param isRetried the isRetried value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setIsRetried(Option<Boolean> isRetried) {
        this.isRetried = isRetried;
        return this;
    }

    /**
     * Set the operationFilterStatus property: Operation status filter.
     *
     * @param operationFilterStatus the operationFilterStatus value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setOperationFilterStatus(Option<OperationFilterStatus> operationFilterStatus) {
        this.operationFilterStatus = operationFilterStatus;
        return this;
    }

    /**
     * Set the tags property: Dictionary of &lt;string&gt;.
     *
     * @param tags the tags value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setTags(Option<Map<String, String>> tags) {
        this.tags = tags;
        return this;
    }

    /**
     * Set the foo property: The foo property.
     *
     * @param foo the foo value to set.
     * @return the DeploymentPatch object itself.
     */
    public DeploymentPatch setFoo(Option<FooPatch> foo) {
        this.foo = foo;
        return this;
    }
}


```


javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+
+                if (property.isAdditionalProperties()) {
+                    classBlock.annotation("JsonAnySetter");
+                    MapType mapType = (MapType) property.getClientType();
+                    classBlock.packagePrivateMethod(String.format("void %s(String key, %s value)", property.getSetterName(), mapType.getValueType()), (methodBlock) -> {
+                        methodBlock.ifBlock(String.format("%s == null", property.getName()), ifBlock ->
+                                ifBlock.line("%s = new HashMap<>();", property.getName()));
+                        methodBlock.line("%s.put(%s, value);", property.getName(), model.getNeedsFlatten() ? "key.replace(\"\\\\.\", \".\")" : "key");
+                    });
+                }
```
weidongxu-microsoft, line 192: We might ignore the `additionalProperties` for now. You will need proper test if you enable this.

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+//            // being serialized twice, once for the class level annotation and again for the property annotation.
+//            if (property.isPolymorphicDiscriminator()) {
+//                classBlock.annotation("JsonTypeId");
+//            }
+
+            if (property.getHeaderCollectionPrefix() != null && !property.getHeaderCollectionPrefix().isEmpty()) {
+                classBlock.annotation("HeaderCollection(\"" + property.getHeaderCollectionPrefix() + "\")");
+            } else if (property.getAnnotationArguments() != null && !property.getAnnotationArguments().isEmpty()) {
+                classBlock.annotation(String.format("JsonProperty(%1$s)", property.getAnnotationArguments()));
+            }
```
weidongxu-microsoft, line 226: Any test or sample generated code on this?

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+
+            if (!property.isAdditionalProperties() && property.getClientType() instanceof MapType && settings.isFluent()) {
+                classBlock.annotation("JsonInclude(value = JsonInclude.Include.NON_NULL, content = JsonInclude.Include.ALWAYS)");
+            }
```
weidongxu-microsoft, line 230: Not sure how this works with `Option`?

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+                    classBlock.method(methodVisibility, null, String.format("%s %s(%s %s)",
+                                    patchClassNameWithBaseType, property.getSetterName(), setterPropertyType, property.getName()),
+                            (methodBlock) -> {
+                                String expression;
+                                if (propertyClientType.equals(ArrayType.ByteArray)) {
+                                    expression = String.format("CoreUtils.clone(%s)", property.getName());
+                                } else {
+                                    expression = property.getName();
+                                }
+                                methodBlock.ifBlock(String.format("%s == null", property.getName()),
```
weidongxu-microsoft, line 174: check this with primitives like int / long (as required property)

weidongxu-microsoft (commented):
I don't think required property should be used as it is in PATCH.

If the model is only for PATCH, none of its properties should be required.

If the model is shared with PUT and PATCH, its required properties should already by set during PUT, and PATCH should treat them all as optional. (there is `x-ms-mutability` but we can check it later)

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
haolingdong-msft: If we have the annotation and set the map as null, like this: `fooPath.setMyMap(null)`, it will be serialized as this: `"body":"{\"map\":null}"`

If we have the annotation and set the map with key and value, it will be serialized as this: "body":"{\"map\":{\"a\":\"124\"}}"

weidongxu-microsoft: This `JsonInclude.Include.NON_NULL` is actually for case like `"map": {"a": "avalue", "b", null"}`

haolingdong-msft: It will be serialized like this: "body":"{\"myMap\":{\"a\":\"avalue\",\"b\":null}}". The annotation works.

haolingdong-msft: Without this annotation, it will be serialized like this: 
`"body":"{\"myMap\":{\"a\":\"124\"}}"`
So seems it's the annotation works

weidongxu-microsoft: Great, thanks for the confirmation.

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
haolingdong-msft: `@HeaderCollection` annotation is used for response header, it does not relate to body patch. https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-header-collection-prefix 
I will delete this `HeaderCollection` part.

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+//        if (settings.getClientFlattenAnnotationTarget() == JavaSettings.ClientFlattenAnnotationTarget.TYPE && model.getNeedsFlatten()) {
+//            javaFile.annotation("JsonFlatten");
+//        }
+
+        ArrayList<JavaModifier> classModifiers = new ArrayList<>();
+        if (!hasDerivedModels && !model.getNeedsFlatten()) {
+            if (!settings.isFluent() || !model.getName().endsWith("Identity")) {    // bug https://github.com/Azure/azure-sdk-for-java/issues/8372
+                classModifiers.add(JavaModifier.Final);
+            }
+        }
```
weidongxu-microsoft, line 136: you don't need this.

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+            // If the import isn't used it will be removed later on.
+            imports.add(Base64.class.getName());
+            imports.add(HashMap.class.getName());
+            imports.add(HttpHeader.class.getName());
+
+            // JacksonAdapter will be removed in the future once model types are converted to using stream-style
+            // serialization. For now, it's needed to handle the rare scenario where the strong type is a non-Java
+            // base type.
+            imports.add(JacksonAdapter.class.getName());
+        }
```
weidongxu-microsoft, line 93: This likely only for response.

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+
+        // If there is client side validation and the model will generate a ClientLogger to log the validation
+        // exceptions add an import of 'com.azure.core.util.logging.ClientLogger' and
+        // 'com.fasterxml.jackson.annotation.JsonIgnore'.
+        //
+        // These are added to support adding the ClientLogger and then to JsonIgnore the ClientLogger so it isn't
+        // included in serialization.
+        if (settings.shouldClientSideValidations() && settings.shouldClientLogger()) {
+            ClassType.ClientLogger.addImportsTo(imports, false);
+        }
```
weidongxu-microsoft, line 70: remove if you don't have validate method.

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+        if (model.getParentModelName() != null) {
+            classNameWithBaseType += String.format(" extends %1$s", model.getParentModelName());
+        }
+        final String patchClassNameWithBaseType = classNameWithBaseType + "Patch";
+
+        if (model.getProperties().stream().anyMatch(p -> !p.getIsReadOnly())
+                || propertyReferences.stream().anyMatch(p -> !p.getIsReadOnly())) {
+            javaFile.annotation("Fluent");
+        } else {
+            javaFile.annotation("Immutable");
```
weidongxu-microsoft, line 148: I don't think you will ever have a Patch model to be `Immutable` (if all its property is read-only, then what's the point of PATCH?).

javagen/src/main/java/com/azure/autorest/template/PatchModelTemplate.java
```diff
@@ -0,0 +1,290 @@
...
+                classBlock.annotation(String.format("JsonProperty(%1$s)", property.getAnnotationArguments()));
+            }
+
+            if (!property.isAdditionalProperties() && property.getClientType() instanceof MapType && settings.isFluent()) {
+                classBlock.annotation("JsonInclude(value = JsonInclude.Include.NON_NULL, content = JsonInclude.Include.ALWAYS)");
+            }
+
+            // handle x-ms-client-default
+            if (property.getDefaultValue() != null) {
+                classBlock.privateMemberVariable(String.format("%1$s %2$s = Option.of(%3$s)", new GenericType("com.azure.core.implementation", "Option", property.getWireType()), property.getName(), property.getDefaultValue()));
```
weidongxu-microsoft, line 234: Why should we have default value for a PATCH? The default should be always `null`?

haolingdong-msft: Agree, default value should not be used in patch model, I will remove this.

haolingdong-msft:
For primitive type with required, currently I plan to handle patch model setter like this, do you think this is expected?
Foo.java:
```java
@Fluent
public final class Foo {
    /*
     * The foo property.
     */
    @JsonProperty(value = "foo", required = true)
    private int foo;
}
```
FooPatch.java -> setter will take primitive type `int`, and the property will use boxed type `Integer` inside `Option`, like this: `Option<Integer>`
```java
@Fluent
public final class FooPatch {
    /*
     * The foo property.
     */
    @JsonProperty(value = "foo", required = true)
    private Option<Integer> foo;

    /**
     * Set the foo property: The foo property.
     *
     * @param foo the foo value to set.
     * @return the FooPatch object itself.
     */
    public FooPatch setFoo(int foo) {
        this.foo = Option.of(foo);
        return this;
    }

```

weidongxu-microsoft (approved):
We might first check it in, since we had good work here.

If we later decided that `OptionModule` is never needed, we can clean it up.

weidongxu-microsoft (approved):

PR #1587, opened 2022-07-29
Lro support relative url using polling strategy

haolingdong-msft:


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -955,7 +955,8 @@ private String getPollingStrategy(ClientMethod clientMethod, String contextParam
             .replace("{context}", contextParam)
             .replace("{serializerAdapter}", clientMethod.getClientReference() + ".getSerializerAdapter()")
             .replace("{intermediate-type}", clientMethod.getMethodPollingDetails().getIntermediateType().toString())
-            .replace("{final-type}", clientMethod.getMethodPollingDetails().getFinalType().toString());
+            .replace("{final-type}", clientMethod.getMethodPollingDetails().getFinalType().toString())
+            .replace("{endpoint}", "\"https://\"+" + clientMethod.getClientReference() + ".getEndpoint()");
```
weidongxu-microsoft, line 959: no, this is not correct. the resolution of the `{endpoint}` need to follow the `x-ms-parameterized-host`. deviceupdate takes this value only because its `x-ms-parameterized-host` is configured as such https://github.com/Azure/azure-rest-api-specs/blob/main/specification/deviceupdate/data-plane/Microsoft.DeviceUpdate/preview/2022-07-01-preview/deviceupdate.json#L9-L10

PR #1589, opened 2022-07-29
Make Custom Strongly-Type Header Deserialization and ResponseBase Defaults

alzimmermsft:
Removes Autorest configurations `custom-strongly-typed-header-deserialization` and `generic-response-type` as the functionality is now the default without ability to configure. Moving forward all strongly-typed header classes will generate a constructor that takes `HttpHeaders` and will handle deserializing headers into their POJO types and all service responses that return strongly-typed headers will use `ResponseBase` with generics (`ResponseBase<Headers, Body>`) instead of creating a subclass that extends `ResponseBase` (`CustomResponse extends ResponseBase<Headers, Body>`). With the standardization of using `ResponseBase` directly there was also a code cleanup to remove properties, getters and setters, and utility methods that handled processing the custom subclasses.


weidongxu-microsoft:
How does it affect management and DPG? For them the models are directly served to customer (unlike vanilla which we likely have another model to wrap the response).

I think strongly-typed header deserialization is fine (we might need to have some tests in mgmt), but ResponseBase seems going to break customer?

alzimmermsft:
> How does it affect management and DPG? For them the models are directly served to customer (unlike vanilla which we likely have another model to wrap the response).
> 
> I think "strongly-typed header deserialization" is fine (we might need to have some tests in mgmt), but "ResponseBase" seems going to break customer (for mgmt, there is lots of GAed libs)? If it breaks, mgmt would like to keep the old behavior.
> 
> PS: one error in fluent CI is fixed in #1590, merged

Good catch, I'm going to close this PR and open a new one with only the change to make strongly-typed header deserialization the default without configuration. Moving forward, should new management and DPG libraries be using `true` for ResponseBase?

weidongxu-microsoft:
We can let DPG use `true` by default, as no one have model for it yet.
We can set `generic-response-type` to `true` in vanilla and DPG readme.md, this should make them default as true, unless disabled in swagger/readme.md

@srnagar for opinion, as this would affect the shape of the convenience method.

```
Mono<ResponseBase<LROsCustomHeadersPutAsyncRetrySucceededHeaders, Product>>
```
vs.
```
Mono<LROsCustomHeadersPutAsyncRetrySucceededResponse>
```
plus `LROsCustomHeadersPutAsyncRetrySucceededResponse.java`
```
public final class LROsCustomHeadersPutAsyncRetrySucceededResponse
        extends ResponseBase<LROsCustomHeadersPutAsyncRetrySucceededHeaders, Product> {
```

---

Is there other benefit (other than performance, and maybe consistency of data-plane and mgmt-plane), for the switch to `ResponseBase`?
Lots of mgmt calls takes minutes as long-running, waiting for backend to provision the Azure resource. Hence performance is likely not an issue for mgmt.
Other than long-running. The usual GET resource API likely not having the headers in response (as they would conform to ARM). Other APIs say `getKey` likely not be called very often.

PR #1595, opened 2022-08-02
dpg, fix getEndpoint from client

weidongxu-microsoft:
At present, this is not enabled.

1. Add `baseUrl` to `ServiceClient`
2. Add `requestParameterName` to `ServiceClientProperty`, similar to `ProxyMethodParameter`
3. `ENDPOINT_TRAIT` will reflect the default value / serialized name of this endpoint parameter from swagger/CADL. Hence the change to `ClientBuilderTrait.getEndpointTrait(ServiceClientProperty)`

Endpoint might like this, for swagger
```json
  "x-ms-parameterized-host": {
    "hostTemplate": "{Endpoint}",
    "useSchemePrefix": true,
    "parameters": [
      {
        "$ref": "#/parameters/EndpointParameter"
      }
    ]
  },
```

```java
    public String getEndpoint() {
        return "https://{Endpoint}".replace("{Endpoint}", this.client.getEndpoint());
    }
```

---

For this, we would have (it is a bit ambiguous whether we should return the baseUrl or the host as `getEndpoint()`)
```json
  "x-ms-parameterized-host": {
    "hostTemplate": "{endpoint}/metricsadvisor/v1.0",
    "useSchemePrefix": false,
    "parameters": [
      {
        "$ref": "#/parameters/Endpoint"
      }
    ]
  }
```

```java
    public String getEndpoint() {
        return "{endpoint}/metricsadvisor/v1.0".replace("{endpoint}", this.serviceClient.getEndpoint());
    }
```


weidongxu-microsoft:
@srnagar 

You might want to review this since I've modified code around `ClientBuilderTrait.ENDPOINT_TRAIT`, to make it take the default value and serialized name from Parameter.

default value is used as usual.
serialized name is used to generate code like this "https://{Endpoint}".replace("{Endpoint}", this.client.getEndpoint())`, as swagger can have `useSchemePrefix=true/false` to control whether the `endpoint` provided by user contains the scheme.

PR #1597, opened 2022-08-02
cadl, support content-type

weidongxu-microsoft:
Handle `content-type` header from CADL, and update m4 output (request.protocol.http.mediaTypes) accordingly.


cadl-tests/cadl/patch.cadl
```diff
@@ -0,0 +1,25 @@
...
+  description: string;
+}
+
+@route("/patch")
+interface Patch {
+  @put
+  op create(@path name: string, @body body: Resource): OkResponse<Resource> | ErrorResponse;
+
+  @patch
+  op update(@header("content-type") contentType: "application/merge-patch+json", @path name: string, @body body: Resource): OkResponse<Resource> | ErrorResponse;
```
weidongxu-microsoft, line 24: @haolingdong-msft `content-type` in CADL can be done like this.

If multiple, it would be `"application/merge-patch+json" | "application/json"` (probably not want to handle this at present)

PR #1605, opened 2022-08-03
cadl, emitter, filter out unwanted templated model from polymorphic

weidongxu-microsoft:
This likely not the final solution. But at present, avoid getting too deep in the template thing.


cadl-tests/cadl/polymorphism.cadl
```diff
@@ -21,6 +22,26 @@ model Dog extends Pet {
...
+  kind: TDiscriminator;
+  name: string;
+  parameters?: TParameters;
+}
+model Parameter {
+}
+model EntitiesTaskParameters extends Parameter {
+  entity?: string;
+}
+model EntityTask is TemplatedTask<"EntityRecognition", EntitiesTaskParameters>;
```
weidongxu-microsoft, line 39: We don't want `TemplatedTask` to show up in the class.

PR #1607, opened 2022-08-04
DPG, refactor createClientMethods to take isProtocolMethod

weidongxu-microsoft:
PR is only refactor, for future work (so that we can create ClientMethod for DPG as both protocol method and convenience method in one pass), logic should be identical.

We had too much dependency on the `JavaSettings.isDataPlaneClient()`, should try to confine its usage, as basically it is just a global static variable.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
@@ -148,7 +148,7 @@ public class ClientMethodMapper implements IMapper<Operation, List<ClientMethod>
                 syncReturnType = createPagedSyncReturnType(elementType);
             }
         } else {
-            asyncRestResponseReturnType = null;
+            asyncRestResponseReturnType = proxyMethodsMap.values().iterator().next().iterator().next().getReturnType().getClientType();
             IType responseBodyType = SchemaUtil.getOperationResponseType(operation, settings);
             if (isSimpleMethod) {
                 if (responseBodyType instanceof ClassType || responseBodyType instanceof ListType || responseBodyType instanceof MapType) {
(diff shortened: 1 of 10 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -124,19 +124,18 @@ public Map<Request, List<ProxyMethod>> map(Operation operation) {
...
+                builder.returnType(createSingleValueAsyncReturnType(clientResponseClassType));
             } else {
-                clientResponseClassType = ClientMapper.getClientResponseClassType(operation, settings);
+                ClassType clientResponseClassType = ClientMapper.getClientResponseClassType(operation, settings);
+                builder.returnType(createAsyncResponseReturnType(clientResponseClassType));
             }
-            builder.returnType(GenericType.Mono(clientResponseClassType));
         } else {
             IType singleValueType;
-            if (responseBodyType.equals(GenericType.FluxByteBuffer)
```
weidongxu-microsoft, line 138: responseBodyType is a sync class, it should never be `Flux`.

weidongxu-microsoft: refactored into `getAsyncRestResponseReturnType` method

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -718,13 +725,13 @@ protected IType createSyncReturnWithResponseType(IType syncReturnType, Operation
...
+                syncReturnWithResponse);
     }
 
-    protected ReturnValue createSimpleAsyncRestResponseReturnValue(Operation operation, ProxyMethod proxyMethod, IType syncReturnType) {
-        return new ReturnValue(returnTypeDescription(operation, proxyMethod.getReturnType().getClientType(), syncReturnType),
-                proxyMethod.getReturnType().getClientType());
+    protected ReturnValue createSimpleAsyncRestResponseReturnValue(Operation operation, IType asyncRestResponseReturnType, IType syncReturnType) {
+        return new ReturnValue(returnTypeDescription(operation, asyncRestResponseReturnType, syncReturnType),
+                asyncRestResponseReturnType);
     }
```
weidongxu-microsoft, line 735: It should be equivalent, as `asyncRestResponseReturnType` variable comes from `ProxyMethodMapper.getAsyncRestResponseReturnType`.

PR #1608, opened 2022-08-04
Better Handling for TypeReference

alzimmermsft:
Updates how `TypeReference` is handled in polling operations to use `TypeReference.createInstance(Class<T>)` when the intermediate or final response type is an array, simple class, enum, or primitive and to generate a static class of `TypeReference<GenericType>` when the intermediate or final response type is a generic type (`Map<String, String>`, `List<Class<T>>`, etc). This change prevents creating numerous `TypeReference` instances for the same generic type and for simple instances to use a singleton that is managed by the `TypeReference` class itself.


javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
```diff
@@ -124,24 +133,25 @@ public static void writeClientMethodsAndHelpers(JavaClass classBlock, List<Clien
...
-            javaBlock.line("new TypeReference%s(), new TypeReference%s());",
-                    clientMethod.getMethodPollingDetails().getIntermediateType(),
-                    clientMethod.getMethodPollingDetails().getFinalType());
-        } else {
-            javaBlock.line("new TypeReference<%s>() {\n// empty\n}, new TypeReference<%s>() {\n// empty\n});",
-                    clientMethod.getMethodPollingDetails().getIntermediateType(),
-                    clientMethod.getMethodPollingDetails().getFinalType());
-        }
+        MethodPollingDetails details = clientMethod.getMethodPollingDetails();
+        javaBlock.line("%s, %s);", getTypeReferenceCreation(details.getIntermediateType()),
```
weidongxu-microsoft, line 137: `%s, %s);` here seems not right. (this line might not be used anywhere in test)

weidongxu-microsoft (approved):
LGTM. The "createInstance" make life a lot easier.

One common on a possible bug.

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
alzimmermsft: This is following what it was using before, being either

`javaBlock.line("new TypeReference%s(), new TypeReference%s());`

or

`javaBlock.line("new TypeReference<%s>() {\n// empty\n}, new TypeReference<%s>() {\n// empty\n});"`

but I agree that this is definitely a spot where a lot of bugs could crop up as it's making an assumption that this is always going to be inserting the last two parameters in an API call.

weidongxu-microsoft: OK, this might be old problem of mine. I will try improve it (make the "writeLongRunningOperationTypeReference" return a good expression)

Issue #1576, opened 2022-07-28
local Cadl.Optional


weidongxu-microsoft:
ref https://github.com/Azure/autorest.java/pull/1573

Cadl.Optional for
https://github.com/Azure/autorest.java/issues/1526
https://github.com/Azure/autorest.java/issues/1524

for both operation parameter and model property

PR #1611, opened 2022-08-05
Cadl add optional tests

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/1576


cadl-tests/generate.sh
```diff
@@ -9,10 +9,10 @@ function generate_remote {
         exit 1
     fi
 
-    cp -rf cadl-output/java/src/ .
```
XiaofeiCao, line 12: On mac, the extra slash will result in copy everthing under `src` but itself. 

Tested on both mac and windows.

cadl-tests/cadl/optional.cadl
```diff
@@ -0,0 +1,51 @@
...
+
+using Cadl.Http;
+
+@serviceTitle("Optional")
+@serviceVersion("1.0.0")
+namespace Cadl.Optional;
+
+model Optional {
+  boolean?: boolean = true;
+  booleanNullable?: boolean;
```
weidongxu-microsoft, line 11: this is not nullable? it is only optional?

weidongxu-microsoft: recommend to do it on int and string (as one primitive and one class).

weidongxu-microsoft: Where is the model class?

cadl-tests/src/main/java/com/cadl/optional/models/WithOptional.java
```diff
@@ -0,0 +1,381 @@
...
+     * The booleanNullable property.
+     */
+    @JsonProperty(value = "booleanNullable")
+    private Boolean booleanNullable;
+
+    /*
+     * The booleanRequired property.
+     */
+    @JsonProperty(value = "booleanRequired")
+    private Boolean booleanRequired;
```
weidongxu-microsoft, line 34: what does `withOptionalProperties` do? why `booleanRequired` here?

XiaofeiCao: It makes all the properties in a model optional: https://github.com/microsoft/cadl/blob/main/docs/tutorial.md#withoptionalproperties

`WithOptional` has the spread properties of `Optional` and `booleanRequired` is defined in `Optional` as a required property. The goal here is to test whether `withOptionalProperties` works as intended.

weidongxu-microsoft: I see.

These wasn't there in tutorial last time...

XiaofeiCao: Yes.. And it has different semantics from `withVisibility` that haoling did test on, which acts like as a filter to filter out properties that don't conform with the visibility specified.

Renamed it to `AllPropertiesOptional` to avoid confusion.

weidongxu-microsoft (approved):
LGTM.

@haolingdong-msft please also take a look.

haolingdong-msft (approved):

Issue #1537, opened 2022-06-24
DPG, generate convenience method for simple API


weidongxu-microsoft:
First step might be simple method, no pageable, no LRO.

weidongxu-microsoft:
Options:
1. Not generate additional ProxyMethod for API with models, and generate a wrapper convenience method in `AsyncSyncClient` to serialize/de-serialize models and optional parameters to the DPG method. This could be hard, and might have performance issue.
2. Generate additional ProxyMethod as we do in vanilla (so we had 2 sets of ProxyMethod, DPG and vanilla), the higher level would be same as vanilla. This should be easy, but it might have impact on telemetry, as ProxyMethod for DPG and vanilla are essentially same.

---

Given CADL versioning, we should be able to generate all overload that matches the history of the API. This might complicate above discussion.
If V1 and V2 have 2 different set of required/optional parameters, should we further generate 2 sets of ProxyMethod for V2?

---

Do we consider `BarOptions` if there is lots of numbers of parameters?

srnagar:
For option 1, the model can easily be converted to BinaryData. We can just use `BinaryData.fromObject()` and `BinaryData.toObject()` to map the model to BinaryData which can then be used to call the protocol method. So, not sure if it would be harder to write but I'd expect the grow-up methods that devs write also to use this same approach. 

Option 2 is acceptable too. This also seems like relatively less work on the generator and we should rather spend resources on CADL integration and go with option 2 now. Since the proxy methods are internal, we can make changes later if required.

Regarding `BarOptions`, yes, we should create an options bag when we have a large number of params.

weidongxu-microsoft:
First we will have some refactor to `ClientMethod` https://github.com/Azure/autorest.java/pull/1607

Plan on next step, I will prototype an additional mapping of client method to convenience method (still as class `ClientMethod` for now) in `AsyncSyncClient`.
E.g. `BinaryData getFoo(RequestOptions)` to `Foo getFoo(expand, Context)`
Then in template, fill the code for conversion of method call.

At present, I will only target simple method (no pageable, no LRO).

weidongxu-microsoft:
Question: do we provide the overloaded method of "required parameter only"?

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/1670

PR #1612, opened 2022-08-05
CADL, convenience method

weidongxu-microsoft:
task https://github.com/Azure/autorest.java/issues/1537

Use `GENERATE_CONVENIENCE_METHODS` env variable to enable it for tests. Will switch to CADL side-car when available.

---

sample

```java
    /**
     * The create operation.
     *
     * @param alertConfigName Unique identifier of an alert configuration. This parameter is case-sensitive.
     * @param body Criteria that determine which anomalies should trigger an alert and via which notification
     *     channel(s). This setting can later be applied to a live inference schedule.
     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
     * @return criteria that determine which anomalies should trigger an alert and via which notification channel(s) on
     *     successful completion of {@link Mono}.
     */
    public Mono<AlertConfig> create(String alertConfigName, AlertConfig body) {
        RequestOptions requestOptions = new RequestOptions();
        return createWithResponse(alertConfigName, BinaryData.fromObject(body), requestOptions)
                .map(Response::getValue)
                .map(r -> r.toObject(AlertConfig.class));
    }

    /**
     * The read operation.
     *
     * @param alertConfigName Unique identifier of an alert configuration. This parameter is case-sensitive.
     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
     * @return criteria that determine which anomalies should trigger an alert and via which notification channel(s) on
     *     successful completion of {@link Mono}.
     */
    public Mono<AlertConfig> read(String alertConfigName) {
        RequestOptions requestOptions = new RequestOptions();
        return readWithResponse(alertConfigName, requestOptions)
                .map(Response::getValue)
                .map(r -> r.toObject(AlertConfig.class));
    }

    /**
     * The list operation.
     *
     * @param skip The number of records to skip from the list of records based on the sorting field and ordering method
     *     specified. By default, records are ranked by descending created time (UTC).
     * @param maxpagesize The maximum number of records to be returned per page. If more records are requested via the
     *     API, @nextLink will contain the link to the next page.
     * @param sortBy The name of the field on which you want to sort records. By default, records are sorted by created
     *     time (UTC).
     * @param orderBy Determines whether the records will be returned in descending or ascending order. By default,
     *     records are ranked in descending order.
     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
     * @return the response body on successful completion of {@link Mono}.
     */
    public Mono<PageAlertConfig> list(Long skip, Long maxpagesize, String sortBy, OrderBy orderBy) {
        RequestOptions requestOptions = new RequestOptions();
        requestOptions.addQueryParam("skip", String.valueOf(skip));
        requestOptions.addQueryParam("maxpagesize", String.valueOf(maxpagesize));
        requestOptions.addQueryParam("sortBy", sortBy);
        requestOptions.addQueryParam("orderBy", orderBy.toString());
        return listWithResponse(requestOptions).map(Response::getValue).map(r -> r.toObject(PageAlertConfig.class));
    }
```


cadl-tests/generate.sh
```diff
@@ -33,6 +33,8 @@ function generate {
 rm -rf src/main/
 rm -rf cadl-output/
 
+export GENERATE_CONVENIENCE_METHODS=true
```
weidongxu-microsoft, line 36: using it to let tests to always generate convenience methods.

javagen/src/main/java/com/azure/autorest/template/ConvenienceAsyncMethodTemplate.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    protected boolean isConvenienceMethod(ClientMethod method) {
+        return isMethodAsync(method) && isMethodVisible(method);
+    }
+
+    @Override
+    protected boolean isConvenienceMethod(ConvenienceMethod method) {
+        return isMethodAsync(method.getClientMethod()) && isMethodVisible(method.getClientMethod())
+                && method.getClientMethod().getMethodParameters().stream().noneMatch(p -> p.getClientType() == ClassType.Context);
```
haolingdong-msft, line 34: Why do we add `Context` for async client method, but not add for sync client method?

weidongxu-microsoft: Async have `Context` in mono/flux.

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -0,0 +1,281 @@
...
+                });
+    }
+
+    /**
+     * Whether the convenience method should be included.
+     *
+     * @param method the convenience method.
+     * @return Whether include the convenience method.
+     */
+    protected abstract boolean isConvenienceMethod(ClientMethod method);
```
XiaofeiCao, line 144: nit, `shouldInclude` would be clearer?

weidongxu-microsoft: changed to `isMethodIncluded`.

haolingdong-msft (approved):

Issue #1623, opened 2022-08-12
CADL, support authentication


weidongxu-microsoft:

weidongxu-microsoft:
https://github.com/Azure/autorest.java/wiki/CADL-to-Java#authentication

PR #1626, opened 2022-08-12
Cadl, support ServiceAuthentication

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1623

for now only support OAuth2 and ApiKey (same as m4)


cadl-tests/cadl/auth.cadl
```diff
@@ -0,0 +1,22 @@
...
+using Cadl.Http;
+
+@serviceTitle("Auth")
+@serviceVersion("1.0.0")
+@useAuth(OAuth2Auth<[AuthFlow]> | ApiKeyAuth<ApiKeyLocation.header, "x-ms-api-key">)
+namespace Cadl.Auth;
+
+model AuthFlow {
+    type: OAuth2FlowType.clientCredentials;
+    tokenUrl: "https://api.example.com/oauth2/token";
```
haolingdong-msft, line 12: are `tokenUrl` and `refreshUrl` used in generated code?

weidongxu-microsoft: No. They are done in azure-identity. These value in CADL won't take effect.

haolingdong-msft (approved):

PR #1628, opened 2022-08-15
cadl, support array in query

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1624

Currently only do as if it is "CSV".
Not aware whether CADL can specify other format or explode.

Support this as guideline has `expand` as `string array`, which means it will very likely appear in data-plane.
https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#query-options


cadl-tests/src/main/java/com/cadl/enumservice/EnumServiceAsyncClient.java
```diff
@@ -318,4 +479,209 @@ public Mono<Operation> getOperation(OperationStateValues state) {
...
+            requestOptions.addQueryParam(
+                    "priorityArrayOpt",
+                    priorityArrayOpt.stream()
+                            .map(Priority::toLong)
+                            .map(String::valueOf)
+                            .collect(Collectors.joining(",")));
+        }
+        return setIntEnumArrayWithResponse(
+                        priorityArray.stream().map(Priority::toLong).map(String::valueOf).collect(Collectors.toList()),
+                        requestOptions)
```
weidongxu-microsoft, line 562: this does not work if there is `null` in list. Need to use something else.

PR #1639, opened 2022-08-19
Cadl, bug fix, ServiceMethod and Generated annotation on convenience method

weidongxu-microsoft:


cadl-tests/src/main/java/cadl/testserver/servicedriven2/DPGClient.java
```diff
@@ -167,8 +167,8 @@ public Response<BinaryData> putRequiredOptionalWithResponse(String requiredParam
     @Generated
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Response<BinaryData> postParametersWithResponse(
-            String contenttype, BinaryData parameter, RequestOptions requestOptions) {
-        return this.client.postParametersWithResponse(contenttype, parameter, requestOptions).block();
+            String contentTypePath, BinaryData parameter, RequestOptions requestOptions) {
+        return this.client.postParametersWithResponse(contentTypePath, parameter, requestOptions).block();
```
weidongxu-microsoft, line 171: seems cadl-ranch updated?

XiaofeiCao: Yes, it's discussed [here](https://github.com/Azure/cadl-ranch/pull/50#discussion_r948583808).

weidongxu-microsoft: ok, if they've released new version, we should be able to enable the models in cadl-ranch.

commit added.

PR #1651, opened 2022-08-29
cadl, load "java.json" and read namespace (temporary solution)

weidongxu-microsoft:
temporary solution (before side-car is ready)

java.json as
```json
{
  "namespace": "com.azure.storage"
}
```

gets to code-model.yaml
```yaml
...
configuration:
  namespace: com.azure.storage
```

### DIFF
--- a/cadl-extension/src/emitter.ts
+++ b/cadl-extension/src/emitter.ts
@@ -2,17 +2,31 @@ import { resolvePath, getNormalizedAbsolutePath, Program } from "@cadl-lang/comp
 import { dump } from "js-yaml";
 import { promisify } from "util";
 import { execFile } from "child_process";
-import { mkdir } from "fs";
+import { promises } from "fs";
 import { CodeModelBuilder } from "./code-model-builder.js";
 
 export async function $onEmit(program: Program) {
   const builder = new CodeModelBuilder(program);
   const codeModel = builder.build();
 
+  const tempConfigurationFile = resolvePath("./java.json");
+  let tempConfigurationFileAccessible = true;
+  await promises.access(tempConfigurationFile).catch((_) => {
+    tempConfigurationFileAccessible = false;
+  });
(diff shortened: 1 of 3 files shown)


extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/CodeModel.java
```diff
@@ -145,4 +145,13 @@ public TestModel getTestModel() {
...
     }
+
+    // TODO (weidxu): side-car
+    private Object configuration;
+    public Object getConfiguration() {
+        return configuration;
+    }
+    public void setConfiguration(Object configuration) {
+        this.configuration = configuration;
+    }
```
weidongxu-microsoft, line 156: We probably not want it to say long (one or two months best). So just make it an `Object` and prepare to get rid of it when side-car comes.

cadl-extension/src/emitter.ts
```diff
@@ -2,27 +2,46 @@ import { resolvePath, getNormalizedAbsolutePath, Program } from "@cadl-lang/comp
...
   });
 
   await program.host.writeFile(codeModelFileName, dump(codeModel));
 
+  program.logger.info(`Code model file written to ${codeModelFileName}`);
+
+  const jarFile = "node_modules/@azure-tools/cadl-java/target/azure-cadl-extension-jar-with-dependencies.jar";
+  program.logger.info(`Exec JAR ${jarFile}`);
+
   const output = await promisify(execFile)("java", [
```
haolingdong-msft, line 42: why some places use `promises`, some places use `promisify`?

weidongxu-microsoft: `fs.promises` is the native promise support in `fs`, the built-in lib.

`promisify` is not built-in lib, but AFAIK, `child_process` does not have something equvalent to `fs.promises`. Let me know if this is not the case.

PR #1652, opened 2022-08-29
cadl, support LRO

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1660

---

likely temporary solution, as cadl-azure-core likely continue evolve and break.

targeting `pollingLocation` decorator
https://github.com/Azure/cadl-azure/blob/main/packages/cadl-azure-core/lib/foundations.cadl#L59

---

We likely need to enhance the LRO from Cadl (for different type in intermidate and final), but might not be this PR.


cadl-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
```diff
@@ -0,0 +1,184 @@
...
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @return the {@link SyncPoller} for polling of long-running operation.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
+    public SyncPoller<BinaryData, BinaryData> beginCreateOrUpdate(String name, RequestOptions requestOptions) {
+        return this.client.beginCreateOrUpdate(name, requestOptions).getSyncPoller();
+    }
```
weidongxu-microsoft, line 63: @haolingdong-msft @XiaofeiCao 

When you check the deviceupdate LRO in create, does intermediate type and final type same?

cadl-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
```diff
@@ -0,0 +1,184 @@
...
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @return the {@link SyncPoller} for polling of status monitor resource for long running operations.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
+    public SyncPoller<BinaryData, BinaryData> beginDelete(String name, RequestOptions requestOptions) {
+        return this.client.beginDelete(name, requestOptions).getSyncPoller();
+    }
```
weidongxu-microsoft, line 117: This likely not correct.

DELETE usually have final type as `Void`.
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/keyvault/azure-security-keyvault-keys/src/main/java/com/azure/security/keyvault/keys/KeyClient.java#L850-L853

cadl-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
XiaofeiCao: From the provided session-record for `testImportUpdates`, I believe they are the same.

weidongxu-microsoft: Emm, OK, thanks. "import" might be different from CRUD (similar to the "importx" and "export" in authoring, as action).

cadl-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
weidongxu-microsoft: I will take the DPG protocol method as it is. We seems always generate `SyncPoller<BinaryData, BinaryData>` for any LRO.

cadl-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
weidongxu-microsoft: For DPG it is simple, as both are `BInaryData`.

It would get complicated when we generate convenience method.

PR #1653, opened 2022-08-29
dpg, add partial update test in cadl-tests

haolingdong-msft:
code change:
1. add PARTIAL_UPDATE global environment variable as configuration.
2. add partial update test.

generate.sh code logic:
1. copy `partialupdate` source code folder to `existingcode` folder
3. run partial update test with `--outputPath` as `existingcode` folder
4. copy generated partial update code in `existingcode` folder to `src` folder


weidongxu-microsoft:
where is your manual code?

cadl-tests/generate.sh
```diff
@@ -15,19 +15,29 @@ function generate {
     rm -rf cadl-output
 }
 
+mkdir existingcode
+cp -rf ./src existingcode
```
weidongxu-microsoft, line 19: just copy src/main/java/com/cadl/partialupdate?

haolingdong-msft: I found it's easier to copy all the code to maintain the `src/main/java/` folder structure, but can `mkdir` the folders and then copy as well.

cadl-tests/src/main/java/com/cadl/partialupdate/PartialUpdateClientBuilder.java
```diff
@@ -0,0 +1,255 @@
...
+     * Builds an instance of PartialUpdateClient class.
+     *
+     * @return an instance of PartialUpdateClient.
+     */
+    @Generated
+    public PartialUpdateClient buildClient() {
+        return new PartialUpdateClient(new PartialUpdateAsyncClient(buildInnerClient().getPartialUpdateOps()));
+    }
+
+    public void test() {}
```
haolingdong-msft, line 254: manual code here. I did not add complex cases, because they are covered in partial update unit tests previously.

weidongxu-microsoft: any one in client?

cadl-tests/generate.sh
weidongxu-microsoft: Please focus on the source file that is required in your test case, and not involve others that not related in your test case.

cadl-tests/src/main/java/com/cadl/partialupdate/PartialUpdateClient.java
```diff
@@ -0,0 +1,83 @@
...
+     * @return the response.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public PartialUpdateModel read() {
+        RequestOptions requestOptions = new RequestOptions();
+        return readWithResponse(requestOptions).getValue().toObject(PartialUpdateModel.class);
+    }
+    /*
+     * Generated convenience method for readWithResponse
```
haolingdong-msft, line 81: There is a bug regarding to this `Generated convenience method` comment place after enabling partial update. I am checking.

haolingdong-msft: The root cause is JavaParser can only treat below javadoc comment as part of the method delaration, but not the additional comment `Generated convenience method for readWithResponse`. So this additional comment is not treated as part of the convenience method, the place of this comment is not together with the convenient method. I  can create an issue to fix this.
```java
/**
     * The read operation.
     *
     * <p><strong>Response Body Schema</strong>
     *
     * <pre>{@code
     * {
     *     boolean: boolean (Required)
     *     string: String (Required)
     *     bytes: byte[] (Required)
     * }
     * }</pre>
     *
     * @return the response body along with {@link Response}.
     */
```

weidongxu-microsoft: depends on how difficult the fix.

we can alternatively move this `Generated convenience method for readWithResponse` comment to inside the method (not before it).

haolingdong-msft: I also think moving this `Generated convenience method for readWithResponse` comment inside the javadoc is easier for now.

weidongxu-microsoft: No, "inside the method", not "inside the javadoc".

Javadoc exposes to customer. Comment inside method does not.

We want it stay as only a comment, not anything user can see.

cadl-tests/src/main/java/com/basicpolymorphicmodels/BasicPolymorphicModelsAsyncClient.java
```diff
@@ -107,11 +107,8 @@ public Mono<Response<BinaryData>> setValueWithPolymorphicPropertyWithResponse(
         return this.serviceClient.setValueWithPolymorphicPropertyWithResponseAsync(input, requestOptions);
     }
 
-    /*
-     * Generated convenience method for setValueWithResponse
-     */
     /**
-     * The setValue operation.
+     * Generated convenience method for setValueWithResponse. The setValue operation.
```
haolingdong-msft, line 111: In code, we put a new line after 'Generated convenience method for setValueWithResponse.'. 
https://github.com/Azure/autorest.java/blob/f2e074a42ed435cbe58daaff719ce2085bbf2b6f/javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java#L49
But after formatting, `Generated convenience method for setValueWithResponse.` and `The setValue operation.` appear in the same line. Please let me know if there is any concern.

weidongxu-microsoft (approved):

Issue #1658, opened 2022-08-30
Cadl, support string with @format


weidongxu-microsoft:
reason https://github.com/Azure/cadl-azure/blob/main/packages/cadl-samples/data-plane/language/authoring/authoring.cadl#L57-L75

https://swagger.io/specification/#data-types

weidongxu-microsoft:
support
- byte
- binary
- date-time
- date-time-rfc1123
- password

password is same as string

PR #1661, opened 2022-08-30
Cadl, support format on Model or ModelProperty

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1658


cadl-tests/src/main/java/com/cadl/builtin/BuiltinClient.java
```diff
@@ -41,6 +41,13 @@ public final class BuiltinClient {
      *
      * <pre>{@code
      * {
+     *     formatString (Required): {
+     *         base64Encoded: byte[] (Required)
+     *         binary: byte[] (Required)
+     *         dateTime: OffsetDateTime (Required)
+     *         dateTimeRfc1123: OffsetDateTime (Required)
```
weidongxu-microsoft, line 48: @haolingdong-msft 

Would you add a bug for yourself?

type of `base64Encoded` and `dateTimeRfc1123` seems not correct.

haolingdong-msft: issue created: https://github.com/Azure/autorest.java/issues/1666

Issue #1671, opened 2022-09-01
Cadl, convenience method for pageable


weidongxu-microsoft:
depend: https://github.com/Azure/cadl-azure/issues/1839

epic: https://github.com/Azure/cadl-azure/issues/1943

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/1741

PR #1676, opened 2022-09-05
Cadl, convenience method for pageable

weidongxu-microsoft:
ref https://github.com/Azure/autorest.java/issues/1671

### DIFF
--- a/cadl-tests/src/main/java/com/cadl/paged/PagedAsyncClient.java
+++ b/cadl-tests/src/main/java/com/cadl/paged/PagedAsyncClient.java
@@ -16,6 +16,7 @@ import com.azure.core.http.rest.PagedFlux;
 import com.azure.core.http.rest.RequestOptions;
 import com.azure.core.util.BinaryData;
 import com.cadl.paged.implementation.PagedOpsImpl;
+import com.cadl.paged.models.Resource;
 
 /** Initializes a new instance of the asynchronous PagedClient type. */
 @ServiceClient(builder = PagedClientBuilder.class, isAsync = true)
@@ -61,5 +62,23 @@ public final class PagedAsyncClient {
     @ServiceMethod(returns = ReturnType.COLLECTION)
     public PagedFlux<BinaryData> list(RequestOptions requestOptions) {
         return this.serviceClient.listAsync(requestOptions);
+    }
+
+    /**
+     * List Resource resources.
+     *
+     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return paged collection of Resource items as paginated response with {@link PagedFlux}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.COLLECTION)
+    public PagedFlux<Resource> list() {
+        // Generated convenience method for list
+        RequestOptions requestOptions = new RequestOptions();
+        return list(requestOptions).mapPage(protocolMethodData -> protocolMethodData.toObject(Resource.class));
     }
 }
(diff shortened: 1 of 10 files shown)


cadl-tests/src/main/java/com/cadl/paged/PagedAsyncClient.java
```diff
@@ -62,4 +63,22 @@ public final class PagedAsyncClient {
...
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return paged collection of Resource items as paginated response with {@link PagedFlux}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.COLLECTION)
+    public PagedFlux<Resource> list() {
+        // Generated convenience method for list
+        RequestOptions requestOptions = new RequestOptions();
+        return list(requestOptions).mapPage(protocolMethodData -> protocolMethodData.toObject(Resource.class));
```
weidongxu-microsoft, line 82: `mapPage` is deprecated in core, we might need to switch to some util code later.

cadl-tests/src/main/java/com/cadl/paged/PagedClient.java
```diff
@@ -61,4 +62,21 @@ public final class PagedClient {
...
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return paged collection of Resource items as paginated response with {@link PagedIterable}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.COLLECTION)
+    public PagedIterable<Resource> list() {
+        // Generated convenience method for list
+        return new PagedIterable<>(client.list());
```
weidongxu-microsoft, line 80: At present, it is fine to just call async method and convert to sync (if we do not need `Context`).

However, it could change when we had dedicated call stack for sync method.

weidongxu-microsoft: Similar to current convenience method on simple method, `Context` overload is not generated.

haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -898,7 +898,9 @@ protected JavaVisibility methodVisibility(ClientMethodType methodType, boolean h
             } else {
                 // at present, only generate convenience method for simple API (no pageable, no LRO)
```
XiaofeiCao, line 899: Comment change?

weidongxu-microsoft: thx, updated

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -56,69 +56,83 @@ public final void write(ConvenienceMethod convenienceMethodObj, JavaClass classB
...
     }
 
+    /**
+     * Write the implementation of the convenience method.
+     *
+     * @param protocolMethod the protocol method.
+     * @param convenienceMethod the convenience method.
+     * @param methodBlock the code block.
+     */
+    protected void writeMethodImplementation(
```
weidongxu-microsoft, line 72: Modification on this class is refactor. To make subclass can override `writeMethodImplementation`.

XiaofeiCao (approved):

cadl-tests/src/main/java/com/cadl/paged/PagedClient.java
weidongxu-microsoft: Because it is pretty hard to convert `PagedIterable<BinaryData>` to `PagedIterable<Foo>`. The `mapPage` in core is buggy (actually `PagedFlux.mapPage` is also buggy, so we had another Utils class for these in mgmt https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/resourcemanager/azure-resourcemanager-resources/src/main/java/com/azure/resourcemanager/resources/fluentcore/utils/PagedConverter.java).

PR #1679, opened 2022-09-06
Updates to HttpHeaders Deserialization

alzimmermsft:
Changes how strongly-typed headers are deserialized with the following changes:

1. Use a local variable to store the fix get when the type is nullable. Retrieving a header may not be cheap depending on the underlying implementation, for example the `HttpHeaders` class in azure-core `toLowercase`'s the name, while individually this is fast is a lot of headers are being deserialized this takes a non-insignificant amount of CPU and it should be assumed that headers are being accessed in a thread-safe way.

Before:

```java
if (headers.getValue("value") != null) {
  this.value = Duration.parse(headers.getValue("value"));
}
```

After:

```java
String value = headers.getValue("value");
if (value != null) {
  this.value = Duration.parse(value);
}
```

In a case where the model access 10 headers and 5 were nullable this change increased performance from 1200 ns/op to 800 ns/op.

2. Header collections, such as metadata or object replication metadata in Storage Blobs, will be handled as a group. If there are a lot of header, which in case of Storage Blob there is 10s, this has a lot of validation being performed just to stream the headers and it is safe to bundle them in a single loop.

Before:

```java
Map<String, String> metadata = new HashMap<>();
for (HttpHeader header : headers) {
  if (header.getName().startsWith("x-ms-meta-")) {
    metadata.put(header.getName().substring(10), header.getValue());
  }
}
this.metadata = metadata;

Map<String, String> objectReplication = new HashMap<>();
for (HttpHeader header : headers) {
  if (header.getName().startsWith("x-ms-or-")) {
    objectReplication.put(header.getName().substring(8), header.getValue());
  }
}
this.objectReplication = objectReplication;
```

After:

```java
Map<String, String> metadata = new HashMap<>();
Map<String, String> objectReplication = new HashMap<>();
for (HttpHeader header : headers) {
  if (header.getName().startsWith("x-ms-meta-")) {
    metadata.put(header.getName().substring(10), header.getValue());
    continue;
  }
  
  if (header.getName().startsWith("x-ms-or-")) {
    objectReplication.put(header.getName().substring(8), header.getValue());
  }
}
this.metadata = metadata;
this.objectReplication = objectReplication;
```

Performance for this hasn't been validated but it will be better as it removes an iteration over the entire set of HTTP headers plus some extra if the first `startsWith` check passes and skips the second `startsWith` check from needing to run.


protocol-tests/src/main/java/fixtures/paging/implementation/PagingsImpl.java
```diff
@@ -860,7 +860,7 @@ Mono<Response<BinaryData>> getPagingModelWithItemNameWithXmsClientNameNext(
      * @return the response body along with {@link PagedResponse} on successful completion of {@link Mono}.
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
-    private Mono<PagedResponse<BinaryData>> getNoItemNamePagesSinglePageAsync(RequestOptions requestOptions) {
+    public Mono<PagedResponse<BinaryData>> getNoItemNamePagesSinglePageAsync(RequestOptions requestOptions) {
```
srnagar, line 863: Why is this generated as `public`?

vanilla-tests/src/main/java/fixtures/nonamedresponsetypes/models/HeadersResponseDurationHeaders.java
```diff
@@ -25,7 +25,8 @@ public final class HeadersResponseDurationHeaders {
      * @param rawHeaders The raw HttpHeaders that will be used to create the property values.
      */
     public HeadersResponseDurationHeaders(HttpHeaders rawHeaders) {
-        if (rawHeaders.getValue("value") != null) {
+        String value = rawHeaders.getValue("value");
+        if (value != null) {
```
srnagar, line 29: This may create unnecessary noise in client library code re-generation showing as diff. We should just keep the previous code as-is. Also, the local variable is not used in the line below, so, `rawHeaders.getValue()` is called still called twice.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -807,23 +809,33 @@ private static void addSetterMethod(IType propertyWireType, IType propertyClient
...
-                ifBlock -> ifBlock.line("continue;"));
-            body.line(collectionName + ".put(header.getName().substring(%d), header.getValue());", headerPrefixLength);
+            for (ClientModelProperty property : properties) {
+                body.ifBlock(String.format("header.getName().startsWith(\"%s\")", property.getHeaderCollectionPrefix()),
+                    ifBlock -> {
+                        ifBlock.line("%sHeaderCollection.put(header.getName().substring(%d), header.getValue());",
+                            property.getName(), property.getHeaderCollectionPrefix().length());
+                        ifBlock.line("break;");
+                    });
+            }
```
srnagar, line 831: Do we have an example/testcase of what was previously generated vs what is generated now? The generated code here is slightly different and I just wanted to check if the two are equivalent.

protocol-tests/src/main/java/fixtures/paging/implementation/PagingsImpl.java
alzimmermsft: I forgot to rebuild after pulling in main, will regenerate

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
alzimmermsft: There isn't a case in the generation Swagger but Storage Blobs has a case where there are multiple header collection in the same response.

srnagar: It may be good to add the before/after code snippets in the PR description. I can create an issue in testserver to add a test case for this.

srnagar (approved):
LGTM!

PR #1687, opened 2022-09-09
cadl-java, find the jar from relative path of the JS file

weidongxu-microsoft:
instead of from CWD

test change due to cadl-ranch

### DIFF
--- a/cadl-extension/src/emitter.ts
+++ b/cadl-extension/src/emitter.ts
@@ -4,8 +4,13 @@ import { promisify } from "util";
 import { execFile } from "child_process";
 import { promises } from "fs";
 import { CodeModelBuilder } from "./code-model-builder.js";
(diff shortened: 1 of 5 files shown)


cadl-extension/src/emitter.ts
```diff
@@ -4,8 +4,13 @@ import { promisify } from "util";
 import { execFile } from "child_process";
 import { promises } from "fs";
 import { CodeModelBuilder } from "./code-model-builder.js";
+import { dirname } from "path";
+import { fileURLToPath } from "url";
 
 export async function $onEmit(program: Program) {
+  const __dirname = dirname(fileURLToPath(import.meta.url));
```
weidongxu-microsoft, line 11: seems a trick in nodejs? https://github.com/nodejs/node/issues/28114

PR #1692, opened 2022-09-12
Update pom.xml

weidongxu-microsoft:
upgrade snakeyaml


alzimmermsft (approved):

weidongxu-microsoft:
seems cadl-ranch gets 0.35, which is good news (had to have every dependency on 0.35 before we can upgrade our cadl dependency).

ignore cadl CI, will increase cadl dependency in next PR.

haolingdong-msft (approved):

PR #1693, opened 2022-09-13
Cadl, upgrade compiler to 0.35

weidongxu-microsoft:


cadl-tests/cadl/lro.cadl
```diff
@@ -13,6 +14,7 @@ model Resource {
   id: string;
 
   @key
+  @segment("resources")
```
weidongxu-microsoft, line 17: new compiler will fail if we do not add `segment`

cadl-tests/src/main/java/com/cadl/enumservice/models/Priority.java
```diff
@@ -10,15 +10,15 @@
 /** Defines values for Priority. */
 public enum Priority {
     /** Enum value 100. */
-    HIGH(100L),
+    HIGH("100"),
 
     /** Enum value 0. */
-    LOW(0L);
+    LOW("0");
```
weidongxu-microsoft, line 16: investigating

weidongxu-microsoft: solved

cadl-extension/src/code-model-builder.ts
```diff
@@ -719,54 +719,44 @@ export class CodeModelBuilder {
     sealed: boolean,
   ): ChoiceSchema | SealedChoiceSchema | ConstantSchema {
     const namespace = getNamespace(type);
-    const isConstant = false; //sealed && type.members.length === 1;
```
weidongxu-microsoft, line 721: removed (as well as the branch code) that we don't use Cadl Enum for Java Constant

Issue #1666, opened 2022-08-31
dpg, properties with specific types not correct in javadoc


haolingdong-msft:
https://github.com/Azure/autorest.java/pull/1661#discussion_r958040086

weidongxu-microsoft:
It is not response body type. Just any property with such class.

PR #1694, opened 2022-09-13
dpg, fix properties with specific types not correct in javadoc

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1666


cadl-tests/.vscode/launch.json
```diff
@@ -12,6 +12,16 @@
                 "<node_internals>/**"
             ],
             "type": "node"
+        },
```
haolingdong-msft, line 15: This is for vs code debugging configuration. With this, you can just click the debug button to debug cadl codegen.

weidongxu-microsoft: so far it is fine.

we might see later if `"runtimeArgs": ["cadl", "compile", "builtin.cadl"],` will affect other usages.

weidongxu-microsoft (approved):

cadl-tests/.vscode/launch.json
haolingdong-msft: You can modify this `runtimeArgs` to debug other cadl files generation

weidongxu-microsoft: yes, but I don't want to either commit it every time it get modified locally, or take the pain to exclude it from commit.

we might later see if put it to ignore.

PR #1695, opened 2022-09-13
cadl, test case for api key

weidongxu-microsoft:
And a bug fix for `Response::getValue` when it returns `null`.

### DIFF
--- a/cadl-tests/src/test/java/com/authentication/apikey/AzureKeyCredentialPolicy.java
+++ b/cadl-tests/src/test/java/com/authentication/apikey/AzureKeyCredentialPolicy.java
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.authentication.apikey;
+
+import com.azure.core.credential.AzureKeyCredential;
+import com.azure.core.http.HttpPipelineCallContext;
+import com.azure.core.http.HttpPipelineNextPolicy;
+import com.azure.core.http.HttpPipelineNextSyncPolicy;
+import com.azure.core.http.HttpResponse;
+import com.azure.core.http.policy.HttpPipelinePolicy;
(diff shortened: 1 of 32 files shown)


javagen/src/main/java/com/azure/autorest/template/ConvenienceAsyncMethodTemplate.java
```diff
@@ -47,7 +47,9 @@ protected void writeInvocationAndConversion(
 
         String returnExpression = (methodType == ClientMethodType.PagingAsync)
                 ? "%1$s(%2$s)%3$s"
-                : "%1$s(%2$s).map(Response::getValue)%3$s";
+                : ((responseBodyType.asNullable() == ClassType.Void)
+                ? "%1$s(%2$s).then()%3$s"   // return type is Mono<Void>, Response::getValue would be null
+                : "%1$s(%2$s).map(Response::getValue)%3$s");
```
weidongxu-microsoft, line 52: Use `then()` when return type is `Mono<Void>` (and `Response.getValue() == null`)

cadl-tests/src/test/java/com/authentication/apikey/AzureKeyCredentialPolicy.java
```diff
@@ -0,0 +1,75 @@
...
+    private final String nameLowerCase;
+    private final AzureKeyCredential credential;
+
+    private final HttpPipelineSyncPolicy inner = new HttpPipelineSyncPolicy() {
+        @Override
+        protected void beforeSendingRequest(HttpPipelineCallContext context) {
+//            if ("http".equals(context.getHttpRequest().getUrl().getProtocol())) {
+//                throw LOGGER.logExceptionAsError(
+//                        new IllegalStateException("Key credentials require HTTPS to prevent leaking the key."));
+//            }
```
weidongxu-microsoft, line 38: maybe we have better way to disable this, but at present use the simplest solution.

PR #1696, opened 2022-09-13
cadl, treat property with @segment as readonly

weidongxu-microsoft:
the property with `@segment` is actually in path, hence we should not include it in body.

I didn't add it to `@key`, as from doc https://github.com/microsoft/cadl/blob/main/docs/tutorial.md#key I cannot see whether it must be read-only.

The next task is to remove read-only properties from ctor (as user should not set these when init object).


cadl-tests/src/main/java/com/cadl/response/models/Resource.java
```diff
@@ -20,9 +20,15 @@ public final class Resource {
     /*
      * The name property.
      */
-    @JsonProperty(value = "name", required = true)
+    @JsonProperty(value = "name", required = true, access = JsonProperty.Access.WRITE_ONLY)
```
weidongxu-microsoft, line 23: Since `name` is `@segment`, now they are all read-only.

PR #1697, opened 2022-09-13
cadl, model ctor without readonly properties

weidongxu-microsoft:
there should be no effect on swagger, as swagger cannot have required and read-only on same property.

however Cadl can have both on same property.


cadl-tests/cadl/response.cadl
```diff
@@ -16,6 +16,7 @@ model Resource {
   @key
   @segment("resources")
   name: string;
+  description?: string;
```
weidongxu-microsoft, line 19: add an optional for future testing of round-trip model.

cadl-tests/cadl/response.cadl
```diff
@@ -46,8 +47,16 @@ interface ResponseOp {
     @header("operation-location")
     operationLocation: ResourceLocation<Resource>;
   };
+}
+
+// operation using Azure.Core
+interface CadlCoreOp {
```
weidongxu-microsoft, line 53: separate the testing of cadl-azure-core related request/response in another interface.

haolingdong-msft (approved):

Issue #1707, opened 2022-09-16
  found: no arguments   reason: actual and formal argument lists differ in length


halkeye:
@halkeye

What's kind of SDK are you generating?

Please remove "--generate-tests" option. The automatically generated tests is still experimental feature (that we are testing in mgmt-plane).

PS: If you can paste the class of `AppsListRegionsResponse`, it will help me fix that test issue (but this is not related to the Headers issue here).

_Originally posted by @weidongxu-microsoft in https://github.com/Azure/autorest.java/issues/1706#issuecomment-1248842579_

halkeye:
> What's kind of SDK are you generating?

Digital ocean api. I'm afraid I don't know what the question means though.
the spec file is https://api-engineering.nyc3.digitaloceanspaces.com/spec-ci/DigitalOcean-public.v2.yaml

I stole the client_gen_config.md from work, so i'll see if i'm allowed to share publically it in a bug issue for the future. (i'll gladly email you the configs if it helps builds a more robust tool)

> The automatically generated tests is still experimental feature (that we are testing in mgmt-plane).

I can do that. I just wanted to find something to confirm things compiled right.

> PS: If you can paste the class of `AppsListRegionsResponse`, it will help me fix that test issue (but this is not related to the Headers issue here).

```
package com.digitalocean.api.models;

import com.azure.core.http.HttpHeaders;
import com.azure.core.http.HttpRequest;
import com.azure.core.http.rest.ResponseBase;

/** Contains all response data for the listRegions operation. */
public final class AppsListRegionsResponse extends ResponseBase<AppsListRegionsHeaders, Object> {
    /**
     * Creates an instance of AppsListRegionsResponse.
     *
     * @param request the request which resulted in this AppsListRegionsResponse.
     * @param statusCode the status code of the HTTP response.
     * @param rawHeaders the raw headers of the HTTP response.
     * @param value the deserialized value of the HTTP response.
     * @param headers the deserialized headers of the HTTP response.
     */
    public AppsListRegionsResponse(
            HttpRequest request, int statusCode, HttpHeaders rawHeaders, Object value, AppsListRegionsHeaders headers) {
        super(request, statusCode, rawHeaders, value, headers);
    }

    /**
     * Gets the deserialized response body.
     *
     * @return the deserialized response body.
     */
    @Override
    public Object getValue() {
        return super.getValue();
    }
}
```

weidongxu-microsoft:
@halkeye

[...]

PR #1709, opened 2022-09-16
Client, deduplicate class name, avoid Java keyword

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1707

Root cause is that the spec has "apps_list_regions_response", hence the model as "AppsListRegionsResponse".

Then there is client response "AppsListRegionsResponse" from "Apps" and "ListRegions".

In usual usage, we recommend `--generic-response-type=true` option.

---

Also fix another issue of conflict with reserved Java keyword ("private" in this case).

Escape name from keyword should happen after camel casing or pascal casing, as changing case would make non-keyword to keyword ("Private" to "private" in this case).


halkeye:
[INFO] Tests run: 1004, Failures: 0, Errors: 0, Skipped: 0

\o/

lgtm

XiaofeiCao (approved):

PR #1712, opened 2022-09-19
Cadl, support roundtrip model

weidongxu-microsoft:


cadl-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
```diff
@@ -40,7 +40,9 @@ public final class LongRunningClient {
      *
      * <pre>{@code
      * {
-     *     type: String (Optional)
+     *     id: String (Required)
+     *     name: String (Required)
+     *     type: String (Required)
```
weidongxu-microsoft, line 45: Javadoc might need some improvement.

id and name property should be read-only, hence not settable in request,

however fix might not be easy, as model had to be cached differently as request and response,

@haolingdong-msft

haolingdong-msft: Shall we remove read-only properties or having some wording indicating that it is read-only in request body?

cadl-tests/src/main/java/com/cadl/response/CadlCoreOpClient.java
```diff
@@ -165,10 +166,10 @@ public PagedIterable<BinaryData> list(RequestOptions requestOptions) {
      */
     @Generated
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public Resource createOrUpdate(String name, ResourceRequest updateableProperties) {
+    public Resource createOrUpdate(String name, Resource resource) {
```
weidongxu-microsoft, line 169: change

PR #1716, opened 2022-09-20
cadl, test case for oauth2

weidongxu-microsoft:


cadl-tests/src/test/java/com/authentication/oauth2/BearerTokenAuthenticationPolicy.java
```diff
@@ -0,0 +1,196 @@
...
+     */
+    public boolean authorizeRequestOnChallengeSync(HttpPipelineCallContext context, HttpResponse response) {
+        return false;
+    }
+
+    @Override
+    public Mono<HttpResponse> process(HttpPipelineCallContext context, HttpPipelineNextPolicy next) {
+//        if ("http".equals(context.getHttpRequest().getUrl().getProtocol())) {
+//            return Mono.error(new RuntimeException("token credentials require a URL using the HTTPS protocol scheme"));
+//        }
```
weidongxu-microsoft, line 99: copied from azure-core with these disabled.

PR #1720, opened 2022-09-20
StreamResponse and Miscellaneous Fixes

alzimmermsft:
A few fixes related to `StreamResponse` and other miscellaneous fixes.

- Recently `StreamResponse` APIs with strongly-typed headers began generating as `ResponseBase<HeaderType, InputStream>` where this should have been `Flux<ByteBuffer>` instead of `InputStream`. This has been fixed to generate as `ResponseBase<HeaderType, Flux<ByteBuffer>>`.
- Changed `StreamResponse::getValue` to `Response::getValue` as `Response::getValue` is more generic and will work for `StreamResponse` returns as well as `ResponseBase<HeaderType, Flux<ByteBuffer>>` returns.
- Prevent `PagedFlux` and `PagedIterable` `ClientMethod`s without `Context` being generated.
- Simplified the logic to determine if `FluxUtil` should be included as an import to just checking if the `ClientMethod` is asynchronous and doesn't have a `Context` parameter.
- Moved static inner class creation for XML wrappers to an earlier point in code generation to fix Autorest from commenting the static inner class instead of the class property.

### DIFF
--- a/fluentgen/src/main/java/com/azure/autorest/fluent/mapper/FluentClientMethodMapper.java
+++ b/fluentgen/src/main/java/com/azure/autorest/fluent/mapper/FluentClientMethodMapper.java
@@ -51,6 +51,11 @@ public class FluentClientMethodMapper extends ClientMethodMapper {
                     visibility = NOT_GENERATE;
                     break;
 
+                case PagingSync:
+                case PagingAsync:
+                    visibility = VISIBLE;
+                    break;
+
                 default:
                     visibility = super.methodVisibility(methodType, false, isProtocolMethod);
                     break;
--- a/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
@@ -841,6 +841,7 @@ public class ClientMethodMapper implements IMapper<Operation, List<ClientMethod>
         if (isProtocolMethod || !responseContainsHeaders) {
             return GenericType.Response(syncReturnType);
         } else if (settings.isGenericResponseTypes()) {
+
             return GenericType.RestResponse(Mappers.getSchemaMapper().map(ClientMapper.parseHeader(operation, settings)),
                 syncReturnType);
         } else {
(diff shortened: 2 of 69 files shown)


fluentgen/src/main/java/com/azure/autorest/fluent/mapper/FluentClientMethodMapper.java
```diff
@@ -51,6 +51,11 @@ protected JavaVisibility methodVisibility(ClientMethodType methodType, boolean h
                     visibility = NOT_GENERATE;
                     break;
 
+                case PagingSync:
+                case PagingAsync:
+                    visibility = VISIBLE;
+                    break;
+
```
weidongxu-microsoft, line 58: I will probably remove it in another PR (after test). Your change of logic in ClientMethodMapper in this PR is correct, and fluent should not generate pageable API with optional but not with Context (it was a bug in the "sync-stack" PR, found yesterday but didn't have time to adjust).

vanilla-tests/src/main/java/fixtures/bodyfile/Files.java
```diff
@@ -118,7 +118,7 @@ public Mono<StreamResponse> getFileWithResponseAsync(Context context) {
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Flux<ByteBuffer> getFileAsync() {
-        return getFileWithResponseAsync().flatMapMany(StreamResponse::getValue);
+        return getFileWithResponseAsync().flatMapMany(Response::getValue);
```
weidongxu-microsoft, line 121: not sure what code change causes this, but I assume they are equivalent?

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1013,6 +1015,12 @@ protected JavaVisibility methodVisibility(ClientMethodType methodType, boolean h
                     : NOT_GENERATE;
             }
         } else {
+            // Don't generate PagedFlux or PagedIterable methods without Context if Context methods are being generated.
+            if ((methodType == ClientMethodType.PagingSync || methodType == ClientMethodType.PagingAsync)
+                && !hasContextParameter && settings.isContextClientMethodParameter()) {
+                return NOT_GENERATE;
+            }
```
weidongxu-microsoft, line 1022: This actually is wrong. The overload of required only parameters still need to be generated.

alzimmermsft: Thanks for catching that and adding tests. `ClientMethodMapper` and a few other locations definitely need more unit testing before we begin separating out configurations in `javagen` to the other generators like `fluentgen`

Issue #1721, opened 2022-09-21
mgmt, problem with sync-stack


weidongxu-microsoft:
1.

overload on pageable is not correct.

---

2.

[INFO] -------------------------------------------------------------
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/autorest.java_ci/autorest.java_ci/azure-sdk-for-java/sdk/appservice/azure-resourcemanager-appservice-generated/src/main/java/com/azure/resourcemanager/appservice/generated/implementation/GlobalsClientImpl.java:[556,9] cannot find symbol
  symbol:   method getSubscriptionOperationWithResponseWithResponse(java.lang.String,java.lang.String,com.azure.core.util.Context)
  location: class com.azure.resourcemanager.appservice.generated.implementation.GlobalsClientImpl
[INFO] 1 error

PR #1723, opened 2022-09-21
mgmt, fix pageable overload

weidongxu-microsoft:
ref https://github.com/Azure/autorest.java/issues/1721


azure-tests/src/main/java/fixtures/paging/AutoRestPagingTestServiceAsyncClient.java
```diff
@@ -0,0 +1,410 @@
...
+     *
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return the paginated response with {@link PagedFlux}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.COLLECTION)
+    public PagedFlux<Product> getNullNextLinkNamePages() {
+        return this.serviceClient.getNullNextLinkNamePagesAsync();
+    }
```
weidongxu-microsoft, line 60: without change in this PR, if `--context-client-method-parameter`, this class would be empty.

javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
```diff
@@ -96,7 +95,6 @@ public static void getAsyncSyncClients(CodeModel codeModel, ServiceClient servic
                                         String methodName = cMethods.iterator().next().getProxyMethod().getName();
                                         methodGroupClient.getClientMethods().stream()
                                                 .filter(m -> methodName.equals(m.getProxyMethod().getName()))
-                                                .filter(m -> m.getType() != ClientMethodType.PagingAsyncSinglePage)
```
weidongxu-microsoft, line 99: cadl path should already be good

PR #1725, opened 2022-09-21
client, fix incorrect method name

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1721


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -801,11 +801,11 @@ private void generateSimpleSyncMethod(ClientMethod clientMethod, JavaType typeBl
             }
             if (clientMethod.getReturnValue().getType().equals(PrimitiveType.Void)) {
                 function.line("%s(%s);",
-                    clientMethod.getProxyMethod().getSimpleAsyncRestResponseMethodName().replace("Async", ""),
```
weidongxu-microsoft, line 804: replacement is wrong if method name contains "Async"

PR #1726, opened 2022-09-22
Client, enable test for sync

weidongxu-microsoft:


vanilla-tests/src/main/java/fixtures/streamstyleserialization/Arrays.java
```diff
@@ -203,8 +257,8 @@ public Response<Void> putValidWithResponse(ArrayWrapper complexBody) {
      * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public void putValid(ArrayWrapper complexBody) {
-        putValidWithResponse(complexBody);
+    public void putValidSync(ArrayWrapper complexBody) {
```
weidongxu-microsoft, line 260: @alzimmermsft 

Should this be `putValid` instead of `putValidSync`? Could it a bug because proxy method name is `putValidSync`?

alzimmermsft: I think so, this may be an interesting bug where the `ProxyMethod` for synchronous calls is appending `Sync`

PR #1729, opened 2022-09-22
client, fix async/sync overload when sync-stack=false

weidongxu-microsoft:
try to make it same as before, if sync-stack is false. (hope we can relax the configuration on vanilla after synapse gets to DPG)

still need more time to group the condition for the method overloads (the case of required and required+optional is complicated, especially when we had the `--required-parameter-client-methods` option).

### DIFF
--- a/azure-tests/src/main/java/fixtures/azurespecials/ApiVersionDefaults.java
+++ b/azure-tests/src/main/java/fixtures/azurespecials/ApiVersionDefaults.java
@@ -145,6 +145,20 @@ public final class ApiVersionDefaults {
      * @throws IllegalArgumentException thrown if parameters fail the validation.
      * @throws ErrorException thrown if the request is rejected by server.
      * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return method with api-version modeled in global settings on successful completion of {@link Mono}.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Mono<Void> getMethodGlobalValidAsync(Context context) {
+        return getMethodGlobalValidWithResponseAsync(context).flatMap(ignored -> Mono.empty());
+    }
+
+    /**
+     * GET method with api-version modeled in global settings.
+     *
+     * @param context The context to associate with this operation.
+     * @throws IllegalArgumentException thrown if parameters fail the validation.
+     * @throws ErrorException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
      * @return method with api-version modeled in global settings along with {@link Response}.
      */
     @ServiceMethod(returns = ReturnType.SINGLE)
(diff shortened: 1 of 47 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -770,15 +772,15 @@ private static void createSimpleClientMethods(JavaSettings settings, List<Client
 
         // Generate a non-WithResponse overload with non-required parameters only if one of fluent, include Context,
         // and only required parameters is false.
-        if (!settings.isFluent() || !settings.isContextClientMethodParameter() || !generateClientMethodWithOnlyRequiredParameters) {
+        if (!isSync || !settings.isFluent() || !settings.isContextClientMethodParameter() || !generateClientMethodWithOnlyRequiredParameters) {
```
weidongxu-microsoft, line 775: async still need this overload

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -750,7 +750,9 @@ private static void createSimpleClientMethods(JavaSettings settings, List<Client
             .methodVisibility(visibilityFunction.apply(true, false));
 
         // Always generate an overload of WithResponse with non-required parameters without Context.
-        methods.add(builder.build());
+        if (!isSync || settings.isSyncStackEnabled()) {
+            methods.add(builder.build());
+        }
```
weidongxu-microsoft, line 755: only for async if sync-stack=false

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -770,15 +772,15 @@ private static void createSimpleClientMethods(JavaSettings settings, List<Client
...
 
         if (generateClientMethodWithOnlyRequiredParameters) {
             methods.add(builder.onlyRequiredParameters(true).build());
         }
 
-        if (settings.isContextClientMethodParameter()) {
+        if (settings.isSyncStackEnabled() && settings.isContextClientMethodParameter()) {
             builder.methodVisibility(visibilityFunction.apply(false, true)).onlyRequiredParameters(false);
             addClientMethodWithContext(methods, builder, parameters, contextParameter);
         }
```
weidongxu-microsoft, line 786: not at all if sync-stack=false

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
weidongxu-microsoft: reverted.

note that it would have effect on inner clients of Fluent Premium.

async overload would be
```java
listKeysAsync(String resourceGroupName, String accountName)

listKeysWithResponseAsync(String resourceGroupName, String accountName, ListKeyExpand expand, Context context)
```

but no
```java
listKeysAsync(String resourceGroupName, String accountName, ListKeyExpand expand)
```

weidongxu-microsoft: this would result in some minor adjustment to handwritten code in premium, on next regen.

alzimmermsft (approved):

azure-tests/src/main/java/fixtures/azurespecials/ApiVersionDefaults.java
```diff
@@ -138,32 +138,6 @@ public Mono<Void> getMethodGlobalValidAsync() {
...
-     * GET method with api-version modeled in global settings.
-     *
-     * @param context The context to associate with this operation.
-     * @throws IllegalArgumentException thrown if parameters fail the validation.
-     * @throws ErrorException thrown if the request is rejected by server.
-     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
-     * @return method with api-version modeled in global settings on successful completion of {@link Mono}.
-     */
-    @ServiceMethod(returns = ReturnType.SINGLE)
-    public Mono<Void> getMethodGlobalValidAsync(Context context) {
```
weidongxu-microsoft, line 151: emm, I need to get this one back.

PR #1736, opened 2022-09-27
Client, refactor method overload

weidongxu-microsoft:


weidongxu-microsoft:
@haolingdong-msft Please review commit https://github.com/Azure/autorest.java/pull/1736/commits/18b7dc0b428b442933261206c2fc028672a31707

We need the merge done only once.

If we had convenience method, this `tryMergeBinaryRequests` will be called again in the convenience method follow. In this pass, it appears skip the 1st merged request and get to 2nd request.

weidongxu-microsoft:
Since I've notified the removal of "context-client-method-parameter" option (which would always be true), I might start remove it in following PR.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1013,18 +1028,25 @@ protected JavaVisibility methodVisibility(ClientMethodType methodType, boolean h
...
-//                        || (methodType == ClientMethodType.SimpleSyncRestResponse && hasContextParameter))
+                    // || (methodType == ClientMethodType.SimpleSyncRestResponse && hasContextParameter))
                     ? VISIBLE
                     : NOT_GENERATE;
             }
         } else {
             if (!settings.isSyncStackEnabled() && methodType == ClientMethodType.SimpleSyncRestResponse && settings.isContextClientMethodParameter() && !hasContextParameter) {
                 return NOT_GENERATE;
+            } else if (!settings.isSyncStackEnabled() && methodType == ClientMethodType.SimpleSync && settings.isContextClientMethodParameter() && hasContextParameter) {
+                return NOT_GENERATE;
```
weidongxu-microsoft, line 1039: sync-stack might need them, or might not, or might need them only as private

vanilla-tests/src/main/java/fixtures/lro/LROs.java
```diff
@@ -508,6 +508,30 @@ public PollerFlux<Product, Product> beginPut200SucceededAsync(Product product) {
...
+    /**
+     * Long running put request, service returns a 200 to the initial request, with an entity that contains
+     * ProvisioningState=’Succeeded’.
+     *
+     * @throws CloudErrorException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return the {@link PollerFlux} for polling of long-running operation.
+     */
+    @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
+    public PollerFlux<Product, Product> beginPut200SucceededAsync() {
```
weidongxu-microsoft, line 520: the minimum overload, as `project` is optional.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -284,22 +275,8 @@ private List<ClientMethod> createClientMethods(Operation operation, boolean isPr
...
-                        .fromClient(false)
-                        .annotations(Collections.emptyList())
-                        .build();
-                    parameters.add(requestOptions);
-                }
-
-                final boolean generateClientMethodWithOnlyRequiredParameters
-                    = settings.getRequiredParameterClientMethods() && hasNonRequiredParameters(parameters);
+                final MethodOverloadType defaultOverloadType = hasNonRequiredParameters(parameters) ? MethodOverloadType.OVERLOAD_MAXIMUM : MethodOverloadType.OVERLOAD_MINIMUM_MAXIMUM;
+                final boolean generateOnlyRequiredParameters = settings.getRequiredParameterClientMethods() && defaultOverloadType == MethodOverloadType.OVERLOAD_MAXIMUM;
```
XiaofeiCao, line 279: settings.isRequiredParameterClientMethods()?

weidongxu-microsoft: which variable do you mean?

weidongxu-microsoft: generateOnlyRequiredParameters only happen when setting sets that, AND that overload is needed (which is not if there is no optional parameters in API)

weidongxu-microsoft: OK, I will rename it. Thx

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -341,13 +318,19 @@ && shouldGeneratePagingMethods()) {
...
+
                     JavaVisibility simpleAsyncMethodVisibility =
-                        methodVisibility(ClientMethodType.SimpleAsyncRestResponse, false, isProtocolMethod);
+                        methodVisibility(ClientMethodType.SimpleAsyncRestResponse, defaultOverloadType, false, isProtocolMethod);
                     JavaVisibility simpleAsyncMethodVisibilityWithContext =
-                        methodVisibility(ClientMethodType.SimpleAsyncRestResponse, true, isProtocolMethod);
+                        methodVisibility(ClientMethodType.SimpleAsyncRestResponse, defaultOverloadType, true, isProtocolMethod);
+                    // for vanilla and fluent, the SimpleAsyncRestResponse is VISIBLE, so that they can be used for possible customization on LRO
+
+                    // there is ambiguity of RestResponse from simple API and from LRO API
```
XiaofeiCao, line 328: nit, maybe we can put this override into `methodVisibility()`? There's also `settings.isDataPlaneClient()` logic.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -341,13 +318,19 @@ && shouldGeneratePagingMethods()) {
...
+                    // e.g. SimpleAsyncRestResponse without Context in simple API should be VISIBLE
+                    // hence override here for DPG
```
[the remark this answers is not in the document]
weidongxu-microsoft: The problem is that type=SimpleAsyncRestResponse can be from a simple API, or a LRO API, we know it here.

But at `methodVisibility()`, it no longer know that type=SimpleAsyncRestResponse  is simple or LRO.

We might later to separate it as e.g. `LroSimpleAsyncRestResponse`. But before that, this seems easier to understand.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -803,12 +765,19 @@ private void addLroMethods(Operation operation, ClientMethod.Builder builder, Li
                 .onlyRequiredParameters(false)
```
XiaofeiCao, line 765: We are still keeping `LongRunningBeginAsync`(and sync) with maximum overload without `Context`? 

This seems to have 3 overloads?
1. Maximum without Context
2. Required with minimum
3. Maximum with Context

XiaofeiCao (approved):

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
weidongxu-microsoft: Yes, vanilla always provides all 3 overloads (similar to pageable). They are more developer facing (as usually vanilla Client is generated in implementation package), and dev would have max flexibility to choose the API they wish to call (as the implementation is not exposed to user).

We then trim down in fluent and dpg.

weidongxu-microsoft: PS: vanilla usually not actually enable LRO as it need to be configured https://github.com/Azure/autorest.java#settings-for-minimal-data-plane-clients

PR #1742, opened 2022-09-27
dpg, cadl, remove customized namespace and generate all cadl ranch tests

haolingdong-msft:


cadl-tests/package.json
```diff
@@ -14,7 +14,8 @@
     "@azure-tools/cadl-ranch-expect": "latest",
     "@azure-tools/cadl-ranch-specs": "latest",
     "@azure-tools/cadl-ranch-api": "latest",
-    "@azure-tools/cadl-java": "file:/../cadl-extension/azure-tools-cadl-java-0.1.0-dev.7.tgz"
+    "@azure-tools/cadl-java": "file:/../cadl-extension/azure-tools-cadl-java-0.1.0-dev.7.tgz",
+    "@cadl-lang/openapi": "latest"
```
haolingdong-msft, line 18: add this because cadl-ranch dev-driven test use it: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/resiliency/dev-driven/main.cadl#L2

cadl-tests/generate.sh
```diff
@@ -41,17 +41,9 @@ rm -rf ./existingcode
 # run cadl ranch tests sources
 cp -rf node_modules/@azure-tools/cadl-ranch-specs/http .
 
-export NAMESPACE=com.cadl.testserver.servicedriven1
-generate ./http/resiliency/srv-driven-1/main.cadl
-
-export NAMESPACE=com.cadl.testserver.servicedriven2
-generate ./http/resiliency/srv-driven-2/main.cadl
-
 unset NAMESPACE
```
weidongxu-microsoft, line 44: Do we have other place setting the env? If not, remove this as well.

cadl-tests/package.json
weidongxu-microsoft, line 18: Please group it to @cadl-lang, and use the ~version, not the latest. We only plan to upgrade it when we upgrade compiler.

weidongxu-microsoft (approved):

XiaofeiCao (approved):

cadl-tests/generate.sh
haolingdong-msft: no other places have NAMESPACE env var.

cadl-tests/package.json
haolingdong-msft: updated

PR #1744, opened 2022-09-27
Use PowerShell Scripts instead of Bash or Batch

alzimmermsft:
Use OS agnostic PowerShell scripts in CI instead of OS specific batch or bash scripts.


cadl-tests/Generate.ps1
```diff
@@ -0,0 +1,69 @@
...
+generate ./http/resiliency/srv-driven-1/main.cadl
+
+$env:NAMESPACE = "com.cadl.testserver.servicedriven2"
+generate ./http/resiliency/srv-driven-2/main.cadl
+
+$env:NAMESPACE = ''
+foreach ($cadlFile in (Get-ChildItem ./http -Filter "*.cadl" -File -Name -Recurse)) {
+    if ($cadlFile -like "*resiliency*") {
+        continue
+    }
```
weidongxu-microsoft, line 66: latest sh removed this.

cadl-tests/Generate.ps1
```diff
@@ -0,0 +1,69 @@
...
+# run cadl ranch tests sources
+Copy-Item -Path node_modules/@azure-tools/cadl-ranch-specs/http -Destination ./ -Recurse -Force
+
+$env:NAMESPACE = "com.cadl.testserver.servicedriven1"
+generate ./http/resiliency/srv-driven-1/main.cadl
+
+$env:NAMESPACE = "com.cadl.testserver.servicedriven2"
+generate ./http/resiliency/srv-driven-2/main.cadl
+
+$env:NAMESPACE = ''
```
weidongxu-microsoft, line 62: latest sh removed this.

weidongxu-microsoft (approved):
commented on the cause of CI failure as latest shell for Cadl changed a bit.

---

I am fine with it, though I am not familiar with Powershell language, so it might take me longer to write something new (copy and paste is no problem).

We can get it in and let people play with it, before delete the duplicate bat/sh.

Anyone else familiar with Powershell?

Issue #1734, opened 2022-09-27
Cadl LRO, emitter


weidongxu-microsoft:
Emitter to include e.g.

```yaml
operationLinks:
  polling: [operation1]
  final: [operation2]
```

---

likely need to copy and modify `Operation` (add above) from https://github.com/Azure/autorest m4, similar to `choice.ts`

PR #1746, opened 2022-09-28
Cadl, operationLinks for LRO

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1734

output
```yaml
        operationLinks:
          polling:
            operation: *ref_19
```


cadl-extension/src/common/operation.ts
```diff
@@ -0,0 +1,161 @@
...
+  constructor(operation: Operation) {
+    this.operation = operation;
+  }
+
+  /** the operation */
+  operation: Operation;
+
+  /** the parameters as Schema */
+  parameters?: Schema;
+}
```
weidongxu-microsoft, line 65: Added.

Rest is same from m4.

Issue #1735, opened 2022-09-27
Cadl LRO, javagen - depends on issues/1734


weidongxu-microsoft:
Parse yaml from https://github.com/Azure/autorest.java/issues/1734

Populate `MethodPollingDetails` for intermittent type and final type.

PR #1748, opened 2022-09-28
Cadl, LRO add types

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1735

---

intermediate type will be taken from response of pollingOperation.
final type will be taken from response of finalOperation. If no finalOperation (but has pollingOperation), it will be taken from LRO response (it is a guess).
DELETE will have final response as void, if not provided otherwise.


cadl-tests/src/main/java/com/cadl/longrunning/implementation/LongRunningOpsImpl.java
```diff
@@ -553,6 +643,71 @@ public SyncPoller<BinaryData, BinaryData> beginDelete(String name, RequestOption
...
+     * @param name The name parameter.
+     * @param requestOptions The options to configure the HTTP request before HTTP client sends it.
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @return the {@link PollerFlux} for polling of status monitor resource for long running operations.
+     */
+    @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
+    public PollerFlux<OperationStatusResourceResource, OperationStatus> beginDeleteWithModelAsync(
```
weidongxu-microsoft, line 668: Emm. Delete should be `Void`.

eng/pipelines/cadl-ci.yaml
```diff
@@ -136,4 +136,4 @@ jobs:
       - publish: $(Build.SourcesDirectory)/cadl-tests/cadl-ranch-coverage-java.json
         artifact: CoverageReport
         displayName: 'Publish Cadl Ranch Coverage Report'
-
+        continueOnError: true
```
weidongxu-microsoft, line 139: not know what's wrong with latest cadl-ranch.

continue if this fails.

Issue #1733, opened 2022-09-26
DPG, potential bug on Javadoc of pageable


weidongxu-microsoft:
https://github.com/Azure/autorest.java/blob/main/cadl-tests/src/main/java/com/cadl/paged/PagedClient.java#L42-L49

Javadoc is the JSON of response, not the JSON of `BinaryData` in `PagedIterable`, which appears not correct.

Sync with other SDK to see if we need to fix this.

PS: see `MethodPageDetails`.

PR #1752, opened 2022-09-30
dpg, display inner model schema for paging in javadoc

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1733

Main logic is in this commit: 51f4dd5a6979ae06c4ea34e16380ae973933c1a3

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
@@ -104,7 +104,7 @@ public abstract class ClientMethodTemplateBase implements IJavaTemplate<ClientMe
                             })
                             .findFirst().orElse(null);
                     if (responseBodyType == null) {
-                        throw new IllegalStateException(String.format("responseBodyType should not be null for paging method. rawResponseType = %s", rawResponseType.toString()));
+                        throw new IllegalStateException(String.format("%s not found in properties of rawResponseType. rawResponseType = ", itemName, rawResponseType.toString()));
                     }
                 } else {
                     responseBodyType = clientMethod.getProxyMethod().getRawResponseBodyType();
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
@@ -93,6 +93,9 @@ public abstract class ClientMethodTemplateBase implements IJavaTemplate<ClientMe
                             .map(clientModelProperty -> clientModelProperty.getClientType())
                             .map(valueListType -> {
                                 // value type is List<T>, we need to get the typeArguments
+                                if (!(valueListType instanceof ListType)) {
+                                    throw new IllegalArgumentException("value type must be list");
+                                }
                                 IType[] listTypeArgs = ((ListType) valueListType).getTypeArguments();
                                 if (listTypeArgs.length == 0) {
                                     throw new IllegalArgumentException("list type arguments' length for paging value should not be 0.");
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
@@ -83,7 +83,7 @@ public abstract class ClientMethodTemplateBase implements IJavaTemplate<ClientMe
                     // rawResponseType has properties: 'value' and 'nextLink'
                     IType rawResponseType = clientMethod.getProxyMethod().getRawResponseBodyType();
                     if (!(rawResponseType instanceof ClassType)) {
-                        throw new IllegalArgumentException("clientMethod.getProxyMethod().getRawResponseBodyType() should be ClassType for paging method.");
+                        throw new IllegalStateException(String.format("clientMethod.getProxyMethod().getRawResponseBodyType() should be ClassType for paging method. rawResponseType = %s", rawResponseType.toString()));
                     }
                     ClientModel model = ClientModelUtil.getClientModel(((ClassType) rawResponseType).getName());
                     List<ClientModelProperty> properties = new ArrayList<>();
@@ -94,15 +94,18 @@ public abstract class ClientMethodTemplateBase implements IJavaTemplate<ClientMe
                             .map(valueListType -> {
                                 // value type is List<T>, we need to get the typeArguments
                                 if (!(valueListType instanceof ListType)) {
-                                    throw new IllegalArgumentException("value type must be list");
+                                    throw new IllegalStateException(String.format("value type must be list for paging method. rawResponseType = %s", rawResponseType.toString()));
                                 }
                                 IType[] listTypeArgs = ((ListType) valueListType).getTypeArguments();
                                 if (listTypeArgs.length == 0) {
-                                    throw new IllegalArgumentException("list type arguments' length for paging value should not be 0.");
+                                    throw new IllegalStateException(String.format("list type arguments' length should not be 0 for paging method. rawResponseType = %s", rawResponseType.toString()));
                                 }
                                 return listTypeArgs[0];
                             })
                             .findFirst().orElse(null);
+                    if (responseBodyType == null) {
+                        throw new IllegalStateException(String.format("responseBodyType should not be null or void for paging method. rawResponseType = %s", rawResponseType.toString()));
+                    }
                 } else {
                     responseBodyType = clientMethod.getProxyMethod().getRawResponseBodyType();
                 }
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
@@ -9,6 +9,7 @@ import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientEnumValue;
 import com.azure.autorest.model.clientmodel.ClientMethod;
 import com.azure.autorest.model.clientmodel.ClientMethodParameter;
+import com.azure.autorest.model.clientmodel.ClientMethodType;
 import com.azure.autorest.model.clientmodel.ClientModel;
 import com.azure.autorest.model.clientmodel.ClientModelProperty;
 import com.azure.autorest.model.clientmodel.EnumType;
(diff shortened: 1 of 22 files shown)


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,30 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
             // Response body
             IType responseBodyType;
             if (JavaSettings.getInstance().isDataPlaneClient()) {
-                responseBodyType = clientMethod.getProxyMethod().getRawResponseBodyType();
+                // special handling for paging method
+                if (clientMethod.getType() == ClientMethodType.PagingSync || clientMethod.getType() == ClientMethodType.PagingAsync || clientMethod.getType() == ClientMethodType.PagingAsyncSinglePage || clientMethod.getType() == ClientMethodType.PagingSyncSinglePage) {
```
weidongxu-microsoft, line 81: no need to include latter 2 SinglePage, their return value is the response, and they are not exposed in DPG in client.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,30 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
+                    responseBodyType = properties.stream()
+                            .filter(property -> property.getName().equals(itemName))
+                            .map(clientModelProperty -> clientModelProperty.getClientType())
+                            .map(valueListType -> {
+                                // value type is List<T>, we need to get the typeArguments
+                                IType[] listTypeArgs = ((ListType) valueListType).getTypeArguments();
+                                if (listTypeArgs.length > 0) {
+                                   return listTypeArgs[0];
+                                } else {
+                                    return null;
```
weidongxu-microsoft, line 97: why return null?

If something should not happen, throw correct exception to fail the codegen.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,30 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
             // Response body
             IType responseBodyType;
             if (JavaSettings.getInstance().isDataPlaneClient()) {
-                responseBodyType = clientMethod.getProxyMethod().getRawResponseBodyType();
+                // special handling for paging method
+                if (clientMethod.getType() == ClientMethodType.PagingSync || clientMethod.getType() == ClientMethodType.PagingAsync || clientMethod.getType() == ClientMethodType.PagingAsyncSinglePage || clientMethod.getType() == ClientMethodType.PagingSyncSinglePage) {
+                    String itemName = clientMethod.getMethodPageDetails().getItemName();
+                    // rawResponseType has properties: 'value' and 'nextLink'
+                    IType rawResponseType = clientMethod.getProxyMethod().getRawResponseBodyType();
+                    ClientModel model = ClientModelUtil.getClientModel(((ClassType) rawResponseType).getName());
```
weidongxu-microsoft, line 85: test the class (do not cast without test), if not expected, throw exception to fail codegen.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,30 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
+                            .map(valueListType -> {
+                                // value type is List<T>, we need to get the typeArguments
+                                IType[] listTypeArgs = ((ListType) valueListType).getTypeArguments();
+                                if (listTypeArgs.length > 0) {
+                                   return listTypeArgs[0];
+                                } else {
+                                    return null;
+                                }
+                            })
+                            .findFirst().orElse(null);
```
weidongxu-microsoft, line 100: how would following code handle this `null`?

same as above, if not expected, fail it; if expected, handle it correctly.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
haolingdong-msft: If we remove the latter two SinglePage checks, the generated `impl` class will have different javadocs for single-page method and non single-page method. like below:
```java
/**
     * A paging operation that finishes on the first call without a nextlink.
     *
     * <p><strong>Response Body Schema</strong>
     *
     * <pre>{@code
     * {
     *     values (Optional): [
     *          (Optional){
     *             properties (Optional): {
     *                 id: Integer (Optional)
     *                 name: String (Optional)
     *             }
     *         }
     *     ]
     *     nextLink: String (Optional)
     * }
     * }</pre>
     *
     * @return the response body along with {@link PagedResponse} on successful completion of {@link Mono}.
     */
    private Mono<PagedResponse<BinaryData>> getSinglePagesSinglePageAsync(RequestOptions requestOptions) {
        return FluxUtil.withContext(
                        context -> service.getSinglePages(this.client.getHost(), accept, requestOptions, context))
                .map(
                        res ->
                                new PagedResponseBase<>(
                                        res.getRequest(),
                                        res.getStatusCode(),
                                        res.getHeaders(),
                                        getValues(res.getValue(), "values"),
                                        getNextLink(res.getValue(), "nextLink"),
                                        null));
    }

    /**
     * A paging operation that finishes on the first call without a nextlink.
     *
     * <p><strong>Response Body Schema</strong>
     *
     * <pre>{@code
     * {
     *     properties (Optional): {
     *         id: Integer (Optional)
     *         name: String (Optional)
     *     }
     * }
     * }</pre>
     *
     * @return the paginated response with {@link PagedFlux}.
     */
    public PagedFlux<BinaryData> getSinglePagesAsync(RequestOptions requestOptions) {
        RequestOptions requestOptionsForNextPage = new RequestOptions();
        requestOptionsForNextPage.setContext(
                requestOptions != null && requestOptions.getContext() != null
                        ? requestOptions.getContext()
                        : Context.NONE);
        return new PagedFlux<>(
                () -> getSinglePagesSinglePageAsync(requestOptions),
                nextLink -> getSinglePagesNextSinglePageAsync(nextLink, requestOptionsForNextPage));
    }
```

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
haolingdong-msft: The handling logic is here:
https://github.com/Azure/autorest.java/blob/624655b5b234bd2ccc84adff2034c58574b15559/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java#L109

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,32 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
+                    }
+                    ClientModel model = ClientModelUtil.getClientModel(((ClassType) rawResponseType).getName());
+                    List<ClientModelProperty> properties = new ArrayList<>();
+                    traverseProperties(model, properties);
+                    responseBodyType = properties.stream()
+                            .filter(property -> property.getName().equals(itemName))
+                            .map(clientModelProperty -> clientModelProperty.getClientType())
+                            .map(valueListType -> {
+                                // value type is List<T>, we need to get the typeArguments
+                                IType[] listTypeArgs = ((ListType) valueListType).getTypeArguments();
```
XiaofeiCao, line 96: Maybe test before cast as well?

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
weidongxu-microsoft: OK, then keep it. `PagedResponse<T>` is `Response<List<T>>` so it is a hybrid of Response and List. And we don't need to be too strict as it is in impl.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,35 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
             IType responseBodyType;
             if (JavaSettings.getInstance().isDataPlaneClient()) {
-                responseBodyType = clientMethod.getProxyMethod().getRawResponseBodyType();
+                // special handling for paging method
+                if (clientMethod.getType() == ClientMethodType.PagingSync || clientMethod.getType() == ClientMethodType.PagingAsync || clientMethod.getType() == ClientMethodType.PagingAsyncSinglePage || clientMethod.getType() == ClientMethodType.PagingSyncSinglePage) {
+                    String itemName = clientMethod.getMethodPageDetails().getItemName();
+                    // rawResponseType has properties: 'value' and 'nextLink'
+                    IType rawResponseType = clientMethod.getProxyMethod().getRawResponseBodyType();
+                    if (!(rawResponseType instanceof ClassType)) {
+                        throw new IllegalArgumentException("clientMethod.getProxyMethod().getRawResponseBodyType() should be ClassType for paging method.");
```
weidongxu-microsoft, line 86: Use `IllegalStateException` as none of these is from user input and hence not argument.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,35 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
+                    ClientModel model = ClientModelUtil.getClientModel(((ClassType) rawResponseType).getName());
+                    List<ClientModelProperty> properties = new ArrayList<>();
+                    traverseProperties(model, properties);
+                    responseBodyType = properties.stream()
+                            .filter(property -> property.getName().equals(itemName))
+                            .map(clientModelProperty -> clientModelProperty.getClientType())
+                            .map(valueListType -> {
+                                // value type is List<T>, we need to get the typeArguments
+                                if (!(valueListType instanceof ListType)) {
+                                    throw new IllegalArgumentException("value type must be list");
```
weidongxu-microsoft, line 97: similar to your text in above exception, mention "paging method" so it is easier to diagnose.

mind the sentence and proper punctuation (I assume I don't need to say it in every PR?).

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
weidongxu-microsoft: Then it will not generate the doc for body?

Personally I would still prefer to throw, if this is not an expected situation (so either dev need to fix if this is bug, or swagger need to fix if it is not valid).

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
weidongxu-microsoft: Please also include the `IType.toString()` of the rawResponseType, for diagnostic in all exceptions.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -76,7 +77,38 @@ protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, J
...
+                                }
+                                IType[] listTypeArgs = ((ListType) valueListType).getTypeArguments();
+                                if (listTypeArgs.length == 0) {
+                                    throw new IllegalStateException(String.format("list type arguments' length should not be 0 for paging method. rawResponseType = %s", rawResponseType.toString()));
+                                }
+                                return listTypeArgs[0];
+                            })
+                            .findFirst().orElse(null);
+                    if (responseBodyType == null) {
+                        throw new IllegalStateException(String.format("responseBodyType should not be null for paging method. rawResponseType = %s", rawResponseType.toString()));
```
weidongxu-microsoft, line 107: say e.g. value (<-- itemName) not found in type

weidongxu-microsoft (approved):

Issue #1737, opened 2022-09-27
Cadl, emitter with Custom configuration options


weidongxu-microsoft:
https://microsoft.github.io/cadl/docs/extending-cadl/emitters-basics/#custom-configuration-options

https://github.com/Azure/cadl-azure/issues/1751

epic: https://github.com/Azure/cadl-azure/issues/1947

XiaofeiCao:
I'm considering adding a `CadlSettings` property to `JavaSettings` for cadl-specific settings like `output-folder`, `generate-convenience-methods`. The rest of the settings go directly to `CadlPlugin`'s `SETTINGS_MAP`.

Will draft a PR for discussion.

weidongxu-microsoft:
For now, we may only want to have `output-folder`.

`generate-convenience-methods` is only for test. I won't expect user to set that.

PR #1753, opened 2022-09-30
cadl,support emitter options

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/1737


cadl-extension/src/main/java/com/azure/cadl/Main.java
```diff
@@ -115,28 +92,58 @@ public static void main(String[] args) throws IOException {
...
-                LOGGER.error("Failed to format file: {}", outputFolderFinal + filePath, e);
+                LOGGER.error("Failed to format file: {}", filePath, e);
                 // but we continue so user can still check the file and see why format fails
             }
             formattedFiles.put(filePath, formattedSource);
         });
 
         // write output
         // java files
-        formattedFiles.forEach((filePath, formattedSource) -> cadlPlugin.writeFile(outputFolderFinal + filePath, formattedSource, null));
```
XiaofeiCao, line 126: Align with autorest, resolve output-folder in `CadlPlugin`.

cadl-extension/src/emitter.ts
```diff
@@ -7,7 +7,24 @@ import { CodeModelBuilder } from "./code-model-builder.js";
...
+
+export const $lib = createCadlLibrary({
+  name: "JavaEmitter",
+  diagnostics: {},
+  emitter: {
+      options: EmitterOptionsSchema,
+  },
+});
+
+export async function $onEmit(program: Program, options: EmitterOptions) {
```
XiaofeiCao, line 27: cadl emitter settings goes into `EmitterOptions`.

cadl-tests/cadl-project.yaml
```diff
@@ -1,2 +1,4 @@
 emitters:
-  '@azure-tools/cadl-java': true
+  '@azure-tools/cadl-java':
+    generate-models: true
```
XiaofeiCao, line 3: The settings will directly go into `CadlPlugin`'s `SETTINGS_MAP`, thus readable by `JavaSettings`.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/CadlSettings.java
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.autorest.extension.base.plugin;
+
+public class CadlSettings {
```
weidongxu-microsoft, line 6: why do we have CadlSetting in shared module? Should it be in cadl-extension module?

cadl-tests/cadl-project.yaml
weidongxu-microsoft: we probably don't want to expose `generate-models` yet (not expect user to use it).

cadl-extension/src/test/java/com/azure/cadl/util/ModelUtilTests.java
```diff
@@ -12,16 +12,19 @@
...
     // sadly ModelUtil.isGeneratingModel queries JavaSettings
     @BeforeAll
     public static void ensurePlugin() {
-        CadlPlugin plugin = new CadlPlugin(
-                new CadlPlugin.Options()
-                        .setNamespace("com.azure.client"));
+        CadlPlugin.Options options = new CadlPlugin.Options().withEmitterOptions(
+                new HashMap<String, Object>() {{
+                    this.put("namespace", "com.azure.client");
+                }});
```
weidongxu-microsoft, line 26: Don't use this `{{` in code. Use standard class.

cadl-extension/src/main/java/com/azure/cadl/Main.java
```diff
@@ -115,28 +92,58 @@ public static void main(String[] args) throws IOException {
...
+                LOGGER.info("Read emitter options failed, emitter options json: {}", emitterOptionsJson);
+            }
+        }
+        if (!emitterOptions.containsKey("namespace")) {
+            if (codeModel.getLanguage().getJava() != null && !CoreUtils.isNullOrEmpty(codeModel.getLanguage().getJava().getNamespace())) {
+                emitterOptions.put("namespace", codeModel.getLanguage().getJava().getNamespace());
+            }
+        }
+        emitterOptions.put("output-folder", outputFolderFinal);
+        return new CadlPlugin.Options().withEmitterOptions(emitterOptions);
```
weidongxu-microsoft, line 144: Just `setEmitterOptions` or `addEmitterOptions`?

---

Why serialize to a `Map<String, Object>` and set the generic map via `withEmitterOptions`? We prefer a model whenever we know the data.
We write to and read from this "emitterOptions", then why keep ourselves in the dark about its possible content?

cadl-extension/src/emitter.ts
```diff
@@ -7,7 +13,24 @@ import { CodeModelBuilder } from "./code-model-builder.js";
 import { dirname } from "path";
 import { fileURLToPath } from "url";
 
-export async function $onEmit(program: Program) {
+export interface EmitterOptions {}
```
weidongxu-microsoft, line 16: Same here, we want exact type of possible options (like the code in tutorial). Not a blank `interface {}`.

weidongxu-microsoft:
Use a model (instead of map) whenever you know/control the data.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
```diff
@@ -130,10 +123,6 @@ public String handlePartialUpdate(String filePath, String generatedContent) {
         SETTINGS_MAP.put("required-parameter-client-methods", true);
         SETTINGS_MAP.put("generic-response-type", true);
         SETTINGS_MAP.put("output-model-immutable", true);
```
weidongxu-microsoft, line 125: If we want to change, these are the things we don't want. In Cadl, we would like to make the `JavaSettings` instance without this map.
However the singleton nature of the `JavaSettings` make it hard.

cadl-extension/src/main/java/com/azure/cadl/Main.java
XiaofeiCao: Yes, that's the part I felt weird about. My initial though was to ease the process of increasing options over time, but the map would confuse the dev about what the possible options could be. And if we want to control what options to accept, yes it's better to use a model.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/CadlSettings.java
XiaofeiCao: I'm thinking about a global way of getting cadl settings like `output-folder`. Previously we had `JavaSettings.getAutorestSettings().getOutputFolder()`. I intended to add `JavaSettings.getCadlSettings().getOutputFolder()`. 

For now, the only place it is used is in `CadlPlugin`, so it seems too much to add a new class for it. Will inline the property in `CadlPlugin` instead.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
XiaofeiCao: I see. I would not change that for now.

cadl-extension/src/main/java/com/azure/cadl/Main.java
```diff
@@ -115,28 +92,101 @@ public static void main(String[] args) throws IOException {
...
+            outputFolder = args[1];
+            if (!outputFolder.endsWith("/")) {
+                outputFolder += "/";
+            }
+        }
+
+        final String outputFolderFinal = outputFolder;
+
+        String emitterOptionsJson = Configuration.getGlobalConfiguration().get("emitterOptions");
+        Map<String, Object> emitterOptions = parseJsonAsMap(emitterOptionsJson);
```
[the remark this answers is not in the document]
weidongxu-microsoft: I mean, would it be easier to just serialize it to a e.g. `EmitterOptions` model with `JsonProperties` annotation (than `getNonEmptyStringOrNull` `getBoolean` etc.)?

And maybe `CadlPlugin.Options` and be merged with it (we likely have `outputFolder` in emitter option)?

cadl-extension/src/emitter.ts
```diff
@@ -7,7 +13,32 @@ import { CodeModelBuilder } from "./code-model-builder.js";
...
+}
+
+const EmitterOptionsSchema: JSONSchemaType<EmitterOptions> = {
+  type: "object",
+  additionalProperties: false,
+  properties: {
+    "namespace": { type: "string", nullable: true },
+    "service-name": { type: "string", nullable: true },
+    "partial-update": { type: "boolean", nullable: true },
+  },
```
weidongxu-microsoft, line 29: we had high probability to add  `outputFolder` as well (I am not sure about the exact name though).

Also, should we use camelCase?

XiaofeiCao: Sure, will add that and change it to camel case.

weidongxu-microsoft: currently seems to be `outputPath` https://github.com/microsoft/cadl/issues/482

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -317,7 +317,7 @@ protected void addSerializationImports(Set<String> imports, JavaSettings setting
     }
 
     /**
-     * Override parent setters if: 1. parent property does not have setter 2. child does not contain property
+     * Override parent setters if: 1. parent property has setter 2. child does not contain property
      * that shadow this parent property, otherwise overridden parent setter methods will collide with child setter
```
XiaofeiCao, line 321: Not relevant, fix old doc.

cadl-extension/src/emitter.ts
XiaofeiCao: Changed to outputPath.

cadl-extension/src/main/java/com/azure/cadl/Main.java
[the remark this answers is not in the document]
XiaofeiCao: Changed. Added `outputPath` and `EmitterOptions`.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
```diff
@@ -189,4 +178,8 @@ public void message(Message message) {
                 break;
         }
     }
+
+    private String getOutputFolder() {
+        return (String) SETTINGS_MAP.get("output-folder");
```
weidongxu-microsoft, line 183: nit, seems easier to put the EmitterOptions in class var?

weidongxu-microsoft (approved):

weidongxu-microsoft:
main should be fine with CI now.

cadl-extension/src/main/java/com/azure/autorest/CadlPlugin.java
XiaofeiCao: sure

PR #1754, opened 2022-10-04
Attempt to Parallelize Swagger Generation

alzimmermsft:


weidongxu-microsoft:
I don't know why, but from CI it seems to be slower (maybe agent on devops has less core)?

alzimmermsft:
> I don't know why, but from CI it seems to be slower (maybe agent on devops has less core)?

I'm not sure what is with the timeouts, a previous commit I had cut generation time from 12 minutes to 7 minutes but this commit times out. I'm going to put this on my backlog until MQ as this isn't something that's required and is nice to have.

weidongxu-microsoft (approved):

PR #1769, opened 2022-10-14
eng, remove script that be replaced by PS script

weidongxu-microsoft:


cadl-tests/package.json
```diff
@@ -11,10 +11,8 @@
     "@cadl-lang/rest": "~0.17.0",
     "@cadl-lang/openapi": "~0.12.0",
     "@azure-tools/cadl-azure-core": "~0.7.0",
-    "@azure-tools/cadl-ranch": "latest",
```
haolingdong-msft, line 14: why remove  "@azure-tools/cadl-ranch" and "@azure-tools/cadl-ranch-api" dependency?

weidongxu-microsoft: They are in others dependency.

If everything works with less dependency, why add them?

haolingdong-msft (approved):

PR #1779, opened 2022-10-18
eng, try to see if "-UseNewRunspace" helps

weidongxu-microsoft:
We seems got > 30% of CI hang around here https://github.com/Azure/autorest.java/blob/main/Generate.ps1#L71-L73 

See if "-UseNewRunspace" helps. So far tried 3 times, 3 pass. Though does not mean it is all good...


weidongxu-microsoft:
Not helping much.

PR #1781, opened 2022-10-18
Fix Missing Context in Async Paging

alzimmermsft:
Fixes a bug when generating async paging APIs that don't have `Context context` as a parameter in the method. Previously, this would call into the overload that takes `Context` with the first page retrieval missing a parameter and the next page attempt to use a parameter that doesn't exist. Now, when generating the async paging API without `Context context` `FluxUtil.withContext` will wrap the first and next page retrieval calls and use the context from the reactive stream.


weidongxu-microsoft:
I think the async does not need to pass the Context?

Since the reactor Context should be with the the whole flux, if user do `.writeContext()` to the `PagedFlux`, the nextLink method without azure-core Context parameter, the method would pick the reactor Context in the flux.

PR #1789, opened 2022-10-19
Cadl, add 1 dict test

weidongxu-microsoft:


cadl-tests/src/test/java/com/dictionary/DictionaryTests.java
```diff
@@ -0,0 +1,32 @@
...
+    @Test
+    public void testBooleanGet() {
+        Map<String, Boolean> response = client.get();
+        Assertions.assertEquals(expected, response);
+    }
+
+    @Test
+    public void testBooleanPut() {
+        client.put(expected);
+    }
```
weidongxu-microsoft, line 31: @haolingdong-msft Added 1 test to verify return type as generic works.

Feel free to add others.

PR #1791, opened 2022-10-19
vanilla, remove add-context-parameter and context-client-method-parameter option

weidongxu-microsoft:
After removal, code behaves as if they be `true`.

Checked SDK repo. All readme.md sets "context-client-method-parameter".

DPG and mgmt always configure them as `true`.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
@@ -682,7 +682,7 @@ public class ClientMethodTemplate extends ClientMethodTemplateBase {
                 String firstPageArgs = clientMethod.getArgumentList();
                 if (clientMethod.getParameters()
                     .stream()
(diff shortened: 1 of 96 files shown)


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -797,7 +797,7 @@ private void generateSimpleSyncMethod(ClientMethod clientMethod, JavaType typeBl
             if (CoreUtils.isNullOrEmpty(argumentList)) {
                 // If there are no arguments the argument is Context.NONE
                 argumentList = "Context.NONE";
-            } else if (!clientMethod.getParameters().contains(ClientMethodParameter.CONTEXT_PARAMETER)) {
+            } else if (clientMethod.getParameters().stream().noneMatch(p -> p.getClientType() == ClassType.Context)) {
```
weidongxu-microsoft, line 800: I am not sure why we have different ClientMethodParameter.CONTEXT_PARAMETER, but it happens (maybe some code clones the parameter).

Issue #1465, opened 2022-05-19
Simplify selectively generating models in public package


srnagar:
To generate models in a different package, today, we have to enumerate the names of all models that have to be moved to public package. This is a bit tedious. We should look into how this can be simplified.

weidongxu-microsoft:
current solution before side-car https://github.com/Azure/autorest.java/issues/1670

PR #1794, opened 2022-10-19
cadl, support convenienceAPI

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1465

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethod.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethod.java
@@ -12,6 +12,7 @@ package com.azure.autorest.model.clientmodel;
 
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.util.CodeNamer;
+import com.azure.autorest.util.MethodNamer;
 import com.azure.core.http.ContentType;
 import com.azure.core.http.HttpMethod;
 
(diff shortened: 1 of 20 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethod.java
```diff
@@ -234,23 +235,23 @@ public final boolean isResumable() {
...
 
     public final String getSimpleAsyncRestResponseMethodName() {
-        return getName() + "WithResponseAsync";
+        return MethodNamer.getSimpleAsyncRestResponseMethodName(getName());
     }
 
     public final String getSimpleRestResponseMethodName() {
-        return getName() + "WithResponse";
+        return MethodNamer.getSimpleRestResponseMethodName(getName());
     }
```
weidongxu-microsoft, line 255: I might need to further clean-up this a bit. There seems not much reason to put these ClientMethod related names in ProxyMethod.

weidongxu-microsoft: OK, actually it is those in ClientMethod needs clean up.

ClientMethod should already have the correct name, there is no point for getting another name from that object.

weidongxu-microsoft: In the presence of sync stack, this part would get complicated.

E.g. the proxy method name could be "listSync". The simple client method be "list". The single page method be "listSinglePage" which then calls "listSync" in proxy.

This case, may of these methods like `getPagingSinglePageMethodName` does not give the expected method name (if based on proxy method name).

weidongxu-microsoft: Maybe we should have a "baseName" in ClientMethod, and use it to derive these names.

PR #1798, opened 2022-10-20
cadl, misc changes to readme/action

weidongxu-microsoft:


.github/workflows/generate-test-code.yml
```diff
@@ -1,6 +1,16 @@
...
+    inputs:
+      scope:
+        type: choice
+        description: scope of the test code to generate
+        required: true
+        default: autorest
+        options:
+          - all
+          - autorest
+          - cadl
```
weidongxu-microsoft, line 13: For now, default to "autorest".

I have yet to get "cadl" working. Script seems fine, but job will hang when invoking "java -jar".

Issue #1766, opened 2022-10-14
client, public EnvironmentsClientBuilder addPolicy(HttpPipelinePolicy customPolicy) {


weidongxu-microsoft:
https://github.com/Azure/azure-sdk-for-java/pull/31300#discussion_r994561670

customPolicy needs to be checked for being non-null

PR #1799, opened 2022-10-20
clientBuilder, check customPolicy not null in addPolicy()

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1766


weidongxu-microsoft (commented):
message like "'customPolicy' cannot be null"? <-- check sdk repo for the proper message

weidongxu-microsoft (approved):

PR #1804, opened 2022-10-26
cadl, add cadl ranch dictionary tests

haolingdong-msft:


cadl-tests/src/test/java/com/dictionary/UnknownValueClientTest.java
```diff
@@ -0,0 +1,28 @@
...
+    void get() {
+       Map<String, Object> response = client.get();
+       Assertions.assertEquals(1, response.get("k1"));
+       Assertions.assertEquals("hello", response.get("k2"));
+       Assertions.assertEquals(null, response.get("k3"));
+    }
+
+    @Test
+    void put() {
+        client.putWithResponse(BinaryData.fromString("{\"k1\":1,\"k2\":\"hello\",\"k3\":null}"), null);
```
weidongxu-microsoft, line 26: there is no other way?

haolingdong-msft: I tried to use `JsonNodeFactory` to build the map, but the problem is that we can't set null value to `ObjectNode`
image

weidongxu-microsoft: Please do some investigate yourself https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/node/NullNode.html

haolingdong-msft: I understand this is doable, I can use `map.put("k3", NullNode.instance);`. But this [`put`](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/node/ObjectNode.html#put(java.lang.String,%20com.fasterxml.jackson.databind.JsonNode)) signature is deprecated. So I did not continue to try this way.

weidongxu-microsoft: please just read it


@Deprecated
public JsonNode put(String fieldName,
                      JsonNode value)
Deprecated. Since 2.4 use either set(String,JsonNode) or replace(String,JsonNode),
Method that will set specified field, replacing old value, if any.

haolingdong-msft: Thanks for pointing out, sorry that I did not download the source but just look at `.class` file, so did not notice this description. 

I can use below way to test the `put` method, but from the test perspective, do you think `BinaryData.fromObject()` bring more benefits than `BinaryData.fromString()`? We now understand we can pass in `ObjectNode` to `BinaryData.fromObject()`, and null value will be serialized. But for test, personally I think either way works well. Using `BinaryData.fromString()`, we can easily build the `BinaryData` through the expected response in the log.

image

weidongxu-microsoft: No from test, yes from knowledge. Please change it.

weidongxu-microsoft:
Please use `NullNode`

"set" method takes `JsonNode`. `null` here is ambiguous.

weidongxu-microsoft (approved):

PR #1810, opened 2022-10-28
dpg, fix lro endpoint expression when there is multiple params

weidongxu-microsoft:


weidongxu-microsoft:
BTW, adding `System.exit(0)` seems not fix the CI hang.

haolingdong-msft (approved):

PR #1811, opened 2022-10-31
Update sync client method names

srnagar:
With the sync method name updates to service client, the wrapper client methods names were getting generated incorrectly. This PR fixes the wrapper client's sync method names.

https://github.com/Azure/azure-sdk-for-java/pull/31540#discussion_r997538019


javagen/src/main/java/com/azure/autorest/template/WrapperClientMethodTemplate.java
```diff
@@ -55,6 +55,10 @@ public void write(ClientMethod clientMethod, JavaType typeBlock) {
         if (clientMethod.getType().name().contains("Async") && methodName.endsWith("Async")) {
             methodName = methodName.substring(0, methodName.length() - "Async".length());
         }
+
+        if (clientMethod.getType().name().contains("Sync")) {
+            methodName = methodName.replace("Sync", "");
+        }
```
weidongxu-microsoft, line 61: That seems to be the wrong place of the solution? I would assume whether or not "sync-stack" is enabled, the name of client method stays the same (without "Sync" in it) even in implementation. <-- why should method name expose its implementation detail?
@alzimmermsft personally I feel this to be a bug in ClientTemplateMapper when "sync-stack" enabled.

weidongxu-microsoft: It is a bit related to the method naming, given now that the name of the proxy method could be different in sync mode.
https://github.com/Azure/autorest.java/pull/1794#discussion_r1000118501

alzimmermsft: Agreed, I don't think `sync-stack` should be adding a suffix of `Sync` to synchronous methods. This wasn't being done before for synchronous methods and may end up being a breaking change in a lot of places.

javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientTemplate.java
```diff
@@ -129,25 +130,28 @@ protected void writeClass(AsyncSyncClient syncClient, JavaClass classBlock, Java
...
 
+    clientMethods.stream()
+        .filter(clientMethod -> clientMethod.getMethodVisibility() == JavaVisibility.Public)
+        .filter(clientMethod -> !clientMethod.isImplementationOnly())
+        .filter(clientMethod -> !clientMethod.getType().name().contains("Async"))
+        // service client methods generate overloads that are with and without context that aren't needed
+        // for wrapper public clients
+        .filter(clientMethod -> (JavaSettings.getInstance().isDataPlaneClient())
+                || ((clientMethod.getName().contains("WithResponse") && clientMethod.getParametersDeclaration().contains("Context context"))
+                || ((!clientMethod.getName().contains("WithResponse")
```
weidongxu-microsoft, line 147: DPG has method "WIthResponse" and "RequestOptions".

Many visibility/hidden is now moved to `ClientMethodMapper.methodVisibility`. You might try adjust there. https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java#L1047-L1052

Particularly, since I was not sure about the visibility of the sync-stack mode, I keep it same as what I saw before the refactor (moving the control of visibility to this place). You can set it to `NOT_VISIBLE` I think it will be private in client impl, and not be shown in wrapper clients.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1044,9 +1044,9 @@ protected JavaVisibility methodVisibility(
                     : NOT_GENERATE;
             }
         } else {
-            if (!settings.isSyncStackEnabled() && methodType == ClientMethodType.SimpleSyncRestResponse && !hasContextParameter) {
+            if (methodType == ClientMethodType.SimpleSyncRestResponse && !hasContextParameter) {
                 return NOT_GENERATE;
-            } else if (!settings.isSyncStackEnabled() && methodType == ClientMethodType.SimpleSync && hasContextParameter) {
+            } else if (methodType == ClientMethodType.SimpleSync && hasContextParameter) {
                 return NOT_GENERATE;
             }
```
weidongxu-microsoft, line 1051: nit, the 2 methods might be needed for internal call within client. You can regen the test and see if some error happens. If they are needed, one could use NOT_VISIBLE.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1277,6 +1277,9 @@ private static MethodNamer resolveMethodNamer(ProxyMethod proxyMethod, Convenien
         if (!isProtocolMethod && convenienceApi != null) {
             return new MethodNamer(convenienceApi.getName());
         } else {
+            if (proxyMethod.isSync()) {
+                return new MethodNamer(proxyMethod.getName().replace("Sync", ""));
```
weidongxu-microsoft, line 1281: It would be safer to only remove the trailing "Sync" from name (or call the `getBaseName`)?

As it might affect the API name if it contains "Sync" e.g. "getSynchronizedResult" (for now just an idea, not a real name, but I guess we would eventually encounter one having "Sync" in it)

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientTemplate.java
```diff
@@ -129,25 +130,21 @@ protected void writeClass(AsyncSyncClient syncClient, JavaClass classBlock, Java
...
 
+    clientMethods.stream()
+        .filter(clientMethod -> clientMethod.getMethodVisibility() == JavaVisibility.Public)
+        .filter(clientMethod -> !clientMethod.isImplementationOnly())
+        .filter(clientMethod -> !clientMethod.getType().name().contains("Async"))
+        .filter(clientMethod -> clientMethod.getMethodVisibility().equals(JavaVisibility.Public))
+        .forEach(clientMethod -> {
+          writeMethod(clientMethod, classBlock);
+        });
+
```
weidongxu-microsoft, line 147: The CI failure in cadl-java maybe caused by this code change. Previously it checks whether it wraps a ServiceClient or MethodGroupClient, and add client methods accordingly.

And you can see in CI diff only happen to the Sync client, not the Async client.

javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientTemplate.java
```diff
@@ -129,25 +130,21 @@ protected void writeClass(AsyncSyncClient syncClient, JavaClass classBlock, Java
...
+    List<ClientMethod> clientMethods = serviceClient.getClientMethods();
+    if (CoreUtils.isNullOrEmpty(clientMethods)) {
+      clientMethods = methodGroupClient.getClientMethods();
     }
 
+    clientMethods.stream()
+        .filter(clientMethod -> clientMethod.getMethodVisibility() == JavaVisibility.Public)
+        .filter(clientMethod -> !clientMethod.isImplementationOnly())
+        .filter(clientMethod -> !clientMethod.getType().name().contains("Async"))
+        .filter(clientMethod -> clientMethod.getMethodVisibility().equals(JavaVisibility.Public))
```
weidongxu-microsoft, line 143: Is it same as line above `        .filter(clientMethod -> clientMethod.getMethodVisibility() == JavaVisibility.Public)`?

javagen/src/main/java/com/azure/autorest/template/ServiceSyncClientTemplate.java
```diff
@@ -129,25 +130,21 @@ protected void writeClass(AsyncSyncClient syncClient, JavaClass classBlock, Java
...
-          .filter(clientMethod -> !clientMethod.isImplementationOnly())
-          .filter(clientMethod -> !clientMethod.getType().name().contains("Async"))
-          .forEach(clientMethod -> {
-            writeMethod(clientMethod, classBlock);
-          });
+
+    List<ClientMethod> clientMethods = serviceClient.getClientMethods();
+    if (CoreUtils.isNullOrEmpty(clientMethods)) {
+      clientMethods = methodGroupClient.getClientMethods();
     }
```
weidongxu-microsoft, line 137: This code will not work in MethodGroupClient, if ServiceClient does contain client methods (it always do, whether the wrapper is on ServiceClient or MethodGroupClient)

srnagar: That seems like a bug. Client methods should either be in ServiceClient or MethodGroupClient and not both (even based on the previous code path, either the service client's methods were used or method group client's methods were used). Anyway, I'll use the previous check to switch but something to fix later.

Issue #1813, opened 2022-11-01
Cadl, support client and operation group in cadl-dpg


weidongxu-microsoft:
Epic https://github.com/Azure/cadl-azure/issues/1942

PR #1814, opened 2022-11-01
Cadl, basic operation group

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1813

It only supports single client and multiple operation groups.


cadl-extension/src/code-model-builder.ts
```diff
@@ -181,9 +185,23 @@ export class CodeModelBuilder {
...
+        this.processRoute("", operation);
+      }
+
+      for (const operationGroup of operationGroups) {
+        const operations = listOperationsInOperationGroup(this.program, operationGroup);
+        for (const operation of operations) {
+          this.processRoute(operationGroup.type.name, operation);
+        }
+      }
+    }
```
weidongxu-microsoft, line 204: major change here.

for now, it still uses m4 format. will update it to a new format that supports multiple clients.

javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
```diff
@@ -63,6 +63,25 @@ public static void getAsyncSyncClients(CodeModel codeModel, ServiceClient servic
...
+                                    List<ClientMethod> cMethods = Mappers.getClientMethodMapper().map(o, false);
+                                    if (!cMethods.isEmpty()) {
+                                        String methodName = cMethods.iterator().next().getProxyMethod().getName();
+                                        serviceClient.getClientMethods().stream()
+                                                .filter(m -> methodName.equals(m.getProxyMethod().getName()))
+                                                .forEach(m -> convenienceMethods.add(new ConvenienceMethod(m, cMethods)));
+                                    }
+                                });
+                    });
+            builder.convenienceMethods(convenienceMethods);
```
weidongxu-microsoft, line 83: this is almost (but slightly different) duplicate code as below on method groups, tolerate it for now.

Issue #1812, opened 2022-11-01
Client, for stream response, use BinaryData by default


weidongxu-microsoft:
and provide an option to revert to previous InputStream/Flux<ByteBuffer>

PR #1816, opened 2022-11-01
client, by default use BinaryData for file/stream

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1812

### DIFF
--- a/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
+++ b/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
@@ -937,7 +937,7 @@ public class JavaSettings {
     private final boolean streamResponseInputStream;
 
     /**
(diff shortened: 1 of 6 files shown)


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -154,7 +154,8 @@ public static JavaSettings getInstance() {
                 getBooleanValue(host, "generic-response-type", false),
                 getBooleanValue(host, "stream-style-serialization", false),
                 getBooleanValue(host, "enable-sync-stack", false),
-                getBooleanValue(host, "output-model-immutable", false)
+                getBooleanValue(host, "output-model-immutable", false),
+                getBooleanValue(host, "stream-response-inputstream", false)
```
weidongxu-microsoft, line 158: let me know if there is better name.

weidongxu-microsoft: Changed to "use-input-stream-for-binary"

PR #1815, opened 2022-11-01
Cadl, emitter for multi-client

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/pull/1815

sample yaml

```yaml
clients:
  - language:
      default:
        name: CoreClient
        description: Cadl Core
    protocol: {}
    operationGroups:
      - language:
          default:
            name: ''
            description: ''
        protocol: {}
        $key: ''
        operations:
          - *ref_20
          - *ref_21
          - *ref_22
          - *ref_23
  - language:
      default:
        name: ResponseClient
        description: Response
    protocol: {}
    operationGroups:
      - language:
          default:
            name: ResponseOp
            description: ''
        protocol: {}
        $key: ResponseOp
        operations:
          - *ref_24
          - *ref_25
          - *ref_26
          - *ref_27
```

Java part will be much more complex. It should switch to process this "clients", instead of from codemodel.

PR #1820, opened 2022-11-04
Cadl, support multi-client

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1813

emitter already done in https://github.com/Azure/autorest.java/pull/1815

some feature not there yet, e.g. ServiceVersion and security definition is same across clients.


preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
```diff
@@ -77,6 +82,40 @@ private void transformSchemas(Schemas schemas) {
...
+            }
+          }
+
+          // paging
+          for (Operation operation : pagingOperations) {
+            if (nonNullNextLink(operation)) {
+              addPagingNextOperation(client, operation.getOperationGroup(), operation);
+            }
+          }
+        }
```
weidongxu-microsoft, line 114: when we have multi-client, we need to update operation group to have correct parent.

javagen/src/main/java/com/azure/autorest/mapper/ServiceClientMapper.java
```diff
@@ -54,8 +54,6 @@ public static ServiceClientMapper getInstance() {
```
weidongxu-microsoft, line 54: this class is mostly refactor, so that methods can be reused by subclass `CadlServiceClientMapper`

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -221,54 +230,64 @@ public Client map(CodeModel codeModel) {
...
-                            ? syncClient.getClassName()
-                            : asyncClient.getClassName().replace("AsyncClient", "Client"));
-                    String clientBuilderName = clientName + builderSuffix;
-                    ClientBuilder clientBuilder = new ClientBuilder(
-                            builderPackage, clientBuilderName, serviceClient,
-                            (syncClient == null) ? Collections.emptyList() : Collections.singletonList(syncClient),
-                            Collections.singletonList(asyncClient));
-
+        List<ClientBuilder> clientBuilders = new ArrayList<>();
+        for (Map.Entry<ServiceClient, ? extends com.azure.autorest.extension.base.model.codemodel.Client> entry : serviceClientsMap.entrySet()) {
```
weidongxu-microsoft, line 234: mostly refactor, move it under a loop to support List of ServiceClient

logic not changed, create a ClientBuilder, list of sync/async wrapper AsyncSyncClient for one ServiceClient.

cadl-tests/src/test/java/com/models/property/optional/CollectionsModelClientTest.java
```diff
@@ -13,7 +13,7 @@
 
 class CollectionsModelClientTest {
 
-    CollectionsModelClient client = new ModelsPropertyOptionalClientBuilder().buildCollectionsModelClient();
+    CollectionsModelClient client = new OptionalClientBuilder().buildCollectionsModelClient();
```
weidongxu-microsoft, line 16: change in tests is mostly for changed logic to get the default client name from `listClients` of cadl-dpg (previous default is the whole namespace, now cad-dpg uses last segment of namespace).

cadl-tests/cadl/response.cadl
```diff
@@ -51,7 +51,6 @@ interface CadlCoreOp {
   name: "ResponseClient",
```
weidongxu-microsoft, line 51: this Cadl is used as test for multi-client

haolingdong-msft:
Do we meet the issue that api of `@azure-tools/cadl-dpg` will direct all the operations into client which will bring name conflict? (the issue mentioned by Yuchao previously, https://github.com/imakhlin/azure-rest-api-specs-pr/pull/2)

weidongxu-microsoft:
> Do we meet the issue that api of `@azure-tools/cadl-dpg` will direct all the operations into client which will bring name conflict? (the issue mentioned by Yuchao previously, [imakhlin/azure-rest-api-specs-pr#2](https://github.com/imakhlin/azure-rest-api-specs-pr/pull/2))

It might. Service is responsible to make it correct.

PR #1823, opened 2022-11-07
client, readme for recommended settings

weidongxu-microsoft:
Add baseline for recommended setting for new handwritten data-plane SDK.

I've got some question during swagger review that service asking about output-only model, hence try start the "recommended" section here.


weidongxu-microsoft:
@alzimmermsft Feel free to directly modify the PR.

readme.md
```diff
@@ -47,6 +47,21 @@ Java files will be generated under `where/to/generate/java/files/src/main/java/s
 
 To debug, add `--java.debugger` to the argument list. The JVM will suspend at the beginning of the execution. Then attach a remote debugger in your IDE to `localhost:5005`. **Make sure you detach the debugger before killing the AutoRest process. Otherwise it will fail to shutdown the JVM and leave it orphaned. (which can be killed in the Task Manager)**
 
+# Recommended Settings
+
+- `client-logger: true`, Exception is required to be thrown via `ClientLogger`.
```
srnagar, line 52: The name of the setting doesn't reflect what it does. Can we instead call the setting as `throw-using-client-logger`? Given that this is a relatively new flag, I don't think any new services would have used it yet and can be updated.

readme.md
```diff
@@ -47,6 +47,21 @@ Java files will be generated under `where/to/generate/java/files/src/main/java/s
...
+- `use-default-http-status-code-to-exception-type-mapping: true`, use subclass of `HttpResponseException` defined in azure-core, when service returns unexpected status code.
+- `generic-response-type: true`, use `ResponseBase<H, T>` for response with HTTP headers.
+- `output-model-immutable: true`, make model classes immutable, if model is used only in response.
+
+Settings for handwritten SDK:
+
+- `generate-client-as-impl: true`, generate Client in implementation package, for customization.
+- `models-subpackage: implementation.models`, generate model classes in implementation package.
+- `custom-types-subpackage: models`, generate selected model classes in models package.
+- `service-interface-as-public: true`, generate the interface consumed by `RestProxy` as public to resolve issues with `SecurityManager`.
```
srnagar, line 62: We should make the default value for this setting `true`. We should also make the service interface public (in the implementation package).

weidongxu-microsoft: I will try to make it default to be true, as usually this client is in implementation package.

readme.md
```diff
@@ -47,6 +47,21 @@ Java files will be generated under `where/to/generate/java/files/src/main/java/s
...
+- `generic-response-type: true`, use `ResponseBase` instead a sub-type of `ResponseBase`, performance and reflective access improvement.
```
weidongxu-microsoft, line 63: It is already in above "- `use-default-http-status-code-to-exception-type-mapping: true`, use subclass of `HttpResponseException` defined in azure-core, when service returns unexpected status code."

Maybe I should merge the items, as dev usually do handwritting (very small number of lib is generated as-it-is but likely that will stay as preview).

readme.md
weidongxu-microsoft: I've remove it for now. If dev uses `client-side-validations`, CI would tell them what's wrong.

PR #1824, opened 2022-11-07
mgmt, support arm-id

weidongxu-microsoft:
Mostly for fixing new issues found on m4 4.25.0 https://github.com/Azure/autorest.java/wiki/Fluent-Codegen-Report/38bddfadd26fb8ebb5c47a19ee34c5402a47b8fc

1. support arm-id (from m4 4.25.0)
2. support URL in example
3. switch to BinaryData for file response

### DIFF
--- a/fluentnamer/readme.md
+++ b/fluentnamer/readme.md
@@ -62,5 +62,4 @@ client-logger: true
 generate-client-interfaces: true
 required-parameter-client-methods: true
 client-flattened-annotation-target: none
-use-input-stream-for-binary: true
 ```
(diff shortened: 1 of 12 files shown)


fluentnamer/readme.md
```diff
@@ -62,5 +62,4 @@ client-logger: true
 generate-client-interfaces: true
 required-parameter-client-methods: true
 client-flattened-annotation-target: none
-use-input-stream-for-binary: true
```
weidongxu-microsoft, line 65: Remove the backward-compatible flag.

If this caused breaking changes to GA lib, we can always add it to swagger readme, for that particular lib.

PR #1825, opened 2022-11-07
Use HttpHeaderName in Strongly Typed Header Classes

alzimmermsft:
Use `HttpHeaderName` when deserializing `HttpHeaders` in strongly typed header classes.


javagen/src/main/java/com/azure/autorest/template/util/ModelTemplateHeaderHelper.java
```diff
@@ -0,0 +1,257 @@
...
+
+    private static void generateHeaderDeserializationFunction(ClientModelProperty property, JavaBlock javaBlock) {
+        IType wireType = property.getWireType();
+        boolean needsNullGuarding = wireType != ClassType.String &&
+            (wireType instanceof ArrayType || wireType instanceof ClassType
+                || wireType instanceof EnumType || wireType instanceof GenericType);
+
+        // No matter the wire type the rawHeaders will need to be accessed.
+        String httpHeaderName = HEADER_TO_KNOWN_HTTPHEADERNAME.containsKey(property.getSerializedName().toLowerCase(Locale.ROOT))
+            ? "HttpHeaderName." + CodeNamer.getEnumMemberName(property.getSerializedName())
```
weidongxu-microsoft, line 131: Could there possible that the name of the property and the constant var name in `HttpHeaderName` is not connected via this `CodeNamer.getEnumMemberName` function?

Since you are doing reflection and created that `HEADER_TO_KNOWN_HTTPHEADERNAME` from `HttpHeaderName`, would `HEADER_TO_KNOWN_HTTPHEADERNAME.get(...)` gets the constant name in `HttpHeaderName`?

weidongxu-microsoft (approved):

Generate.ps1
```diff
@@ -5,6 +5,11 @@ $AZURE_ARGUMENTS = "--version=$AUTOREST_CORE_VERSION --java --use=. --output-fol
 $PROTOCOL_ARGUMENTS = "--version=$AUTOREST_CORE_VERSION --java --use=. --output-folder=protocol-tests --data-plane --generate-samples"
 $PROTOCOL_RESILIENCE_ARGUMENTS = "--version=$AUTOREST_CORE_VERSION --java --use=. --data-plane"
 $SWAGGER_PATH = "node_modules/@microsoft.azure/autorest.testserver/swagger"
+$PARALLELIZATION = 3
```
weidongxu-microsoft, line 8: I think default is 5.

Set to a smaller number as we still see sometimes the CI will stop during the parallelism, still no idea why.

weidongxu-microsoft:
nit, it might cause SDK be more dependent on later version of azure-core.

alzimmermsft:
> nit, it might cause SDK be more dependent on later version of azure-core (for pre-defined constants).

It will cause generated code to be dependent on azure-core 1.34.0+, is there anywhere version requirements are being documented?

weidongxu-microsoft:
> It will cause generated code to be dependent on azure-core 1.34.0+, is there anywhere version requirements are being documented?

At present, I only write it in release info https://github.com/Azure/autorest.java/releases

Shouldn't be a big deal, as GA data-plane lib should be in the big pom and they've aligned on azure-core version. And we had doc for solving dependency conflict.

protocol-tests/src/main/java/fixtures/bodycomplex/ArrayAsyncClient.java
```diff
@@ -167,4 +168,106 @@ public Mono<Response<Void>> putEmptyWithResponse(BinaryData complexBody, Request
...
+     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return complex types with array property on successful completion of {@link Mono}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Mono<ArrayWrapper> getValid() {
```
weidongxu-microsoft, line 184: model should not be in DPG.

I don't know what's wrong. It is overridden (temporarily) by GENERATE_CONVENIENCE_METHODS env var in ClientModelUtil.

weidongxu-microsoft: OK, I am removing the usage of dev environment variable in PR https://github.com/Azure/autorest.java/pull/1835

PR #1832, opened 2022-11-09
Cadl, update test case for operation group

weidongxu-microsoft:


cadl-tests/cadl/visibility.cadl
```diff
@@ -45,7 +45,6 @@ interface VisibilityWrite {
   create(@body dog: WriteDog): Dog;
 }
 
-@operationGroup
```
weidongxu-microsoft, line 48: make a hybrid test for with/without operation group

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1287,7 +1287,7 @@ private static MethodNamer resolveMethodNamer(ProxyMethod proxyMethod, Convenien
             return new MethodNamer(convenienceApi.getName());
         } else {
             if (proxyMethod.isSync()) {
-                return new MethodNamer(proxyMethod.getName().replace("Sync", ""));
+                return new MethodNamer(proxyMethod.getBaseName());
```
weidongxu-microsoft, line 1290: I think we should use the baseName now we had it.

PR #1835, opened 2022-11-09
cadl, use dev-options

weidongxu-microsoft:


weidongxu-microsoft:
@XiaofeiCao @haolingdong-msft 
This removed the dev configure via env variable. We will use emitter options, but for now not document it as we don't want user uses it.

haolingdong-msft (approved):

PR #1843, opened 2022-11-14
Add responses without custom headers

srnagar:


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -232,6 +232,23 @@ public Map<Request, List<ProxyMethod>> map(Operation operation) {
             ProxyMethod proxyMethod = builder.build();
             proxyMethods.add(proxyMethod);
 
+            if(asyncRestResponseReturnType instanceof GenericType
+                    && ((GenericType) asyncRestResponseReturnType).getTypeArguments()[0] instanceof GenericType
+                    && ((GenericType) ((GenericType) asyncRestResponseReturnType).getTypeArguments()[0]).getName().equals("ResponseBase")) {
+                IType asyncResponseWithNoHeaders = getAsyncRestResponseReturnType(operation, responseBodyType,
+                        settings.isDataPlaneClient(), settings, true);
+                builder.returnType(asyncResponseWithNoHeaders);
+                builder.name(operationName + "NoCustomHeaders");
```
weidongxu-microsoft, line 241: Do we want a flag for it? Do we expect it be direct exposed to user through sync/async wrapper?

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -251,6 +268,23 @@ public Map<Request, List<ProxyMethod>> map(Operation operation) {
...
+                    builder.name(operationName + "NoCustomHeaders");
+                    builder.customHeaderIgnored(true);
+
+                    proxyMethods.add(builder.build());
+
+                    // reset builder state
+                    builder.returnType(asyncRestResponseReturnType);
+                    builder.name(operationName);
+                    builder.customHeaderIgnored(false);
+                }
```
weidongxu-microsoft, line 287: If the code is same as above (line 235), it better to be extracted to a function (or lambda if that's easier).

weidongxu-microsoft (approved):

alzimmermsft (approved):

PR #1854, opened 2022-11-17
cadl, improve devOptions

weidongxu-microsoft:
Let the emitter have full control over convenienceAPI.


javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -420,9 +420,12 @@ static ObjectSchema parseHeader(Operation operation, JavaSettings settings) {
         headerSchema.setProperties(new ArrayList<>());
         headerSchema.setStronglyTypedHeader(true);
         headerSchema.setUsage(new HashSet<>(Collections.singletonList(SchemaContext.OUTPUT)));
-        if (operation.getConvenienceApi() != null) {
-            headerSchema.getUsage().add(SchemaContext.CONVENIENCE_API);
-        }
+
+        // TODO (weidxu): at present we do not generate convenience API with Header model
+//        if (operation.getConvenienceApi() != null) {
+//            headerSchema.getUsage().add(SchemaContext.CONVENIENCE_API);
```
weidongxu-microsoft, line 426: The `Headers` class is current not used by convenience API, so not generate them for now.

PR #1864, opened 2022-11-18
cadl, apply namer to ConvenienceApi and Client

weidongxu-microsoft:


cadl-tests/cadl/paged.cadl
```diff
@@ -25,6 +25,6 @@ model Resource {
 
 @route("/paged")
 interface PagedOp {
-  @convenienceAPI("listConvenience")
+  @convenienceAPI("ListCONVENIENCE")
```
weidongxu-microsoft, line 28: expect the namer to fix it, so that there is no difference in generated code.

PR #1865, opened 2022-11-21
Cadl, improve async convenience api

weidongxu-microsoft:
use `.flatMap(FluxUtil::toMono)` instead of `.map(Response::getValue)` (to handle null in response value)

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ConvenienceAsyncMethodTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ConvenienceAsyncMethodTemplate.java
@@ -3,6 +3,7 @@
 
 package com.azure.autorest.template;
 
+import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientMethod;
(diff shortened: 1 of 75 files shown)


javagen/src/main/java/com/azure/autorest/template/ConvenienceAsyncMethodTemplate.java
```diff
@@ -73,13 +111,8 @@ protected void writeInvocationAndConversion(
...
             methodBlock.methodReturn(String.format("serviceClient.%1$s(%2$s)", methodName, invocationExpression));
         } else {
-            // return type is Mono<Void>, Response::getValue would be null
-            String returnExpression = responseBodyType.asNullable() == ClassType.Void
-                    ? "%1$s(%2$s).then()%3$s"   // return type is Mono<Void>, Response::getValue would be null
-                    : "%1$s(%2$s).map(Response::getValue)%3$s";
-
             methodBlock.methodReturn(
-                    String.format(returnExpression,
+                    String.format("%1$s(%2$s).flatMap(FluxUtil::toMono)%3$s",
```
weidongxu-microsoft, line 115: The only change is here. Rest is refactor.

Use `FluxUtil::toMono` so it will handle the case of `Response::getValue` is null.

Issue #1834, opened 2022-11-09
Cadl, bug of partial-update on method overload


weidongxu-microsoft:
E.g. first generate DPG without model and convenience API
```java
    @Generated
    @ServiceMethod(returns = ReturnType.COLLECTION)
    public PagedIterable<BinaryData> listMultivariateModels(RequestOptions requestOptions) {
        return new PagedIterable<>(this.client.listMultivariateModels(requestOptions));
    }
```

After enable convenience API for the operation, we'd expect 1 or 2 more methods
```java
    /**
     * List Multivariate Models
     *
     * <p>List models of a resource.
     *
     * @param skip Skip indicates how many models will be skipped.
     * @param top Top indicates how many models will be fetched.
     * @throws IllegalArgumentException thrown if parameters fail the validation.
     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
     * @return response of listing models as paginated response with {@link PagedIterable}.
     */
    @Generated
    @ServiceMethod(returns = ReturnType.COLLECTION)
    public PagedIterable<Model> listMultivariateModels(Integer skip, Integer top) {
        // Generated convenience method for listMultivariateModels
        return new PagedIterable<>(client.listMultivariateModels(skip, top));
    }

    /**
     * List Multivariate Models
     *
     * <p>List models of a resource.
     *
     * @throws com.azure.core.exception.HttpResponseException thrown if the request is rejected by server.
     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
     * @return response of listing models as paginated response with {@link PagedIterable}.
     */
    @Generated
    @ServiceMethod(returns = ReturnType.COLLECTION)
    public PagedIterable<Model> listMultivariateModels() {
        // Generated convenience method for listMultivariateModels
        return new PagedIterable<>(client.listMultivariateModels());
    }
```

[...]

PR #1866, opened 2022-11-22
partial update, fix bug on method overload

haolingdong-msft:
fix https://github.com/Azure/autorest.java/issues/1834

still working on testing the fix. Found a potential  bug with partial update when regen ADP code. Still working on investigating.
Please on hold review. 

Update:
The bug found during regen ADP is caused by ADP defined methods with same name, when `@client` is not applied, the same name methods will be written in the same class and have compilation error.

## Root cause of the issue:
In partial update handler, we will make the existing file as the base file and do below three steps:
1. keep the manually written method
2. replace generated method with the newly generated ones 
3. adding additional newly generated methods. 

In order to avoid overwrite the manual written methods, when adding newly generated methods, we only add the methods that does have the same name with existing methods. But the bug happens when there is existing generated protocol method, and newly generated convenience method, since they have the same name, the convenience method can not be added. 

## Fix:
In step 3, When deciding adding back additional newly generated method, If there is an existing method whose  name is the same as the generated method and is manually written, or if the generated method and the existing method have the same method signature. We won't add the generated member, otherwise, we will add the generated method.


partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -122,23 +122,32 @@ private static String handlePartialUpdateForClassOrInterfaceFile(String generate
...
         for (BodyDeclaration<?> generatedMember : generatedFileMembers) {
-            boolean hasMembersWithSameName = false;
+            boolean needToAddToUpdateMembersList = true;
             for (BodyDeclaration<?> existingMember : updatedMembersList) {
-                if (isMembersWithSameName(existingMember, generatedMember)) {
-                    hasMembersWithSameName = true;
+                // If there is an existing member who has the same name as the generated member and is manually written,
+                // Or if the generated method and the existing method have the same method signature.
+                // Then we don't put the generated member to the updatedMembersList
+                if (isMembersWithSameName(existingMember, generatedMember) && !hasGeneratedAnnotation(existingMember) || isMembersCorresponding(existingMember, generatedMember)) {
```
weidongxu-microsoft, line 146: `isMembersCorresponding(existingMember, generatedMember)` means the case that both has same signature. Does it need to test for `!hasGeneratedAnnotation(existingMember)` as well? (else you update the manual method?)

weidongxu-microsoft, line 146: Move `isMembersCorresponding` as first condition (and maybe rename it to isMemebersSameSignature).
(as assume `isMembersWithSameName(existingMember, generatedMember) && !hasGeneratedAnnotation(existingMember)` is solely for method, and I am not sure whether this condition be to unpredictable for user)

partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -122,23 +122,32 @@ private static String handlePartialUpdateForClassOrInterfaceFile(String generate
...
+                    needToAddToUpdateMembersList = false;
```
weidongxu-microsoft, line 147: I think you can break after setting it to false?

partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -125,33 +129,72 @@ private static String handlePartialUpdateForClassOrInterfaceFile(String generate
...
+        // 1. Verify there is no duplicate methods (methods with same signature are considered duplicate methods)
+        NodeList<CallableDeclaration<?>> methodList = new NodeList<>();
+        for (BodyDeclaration<?> generatedMember : generatedFileMembers) {
+            if (generatedMember.isCallableDeclaration()) {
+                if (isMethodExistsInMethodList(generatedMember.asCallableDeclaration(), methodList)) {
+                    throw new RuntimeException(String.format("Found duplicate methods in the generated file."));
+                }
+                methodList.add(generatedMember.asCallableDeclaration());
+            }
+        }
```
weidongxu-microsoft, line 179: there should be more efficient way to do this
1. use a Set<String> for signatures
2. use a List<String> for signatures, and sort and check duplicate

weidongxu-microsoft (approved):

PR #1869, opened 2022-11-28
cadl, bug fix on LRO without convenience API

weidongxu-microsoft:
1. modify the test to first generate without convenience API and do a compile without tests (to verify the code compiles)
2. update a test
3. fix the bug

### DIFF
--- a/cadl-tests/cadl/lro.cadl
+++ b/cadl-tests/cadl/lro.cadl
@@ -1,11 +1,9 @@
 import "@cadl-lang/rest";
 import "@azure-tools/cadl-azure-core";
-import "@azure-tools/cadl-dpg";
 
 using Cadl.Http;
 using Cadl.Rest;
 using Azure.Core;
-using Azure.DPG;
 
 @service({
   title: "LongRunning",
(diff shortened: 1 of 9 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1251,7 +1251,7 @@ private static MethodPollingDetails methodPollingDetailsFromOperationLinks(
             JavaSettings.PollingDetails pollingDetails,
             JavaSettings settings) {
 
-        if (operation.getOperationLinks() == null || pollingDetails == null) {
+        if (operation.getOperationLinks() == null || pollingDetails == null || operation.getConvenienceApi() == null) {
```
weidongxu-microsoft, line 1254: Only generate the "WithModel" impl method, if convenienceAPI enabled.

cadl-tests/cadl/lro.cadl
```diff
@@ -35,12 +33,14 @@ model ExportedResource {
 
 @route("/long-running")
 interface LongRunningOp {
-  @convenienceAPI("statusMonitorConvenience")
   statusMonitor is CustomCore.PollingOperation<Resource>;
 
   @pollingOperation(LongRunningOp.statusMonitor)
   createOrUpdate is LongRunningResourceCreateOrUpdate<Resource>;
```
haolingdong-msft, line 39: this method does not have convenience method?

weidongxu-microsoft: currently we do not generate model or convenience method for JSON merge patch, as we haven't decided on final solution.

haolingdong-msft (approved):

PR #1871, opened 2022-11-29
cadl, test case on friendlyName on property

weidongxu-microsoft:


cadl-tests/src/main/java/com/cadl/naming/models/DataResponse.java
```diff
@@ -37,7 +37,7 @@ public final class DataResponse {
      * description of type property
      */
     @JsonProperty(value = "type", required = true)
-    private TypesModel type;
+    private TypesModel dataType;
```
weidongxu-microsoft, line 40: This should be correct. We had wire name unchanged, but SDK name changed.

PR #1872, opened 2022-11-30
Updates to Javagen

alzimmermsft:
Adds a new configuration `include-read-only-in-constructor-args` to support cases where read-only properties were previously included in constructors when the property was also required.

Changes a few code patterns when generating stream style serialization.


weidongxu-microsoft (approved):
I assume this is mostly for backward compatible?
User likely do not need read-only property in ctor, if the model class is used as input?

I am not sure why CI fails.

alzimmermsft:
> I assume this is mostly for backward compatible? User likely does not need read-only property in ctor, if the model class is used as input?
> 
> I am not sure why CI fails.

Yeah, the new configuration is for backwards compatibility when a library was generating with

`required-fields-as-ctor-args: true`

and the Swagger defined required, read-only properties. Without this new configuration regenerating would lose a constructor argument and if that was exposed as public API it would be breaking.

I'll follow-up with the CI failures shortly.

Issue #1800, opened 2022-10-20
Generate Java client from Autonomous Development Platform Cadl API description


lirenhe:
This issue tracks validating that a DPG generator can generate the ADP client, models, convinience APIs and protocol methods from https://github.com/Azure/azure-rest-api-specs-pr/pull/9236/files, to achieve parity with what DPG v1.0 can generate from ADP OpenAPI description from the same PR

Inputs:
https://github.com/Azure/azure-rest-api-specs-pr/pull/9236/files

weidongxu-microsoft:
Generated Java with unreleased codegen (model/convenienceAPI via dev option).
[https://github.com/Azure/azure-rest-api-specs-pr/tree/adp-java-sdk/specification/adp/data-plane/cadl](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FAzure%2Fazure-rest-api-specs-pr%2Ftree%2Fadp-java-sdk%2Fspecification%2Fadp%2Fdata-plane%2Fcadl&data=05%7C01%7CWeidong.Xu%40microsoft.com%7C72ce90b2882b40927fef08dab26bd35a%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638018473446972021%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=F%2FBIUXzdCgKsp530kG5fNcc0ZoPvtlK62wKUGV19ynw%3D&reserved=0)

Build pass. Didn’t go any deeper.


There is pretty strange configure in Cadl

    #suppress "@azure-tools/cadl-azure-core/use-standard-operations" "ADP data-plane API uses custom LRO response template the LRO actions."
    @doc("Initiates the process of generating SAS signed URIs for uploading special files for the discovery.")
    @example(
      "./examples/public/ingestion/discovery/Discovery_GenerateSpecialFilesUploadInfo.json",
      "Initiates the process of generating SAS signed URIs for uploading special files for the discovery."
    )
    @extension("x-ms-long-running-operation", true)
    @asyncOperationOptions("location")
    @pollingOperation(ADP.Common.Operations.LongRunningOperationStatusLocation)
    generateSpecialFilesUploadInfo is CustomLongRunningResourceAction<
      Discovery,
      {},
[...]

PR #1897, opened 2022-12-15
fetch operation and add to operationCache if not hit

haolingdong-msft:
Fix the bug mentioned for ADP: https://github.com/Azure/autorest.java/issues/1800#issuecomment-1348125787

ADP API view: https://apiview.dev/Assemblies/Review/3b77bee6b3b940ad9c85df75cee51df1#com.adp.datamanagement.DataManagementClient

ADP generated code: https://github.com/haolingdong-msft/autorest.java/blob/adp-new-codegen/cadl-tests/cadl/adp/DataManagement/cadl-output/src/main/java/com/adp/datamanagement/DataManagementClient.java

### DIFF
--- a/cadl-extension/src/code-model-builder.ts
+++ b/cadl-extension/src/code-model-builder.ts
@@ -433,7 +433,7 @@ export class CodeModelBuilder {
     op.responses.map((it) => this.processResponse(codeModelOperation, it));
 
     this.processRouteForPaged(codeModelOperation, op.responses);
-    this.processRouteForLongRunning(codeModelOperation, operation, op.responses);
+    this.processRouteForLongRunning(codeModelOperation, operation, op.responses, groupName);
 
     operationGroup.addOperation(codeModelOperation);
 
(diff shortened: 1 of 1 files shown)


cadl-extension/src/code-model-builder.ts
```diff
@@ -467,7 +467,7 @@ export class CodeModelBuilder {
     }
   }
 
-  private processRouteForLongRunning(op: CodeModelOperation, operation: Operation, responses: HttpOperationResponse[]) {
+  private processRouteForLongRunning(op: CodeModelOperation, operation: Operation, responses: HttpOperationResponse[], groupName: string) {
```
weidongxu-microsoft, line 470: Move groupName before operation

cadl-extension/src/code-model-builder.ts
```diff
@@ -480,6 +480,9 @@ export class CodeModelBuilder {
 
         if (linkOperation.linkedOperation) {
           // Cadl requires linked operation written before
+          if (!this.operationCache.get(linkOperation.linkedOperation)) {
+            this.processRoute(groupName, linkOperation.linkedOperation);
```
weidongxu-microsoft, line 484: I assume this linkedOperation (processed here) will not appear in API seen by user, as it not added to a client?

weidongxu-microsoft (approved):

cadl-extension/src/code-model-builder.ts
haolingdong-msft: it will not appear to user, I've checked in ADP generated code.

Issue #1863, opened 2022-11-18
cadl, keep updated with cadl-ranch updates


weidongxu-microsoft:

haolingdong-msft:
Share from Laurent: https://teams.microsoft.com/l/message/19:16f8f1ed70d044139be153707bdaee0e@thread.skype/1670954749275?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1670954749275&teamName=Azure%20SDK&channelName=Cadl%20Engineering&createdTime=1670954749275&allowXTenantAccess=false
 
Cadl Ranch was updated to use a storage account in the prod subscription this means you need to update your pipeline step to upload to teh right place
1.	Update the service connnection to use "Cadl Ranch Storage" 
2.	azureSubscription: "Cadl Ranch Storage"
2. Update the storageAccountName flag to --storageAccountName azuresdkcadlranch

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/2057

PR #1906, opened 2022-12-26
cadl, upgrade cadl-ranch to 0.6.3 and add more tests

haolingdong-msft:
link https://github.com/Azure/autorest.java/issues/1863


cadl-tests/src/test/java/com/enums/extensible/ExtensibleClientTest.java
```diff
@@ -1,16 +1,17 @@
...
-package com.extensibleenums;
+package com.enums.extensible;
 
-import com.extensibleenums.models.DaysOfWeekExtensibleEnum;
+import com.enums.extensible.models.DaysOfWeekExtensibleEnum;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
-class ExtensibleEnumsClientTest {
+import static org.junit.jupiter.api.Assertions.*;
```
weidongxu-microsoft, line 10: avoid import *

weidongxu-microsoft (approved):

cadl-tests/src/test/java/com/enums/extensible/ExtensibleClientTest.java
haolingdong-msft: Thanks, updated, it's strange because I already configured avoid import * in intellij. I will check..

Issue #1898, opened 2022-12-19
cadl, support flatten payload in method signature


weidongxu-microsoft:
cadl
https://github.com/Azure/azure-rest-api-specs/blob/feature/cognitiveservices/openai-cadl/specification/cognitiveservices/OpenAI.Inference/routes.cadl#L15
https://github.com/Azure/azure-rest-api-specs/blob/feature/cognitiveservices/openai-cadl/specification/cognitiveservices/OpenAI.Inference/models/embeddings.create.cadl

target Java convenience API signature
`embeddings(String, Union, String, String, String)`

weidongxu-microsoft:
first we may see if we can avoid create a class

```java
Map<String, Object> bodyObject = new HashMap<>();
bodyObject.put("input", input);
...
BinaryData body = BinaryData.fromObject(bodyObject);
```

weidongxu-microsoft:
emitter https://github.com/Azure/autorest.java/pull/1907

javagen part, we likely reuse MethodTransformationDetail for client model in client method.
However, this structure might not be good for sample code generation (it likely be fine for conversion code generation).

PR #1907, opened 2022-12-26
cadl, flatten payload, emitter

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1898

case https://github.com/Azure/autorest.java/blob/9d916ded53c2af25deacdd1f4ed50646312a425c/cadl-tests/cadl/flatten.cadl#L33-L36


cadl-extension/src/code-model-builder.ts
```diff
@@ -624,6 +625,56 @@ export class CodeModelBuilder {
...
+                  },
+                ),
+              );
+            }
+          }
+        }
+      }
+
+      // TODO remove the line after full support on anonymous model
+      op.convenienceApi = undefined;
```
weidongxu-microsoft, line 676: convenience API disable before full support.

protocol API is good.

weidongxu-microsoft:
javagen part, we likely reuse `MethodTransformationDetail` for client model in client method.

However, this structure might not be good for sample code generate.

cadl-extension/src/code-model-builder.ts
haolingdong-msft: ok, so the alias name will not be used? in our case, the name `Request`.
```
 alias Request = { 
   user?: User; 
   input: string; 
 }; 
```

weidongxu-microsoft: alias is only exist when writing Cadl source.

It is not visible from emitter.

PR #1910, opened 2023-01-03
Cadl, premiliary on payload flatten

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1898

current impl likely have more bugs. e.g. when body is optional.

### DIFF
--- a/cadl-tests/src/main/java/com/cadl/flatten/FlattenAsyncClient.java
+++ b/cadl-tests/src/main/java/com/cadl/flatten/FlattenAsyncClient.java
@@ -15,7 +15,11 @@ import com.azure.core.exception.ResourceNotFoundException;
 import com.azure.core.http.rest.RequestOptions;
 import com.azure.core.http.rest.Response;
 import com.azure.core.util.BinaryData;
+import com.azure.core.util.FluxUtil;
 import com.cadl.flatten.implementation.FlattenClientImpl;
+import com.cadl.flatten.models.User;
+import java.util.HashMap;
+import java.util.Map;
 import reactor.core.publisher.Mono;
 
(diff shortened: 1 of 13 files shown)


cadl-tests/src/main/java/com/cadl/flatten/FlattenAsyncClient.java
```diff
@@ -61,4 +65,52 @@ public final class FlattenAsyncClient {
...
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Mono<Void> send(String id, User user, String input) {
+        // Generated convenience method for sendWithResponse
+        RequestOptions requestOptions = new RequestOptions();
+        Map<String, Object> requestObj = new HashMap<>();
+        requestObj.put("user", user);
+        requestObj.put("input", input);
+        BinaryData request = BinaryData.fromObject(requestObj);
```
weidongxu-microsoft, line 91: currently, use Map without model.

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -138,6 +151,46 @@ protected void writeMethodImplementation(
         writeInvocationAndConversion(convenienceMethod, protocolMethod, invocationExpression, methodBlock, typeReferenceStaticClasses);
     }
 
+    private static void writeParameterTransformation(MethodTransformationDetail detail, ClientMethod convenienceMethod, JavaBlock methodBlock) {
```
weidongxu-microsoft, line 154: Long term, `MethodTransformationDetail` might not be the best way to handle the flatten/grouping (or something more advanced).

PR #1911, opened 2023-01-04
add option rename-operation-group

XiaofeiCao:


fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/FluentTransformer.java
```diff
@@ -43,6 +43,7 @@ public CodeModel preTransform(CodeModel codeModel) {
     public CodeModel postTransform(CodeModel codeModel) {
         codeModel = new OperationGroupFilter(fluentJavaSettings.getJavaNamesForRemoveOperationGroup()).process(codeModel);
         codeModel = new NamingConflictResolver().process(codeModel);
+        codeModel = new OperationGroupRenamer(fluentJavaSettings.getJavaNamesForRenameOperationGroup()).process(codeModel);
```
weidongxu-microsoft, line 46: should be before `NamingConflictResolver`?

XiaofeiCao: Yeah, makes sense. If we have two conflict operation groups and want to rename, we probably want to rename both.

weidongxu-microsoft (approved):

Issue #1912, opened 2023-01-03
[Sample Issue]


dvoet:
### Library name and version

azure-resourcemanager-monitor 2.22.0

### Language of the Sample

- [ ] C#/.NET
- [X] Java
- [ ] JavaScript/TypedScript
- [ ] Python
- [ ] Golang
- [ ] Other - Please specify in Issue details field

### Sample Issue Type

- [X] Sample not working
- [ ] Sample missing
- [ ] Do not understand sample

### Issue details

when I try the sample to create a data collection rule [here](https://learn.microsoft.com/en-us/rest/api/monitor/data-collection-rules/create?tabs=Java) I get the error `Invalid resource payload: 'properties' are missing`.

### Expected behavior

the sample code creates a data collection rule

### Actual behavior

error

### Reproduction Steps

follow the sample code

```
import com.azure.core.util.Context;
import com.azure.resourcemanager.monitor.fluent.models.DataCollectionRuleResourceInner;

/** Samples for DataCollectionRules Create. */
public final class Main {
    /*
     * x-ms-original-file: specification/monitor/resource-manager/Microsoft.Insights/preview/2021-09-01-preview/examples/DataCollectionRulesCreate.json
     */
    /**
     * Sample code: Create or update data collection rule.
     *
     * @param azure The entry point for accessing resource management APIs in Azure.
     */
    public static void createOrUpdateDataCollectionRule(com.azure.resourcemanager.AzureResourceManager azure) {
        azure
            .diagnosticSettings()
            .manager()
            .serviceClient()
            .getDataCollectionRules()
            .createWithResponse(
                "myResourceGroup",
                "myCollectionRule",
                new DataCollectionRuleResourceInner().withLocation("eastus"),
                Context.NONE);
    }
}
```

weidongxu-microsoft:
codegen bug, `DataCollectionRuleResourceProperties` inherit `DataCollectionRule`. but the properies from `DataCollectionRule` didn't get flattened into `DataCollectionRuleResourceInner`.

weidongxu-microsoft:
fix in https://github.com/Azure/azure-sdk-for-java/pull/32767

release would be mid Jan. 2.23.0

[...]

PR #1913, opened 2023-01-04
mgmt, bug fix, flatten the empty model with parent

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1912

root cause is codegen see the empty model and skip, without checking that the parent of that empty model has properties.


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -433,7 +433,7 @@ private List<ClientModelPropertyReference> collectPropertiesFromFlattenedModel(
         ObjectSchema targetModelSchema = (ObjectSchema) property.getSchema();
         String originalFlattenedPropertyName = property.getLanguage().getJava().getName();  // not modelProperty.getName()
         ClientModel targetModel = this.map(targetModelSchema);
-        if (targetModel != null && !CoreUtils.isNullOrEmpty(targetModel.getProperties())) {
+        if (targetModel != null && targetModel.getProperties() != null) {
```
XiaofeiCao, line 436: `targetModel.getProperties() != null` seems redundant?

XiaofeiCao (approved):

javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
weidongxu-microsoft: technically not redundant, as `getProperties()` could be null?

though in reality if that be `null`, we probably just get lots of other places throwing NPE...

Issue #1846, opened 2022-11-16
cadl, option bag


weidongxu-microsoft:
https://azure.github.io/azure-sdk/java_introduction.html#option-parameters

Option bag for multiple optional parameters in query/header (may include required parameters as well).

It could be a threshold as well. E.g. group it to Options if there is more than 6 parameters.

This likely to be in sync with .NET.

epic https://github.com/Azure/cadl-azure/issues/1944

weidongxu-microsoft:
Name can be `[OperationName]Options`

weidongxu-microsoft:
Task is relate to method signature of convenience API https://github.com/Azure/autorest.java/issues/1686

weidongxu-microsoft:
If the operation does not have any alias for body, then all params go into a single optional bag. If not, we'll split based on the alias.

PR #1917, opened 2023-01-06
Cadl, preliminary on option bag

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1846

sample on openai https://apiview.dev/Assemblies/Review/cff7890772ed40e8a5785e5efda9d2e2


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -506,6 +487,50 @@ && shouldGeneratePagingMethods()) {
...
+        FLATTEN,
+        GROUPING
+    }
+
+    private void processParameterTransformations(
+            ParameterTransformationType type,
+            List<MethodTransformationDetail> methodTransformationDetails,
+            Set<Parameter> originalParameters,
+            Parameter parameter, ClientMethodParameter clientMethodParameter,
+            boolean isProtocolMethod) {
```
weidongxu-microsoft, line 500: Changes to this class is mostly refactor.

The transformation moved here, so that it can be called on same parameter to either GROUPING or FLATTEN.

The case is that the body properties get flattened to parameters, and then these parameters get grouped to an option bag.

cadl-extension/src/code-model-builder.ts
```diff
@@ -679,6 +701,55 @@ export class CodeModelBuilder {
           }
         }
         request.signatureParameters = request.parameters;
+
+        if (request.signatureParameters.length > 6) {
```
weidongxu-microsoft, line 705: Currently, just hardcode a "6". And this at present only apply when flatten happens.

cadl-extension/src/code-model-builder.ts
```diff
@@ -679,6 +701,55 @@ export class CodeModelBuilder {
           }
         }
         request.signatureParameters = request.parameters;
+
+        if (request.signatureParameters.length >= 5) {
```
weidongxu-microsoft, line 705: Currently, just hardcode a "5".

Likely change to emitter option in future.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -259,36 +259,10 @@ private List<ClientMethod> createClientMethods(Operation operation, boolean isPr
...
-                        }
-                        if (parameter.getOriginalParameter() != null) {
-                            mapping.setOutputParameterProperty(Mappers.getModelPropertyMapper().map(parameter.getTargetProperty()));
-                            mapping.setOutputParameterPropertyName(parameter.getTargetProperty().getLanguage().getJava().getName());
-                        }
-                        detail.getParameterMappings().add(mapping);
+
+                        processParameterTransformations(
+                                methodTransformationDetails, originalParameters,
+                                parameter, clientMethodParameter, isProtocolMethod);
```
weidongxu-microsoft, line 265: Only refactor in this class.

PR #1922, opened 2023-01-10
cadl, fallback to 200 in LRO

weidongxu-microsoft:
The Javadoc of protocol method would still say `Object`.

For now it is for ADP, whose Cadl defines 200 as model1 (`finalType`) and 202 as model2 (`intermediateType`).

For best, we'd hope Cadl source could explicitly define `finalOperation` so that codegen can get the `finalType` unambiguously.


haolingdong-msft:
It's related to this issue: https://github.com/Azure/autorest.java/issues/1800#issuecomment-1330257158

haolingdong-msft (approved):

Issue #1891, opened 2022-12-12
Support the union of types for a property in model definition in CADL


lirenhe:
Latest design https://github.com/Azure/autorest.java/issues/1891#issuecomment-1381285183

---

in https://github.com/Azure/azure-rest-api-specs/blob/feature/cognitiveservices/openai-cadl/specification/cognitiveservices/OpenAI.Inference/models/embeddings.create.cadl#L22, we have the model defined as

input: string | string[] | integer[] | integer[][]; in the follow model.

```
alias EmbeddingsRequest = {
    @doc("The ID of the end-user, for use in tracking and rate-limiting.")
    user?: string;

    @doc("input type of embedding search to use")
    input_type?: string;

    @doc("ID of the model to use")
    "model"?: string;

    @doc("""
    An input to embed, encoded as a string, a list of strings, or a list of token
    lists
    """)
    input: string | string[] | integer[] | integer[][];
};
```
We need to support union of types for a property in model definition in CADL

This concept is similar to AnyOf in openapi 3 so I believe this is a valid case.

weidongxu-microsoft:
anyOf is also not supported

weidongxu-microsoft:
In this case of Union in request, a possible model would be

```java
EmbeddingsRequest setInput(String);

EmbeddingsRequest setInput(List<String>);

EmbeddingsRequest setInput(List<Integer>);

EmbeddingsRequest setInput(List<List<Integer>>);

Object getInput();
```

However, the `setInput(List<Integer>)` already conflict with `setInput(List<List<Integer>>)`, as Java erases the generic.

---

Union likely won't be supported if it is in response.

---

At present, Union is generated as Object (Unknown).

johanste:
My question is what you would want the API to look like if it was written by hand?

weidongxu-microsoft:
Another way maybe a Generic like this https://github.com/pakoito/JavaSealedUnions

This might solve the general problem better, but likely not very easy to use.
Also given the Jackson, likely it and the related Jackson classes need to be in core.

weidongxu-microsoft:
For overload:

If Union appears on API signature, one might prefer to move it to option bag.

- It is simpler to overload a single setter in option bag (aka model), than overload the API multiple times
- If there is multiple Union on API signature, the overload could cause combinatorial explosion

weidongxu-microsoft:
Currently solution

```java
class InputModel {
   private inputString;
   private inputStringArray;
   private inputIntegerArray;
[...]

PR #1925, opened 2023-01-12
Cadl, support union

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1891

design changed, let me merge and leave a history.


cadl-extension/src/code-model-builder.ts
```diff
@@ -1411,27 +1412,41 @@ export class CodeModelBuilder {
...
+
+  private getUnionVariantName(prefix: string, type: Type): string {
+    switch (type.kind) {
+      case "Scalar":
+        return prefix + pascalCase(type.name);
+      case "Enum":
+        return prefix + pascalCase(type.name);
+      case "Model":
+        if (isArrayModelType(this.program, type)) {
+          return this.getUnionVariantName(prefix, type.indexer.value) + "Array";
```
haolingdong-msft, line 1442: This may need to be updapted according to Srikanta's comment ?https://github.com/Azure/autorest.java/issues/1891#issuecomment-1379424119
```java
private List<List<Integer>> inputListOfIntegerList; // naming conventions may change but should use List instead of Array
```

weidongxu-microsoft: I see.

Then `List<List<List<Integer>>> inputListOfListOfIntegerList`?

haolingdong-msft: I think so. Though a bit hard to read..

weidongxu-microsoft: fixed

cadl-extension/src/code-model-builder.ts
```diff
@@ -1763,6 +1783,14 @@ function operationContainsJsonMergePatch(op: HttpOperation): boolean {
   return false;
 }
 
+function pascalCase(name: string): string {
+  if (name.length > 0) {
```
XiaofeiCao, line 1809: Maybe a null check?

weidongxu-microsoft: should be ok; if allow null, the signature would be `string | undefined` or `string | null`

PR #1928, opened 2023-01-16
cadl, upgrade to cadl compiler 0.39

haolingdong-msft:
1. update cadl/compiler to 0.39 in cadl-tests and cadl-extension.
2. fix deprecated API issues.
3. update related tests.


cadl-tests/cadl-project.yaml
```diff
@@ -5,5 +5,3 @@ options:
     emitter-output-dir: "{project-root}/cadl-output"
     partial-update: true
     namer: true
-    dev-options:
-      generate-code-model: true
```
weidongxu-microsoft, line 9: This still needed. (I say "dev-options.generate-convenience-apis" not needed https://github.com/Azure/autorest.java/blob/5e200a48accd521d95f290ad2f0f72035b1799dc/cadl-extension/src/code-model-builder.ts#L390)

cadl-tests/cadl/paged.cadl
```diff
@@ -25,6 +25,6 @@ model Resource {
 
 @route("/paged")
 interface PagedOp {
-  @convenienceAPI("ListCONVENIENCE")
+  @convenientAPI("ListCONVENIENCE")
```
weidongxu-microsoft, line 28: Does `convenientAPI` still take a string? If not, just remove the rename part.

You can double check with Pan.

haolingdong-msft: I checked the code previously, it takes an operation and a boolean value, I am still checking the test to learn a sample, then update our cadl if needed.

weidongxu-microsoft: Operation is just what it decorates. The boolean should be true/false that controls whether to generate.

Make sure that `@convenientAPI(false)` can turn off the convenience API.

haolingdong-msft: sure, will check.

cadl-extension/src/code-model-builder.ts
```diff
@@ -133,6 +135,7 @@ export class CodeModelBuilder {
   public constructor(program1: Program, options: EmitterOptions) {
     this.options = options;
     this.program = program1;
+    this.dpgContext = { program: this.program, generateProtocolMethods: true, generateConvenienceMethods: true };
```
weidongxu-microsoft, line 138: Does not need to be in this PR, but cadl-dpg might want us to read the 2 configures from emitter options. Please double check with Pan.

cadl-tests/cadl/patch.cadl
```diff
@@ -27,7 +27,7 @@ model Resource {
 @route("/patch")
 interface Patch {
   // this convenienceAPI should be ignored
-  @convenienceAPI
+  @convenientAPI(true)
```
weidongxu-microsoft, line 30: maybe just drop it, as it is the default.

cadl-tests/src/main/java/com/cadl/specialheaders/SpecialHeadersAsyncClient.java
```diff
@@ -169,17 +169,6 @@ public Mono<Response<BinaryData>> postWithResponse(String name, BinaryData body,
     /**
      * Send a LRO request with header Repeatability-Request-ID and Repeatability-First-Sent.
      *
-     * <p><strong>Header Parameters</strong>
-     *
-     * <table border="1">
-     *     <caption>Header Parameters</caption>
-     *     <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
-     *     <tr><td>repeatability-request-id</td><td>String</td><td>No</td><td>Repeatability request ID header</td></tr>
-     *     <tr><td>repeatability-first-sent</td><td>String</td><td>No</td><td>Repeatability first sent header as HTTP-date</td></tr>
```
weidongxu-microsoft, line 178: Please log an issue to investigate.

weidongxu-microsoft (approved):

cadl-extension/src/code-model-builder.ts
haolingdong-msft: Sure, I will double check with Pan, reading the configurations from emitter makes more sense. 

PS: I checked with Pan before, cadl-dpg has [createDpgContext](https://github.com/Azure/cadl-azure/blob/main/packages/cadl-dpg/src/decorators.ts#L250) API,  it takes `EmitContext<DpgEmitterOptions>`, but we don't pass in `EmitContext` to code-model-builder.ts, we need to update the constructor to pass in EmitContext if we want to use `createDpgContext`. Currently for conveniency, I just create DpgContext directly. Please let me know your thoughts on it.

cadl-tests/src/main/java/com/cadl/specialheaders/SpecialHeadersAsyncClient.java
haolingdong-msft: issue created: https://github.com/Azure/autorest.java/issues/1933

cadl-extension/src/code-model-builder.ts
weidongxu-microsoft: Yes, you can pass `context: EmitContext<EmitterOptions>` from emitter to ctor of CodeModelBuilder (replace that of `options: EmitterOptions` which is just `context.options`.

cadl-extension/src/code-model-builder.ts
```diff
@@ -132,11 +134,20 @@ export class CodeModelBuilder {
...
+  
+  public constructor(program1: Program, context: EmitContext<EmitterOptions>) {
+    this.options = context.options;
     this.program = program1;
-    this.dpgContext = { program: this.program, generateProtocolMethods: true, generateConvenienceMethods: true };
 
+    const dpgEmitterOptions = {
+      "generate-protocol-methods": true,
+      "generate-convenience-methods": this.options["dev-options"]?.["generate-convenience-apis"] ?? true
+    };
```
weidongxu-microsoft, line 144: I don't get it. I thought the point is that cadl-dpg can use their code to get the configure, instead of we set the configure https://github.com/Azure/cadl-azure/blob/main/packages/cadl-dpg/src/decorators.ts#L245-L246

PR #1930, opened 2023-01-18
mgmt, bug fix, nextLink in parent model

weidongxu-microsoft:
code is in javagen, but only fluentgen would try to distinguish `nextLink().toString()` for URL and `nextLink()` for String in template.

Though now the default flag is treat URL as String.


haolingdong-msft (approved):
LGTM.

PR #1934, opened 2023-01-20
cadl, use isApiVersion in cadl-dpg

haolingdong-msft:
WIP tests part.


cadl-extension/src/code-model-builder.ts
```diff
@@ -242,7 +243,7 @@ export class CodeModelBuilder {
       server.parameters.forEach((it) => {
         let parameter;
 
-        if (it.name === "ApiVersion") {
+        if (isApiVersion(this.program, it as any)) {
```
haolingdong-msft, line 246: This part I referred to [cadl-python](https://github.com/Azure/autorest.python/blob/main/packages/cadl-python/src/emitter.ts#L1143). I know that using `as any` is not a good practice, I saw Isabella created an issue in cadl-dpg, so I leave a question there https://github.com/Azure/cadl-azure/issues/2239#issuecomment-1398055201, I will keep our code updated with python-cadl.

weidongxu-microsoft: You belongs to the cadl-dpg

I would assume we modify it to take `HttpOperationParameter | ModelProperty`
https://github.com/Azure/cadl-azure/blob/fb8b612/packages/cadl-dpg/src/utils.ts#L38-L42

weidongxu-microsoft (approved):

cadl-tests/cadl/server.cadl
```diff
@@ -1,6 +1,8 @@
 import "@cadl-lang/rest";
+import "@cadl-lang/versioning";
 
 using Cadl.Http;
+using Cadl.Versioning;
```
weidongxu-microsoft, line 5: seems not need Cadl.Versioning

weidongxu-microsoft (approved):

cadl-tests/cadl/server.cadl
```diff
@@ -15,10 +17,21 @@ using Cadl.Http;
 
     @doc("top-level domain, use org")
     tld?: string = "org",
+
+    @doc("Api Version")
+    @path
```
weidongxu-microsoft, line 22: Do we need `@path`?

cadl-extension/src/code-model-builder.ts
haolingdong-msft: I can take this issue as well. I will merge this pr first.

XiaofeiCao (approved):

cadl-tests/cadl/apiversion.cadl
```diff
@@ -0,0 +1,29 @@
...
+)
+namespace Cadl.ApiVersion;
+
+enum APIVersions {
+  v1_0: "v1.0",
+}
+
+@route("/apiversion")
+interface ApiVersionOp {
+  read(): string;
```
XiaofeiCao, line 28: What does this operation do? Should we can add a query parameter named `api-version`?

haolingdong-msft: We tested api version in path param. https://github.com/Azure/autorest.java/blob/c3d52e44fd85c31c835e3b176828b3117ca16a47/cadl-tests/src/main/java/com/cadl/apiversion/implementation/ApiVersionClientImpl.java#L140

PR #1935, opened 2023-01-21
cadl, createDpgContext

weidongxu-microsoft:


cadl-extension/src/code-model-builder.ts
```diff
@@ -138,15 +139,7 @@ export class CodeModelBuilder {
...
-    const dpgEmitterOptions = {
-      "generate-protocol-methods": true,
-      "generate-convenience-methods": this.options["dev-options"]?.["generate-convenience-apis"] ?? true,
-    };
-    const dpgEmitterContext = {
-      ...context,
-      options: dpgEmitterOptions,
-    };
-    this.dpgContext = createDpgContext(dpgEmitterContext);
+    this.dpgContext = createDpgContext(context as EmitContext<DpgEmitterOptions>);
```
weidongxu-microsoft, line 142: we already allow additional properties, so cadl-dpg can just get whatever they want from `EmitterOptions`.
https://github.com/Azure/autorest.java/blob/22b939105798cb9bf894e46bf7aafb4b19a93b61/cadl-extension/src/emitter.ts#L34-L37

PR #1936, opened 2023-01-25
Generate tracing config in template

lmolkova:
Generating tracing configuration. It will:
- replace per-request context key to pass Azure resource provider namespace
- ensure that client libraries configure it properly

TODO:
- [ ] regenerate tests


weidongxu-microsoft:
You can first let Alan and Srikanta take a first pass, then get down to update the tests (and some default versions).

Also we might want to decide whether to have it in mgmt, and whether we'd able to deduce a default value if not provided (e.g. from API path), as many DPG/Cadl lib may not involve a handwritten configure.

PR #1937, opened 2023-01-30
cadl, update some tests to use trait

weidongxu-microsoft:


cadl-tests/cadl/special-headers.cadl
```diff
@@ -23,30 +25,23 @@ model Resource {
...
-model SpecialHeaderParameters {
-  @header("Repeatability-Request-ID") requestId?: string;
-  @header("Repeatability-First-Sent") firstSent?: zonedDateTime;
-}
-
-model Customizations {
-  parameters: SpecialHeaderParameters;
-}
+interface ResourceOperations
+  extends Azure.Core.ResourceOperations<SupportsRepeatableRequests & NoConditionalRequests & NoClientRequestId> {}
```
weidongxu-microsoft, line 29: Use SupportsRepeatableRequests trait

cadl-tests/cadl/response.cadl
```diff
@@ -35,24 +39,16 @@ model ResourceArray is Resource[];
...
 
   @doc("Deletes a resource.")
-  delete is ResourceDelete<Resource>;
+  delete is ResourceOperations.ResourceDelete<Resource>;
 
   @doc("Lists the existing resources.")
-  list is ResourceList<Resource>;
-
-  @doc("Creates a new resource with service provided name.")
-  createResource is ResourceCreateWithServiceProvidedName<
```
weidongxu-microsoft, line 50: removed as this is no longer a valid usage.

source should use trait in Azure.Core.Trait instead.

cadl-tests/cadl/response.cadl
```diff
@@ -26,6 +27,9 @@ model Resource {
 
 model ResourceArray is Resource[];
 
+interface ResourceOperations
+  extends Azure.Core.ResourceOperations<NoRepeatableRequests & NoConditionalRequests & NoClientRequestId> {}
```
weidongxu-microsoft, line 31: This would be same as StandardResourceOperations, but that likely will be removed from core
https://github.com/Azure/cadl-azure/blob/main/packages/cadl-azure-core/lib/operations.cadl#L308

cadl-tests/cadl/special-headers.cadl
```diff
@@ -23,30 +25,23 @@ model Resource {
...
 
 @route("/special-headers")
 interface SpecialHeadersOp {
   @get
-  @doc("Send a get request without header Repeatability-Request-ID and Repeatability-First-Sent")
```
weidongxu-microsoft, line 38: GET actually won't have these.

cadl-tests/cadl/special-headers.cadl
weidongxu-microsoft: https://github.com/Azure/cadl-azure/blob/main/packages/cadl-azure-core/lib/traits.cadl#L143-L156

haolingdong-msft (approved):

cadl-tests/cadl/special-headers.cadl
haolingdong-msft: I checked trait related [doc](https://azure.github.io/cadl-azure/docs/getstarted/azure-core/step09). From my understanding, `SupportsRepeatableRequests` trait is used to make sure `repeatability-request-id` and `repeatability-first-sent` exist in request and response header, in other word, we will treat these two header params as required i.e. putting them to the method signature. Please correct me if I am wrong.

weidongxu-microsoft: This 2 param is special. They are generated by SDK instead (a random GUID and current time) for this specific purpose https://docs.oasis-open.org/odata/repeatable-requests/v1.0/cs01/repeatable-requests-v1.0-cs01.html
(hence this is `special-headers.cadl`. this test is not really relates to trait, it is to test that these header is indeed provided by SDK, not suer)

Issue #1946, opened 2023-02-03
Generate TestBase for DPG/Cadl, even if there is no example


weidongxu-microsoft:
Add `generate-tests` and `generate-samples` to `CadlPlugin`.

If there is no `TestBase` generated, update codegen to generate one.

Reason: the Interceptor in TestBase is likely hard to write, for dev not familiar with Java.
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/devcenter/azure-developer-devcenter/src/test/java/com/azure/developer/devcenter/generated/DevCenterClientTestBase.java

PR #1953, opened 2023-02-07
cadl dpg, add test base

XiaofeiCao:
issue: https://github.com/Azure/autorest.java/issues/1946

Always generate TestBase under `generated` package for dpg.


cadl-extension/src/main/java/com/azure/cadl/model/EmitterOptions.java
```diff
@@ -16,18 +16,27 @@ public class EmitterOptions {
...
     @JsonProperty(value="partial-update")
     private Boolean partialUpdate;
 
     @JsonProperty(value="service-versions")
     private List<String> serviceVersions;
 
+    @JsonProperty(value = "generate-tests")
+    private Boolean generateTests;
+
+    @JsonProperty(value = "generate-examples")
```
weidongxu-microsoft, line 37: generate-samples ?

javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -214,15 +214,15 @@ JavaPackage writeToTemplates(CodeModel codeModel, Client client, JavaSettings se
         }
 
         // Test
-        if (settings.isDataPlaneClient() && settings.isGenerateTests()) {
+        if (settings.isDataPlaneClient()) {
```
weidongxu-microsoft, line 217: I think `settings.isGenerateTests()` still needed?

I mean, default `generate-samples` and `generate-tests` to true in CadlPlugin or EmitterOptions.

User can still disable by explicitly setting it to false.

cadl-tests/pom.xml
```diff
@@ -66,5 +66,17 @@
...
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core-test</artifactId>
+      <version>1.14.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-identity</artifactId>
+      <version>1.7.3</version>
```
weidongxu-microsoft, line 78: use 1.8.0

cadl-extension/src/main/java/com/azure/cadl/model/EmitterOptions.java
XiaofeiCao: Right, thanks...

javagen/src/main/java/com/azure/autorest/Javagen.java
XiaofeiCao: You mean if user sets `generate-tests` to false, then the TestBase won't be generated as well? Makes sense, will do that.

weidongxu-microsoft: Yes, `TestBase` still follow the `generate-tests`. The change is when there is no tests, still generate the TestBase (so dev can add a new test class extends it, avoid writing these interceptorManager).

weidongxu-microsoft (approved):

PR #1961, opened 2023-02-09
cadl, upgrade compiler 0.40

haolingdong-msft:


weidongxu-microsoft (approved):
update readme as well.

PR #1968, opened 2023-02-10
partialupdate, fix module-info.java does not include comments

haolingdong-msft:
Currently if we set partial-update to true. The comments in `module-info.java` are missing. The reason is `compilationUnitForGeneratedFile.toString()` does not contain comments, but `compilationUnitForGeneratedFile. getOrphanComments != null`. I think this is JavaPaser's bug for module-info.java. This pr fix it by adding comments from generated file in front of `compilationUnitForGeneratedFile.toString()`.
image


weidongxu-microsoft (approved):

XiaofeiCao (approved):

partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -272,7 +271,13 @@ private static String mergeModuleFileContent(String generatedFileContent, String
...
         compilationUnitForGeneratedFile.setModule(moduleDeclaration);
 
-        return compilationUnitForGeneratedFile.toString();
+        // add comments as compilationUnitForGeneratedFile.toString() does not include comments
+        StringBuilder comments = new StringBuilder();
+        for (Comment comment : compilationUnitForGeneratedFile.getOrphanComments()) {
+            comments.append(comment.toString());
+        }
+
+        return comments.toString() + compilationUnitForGeneratedFile.toString();
```
XiaofeiCao, line 280: nit, orphan comments at the end of the file would get moved to the top

PR #1975, opened 2023-02-14
cadl, add more cadl-ranch tests

haolingdong-msft:
1. Support get serialized name from projected name
2. Add more cadl-ranch tests
PS: I tried  to add visibility tests using protocol method, the tests passed, let me know if you think it's better to remove those tests.


cadl-extension/src/code-model-builder.ts
```diff
@@ -1469,7 +1469,7 @@ export class CodeModelBuilder {
       nullable: nullable,
       readOnly: this.isReadOnly(prop),
       // clientDefaultValue: this.getDefaultValue(prop.default),
-      serializedName: prop.name,
+      serializedName: this.getSerializedName(prop),
```
weidongxu-microsoft, line 1472: Could you check other places?

weidongxu-microsoft: And what about operation parameters? Do the `projectedName(.., "json")` apply for it?
https://github.com/Azure/autorest.java/blob/ca3fa9c275f048f35737416296134c9d316552d0/cadl-extension/src/code-model-builder.ts#L601

cadl-tests/src/test/java/com/authentication/union/AzureKeyCredentialPolicy.java
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.authentication.union;
```
weidongxu-microsoft, line 4: Are they same as the other case?

Can we move it to e.g. util package?

cadl-tests/src/test/java/com/models/visibility/automatic/AutomaticClientTest.java
```diff
@@ -0,0 +1,44 @@
...
+package com.models.visibility.automatic;
+
+import com.azure.core.util.BinaryData;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class AutomaticClientTest {
+
+    AutomaticClient client = new AutomaticClientBuilder().buildClient();
```
weidongxu-microsoft, line 13: add a comment to mention that we just do this to make coverage.

weidongxu-microsoft (commented):
Most LGTM.

The projection to json is confusing from Cadl.

And cadl-autorest does not support it
https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAY2FkbC1sYW5nL3Jlc3QiOwrTGnZlcnNpb25pbmfMIGF6dXJlLXRvb2xzL8UsxhFjb3JlIjsKCnVzaW5nIENhZGwuSHR0cDvMEVJlc3TNEVbJYsgXQcRNLkNvcmXSEi5UcmFpdHM7CgpAc2VydmljZSh7CiAgdGl0bGU6ICJDb250b3NvIFdpZGdldCBNYW5hZ2VyIiwKfSkKQOcAzGVkKMcnLsYnxyboAIpzKQpuYW1lc3BhY2Ug1So7CgplbnVtIMgxIOQAg0B1c2VEZXBlbmRlbmN5KOsAs8gnLnYxXzBfUHJldmlld18yKQogIHYyMDIyXzA4XzMxOiAixA0tMDgtMzHkALIKCkBkb2MoIlByb2plY3Rpb24iKQptb2RlbCDHE8Z4xSVPbmx5IHZhbGlkxAZ1ZSBpcyAnRFBHJyLEZ0BwxjFlZE5hbWUoImpzb24iLCAiY29kZWdlxFggIHByb2R1Y2VkQnk%2FOiBzdHLlAZHfYNhgY2xpZW50xWJyZWF0xFXFKGJ1aWx0ZnJvbd9jymNjdXN0b21lcnPfaSJEb250VXNlTWVBc8QbIikgLy8gU2hvdWxkIGJlIGlnbm9yZWTUQ3NoYXJwxENNYWRlRm9yQ1PVa2phdmHLJkrED9goc2NyaXDlAJfHLlTWVHB5dGjmAYptYWRlX2Zvcl%2FHE8Qud2Fzxz%2FrAS%2F2AhdAcm91dGUoIi%2FlAdopCkBwb3N0Cm9wIMQQyiooLi4uxw4pOiB2b2lkOwo%3D

cadl-extension/src/code-model-builder.ts
haolingdong-msft: yeah, parameter don't need projected name to get serialiezed name. 
I checked both our code and cadl-dpg code, currently we just support getting projected name for model property. There are two places where we set serialized name for model property. One is here, another is for [discriminator](https://github.com/Azure/autorest.java/blob/main/cadl-extension/src/code-model-builder.ts#L1360). I don't change for discriminator because we don't have such use case and it makes the code complex. Please let me know your thoughts.

cadl-tests/src/test/java/com/authentication/union/AzureKeyCredentialPolicy.java
haolingdong-msft: Yes, they are the same, I will create a util package.

weidongxu-microsoft (approved):

PR #1982, opened 2023-02-16
cadl, refactor for multiple server

weidongxu-microsoft:


cadl-extension/src/code-model-builder.ts
```diff
@@ -183,16 +182,8 @@ export class CodeModelBuilder {
...
-    // TODO: multiple servers
-    if (servers && servers.length === 1) {
-      this.baseUri = servers[0].url;
-    }
-    this.hostParameters = [];
-    this.processHost(servers?.length === 1 ? servers[0] : undefined);
-
+    // auth
+    // TODO: it is not very likely, but different client could have different auth
     const auth = getAuthentication(this.program, serviceNamespace);
```
weidongxu-microsoft, line 187: auth can be different, but not very likely.

cadl-tests/src/main/java/com/cadl/server/AnotherServerClientBuilder.java
```diff
@@ -0,0 +1,268 @@
...
+    @Override
+    public AnotherServerClientBuilder endpoint(String endpoint) {
+        this.endpoint = endpoint;
+        return this;
+    }
+
+    /*
+     * Service version
+     */
+    @Generated private ServerServiceVersion serviceVersion;
```
weidongxu-microsoft, line 163: The `ServerServiceVersion` is wrong for now. Will handle in follow-up PR. This requires us to figure out the name of more than one ServiceVersion.

weidongxu-microsoft:
TODO: may see if I can move some function/class out of code-model-builder. The file already too long.

Issue #1976, opened 2023-02-14
mgmt premium, avoid UUID subscriptionId in ManagementClient(Impl) and Client(Impl)


XiaofeiCao:
relate: https://github.com/Azure/azure-sdk-for-java/pull/33496#discussion_r1105437233

`SubscriptionIdParameter` from `specifications/common-types/v4` has format `uuid`. Current codegen will translate it into UUID in xxManagementClient and xxClient, resulting in API change for services starting to adopt this parameter in their swagger.

PR #1983, opened 2023-02-16
mgmt, treat subscription id as string

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/1976

Solution is to map parameter type to String in both ManagementClient and proxy methods.

### DIFF
--- a/fluent-tests/prepare-tests.bat
+++ b/fluent-tests/prepare-tests.bat
@@ -67,10 +67,6 @@ REM swagger customed Resource and ProxyResource
 CALL autorest --version=%AUTOREST_CORE_VERSION% %FLUENT_ARGUMENTS% --input-file=https://github.com/Azure/azure-rest-api-specs/blob/main/specification/trafficmanager/resource-manager/Microsoft.Network/stable/2018-08-01/trafficmanager.json --namespace=com.azure.mgmttest.trafficmanager
 if %errorlevel% neq 0 exit /b %errorlevel%
 
-REM fluent premium uuid subscriptionId
-CALL autorest --version=%AUTOREST_CORE_VERSION% %FLUENT_ARGUMENTS% --input-file=https://github.com/Azure/azure-rest-api-specs/blob/0a2eb0d14f5132fcfd30222d584acf67713332ea/specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/stable/2022-12-01/containerregistry.json --namespace=com.azure.mgmttest.containerregistry
-if %errorlevel% neq 0 exit /b %errorlevel%
-
 REM fluent lite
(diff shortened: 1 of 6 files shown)


fluent-tests/prepare-tests.bat
```diff
@@ -83,6 +87,10 @@ if %errorlevel% neq 0 exit /b %errorlevel%
 CALL autorest --version=%AUTOREST_CORE_VERSION% %FLUENTLITE_ARGUMENTS% --regenerate-pom=false https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/resources/resource-manager/readme.md --tag=package-policy-2020-09 --java.namespace=com.azure.mgmtlitetest.policy
 if %errorlevel% neq 0 exit /b %errorlevel%
 
+REM fluent lite uuid subscriptionId
+CALL autorest --version=%AUTOREST_CORE_VERSION% %FLUENTLITE_ARGUMENTS% --input-file=https://github.com/Azure/azure-rest-api-specs/blob/0a2eb0d14f5132fcfd30222d584acf67713332ea/specification/containerregistry/resource-manager/Microsoft.ContainerRegistry/stable/2022-12-01/containerregistry.json --namespace=com.azure.mgmtlitetest.containerregistrylite
+if %errorlevel% neq 0 exit /b %errorlevel%
+
```
weidongxu-microsoft, line 93: nit, Lite should be enough. That should cover premium.

javagen/src/main/java/com/azure/autorest/mapper/ProxyParameterMapper.java
```diff
@@ -58,6 +58,10 @@ public ProxyMethodParameter map(Parameter parameter) {
 
         Schema parameterJvWireType = parameter.getSchema();
         IType wireType = Mappers.getSchemaMapper().map(parameterJvWireType);
+        // fluent premium treats subscriptionId as string in order not to break ManagementClients and client methods
+        if (settings.isFluentPremium() && "subscriptionId".equals(CodeNamer.getPropertyName(name))) {
```
weidongxu-microsoft, line 62: why `isFluentPremium`?

weidongxu-microsoft (commented):
Not sure if `FluentTransformer` be a better place? E.g. only check those under `codeModel.globalParameters`. If so, modify the schema (may need to create a StringSchema).

We don't want to match by name for all parameters. Who knows whether some one want to have a valid `subscriptionId` param in api.

javagen/src/main/java/com/azure/autorest/mapper/ProxyParameterMapper.java
XiaofeiCao: Thought we may break existing managementClient in lite?

weidongxu-microsoft: It does not break. Officially ClientManagement is not included in revapi.

XiaofeiCao: OK. FluentTransformer should be a better place to make the change then.

PR #1984, opened 2023-02-17
cadl, upgrade ranch

weidongxu-microsoft:
also bump the autorest.java version


cadl-tests/src/test/java/com/unions/generated/unions/UnionTests.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class UnionTests {
+
+    private UnionsClient client = new UnionsClientBuilder().buildClient();
+
+    @Test
+    public void testNamed() {
+        client.sendFirstNamedUnionValue(new ModelWithNamedUnionProperty(
+                new Model1NamedUnionModel(new Model1("model1", 1))));
+    }
```
weidongxu-microsoft, line 21: I only added one case to validate.

PR #1994, opened 2023-02-21
Inline Stream-Style JSON Deserialization when Constructor has no Arguments

alzimmermsft:
Inlines setting property values on a deserialized object when it has no constructor arguments.


vanilla-tests/src/main/java/fixtures/streamstyleserializationctorargs/models/Shark.java
```diff
@@ -0,0 +1,240 @@
...
+                        }
+                    }
+                    if (lengthFound && birthdayFound) {
+                        Shark deserializedShark = new Shark(length, birthday);
+                        deserializedShark.setSpecies(species);
+                        deserializedShark.setSiblings(siblings);
+                        deserializedShark.age = age;
+
+                        return deserializedShark;
+                    }
```
weidongxu-microsoft, line 227: I assume changes are about this part?

Note that `--output-model-immutable` could make the ctor private. Though I think code will not be affected, as here it can call private ctor.

alzimmermsft: The change being made is so that when there are no required constructor arguments deserialization will create an instance of the model eagerly and read JSON into the model instead of making references for each field and finally creating and populating the model.

This shouldn't have any issues with `--output-model-immutable` as that constructor would be accessible to the model itself, but I need to ask whether the constructor is accessible to sub-types.

weidongxu-microsoft (approved):

vanilla-tests/src/main/java/fixtures/streamstyleserializationctorargs/models/Shark.java
weidongxu-microsoft: Seems to be `protected` https://github.com/Azure/autorest.java/blob/main/vanilla-tests/src/main/java/fixtures/bodycomplex/models/DotFish.java#L32

So should be good?

alzimmermsft: Yeah, that should be good. Was less wondering for this PR specifically and more in general in the case models were spread across multiple packages.

PR #1996, opened 2023-02-21
Changing Script to Prevent Deadlocking

alzimmermsft:
Attempt to prevent `Generate.ps1` from deadlocking randomly.

### DIFF
--- a/Generate.ps1
+++ b/Generate.ps1
@@ -39,6 +39,27 @@ SUCCEEDED
     }
 }
 
+function singleThreadGenerate($arguments) {
+    $generateOutput = Invoke-Expression "autorest $arguments"
+    $global:ExitCode = $global:ExitCode -bor $LASTEXITCODE
+
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "
+========================
+autorest $arguments
+========================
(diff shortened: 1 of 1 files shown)


Generate.ps1
```diff
@@ -7,7 +7,7 @@ $SWAGGER_PATH = "node_modules/@microsoft.azure/autorest.testserver/swagger"
 $AZURE_DATAPLANE_ARGUMENTS = "--use=./ --output-folder=./azure-dataplane-tests"
 $AZURE_DATAPLANE_PATH = "azure-dataplane-tests/swagger"
 $AZURE_SDK_FOR_JAVA = "https://github.com/Azure/azure-sdk-for-java/blob/main/sdk"
-$PARALLELIZATION = 5
+$PARALLELIZATION = 4
```
srnagar, line 10: It's good to document the reason for changing this from 5 to 4.

srnagar (approved):

Generate.ps1
weidongxu-microsoft: I think I once changed it to 1 but still see it happen.

Anyway, let's try.

weidongxu-microsoft (approved):

PR #1997, opened 2023-02-22
mgmt, bug fix on multiple error response

weidongxu-microsoft:
Re-use logic in `ProxyMethodMapper`.

Change is for built-in `ManagementError`, use built-in `ManagementException`.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
@@ -63,6 +63,7 @@ public class ProxyMethodMapper implements IMapper<Operation, Map<Request, List<P
     private static final Pattern APOSTROPHE = Pattern.compile("'");
 
     private final Map<Request, List<ProxyMethod>> parsed = new ConcurrentHashMap<>();
+
     protected ProxyMethodMapper() {
     }
 
@@ -492,7 +493,7 @@ public class ProxyMethodMapper implements IMapper<Operation, Map<Request, List<P
         }
     }
 
-    private static SwaggerExceptionDefinitions getSwaggerExceptionDefinitions(Operation operation,
+    private SwaggerExceptionDefinitions getSwaggerExceptionDefinitions(Operation operation,
         JavaSettings settings) {
 
         SwaggerExceptionDefinitions exceptionDefinitions = new SwaggerExceptionDefinitions();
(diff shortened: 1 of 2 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -538,7 +539,10 @@ private static SwaggerExceptionDefinitions getSwaggerExceptionDefinitions(Operat
...
                 }
 
-                if (swaggerDefaultExceptionType == null) {
+                if (swaggerDefaultExceptionType == null
+                        && !CoreUtils.isNullOrEmpty(operation.getExceptions())
+                        // m4 could return Response without schema, when the Swagger uses e.g. "produces: [ application/x-rdp ]"
+                        && operation.getExceptions().get(0).getSchema() != null) {
                     // no default error, use the 1st to keep backward compatibility
                     swaggerDefaultExceptionType = processExceptionClassType(
                             (ClassType) Mappers.getSchemaMapper().map(operation.getExceptions().get(0).getSchema()), settings);
```
weidongxu-microsoft, line 548: Add a check on whether the `exceptions.[0].schema` is `null`. Skip the SchemaMapper if schema is null.
(if we do the map, `swaggerDefaultExceptionType` will still be `null`)

Existing code will fill the blank
https://github.com/Azure/autorest.java/blob/68190ca3926755c36e10a0b9165f7584b7bc31ad/javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java#L458-L462

javagen/src/main/java/com/azure/autorest/mapper/ProxyMethodMapper.java
```diff
@@ -570,7 +574,14 @@ private static ClassType getExceptionType(Response exception, JavaSettings setti
...
 
-    private static ClassType processExceptionClassType(ClassType errorType, JavaSettings settings) {
+    /**
+     * Extension for map error ClassType to exception ClassType.
+     *
+     * @param errorType the error class.
+     * @param settings the Java settings.
+     * @return the exception ClassType.
+     */
+    protected ClassType processExceptionClassType(ClassType errorType, JavaSettings settings) {
```
weidongxu-microsoft, line 584: changed to `protected` and non-`static` for override.

weidongxu-microsoft (commented):
javagen change only in ProxyMethodMapper.java

fluentgen change only in FluentProxyMethodMapper.java

Rest is actually a fix of inconsistency of Javadoc (whether use full name or simple name for Exception) that affect cadl-java and its tests.

alzimmermsft (approved):

PR #1998, opened 2023-02-22
Cadl, support skip url encoding

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1636


cadl-tests/src/main/java/com/cadl/builtin/implementation/BuiltinClientImpl.java
```diff
@@ -133,6 +134,8 @@ public interface BuiltinClientService {
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<BinaryData>> read(
                 @HostParam("endpoint") String endpoint,
+                @QueryParam("query") String queryParam,
+                @QueryParam(value = "query-encoded", encoded = true) String queryParamEncoded,
```
weidongxu-microsoft, line 138: This would be vanilla logic. If query parameter is already encoded, use the `encoded = true`.

cadl-tests/src/main/java/com/cadl/builtin/BuiltinClient.java
```diff
@@ -172,21 +181,36 @@ public Response<Void> writeWithResponse(BinaryData body, RequestOptions requestO
...
-            requestOptions.addQueryParam("filter", filter);
+            requestOptions.addQueryParam("filter", filter, false);
         }
-        return readWithResponse(requestOptions).getValue().toObject(Builtin.class);
+        if (queryParamOptional != null) {
+            requestOptions.addQueryParam("query-opt", queryParamOptional, false);
+        }
+        if (queryParamOptionalEncoded != null) {
+            requestOptions.addQueryParam("query-opt-encoded", queryParamOptionalEncoded, true);
+        }
```
weidongxu-microsoft, line 204: Here is the Cadl logic in convenience API.

If query parameter is already URL encoded, use the 3rd parameter to set `encoded` to `true` on `RequestOptions.addQueryParam`.

PR #1999, opened 2023-02-22
Add Logging when a Job takes too Long

alzimmermsft:
Adds job status checking when timeout elapses and logs a message indicating whether a job wasn't able to complete in time.


weidongxu-microsoft (approved):
Not sure why the code in azure-dataplane-tests changed. I didn't see other code change or option change?

PR #2001, opened 2023-02-23
DPG, enable sync stack

weidongxu-microsoft:
`enable-sync-stack` by default for DPG (not Cadl).

So that future SDK via DPG can be generated with sync stack.

Bug fix https://github.com/Azure/autorest.java/pull/2001/commits/836289aa5523678798e71944b378c03f011d5581


weidongxu-microsoft:
@haolingdong-msft 
I need to regen the partial-update-test.

Please check if this step (manual change to mock a grow-up) is correct https://github.com/Azure/autorest.java/pull/2001/commits/36feb903977c0fc2315c20b3293dc2bc2337958f

haolingdong-msft:
I think it's correct.

weidongxu-microsoft:
> I think it's correct.

I seems to miss this one https://github.com/Azure/autorest.java/pull/2001/commits/d1ffff14e929ca04822d918b1047ea05ee90699c

PR #2003, opened 2023-02-24
cadl, support @protocolApi  @convenientApi (codegen part)

haolingdong-msft:
This pr includes:
1. ClientMethod.java: add boolean `methodVisibilityInWrapperClient` to `ClientMethod`
2. ClientMethodMapper.java: set `methodVisibilityInWrapperClient` from `operation.getIsGenerateProtocolApi()`
3. WrapperClientMethodTemplate.java: support generating package-private protocol methods
4. Operation.java: use `Boolean` type for `methodVisibilityInWrapperClient`.


javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientMethod.java
```diff
@@ -136,7 +138,7 @@ protected ClientMethod(String description, ReturnValue returnValue, String name,
                            boolean isGroupedParameterRequired, String groupedParameterTypeName,
                            MethodPageDetails methodPageDetails,
                            List<MethodTransformationDetail> methodTransformationDetails,
-                           JavaVisibility methodVisibility, ImplementationDetails implementationDetails,
+                           JavaVisibility methodVisibility, JavaVisibility methodVisibilityInWrapperClient, ImplementationDetails implementationDetails,
```
weidongxu-microsoft, line 141: put it to ImplementationDetails ?

javagen/src/main/java/com/azure/autorest/template/WrapperClientMethodTemplate.java
```diff
@@ -68,7 +71,14 @@ public void write(ClientMethod clientMethod, JavaType typeBlock) {
...
 
             writeMethodInvocation(clientMethod, function, shouldReturn);
-        });
+        };
+        if (clientMethod.getMethodVisibilityInWrapperClient() == JavaVisibility.Public) {
+            typeBlock.publicMethod(declaration, method);
+        } else if (typeBlock instanceof JavaClass && clientMethod.getMethodVisibilityInWrapperClient() == JavaVisibility.PackagePrivate) {
+            JavaClass classBlock = (JavaClass) typeBlock;
+            classBlock.packagePrivateMethod(declaration, method);
+        }
```
weidongxu-microsoft, line 80: this "if" should be able to be unified.

there should be a `method` that takes `JavaVisibility`.

haolingdong-msft: You mean something like this?
```java
 if (typeBlock instanceof JavaClass) {
            JavaClass classBlock = (JavaClass) typeBlock;
            classBlock.method(clientMethod.getMethodVisibilityInWrapperClient(), null, declaration, method);
 }
```

extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Operation.java
```diff
@@ -82,9 +82,9 @@ public class Operation extends Metadata {
     private ConvenienceApi convenienceApi;
 
     /**
-     * a boolean to represent should generate protocol api or not
+     * a boolean to represent should generate protocol api or not, default is true
      */
-    private boolean generateProtocolApi;
+    private boolean generateProtocolApi = true;
```
weidongxu-microsoft, line 87: For consistency, use `Boolean`.

This is code model.
Whether null means true/false the logic of Mapper.

javagen/src/main/java/com/azure/autorest/template/WrapperClientMethodTemplate.java
weidongxu-microsoft: Yes.

The point of having a JavaVisibility instead of boolean, is to use it.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -283,6 +283,7 @@ private List<ClientMethod> createClientMethods(Operation operation, boolean isPr
                     .requiredNullableParameterExpressions(requiredParameterExpressions)
                     .validateExpressions(validateExpressions)
                     .methodTransformationDetails(methodTransformationDetails)
+                    .methodVisibilityInWrapperClient(isProtocolMethod && operation.isGenerateProtocolApi() != null && operation.isGenerateProtocolApi() == Boolean.FALSE ? JavaVisibility.PackagePrivate : JavaVisibility.Public)
```
weidongxu-microsoft, line 286: I think `operation.isGenerateProtocolApi() == Boolean.FALSE` covers `!= null`?

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
haolingdong-msft: Yes, should be covered, just adding `!=null` to make sure we consider the case where `isGenerateProtocolApi` not set is covered. I can remove that.

weidongxu-microsoft: Yes, please remove that.

Issue #1958, opened 2023-02-08
tsp, support multiple content types


weidongxu-microsoft:
epic: https://github.com/Azure/cadl-azure/issues/1943

weidongxu-microsoft:
pending on Cadl source

haolingdong-msft:
I wrote a draft design [doc](https://github.com/haolingdong-msft/working-docs/blob/main/cadl/multi-content-type.md) on it. I first list different user cases with the proposed generated SDK. Once we agree on what should SDK look like, I will start design on the codegen implementation part. Please take a look and let me know your thoughts.
/cc @srnagar @weidongxu-microsoft @XiaofeiCao

weidongxu-microsoft:
### Case 1

For multi content-type without overload, inclined to only generate protocol API.
In swagger, there is no way to properly define what method signature it want to generate, hence it is kind of reasonable for modelerfour to make up a solution (likely sperate JSON / text / binary types).

Given now Cadl provide the "overload", we'd expect user use that.

Multi content-type is not a common case, I don't think adding a few "overload"s a big effort on service (and it applies to all SDKs, not only Java).

### Case 2

I think SDK should not use `byte[]` in customer-facing API. It should still be `upload(BinaryData data, ContentType contentType)`.

Agree that we'd need some consensus (maybe across language) on what "overload" is allowed.
Probably not want to see this https://gist.github.com/msyyc/87a79b917deec25639f9f10d20589e55?permalink_comment_id=4472749#gistcomment-4472749

haolingdong-msft:
Metting notes at 2023-02-22:

Case 1:

This case is not valid. This is agreed by all languages' archs. We will have lint rule later to prevent the definition which defines both union types in body and union types in header. User should define overload if they define body type as union and header type as union.
Before that, we can throw error when finding this case in emitter or use option1 to generate protocol method only.

Case 2:
- `uploadStringOrResource` is not valid, similar to case1, should prevent in the linter.
- We agree on the overload methods should be the same name.
- For the body types that do not define overload method, we will not generate convenient methods for them, in this case, `Resource` type with `application/json`. Later we can also consider a lint rule to make sure all the types in the union body have defined overload operations.
- For overload with same signature case, we will generate two overload methods with compilation error.
- We should use `BinaryData` instead of `byte[]`

Other cases:

[...]

PR #2007, opened 2023-03-03
cadl, support multiple content types - add test case first

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/1958

First add two cases, to verify existing behavior.
1. data type is not union, contentType is union.
2. data type is union, contentType is union, with overloads to define the mapping.

Other cases worthwhile to add:
- multi content type with LRO and paging
- multi content type with @convenientApi and @protocolApi
- shared route


weidongxu-microsoft:
The protocol method seems good. But please still do a verification of make REST call with them and inspect the payload.

The overload part is a bit unexpected, need more info on how emitter handles it and how we had/should handle afterwards.

We might need to check with Srikanta whether we should map all Cadl bytes to `BinaryData` (instead of `byte[]`).

You can actually set convenientAPI=false in Cadl source and check it in, if the protocol method is good.

weidongxu-microsoft:
though you probably need to handle typespec upgrade/rename first...

haolingdong-msft:
For the issues you mentioned above, I have below thoughts, please let me know your comments/suggestions.

1. map `byte[]` to `BinaryData`:
I think we can map for multiple content types case to `BinaryData` only, e.g. emitter can set the schema type as `binary` in code-model.yaml like below. We can then add logic to set return type as `BinaryData` for [`Schema.AllSchemaTypes.BINARY`](https://github.com/Azure/autorest.java/blob/main/extension-base/src/main/java/com/azure/autorest/extension/base/model/codemodel/Schema.java#L276) type in [CadlPrimitiveMapper.java](https://github.com/Azure/autorest.java/blob/main/cadl-extension/src/main/java/com/azure/cadl/mapper/CadlPrimitiveMapper.java#L22)
image

2. expect upload in proxy method, but we got 3:
I think we can add a flag in `code-model.yaml` called `overload: boolean` to indicate the operation is overload operation or not. If it is overload operation, in codegen, we don't need to generate protocol method and proxy method for it.
```yaml
...
generateProtocolApi: true
overload: true
```

weidongxu-microsoft:
1.
That might do, does m4 do this to request schema?

2.
Probably further discussion is needed. `overload` won't tell you which API to override.
Alternative could be customize the convenicenceAPI.requests (it can put multiple requests as different overload to convenience API)

haolingdong-msft:
1. M4 has `requestMediaTypes`. This is an example m4 output:
image
image


2. I understand your alternative. The structure of code-model would better describe the relationship between major operation and overload operations. I will think more of this alternative.
What I try to propose previously is: If we have 1 major operation with 3 overload operations. The compiler will simply return 4 operations regardless it's overload or not. That's why we generate both protocol methods and convenient methods for the 4 operations. I'm thinking if we could add `overload: true` to the 3 overload operations in `code-model.yaml`. So the codegen can disable generation for proxy methods and protocol methods for those operations. We will also put name (in this case `upload`) of the operation into `convenicenceAPI`, so the convenient method name will be changed.

weidongxu-microsoft:
The problem is when you disable the proxy method and protocol method, the previous logic of convenience method calling corresponding protocol method will likely won't work (of course, even if we reuse it, it might still need some enhancement). So this might result in you writing a new logic of convenience method calling whatever, for overload.

weidongxu-microsoft:
For 1, you can first check/fix in in a new PR, if this happens to single content-type=octect-stream as well.

cadl-extension/src/code-model-builder.ts
```diff
@@ -701,7 +701,16 @@ export class CodeModelBuilder {
...
+    let schema = this.processSchema(body.kind === "Model" ? body : body.type, body.name);
+    // handle for binary body
+    if (
+      parameters.properties.has("contentType") &&
+      body.kind === "ModelProperty" &&
+      body.type.kind === "Scalar" &&
+      body.type.name === "bytes"
+    ) {
+      schema = new BinarySchema(body.name);
+    }
```
weidongxu-microsoft, line 713: 1. should be an `if else` for the schema
2. do we need to test "contentType", probably just drop it (and even if we need to test, "contentType" is not a wire name; also then we would need to test again "application/json" or anything relates to JSON, as it is totally valid for user to write `content-type: application/json` even it is not necessary)?

weidongxu-microsoft: If user uses `application/json` and implicit body (as we most use here), the body would be `kind=Model`

(want to make sure we still can distinguish `@body data: bytes`, from `data: bytes` in an implicit Model for JSON)

cadl-extension/src/main/java/com/azure/cadl/mapper/CadlPrimitiveMapper.java
```diff
@@ -21,6 +21,8 @@ public static PrimitiveMapper getInstance() {
     protected IType createPrimitiveType(PrimitiveSchema primaryType) {
         if (primaryType.getType() == Schema.AllSchemaTypes.DATE) {
             return ClassType.LocalDate;
+        } else if (primaryType.getType() == Schema.AllSchemaTypes.BINARY){
+            return ClassType.BinaryData;
```
weidongxu-microsoft, line 25: `Schema.AllSchemaTypes.BINARY` seems not processed in `PrimitiveMapper`. How does it handle binary date from m4?

haolingdong-msft: This is M4 example, it has `requestMediaTypes` on each operation, `requestMediaTypes` will map different content types to different input schema.
image
image
image

It is processed here: https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/mapper/MethodGroupMapper.java#L89

weidongxu-microsoft: How it process that, I cannot find code `requestMediaTypes` in the whole project?

weidongxu-microsoft: E.g. in vanilla mediatypes test, it has
```java
    public String analyzeBody(ContentType contentType, Flux<ByteBuffer> input, Long contentLength) {
        return analyzeBodyWithResponse(contentType, input, contentLength, Context.NONE).getValue();
    }
```

Ignore the `contentLength` (DPG does not have it), it has the `Flux<ByteBuffer> input`. I assumed something mapped to `Flux<ByteBuffer>`?

weidongxu-microsoft: Also, we are talking about single content-type with binary upload.

We can avoid discussion of multiple content-type for now.

weidongxu-microsoft: > image requests refer to the item in requestMediaTypes, we proccess requests.

I see. But still this question. How vanilla process this without mapping in `PrimitiveMapper`? Your code is DPG only (which map any payload to BinaryData)

haolingdong-msft: This is where we map `BinarySchema` to `FluxByteBuffer`
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/mapper/BinaryMapper.java#L28

call stack:
image

weidongxu-microsoft: So this change here is not needed?

haolingdong-msft: I think so..I'm testing..

cadl-extension/src/code-model-builder.ts
weidongxu-microsoft: Use `if else`. Don't process schema you do not use.

cadl-extension/src/code-model-builder.ts
```diff
@@ -701,7 +701,13 @@ export class CodeModelBuilder {
   }
 
   private processParameterBody(op: CodeModelOperation, body: ModelProperty | Model, parameters: Model) {
-    const schema = this.processSchema(body.kind === "Model" ? body : body.type, body.name);
+    let schema: Schema;
+    if (body.kind === "ModelProperty" && body.type.kind === "Scalar" && body.type.name === "bytes") {
+      //handle for binary body
+      schema = new BinarySchema(body.name);
```
weidongxu-microsoft, line 707: Ideally this should be a `processBinary` which will add it to `codeModel.schemas` (that's the side-effect)
https://github.com/Azure/autorest.java/blob/54b1738de6d48b23a298fda4581198c2b3af2815/cadl-extension/src/code-model-builder.ts#L1142-L1148

Recommend to do this.

weidongxu-microsoft (approved):

Issue #2008, opened 2023-03-03
Cadl, validate for TraitOverride<VersionParameterTrait<ApiVersionPathParameter>>


weidongxu-microsoft:
cadl-ranch PR https://github.com/Azure/cadl-ranch/pull/214

Java code seems not correct

```java
        @Delete("/azure/traits/api/{apiVersion}/user/{id}")
        @ExpectedResponses({204})
        @UnexpectedResponseExceptionType(
                value = ClientAuthenticationException.class,
                code = {401})
        @UnexpectedResponseExceptionType(
                value = ResourceNotFoundException.class,
                code = {404})
        @UnexpectedResponseExceptionType(
                value = ResourceModifiedException.class,
                code = {409})
        @UnexpectedResponseExceptionType(HttpResponseException.class)
        Mono<Response<Void>> delete(
                @QueryParam("api-version") String apiVersion,
                @PathParam("id") int id,
                @HeaderParam("accept") String accept,
                RequestOptions requestOptions,
                Context context);
```

If in path, param should be `@PathParam("apiVersion") String apiVersion`

weidongxu-microsoft:
https://github.com/Azure/autorest.java/pull/2009

PR #2009, opened 2023-03-03
cadl, update test for core/traits

weidongxu-microsoft:
one test disabled, not sure whether `apiVersion` is the cause, or something else.


cadl-tests/src/main/java/com/_specs_/azure/traits/implementation/TraitsClientImpl.java
```diff
@@ -0,0 +1,314 @@
...
+        @UnexpectedResponseExceptionType(
+                value = ResourceModifiedException.class,
+                code = {409})
+        @UnexpectedResponseExceptionType(HttpResponseException.class)
+        Mono<Response<Void>> delete(
+                @QueryParam("api-version") String apiVersion,
+                @PathParam("id") int id,
+                @HeaderParam("accept") String accept,
+                RequestOptions requestOptions,
+                Context context);
```
weidongxu-microsoft, line 164: https://github.com/Azure/autorest.java/issues/2008

PR #2012, opened 2023-03-06
Cadl, fix for api-version in path via TraitOverride<VersionParameterTrait<>>

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2008

### DIFF
--- a/cadl-extension/src/code-model-builder.ts
+++ b/cadl-extension/src/code-model-builder.ts
@@ -211,26 +211,7 @@ export class CodeModelBuilder {
         let parameter;
 
         if (isApiVersion(this.dpgContext, it)) {
-          const schema = this.codeModel.schemas.add(
-            new ConstantSchema(it.name, "api-version", {
-              valueType: this.stringSchema,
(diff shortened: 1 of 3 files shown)


cadl-extension/src/code-model-builder.ts
```diff
@@ -1798,37 +1779,50 @@ export class CodeModelBuilder {
...
+      (this._apiVersionParameter = this.createApiVersionParameter("api-version", ParameterLocation.Query))
+    );
+  }
+
+  private _apiVersionParameterInPath?: Parameter;
+  get apiVersionParameterInPath(): Parameter {
+    return (
+      this._apiVersionParameterInPath ||
+      // TODO: hardcode as "apiVersion", as it is what we get from compiler
+      (this._apiVersionParameterInPath = this.createApiVersionParameter("apiVersion", ParameterLocation.Path))
```
weidongxu-microsoft, line 1825: hardcoded as "apiVersion" for now.

If emitter logic changes, code will fail the test and we shall change this.

Issue #2017, opened 2023-03-07
Update TypeSpec to 0.41 (including rename cadl to typespec)


haolingdong-msft:
1. @typespec/  from @cadl-lang
   compiler `0.41.0`
   rest
   openapi 
2. @azure-tools/ scope name unchanged
   Package names changed as follows:
     typespec-azure-xxx
     typespec-client-generator-core, namespace Azure.ClientGenerator.Core  from  cadl-dpg, 
3. SDK packages:  typespec-csharp|python   cadl->typespec
4. Extension:  .tsp   from .cadl,   .ts vs .tsp
5. cli: tsp   from cadl   tsc
6. Config file:  tspconfig.yaml   ->  cadl-project.yaml
7. Compiler exported artifacts:  17 xxTypeSpecXX
8. Back compat:
     .cadl, cadl-project.yaml, re-exported as xxCadlxxx but has @deprecate
9. @typespec/migrate is coming

haolingdong-msft:
Follow up issue on internal code rename: https://github.com/Azure/autorest.java/issues/2025

PR #2018, opened 2023-03-07
typespec, upgrade to typespec 0.41

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/2017

This pr includes:
1. update typespec and related dependencies to 0.41 version
2. update readme
3. update scripts
4. update cadl definitions and rename .cadl to .tsp
5. disable cadl ranch for now as cadl-ranch tests are not upgraded yet

Hi @weidongxu-microsoft @XiaofeiCao , as we rename .cadl to .tsp, it might cause the existing cadl vscode extension not highlight and navigate code for `.tsp` file, I installed tsp vscode extension again to make the highlight and code navigation work.

Furthur tasks on renaming: 
1. folder name change, e.g. cadl-output, cadl-extension etc.
2. cadl-java extension name change
3. internal source code to remove the term 'cadl'
4. automation related in sdk repo, e.g. cadl-location.yaml etc.


weidongxu-microsoft (commented):
LGTM

Can we disable the 'Git Diff' step in CI, instead of deleting all code/test from cadl-ranch? (likely also need to disable the clean-up part in script)

Would like to avoid deleting and re-creating these in repo. The least it disrupt the history.

cadl-tests/cadl/lro.tsp
```diff
@@ -15,12 +15,13 @@ using Azure.Core.Traits;
...
 namespace Cadl.LongRunning;
 
+@resource("resources")
 model Resource {
   @visibility("read")
   id: string;
 
   @key
-  @segment("resources")
+  @visibility("create", "update")
```
weidongxu-microsoft, line 24: is `segement` be deprecated?

weidongxu-microsoft: I thought previous `@segment` actually means `name` is read-only in JSON.

It returns in response, but not settable in request (as name is set to path, not JSON).

haolingdong-msft: `@segment` will be inserted into the operation route before the path parameter's name field.
https://microsoft.github.io/typespec/standard-library/rest/reference/decorators#@TypeSpec.Rest.segment
`@segment` is not deprecated.

tsp compiler throw error when a property defines a key, but does not define visiblity. 
It also reports error if we don't set `@segment` or `@resource`. But it recommends using `@resource` in error message.
```
Model type 'Cadl.LongRunning.Resource' is not a valid resource type. It must be decorated with the '@resource' decorator.
```
[playground](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvcmVzdCI7DQrJGmF6dXJlLXRvb2xzL8gmLcYVY29yZcQsDQp1c2luZyBUeXBlU3BlYy5IdHRwO9EWUmVzdNIWVmVyc2lvbmluZ8kcQcReLkNvcmXTEy5Gb3VuZGF0aW9uc9QfVHJhaXTEGg0KQHNlcnZpY2Uoew0KICB0aXRsZTogIkxvbmdSdW7EciIsxBl25gCCOiAiMS4wLjDEFX0pDQpAdXNlRGVwZW5kZW5jeSjLY%2BcAs3MudjFfMF9QcmV2aWV3XzIpDQpuYW1lc3BhY2UgQ2FkbC7La%2BUAkQ0KbW9kZWwgUmVzb3VyY2Ug5QCZQHZpc2liaWxpdHkoInJlYWQiKcQXaWQ6IHN0csg8ICBAa2V50TBjcmVhdGUiLCAidXBkxArFPOQAiMs%2BICDkAbrLEX3qAIxFeOQB6VBhcmFtc%2BcAkHF1ZXLFanByb2plY3RGaWxl5wD03ERlZO4A2O0AwSAgcscdVXJp0EZpbnRlcmbkATfIQk9wZXLmAeTEO2V4dGVuZHPsAefSKTxOb1JlcGVhdGFibGVSZXF1ZXN0cyAmIE5vQ29uZGnEJGFszhhsaWVudMcTSWQ%2BIHvlAI5Acm91dGUoIi9sb25nLXLnAisp7ACn6wHZT3DmAO7mAYlPclXlAYcgaXPzAMzpALZDzS08yBc%2B6AHqcG9sbMVcxzoo00xHZXTRFlN0YXR1c8lQLCBuZXZlcj7lAhToAKFSZXBs5ADE9gCi6wDa8ACtxznxAK5nZXTXTcgTUmVhZP8A5f8A5f8A5ekA5WRlbGX5AX7zANxExTD%2FAJz%2FAJz9AJzwAxvmAKdlxRb%2FAKfrAKdBY8Qb0VfmA7bTZeQDWQ%3D%3D)

weidongxu-microsoft: OK, what about "read" vs. "create", "update"?

haolingdong-msft: Yeah, in this case `read` would be better.

weidongxu-microsoft: OK, could you use `read`?

Wondering why this does not affect generated code>?

haolingdong-msft: I will update.

weidongxu-microsoft (approved):

Issue #2010, opened 2023-03-03
[dpg] Compile error when generating DPG client enabling sync stack 


samvaity:
Ref: https://github.com/Azure/azure-sdk-for-java/pull/33719/files#r1123840809

weidongxu-microsoft:
We've fixed a few issues in DPG with sync stack mode https://github.com/Azure/autorest.java/pull/2001

But we haven't released a autorest.java with the fix. Plan to release one after azure-core/identity releases in March.

We can regen the purview share with it to see how it works.

weidongxu-microsoft:
@samvaity 

One question, why they have models package?

kevinmichaelbowersox:
It was recommended that we create models for the samples/tests.

srnagar:
@kevinmichaelbowersox the models should not be included in the package we ship. The models should be created manually to show how to test/use `BinaryData.fromObject()` in `src/samples` directory (not `src/main`).

weidongxu-microsoft:
@kevinmichaelbowersox 
If you need to ship the models, the minimal requirement is that they need to be reviewed as well. One quick look can see some of them having naming problems, e.g. AdlsGen2AccountSink.

weidongxu-microsoft:
@samvaity 

4.1.15 released

Do you need me to regen for purview sharing?

samvaity:
I think @kevinmichaelbowersox is taking care of that here - https://github.com/Azure/azure-sdk-for-java/pull/33878

samvaity:
> @samvaity
> 
> 4.1.15 released
> 
> Do you need me to regen for purview sharing?

Yes, I think we should do that and check sync stack generation works fine for DPG clients.

weidongxu-microsoft:
still got one in catalog
https://dev.azure.com/azure-sdk/public/_build/results?buildId=2614036&view=logs&jobId=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=ec247c96-ff2a-56b3-43b7-8e8a4c52422f
(test PR https://github.com/Azure/azure-sdk-for-java/pull/33921/checks?check_run_id=11841197114)

```java
        @Post("/glossary/{glossaryGuid}/terms/export")
        @ExpectedResponses({200})
        @UnexpectedResponseExceptionType(
                value = ClientAuthenticationException.class,
                code = {401})
        @UnexpectedResponseExceptionType(
[...]

PR #2022, opened 2023-03-09
dpg, bug fix on SchemaResponse with binary

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2010

`produce: text/csv` makes the operation not having a `BinaryResponse`. It would be a `SchemaResponse` with schema being binary (bytes).

Previous Mapper maps it to `GenericType.FluxByteBuffer`, and this type is not properly mapped to `BinaryData` in DPG case.


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -560,7 +560,10 @@ private ReturnTypeHolder getReturnTypes(Operation operation, boolean isProtocolM
 
         IType responseBodyType = SchemaUtil.getOperationResponseType(operation, settings);
         if (isProtocolMethod) {
-            if (responseBodyType instanceof ClassType || responseBodyType instanceof ListType || responseBodyType instanceof MapType) {
+            if (responseBodyType instanceof ClassType
+                    || responseBodyType instanceof ListType
+                    || responseBodyType instanceof MapType
+                    || responseBodyType == GenericType.FluxByteBuffer) {
                 responseBodyType = ClassType.BinaryData;
```
weidongxu-microsoft, line 567: `SchemaUtil.getOperationResponseType` handles `BinaryResponse`.

But m4 does not use `BinaryResponse` for `text/csv`.

PR #2031, opened 2023-03-13
Update SnakeYaml

alzimmermsft:
Update to `snakeyaml` `2.0`


javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -155,6 +156,7 @@ protected NodeTuple representJavaBeanProperty(Object javaBean, Property property
         loaderOptions.setCodePointLimit(50 * 1024 * 1024);
         loaderOptions.setMaxAliasesForCollections(Integer.MAX_VALUE);
         loaderOptions.setNestingDepthLimit(Integer.MAX_VALUE);
+        loaderOptions.setTagInspector(new TrustedTagInspector());
```
alzimmermsft, line 159: There is another type of `TagInspector` that allows for trusted prefixes to be used, should determine if that should be used instead of this.

weidongxu-microsoft (approved):
Thanks.

Seems non-trival?

PR #2033, opened 2023-03-14
mgmt, bug fix on proxy method default Exception

weidongxu-microsoft:
https://github.com/Azure/autorest.java/pull/2033/commits/871a245fd438c7bfaa5f2a1b786b2f3f13ef5064

The case when swagger write 404 without schema.

data-plane and mgmt-plane would fallback to different Exception class.

---

Also disable `fixOverlongStringLiteral` (was for overlong string like URL)


weidongxu-microsoft:
btw that containerregistry changed again https://github.com/Azure/autorest.java/pull/2033/commits/ea285d0d2327100f7370e2065b4de3db26fe1a1b

Issue #1876, opened 2022-12-05
cadl, multi-client on different api-version or different server


weidongxu-microsoft:
Plan is move version / url / host param to client level (avoid global level).

Version could be a problem, need to determine whether different client shared a same version (or `@versioned` list). So this would be likely the last to solve.

weidongxu-microsoft:
server done in https://github.com/Azure/autorest.java/pull/1982

PR #2034, opened 2023-03-14
Cadl, support ServiceVersion per Client

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1876

May not be the final solution. We've not handles the change from same version in clients to different version in different client, or vice versa.

Current logic is if all client has same api-version (the usual case), go with existing logic of a single ServiceVersion class.

If multiple sets of api-version exists, use ServiceVersion per Client.

---

For those tests that api-version is not used, now it will not generate ServiceVersion class.


cadl-tests/cadl/multiple-apiversion.tsp
weidongxu-microsoft, line 1: the new test for client on different sets of api-version

PR #2035, opened 2023-03-15
mgmt, remove code no longer needed

weidongxu-microsoft:
mgmt no longer need this after removing JsonFlatten


weidongxu-microsoft:
still not getting the autorest.java run...

PR #2046, opened 2023-03-28
Log warning for unsupported auth options

srnagar:


cadl-extension/src/code-model-builder.ts
```diff
@@ -285,6 +285,10 @@ export class CodeModelBuilder {
               securitySchemes.push(keyScheme);
             }
             break;
+          case "http":
+          {
+            this.logWarning(scheme.scheme + " auth method is currently not supported.");
+          }
```
haolingdong-msft, line 291: do we need to add `break;` at the end?

srnagar: Added `break` for future proofing this when we add more cases below `http`.

haolingdong-msft (approved):

PR #2048, opened 2023-03-29
tsp, rename folder

weidongxu-microsoft:
acr changed again, regening tests. https://github.com/Azure/azure-sdk-for-java/pull/34070


eng/pipelines/ci-cadl-java.yaml
```diff
@@ -71,14 +71,14 @@ jobs:
       - script: |
           node node_modules/\@azure-tools/cadl-ranch/dist/cli/cli.js serve node_modules/\@azure-tools/cadl-ranch-specs/http/ --coverageFile ./cadl-ranch-coverage-java.json &
```
haolingdong-msft, line 72: do we want do change the file name to ci-typespec-java.yaml?

typespec-extension/src/main/java/com/azure/cadl/Main.java
```diff
@@ -50,11 +50,11 @@ public class Main {
                 .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
     }
```
haolingdong-msft, line 51: Do we want to change the namespace to `package com.azure.typespec;`?

haolingdong-msft (approved):

typespec-extension/src/main/java/com/azure/cadl/Main.java
weidongxu-microsoft: java namespace currently is not exposed anywhere, so no change for now (include the name of `CadlPlugin`)

eng/pipelines/ci-cadl-java.yaml
weidongxu-microsoft: renamed, hope CI still works

PR #2049, opened 2023-03-30
Rename typespec maven profile

srnagar:


weidongxu-microsoft (commented):
nit, not sure if "tsp" be simpler (less type when run mvn)

There is a few other places (github action) as well
https://github.com/Azure/autorest.java/blob/main/.github/workflows/pre-release-typespec.yml#L36
https://github.com/Azure/autorest.java/blob/main/.github/workflows/generate-test-code.yml#L37

However not sure why CI fails.

weidongxu-microsoft (approved):

srnagar:
> nit, not sure if "tsp" be simpler (less type when run mvn)

We had not used `tsp` in any other place. So, I went with the full name but since we'll use this multiple times to build locally, it might be easier to just use `tsp`. Updated the PR.

PR #2050, opened 2023-03-30
tsp, test case for literal

weidongxu-microsoft:
I don't expect optional Literal get used. Even Literal is pretty rare, and mostly only make sense as discriminator value (parameter value may make some sense, e.g. "/resources/{default}" where only "default" is valid).


typespec-tests/src/main/java/com/cadl/literalservice/models/OptionalLiteral.java
weidongxu-microsoft, line 1: There is pending question that need to be answered in tsgc, like what class name should it use.

typespec-tests/src/main/java/com/cadl/literalservice/models/Model.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Get the literal property: The literal property.
+     *
+     * @return the literal value.
+     */
+    public String getLiteral() {
+        return this.literal;
+    }
```
weidongxu-microsoft, line 37: In client, we may need to later refine whether this should be an `Enum Literal` (instead of a String).

PR #2062, opened 2023-04-05
Attempt to fix TypeSpec-Java dependencies

alzimmermsft:
While attempting to generate TypeSpec with `typespec-java` it's possible to have missing imports due to the dependencies used by `typespec-java`. This is updating the dependencies to be similar to what [dotnet does](https://github.com/Azure/autorest.csharp/blob/feature/v3/src/CADL.Extension/Emitter.Csharp/package.json).

### DIFF
--- a/typespec-extension/package.json
+++ b/typespec-extension/package.json
@@ -44,15 +44,20 @@
     "!dist/test/**",
     "target/azure-typespec-extension-jar-with-dependencies.jar"
   ],
-  "peerDependencies": {
-    "@typespec/compiler": "~0.42.0",
-    "@typespec/rest": "~0.42.0",
-    "@typespec/versioning": "~0.42.0",
-    "@azure-tools/typespec-azure-core": "~0.28.0",
-    "@azure-tools/typespec-client-generator-core": "~0.28.0"
-  },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


typespec-extension/package.json
```diff
@@ -44,15 +44,20 @@
...
-  "peerDependencies": {
-    "@typespec/compiler": "~0.42.0",
-    "@typespec/rest": "~0.42.0",
-    "@typespec/versioning": "~0.42.0",
-    "@azure-tools/typespec-azure-core": "~0.28.0",
-    "@azure-tools/typespec-client-generator-core": "~0.28.0"
-  },
   "dependencies": {
     "@autorest/codemodel": "~4.19.2",
+    "@autorest/java": "4.1.16",
```
haolingdong-msft, line 49: why do we add autorest/java? (I don't think we publish autorest/java to npm, please correct me if I am wrong.)

weidongxu-microsoft: I think the recommendation from Wes (maybe the outdated), is not adding unneeded dependency into the typespec-java lib.

We aware .NET has it, but it seems yet to be an agreement across emitters (I think python does not do this).

But anyway, not autorest/java. It maybe typespec-autorest etc.

PR #2084, opened 2023-04-18
partial-update, add static field test

haolingdong-msft:
Text translator has static fields generated, we add unit test on validating static fields


partial-update/src/test/resources/partialupdate/StringOperationWithAddedMemberGeneratedClient.java
```diff
@@ -11,6 +11,7 @@
 import com.azure.core.exception.HttpResponseException;
 import com.azure.core.http.rest.RequestOptions;
 import com.azure.core.http.rest.Response;
+import com.github.javaparser.ast.Generated;
```
weidongxu-microsoft, line 14: ???

weidongxu-microsoft (approved):

XiaofeiCao (approved):

PR #2093, opened 2023-04-20
Add Custom package-info Descriptions Configuration

alzimmermsft:
Adds a new Autorest configuration `custom-package-info-descriptions` which allows for a map of packageName-packageInfoDescription to be passed to code generation which overrides the default `package-info` javadoc descriptions.

Also fixes a bug with stream-style serialization for Unix long times and Unix datetime times where they didn't have serialization mappings configured.


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -161,7 +160,8 @@ public static JavaSettings getInstance() {
                 getBooleanValue(host, "no-custom-headers", false),
                 getBooleanValue(host, "include-read-only-in-constructor-args", false),
                 // setting the default as true as the Java design guideline recommends using String for URLs.
-                getBooleanValue(host, "url-as-string", true)
+                getBooleanValue(host, "url-as-string", true),
+                host.getValue(new TypeReference<Map<String, String>>() {}.getType(), "custom-package-info-descriptions")
```
srnagar, line 164: @alzimmermsft - can customization be used to update the package info javadocs?

We should solve the docs customization issue holistically. As of now, using customization is the recommended way.

weidongxu-microsoft (approved):

readme.md
```diff
@@ -94,6 +94,7 @@ Settings can be provided on the command line through `--name:value` or in a READ
 |`--required-fields-as-ctor-args`|Indicates that class models have constructor taking required properties.|
 |`--output-model-immutable`|Indicates that output-only models be generated as immutable, and without public constructor.|
 |`--use-input-stream-for-binary`|Indicates that `InputStream` is used for binary response body. By default, `BinaryData` is used.|
+|`--custom-package-info-descriptions`|Custom Javadoc descriptions for package-infos.|
```
weidongxu-microsoft, line 97: We may need a bit more description to tell user that the content of it be the dict of package name to package description.

extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
alzimmermsft: It can be done via code customizations but that can get confusing when a library uses multiple Swaggers as part of code generation.

At the same time, I do agree that this is more a temporary solution than a long-term solution. But, this is better than the provided hardcoded `package-info.java` javadoc descriptions and is a low cost add.

readme.md
alzimmermsft: I'll update the README to give an example of this similar to what was done for HTTP status code exception mapping.

alzimmermsft:
Closing as this won't be done and code customizations will be used instead. In the future there are plans to make this a formal part of code generation instead of a one-off configuration.

PR #2105, opened 2023-04-24
mgtm, fix ResourceTypeNormalization on subclass same name as superclass

weidongxu-microsoft:
This could be a risky change.

I will track the report after merge.
https://github.com/Azure/autorest.java/wiki/Fluent-Codegen-Report


fluentgen/src/test/java/com/azure/autorest/fluent/TestUtils.java
```diff
@@ -55,7 +55,7 @@ public static class MockFluentGen extends FluentGen {
             DEFAULT_SETTINGS.put("required-parameter-client-methods", true);
             DEFAULT_SETTINGS.put("generate-samples", true);
             DEFAULT_SETTINGS.put("generate-tests", true);
-            //DEFAULT_SETTINGS.put("client-flattened-annotation-target", "NONE");
+            DEFAULT_SETTINGS.put("client-flattened-annotation-target", "NONE");
```
weidongxu-microsoft, line 58: new record has client-flattened-annotation-target=NONE, so we can now enable it here as well.

fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/ResourceTypeNormalization.java
```diff
@@ -295,6 +291,35 @@ private static void adaptForParentSchema(ObjectSchema compositeType, ResourceTyp
 
             LOGGER.info("Change parent from '{}' to '{}', for '{}'", Utils.getJavaName(parentType), type.getClassName(), Utils.getJavaName(compositeType));
         }
+
+        if (type.getClassName().equals(Utils.getJavaName(compositeType))) {
```
weidongxu-microsoft, line 295: This is to handle that current model be `ProxyResource`, and its parent also be `ProxyResource`.

On ModelMapper, it would be regarded as "external model".

But when generating sample/test, the util will get into infinite loop on the parent class...

weidongxu-microsoft: One concern (which is not related to this change) is that what if this model is called `ProxyResource`, but having some property not in a standard `ProxyResource`.

PR #2107, opened 2023-04-25
tsp, bug fix on secret model property

weidongxu-microsoft:
when visibility does not contains "read".

This also apply to Swagger `x-ms-secret`.

---

I will probably merge it soon and release typespec-java 0.6.2 to unblock Azure Batch.

If there is defect or concern on this change, please do add comment, and I will fix ASAP (this PR or follow-up PR).


typespec-tests/src/main/java/com/cadl/visibility/models/RoundTripModel.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param name the name value to set.
+     * @param secretName the secretName value to set.
+     */
+    @JsonCreator
+    public RoundTripModel(
+            @JsonProperty(value = "name", required = true) String name,
+            @JsonProperty(value = "secretName") String secretName) {
+        this.name = name;
+        this.secretName = secretName;
+    }
```
weidongxu-microsoft, line 38: Most change on generated code happens here.

Option `required = true` in annotation is removed from secret property, so that Jackson will not report error when de-serialize from response (which does not contain the secret property).

javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
```diff
@@ -129,7 +136,7 @@ public ClientModelProperty map(Property property) {
         List<String> annotationArgumentList = new ArrayList<String>() {{
             add(String.format("value = \"%s\"", xmlParamName));
         }};
-        if (property.isRequired()) {
+        if (property.isRequired() && !propertyIsSecret) {
             annotationArgumentList.add("required = true");
         }
```
weidongxu-microsoft, line 141: Code happens here.

Note it apply to both Typespec and Swagger (from m4). Let me know if this be a concern for Swagger.

typespec-tests/tsp/visibility.tsp
```diff
@@ -18,6 +18,11 @@ model Dog {
   name: string;
 }
 
+model RoundTripModel {
+  name: string;
+  @visibility("create", "update") secretName: string;
```
weidongxu-microsoft, line 23: I assume this regard same as a `x-ms-secret` for Swagger (when visibility contains create or update, but not contain read).

typespec-extension/changelog.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 0.6.1 (2023-04-25)
```
XiaofeiCao, line 3: 0.6.2?

XiaofeiCao (approved):

PR #2108, opened 2023-04-25
Fix Stream Style Serialization Generation for Unix Times

alzimmermsft:
Fixes stream-style serialization generation for Unix times. They were missing configuration on how they should be handled and resulted in an exception being thrown about them not being supported when attempting to generate Key Vault Administration with stream-style serialization.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
@@ -170,7 +170,13 @@ public class ClassType implements IType {
         .build();
 
     public static final ClassType UnixTimeDateTime = new ClassType.Builder(false)
+        .defaultValueExpressionConverter(defaultValueExpression -> java.lang.String.format("OffsetDateTime.parse(\"%1$s\")", defaultValueExpression))
         .knownClass(java.time.OffsetDateTime.class)
+        .wrapSerializationWithObjectsToString(true)
+        .jsonDeserializationMethod("getNullable(nonNullReader -> OffsetDateTime.parse(nonNullReader.getString()))")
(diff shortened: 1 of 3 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
```diff
@@ -681,22 +691,26 @@ public ClassType build() {
...
-        put('\b', "\\b");
-        put('\n', "\\n");
-        put('\r', "\\r");
-        put('\f', "\\f");
-        put('\"', "\\\"");
-    }};
+    private static final String[] ESCAPE_REPLACEMENT;
+
+    static {
+        ESCAPE_REPLACEMENT = new String[256];
```
srnagar, line 697: Do we need to allocate 256 size array for only 7 values? Can we continue to use the map instead?

srnagar (approved):

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
alzimmermsft: Create a 256-index array makes this simpler, and even with the sparseness it won't be too large in memory and the performance improvement over `Map` is substantial.

srnagar: Quick benchmark for this confirms the perf improvements when using an array over map.

```
Benchmark                                       Mode  Cnt           Score   Error  Units
MapAndArrayBenchmarks.lookupArray              thrpt    2  3212549857.265          ops/s
MapAndArrayBenchmarks.lookupMap                thrpt    2   121003281.154          ops/s
```

PR #2109, opened 2023-04-26
tsp, add tests

weidongxu-microsoft:


typespec-tests/src/test/java/com/projection/projectedname/ProjectedNameClientTest.java
```diff
@@ -30,4 +30,14 @@ void languageProjection() {
...
         propertyClient.language(project);
     }
+
+    @Test
+    public void testClient() {
+
+        // method name be clientName
+        client.clientName();
+
+        // parameter name be clientName
```
XiaofeiCao, line 40:
```suggestion
        // parameter name be parameter
```

XiaofeiCao (approved):

typespec-tests/src/test/java/com/projection/projectedname/ProjectedNameClientTest.java
weidongxu-microsoft: I mean this
https://github.com/Azure/autorest.java/blob/dee69aa0273b484289ad9718d6907bcf43172fb3/typespec-tests/src/main/java/com/projection/projectedname/ProjectedNameClient.java#L97

It not reflect in code, but the `String clientName`

PR #2110, opened 2023-04-26
Prepare Autorest JARs for release

alzimmermsft:


srnagar (approved):

typespec-tests/src/test/java/com/_specs_/azure/clientgenerator/core/internal/InternalTests.java
```diff
@@ -1,49 +0,0 @@
...
-import com.azure.core.http.rest.RequestOptions;
-import com.azure.core.http.rest.Response;
-import com.azure.core.util.BinaryData;
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Test;
-
-import java.util.HashMap;
-import java.util.Map;
-
-public class InternalTests {
```
weidongxu-microsoft, line 15: @XiaofeiCao 

I think Alan deleted it because there is large change on cadl-ranch test case on internal. Would you add it back later with the updated tests?

weidongxu-microsoft (approved):

typespec-tests/src/test/java/com/_specs_/azure/clientgenerator/core/internal/InternalTests.java
alzimmermsft: Yes, the case was what this was calling into changed, failed compile, and didn't have a replacement for the calls it was making before.

PR #2112, opened 2023-04-27
tsp, add test for resiliency case (that works without actually doing the resiliency)

weidongxu-microsoft:
We not actually support `@added` yet.

But the case of adding optional parameter to no param or only required param would work anyway...

### DIFF
--- a/typespec-tests/src/test/java/com/resiliency/servicedriven/ServiceDrivenTests.java
+++ b/typespec-tests/src/test/java/com/resiliency/servicedriven/ServiceDrivenTests.java
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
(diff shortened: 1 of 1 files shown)


typespec-tests/src/test/java/com/resiliency/servicedriven/ServiceDrivenTests.java
```diff
@@ -0,0 +1,44 @@
...
+        client2.fromOneRequired("required", "new");
+    }
+
+    // this case does not work without resiliency on "added"
+    // at present, v2 client does not have overload "fromOneOptional(String)"
+//    @Test
+//    public void testAddOptionalParamFromOneOptional() {
+//        client1.fromOneOptional("optional");
+//        client2.fromOneOptional("optional", "new");
+//    }
```
weidongxu-microsoft, line 38: This case require codegen actually handle `@added`

PR #2114, opened 2023-04-27
Guard file formatting

alzimmermsft:
Adds guards to prevent duplicative file formatting


partial-update-tests/generated/src/main/java/fixtures/bodystring/StringOperationClient.java
```diff
@@ -14,26 +15,11 @@
...
-
-    /**
-     * 1. manually add one constructor
-     *
-     * @param serviceClient the async client.
-     * @param test the parameter.
-     */
-    StringOperationClient(StringOperationsImpl serviceClient, String test) {
-        this.serviceClient = serviceClient;
-    }
```
weidongxu-microsoft, line 35: Wondering why the change in PR would affect partial-update test (with does not involve customization class).

+ @haolingdong-msft

weidongxu-microsoft (commented):
Also not sure why new files under azure-tests

Issue #2036, opened 2023-03-15
Add `@Generated` to models generated by TypeSpec


weidongxu-microsoft:
Add for all models (aka classes for JSON).

May include DPG, as it only generates model with `--generate-models` option.

weidongxu-microsoft:
Use a dedicated `addGenerated` method, like other Template.

We might need to do this only for TSP, if it interfere with vanilla and its customization via language server.

PR #2117, opened 2023-04-28
Add generated decorator to models and enums

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/2036

I added `@Generated` to Enums as well as enums are also under `models` package, please let me know if this is not needed.


javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -941,6 +957,18 @@ protected void writeStreamStyleSerialization(JavaClass classBlock, ClientModel m
...
+        if (JavaSettings.getInstance().isDataPlaneClient()) {
+            imports.add(Generated.class.getName());
+        }
+    }
+
+    protected void addGeneratedAnnotation(JavaContext classBlock) {
+        if (JavaSettings.getInstance().isDataPlaneClient()) {
+            classBlock.annotation("Generated");
+        }
+    }
```
weidongxu-microsoft, line 970: If we only do this for DPG, we might instead have a `TypeSpecModelMapper` to do the override in typespec-extension.

Wait for opinion from Srikanta and Alan.

vanilla-tests/src/main/java/fixtures/url/multi/Queries.java
```diff
@@ -64,7 +63,7 @@ Mono<Response<Void>> arrayStringMultiNull(
         @UnexpectedResponseExceptionType(ErrorException.class)
         Mono<Response<Void>> arrayStringMultiEmpty(
                 @HostParam("$host") String host,
-                @QueryParam(value = "arrayQuery", multipleQueryParams = true) List<String> arrayQuery,
+                @QueryParam("arrayQuery") String arrayQuery,
```
weidongxu-microsoft, line 66: why this changed?

weidongxu-microsoft (commented):
LGTM, except the change on vanilla-tests

When you ready, add Srikanta and Alan to reviewer.

Also plan to add a UT for simple customization on model (e.g. modify a getter).

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -693,6 +707,8 @@ private void addModelConstructor(ClientModel model, JavaVisibility constructorVi
         // Add the Javadocs for the constructor.
         classBlock.javadocComment(settings.getMaximumJavadocCommentWidth(), javadocCommentConsumer);
 
+
```
weidongxu-microsoft, line 710: avoid that empty new line

fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentModelTemplate.java
```diff
@@ -119,6 +120,14 @@ protected List<ClientModelPropertyReference> getClientModelPropertyReferences(Cl
...
     }
 
+    @Override
+    protected void addGeneratedImport(Set<String> imports) {
+    }
+
+    @Override
+    protected void addGeneratedAnnotation(JavaContext classBlock) {
+    }
+
```
weidongxu-microsoft, line 130: If you do this for Fluent, you probably also need to do this for Android.

vanilla-tests/src/main/java/fixtures/url/multi/Queries.java
haolingdong-msft: Thanks, I also notice this, am checking

weidongxu-microsoft (approved):

weidongxu-microsoft:
Please ping Srikanta for a quick check as well.

How's the UT? Is it work as expected (that we don't need any logic change on partial-update code)?

haolingdong-msft:
> Please ping Srikanta for a quick check as well.
> 
> How's the UT? Is it work as expected (that we don't need any logic change on partial-update code)?

Partial update code checks  `@Generated` annotation for a class, if annotation exists, it will do partial update, so we don't need to do code change. For UT, I did not add a UT to test customization on models currently, do we have example on how MVAD customization on models look like? I can add a UT to test it.

weidongxu-microsoft:
It may not in their current code.
A similar one would be https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/anomalydetector/azure-ai-anomalydetector/src/main/java/com/azure/ai/anomalydetector/models/MultivariateLastDetectionOptions.java
but with one required property and one optional property (not 2 required here).

Customization would involve adding a new required property (not JSON) to the first of ctor.

In summary, previous
```java
private required1;
private optional1;

public ctor(required1, optional1);

getter for 2
setter for 1
```

After customize
``java
private required0;
private required1;
private optional1;

public ctor(required0, required1, optional1);
ctor(required1, optional1);

getter for 3
setter for 1
```

So a new property and getter, a new ctor (hide the old ctor).

**Comment by haolingdong-msft** [comment]
> It may not in their current code. A similar one would be https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/anomalydetector/azure-ai-anomalydetector/src/main/java/com/azure/ai/anomalydetector/models/MultivariateLastDetectionOptions.java but with one required property and one optional property (not 2 required here). https://apiview.dev/Assemblies/Conversation/66bfc5cd130a4eb4a4e05b9d192c7e2b#3b0999fbf4bb4752af7fb60c958b129e
> 
> Customization would involve adding a new required property (e.g. `String modelId`, not JSON property) to the first of ctor.
> 
> In summary, previous
> 
> ```java
> private required1;
> private optional1;
> 
> public ctor(required1, optional1);
> 
> getter for 2
> setter for 1
> ```
> 
> After customize
> 
> ```java
> private required0;
> private required1;
> private optional1;
> 
> public ctor(required0, required1, optional1);
> ctor(required1, optional1);
> 
> getter for 3
> setter for 1
> ```
> 
> So a new property and getter, a new ctor (hide the old ctor).

Thanks! I will add a UT.

File: javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java [hunk 5]
```diff
@@ -195,4 +202,16 @@ protected String createEnumJsonCreatorIfCheck(EnumType enumType) {
             return String.format("item.%s().equals(value)", toJsonMethodName);
         }
     }
+
+    protected void addGeneratedImport(Set<String> imports) {
+        if (JavaSettings.getInstance().isDataPlaneClient()) {
```
srnagar, line 207: Do we need this check? We should enable partial-update for non-DPG clients as well. But if we don't have this supported for clients and client builders, we can defer this to later. Eventually, all DPG and non-DPG generations should support partial updates. 
(No change for this PR but likely a GitHub issue to follow up).

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -941,6 +957,18 @@ protected void writeStreamStyleSerialization(JavaClass classBlock, ClientModel m
...
 
+    protected void addGeneratedImport(Set<String> imports) {
+        if (JavaSettings.getInstance().isDataPlaneClient()) {
+            imports.add(Generated.class.getName());
+        }
+    }
+
+    protected void addGeneratedAnnotation(JavaContext classBlock) {
+        if (JavaSettings.getInstance().isDataPlaneClient()) {
+            classBlock.annotation("Generated");
```
srnagar, line 968: The string "Generated" is used in several places. We should have a constant instead.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
srnagar: This annotation should be extended to non-DPG and non-TypeSpec models too in the future. So, we should not make this TypeSpec specific.

partial-update/src/test/java/com/azure/autorest/partialupdate/util/PartialUpdateHandlerTest.java
```diff
@@ -189,6 +189,23 @@ public void testClassOrInterfaceFile_VerifyGeneratedFileShouldNotContainDuplicat
         assertTrue(actualMessage.contains(expectedMessage));
 
     }
+
+    @Test
+    public void testClassOrInterfaceFile_ModelCustomization() throws URISyntaxException, IOException {
```
srnagar, line 194: nit: we shouldn't use `_` in Java method names.

partial-update/src/test/resources/partialupdate/GeneratedModel.java
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com.azure.autorest.partialupdate.util;
+
+import com.azure.core.annotation.Generated;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public final class Model {
```
srnagar, line 11: Is this right? The file name is `GeneratedModel.java` but the class name is `Model`. The file name and classname should match.

protocol-tests/src/main/java/fixtures/dpgcustomization/models/ProductReceived.java
```diff
@@ -4,23 +4,25 @@
...
+import com.azure.core.annotation.Generated;
 import com.azure.core.util.ExpandableStringEnum;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import java.util.Collection;
 
 /** Defines values for ProductReceived. */
 public final class ProductReceived extends ExpandableStringEnum<ProductReceived> {
     /** Static value raw for ProductReceived. */
-    public static final ProductReceived RAW = fromString("raw");
+    @Generated public static final ProductReceived RAW = fromString("raw");
```
srnagar, line 15: We should have the annotation in a separate line. 

```suggestion
    @Generated 
    public static final ProductReceived RAW = fromString("raw");
```

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
```diff
@@ -47,6 +49,7 @@ private void writeExpandableStringEnum(EnumType enumType, JavaFile javaFile, Jav
         if (!settings.isStreamStyleSerialization()) {
             imports.add("com.fasterxml.jackson.annotation.JsonCreator");
         }
+        addGeneratedImport(imports);
```
srnagar, line 52: We should do the same for `writeEnum` method too in this class. We should add `@Generated` annotation to enum values and any public methods on the enum type.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -128,11 +130,14 @@ public final void write(ClientModel model, JavaFile javaFile) {
             if (model.getProperties().stream().anyMatch(ClientModelProperty::isAdditionalProperties)
                 && model.getNeedsFlatten()
                 && !settings.isStreamStyleSerialization()) {
+                addGeneratedAnnotation(classBlock);
```
srnagar, line 133: We should also add `@Generated` annotation in `ServiceVersionTemplate`.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
weidongxu-microsoft: Sure.

@alzimmermsft Do you have concern if `@Generated` annotation added to vanilla (non-DPG) as well (it might affect customization jar; I assume risk is low, but not able to say no risk).

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
haolingdong-msft: ok, I will add to enums as well. (Previously I thought this is to support models, so I only changes enums under `models` package. I did not change for enums.)

protocol-tests/src/main/java/fixtures/dpgcustomization/models/ProductReceived.java
haolingdong-msft: This is due to the java formatter, we have the same issue in the first partial update pr as well: https://github.com/Azure/autorest.java/pull/1273#discussion_r764423408

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
haolingdong-msft: Vanilla also supports @Generated for `clients` and `builders`,  I add this check after [discussing](https://github.com/Azure/autorest.java/issues/2036#issuecomment-1519567045) with Weidong to reduce the impact,I can remove this check and test in next pr.

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
weidongxu-microsoft: I assume Java `Enum` also generated under models package?

Though I've currently no idea how we are going to customize the `Enum` (but guess adding it won't hurt).

haolingdong-msft: Sorry I did not say it clearly, enums are generated in `models` package, but I did not see the usage in DPG/TypeSpec. We are using enums for `ServiceVersion`, it is not under model, so I did not add Generated annotation previously.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
alzimmermsft: No concerns from me with regards to the customization framework. There may be some customizations that run into minor issues but those would have likely failed due to any change in annotation generation which should be fixed anyways.

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
srnagar: This can be done in a separate PR but please file an issue to track this.

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
weidongxu-microsoft: Usage is that `@fixed` over tsp enum. We should have such case in both cadl-ranch test and local test.

We may add some method to the `Enum`, e.g. `getValueAsInt` that convert String on wire to Int to customer. But this is a very hypothesis use case.

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
haolingdong-msft: Issue created: https://github.com/Azure/autorest.java/issues/2130

androidgen/src/main/java/com/azure/autorest/android/template/AndroidEnumTemplate.java
```diff
@@ -29,7 +30,10 @@ protected void addGeneratedImport(Set<String> imports) {
     }
 
     @Override
-    protected void addGeneratedAnnotation(JavaContext classBlock) {
+    protected void addGeneratedAnnotationForExpandableStringEnum(JavaContext context) {
     }
 
+    @Override
+    protected void addGeneratedAnnotationForEnum(JavaEnum context) {
+    }
```
weidongxu-microsoft: nit, probably OK with same name as overload `addGeneratedAnnotation`.

haolingdong-msft:
```java
protected void addGeneratedAnnotationForExpandableStringEnum(JavaContext context) {
    }


  protected void addGeneratedAnnotationForEnum(JavaEnum context) {
  }
```
The two methods takes different input parameter type, so I add the suffix  to make it clearer. Do you mean we call the two methods `addGeneratedAnnotationForExpandableStringEnum` and `addGeneratedAnnotation`?

weidongxu-microsoft: I mean we probably call both `addGeneratedAnnotation`. (just different parameter)

Ideally I feel `JavaEnum` should also inherit from `JavaContext`, so it would be a single method. But no need to do it here.

weidongxu-microsoft:
What is the reason annotation cannot be add to Enum value?

I seems have no error on 
```java
public enum JavaModifier {
    @Generated

    Final("final"),
    @Generated

    Static("static"),
    @Generated
    Abstract("abstract");
```

haolingdong-msft:
> What is the reason annotation cannot be add to Enum value?
> 
> I seems have no error on
> 
> ```java
> public enum JavaModifier {
>     @Generated
> 
>     Final("final"),
>     @Generated
> 
>     Static("static"),
>     @Generated
>     Abstract("abstract");
> ```
> 
> (though you likely need to test partial-update on this case aka enum value, if annotation are required)

This is because to add annotation to enum value, we need to add the `addGeneratedAnnotation` logic into [`JavaEnum`](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/javamodel/JavaEnum.java#L63).  If we add the logic in [`EnumTempalte`](https://github.com/Azure/autorest.java/blob/3a05fd2874be1bca8757e1c90de440a75a828aed/javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java#L126), the generated code will be like below, it will 1. have @Generated annotation before java doc 2. have `;` insteand of `,`at the end of each value which causes compile error.
```java
public enum ContentType {
    @Generated
    /**
     * Enum value application/octet-stream.
     */
    APPLICATION_OCTET_STREAM("application/octet-stream");

    @Generated
    /**
     * Enum value image/jpeg.
     */
    IMAGE_JPEG("image/jpeg");

    @Generated
    /**
     * Enum value image/png.
     */
    IMAGE_PNG("image/png");
```
So I remove this to get CI pass first and want to cofirm whether we need to support partial update on enum values before updating `JavaEnum` class.

haolingdong-msft:
Hi @srnagar, could you please take a look at this comment and help suggest?https://github.com/Azure/autorest.java/pull/2117#issuecomment-1541463786

srnagar (approved):

srnagar:
> > What is the reason annotation cannot be add to Enum value?
> > I seems have no error on
> > ```java
> > public enum JavaModifier {
> >     @Generated
> > 
> >     Final("final"),
> >     @Generated
> > 
> >     Static("static"),
> >     @Generated
> >     Abstract("abstract");
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (though you likely need to test partial-update on this case aka enum value, if annotation are required)
> 
> This is because to add annotation to enum value, we need to add the `addGeneratedAnnotation` logic into [`JavaEnum`](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/javamodel/JavaEnum.java#L63).
> 
> If we add the logic in [`EnumTempalte`](https://github.com/Azure/autorest.java/blob/3a05fd2874be1bca8757e1c90de440a75a828aed/javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java#L126), the generated code will be like below, it will 1. have @generated annotation before java doc 2. have `;` insteand of `,`at the end of each value which causes compile error.
> 
> ```java
> public enum ContentType {
>     @Generated
>     /**
>      * Enum value application/octet-stream.
>      */
>     APPLICATION_OCTET_STREAM("application/octet-stream");
> 
>     @Generated
>     /**
>      * Enum value image/jpeg.
>      */
>     IMAGE_JPEG("image/jpeg");
> 
>     @Generated
>     /**
>      * Enum value image/png.
>      */
>     IMAGE_PNG("image/png");
> ```
> 
> So I remove annotation on enum value to get CI pass first and want to cofirm whether we need to support partial update on enum values before updating `JavaEnum` class.

For now, we can skip adding the `@Generated` annotation to enum if this is still an issue. We should still look into why this happening and we will need partial updates for enums too but not immediately needed.

weidongxu-microsoft:
If you go merge it, please add a 0.7.1 to changelog with current date.

haolingdong-msft:
> 
> For now, we can skip adding the `@Generated` annotation to enum if this is still an issue. We should still look into why this happening and we will need partial updates for enums too but not immediately needed.

Hi @srnagar , I listed why this is happening below.

**Generated enum:**
```java
public enum ContentType {
    @Generated
    /**
     * Enum value application/octet-stream.
     */
    APPLICATION_OCTET_STREAM("application/octet-stream");

    @Generated
    /**
     * Enum value image/jpeg.
     */
    IMAGE_JPEG("image/jpeg");

    @Generated
    /**
     * Enum value image/png.
     */
    IMAGE_PNG("image/png");
```
**The reason for the generation error is:**
 1. have `@generated` annotation before java doc. -> This can pass compile.
 
It is because to make the code more decoupled and consistent, it's better to add `@Generated` annotation in [`EnumTemplate.java`](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java#L117) like below, and `enumBlock.value()` in [JavaEnum.java](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/javamodel/JavaEnum.java#L57) contains code writing javadoc.
 
EnumTemplate.java
 ```java
for (ClientEnumValue value : enumType.getValues()) {
      addGeneratedAnnotation(enumBlock);
      enumBlock.value(value.getName(), value.getValue(), value.getDescription(), elementType);
  }
 ```
JavaEnum.java
```java
  public final void value(String name, String value) {
      addExpectedCommaAndNewLine();
      contents.javadocComment("Enum value " + value + ".");
      contents.text(name + "(\"" + value + "\")");
      previouslyAddedValue = true;
      addNewLine = true;
  }
```

I had a discussion with @weidongxu-microsoft on this, we thought if the compilation passes and partial update works, it's OK.
@srnagar Welcome your comments and suggestions.
If we want to solve this, we can have another overload API like this `public final void value(String name, String value, boolean addGeneratedAnnotation) {}` and add `@Generated` annotation inside `value()`.


 2. have `;` insteand of `,` at the end of each value which causes compile error.
 
It is because [`addGeneratedAnnotation(enumBlock);`](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/javamodel/JavaEnum.java#L110) will add semicolon at the end of the previous line

EnumTemplate.java
```java
    public final void annotation(String... annotations) {
        addExpectedSemicolonAndNewLine();
        contents.annotation(annotations);
    }
```

To solve this, I will delete [` addExpectedSemicolonAndNewLine();`](https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/javamodel/JavaEnum.java#L110) first to see if there is any test failures.

haolingdong-msft:
Hi @weidongxu-microsoft . @srnagar , I removed the partial update support for enums in this pr to avoid having incomplete logic, I would like to merget this pr first  and create a seperate pr to support partial update for enums as the logic is a bit complex and it would be clearer to look in a new pr. Please let me know if you have any concern.

PR #2127, opened 2023-05-08
tsp, fix emitter name

weidongxu-microsoft:


typespec-extension/src/emitter.ts
```diff
@@ -59,7 +59,7 @@ const EmitterOptionsSchema: JSONSchemaType<EmitterOptions> = {
 };
 
 export const $lib = createTypeSpecLibrary({
-  name: "JavaEmitter",
+  name: "@azure-tools/typespec-java",
```
haolingdong-msft, line 62: What is this change's impact?

weidongxu-microsoft: I've sent an email about this "FW: Steps to Add Nightly Pipeline Testing Latest Tsp Releases".

Change in (dev) compiler make this name matters now.

weidongxu-microsoft: Though nightly still fails (at a later place...)

Issue #2068, opened 2023-04-07
tsp, generate sample/test for convenient API


weidongxu-microsoft:

weidongxu-microsoft:
thought on impl

1. sample first
2. can temporary ignore spread/options (parameter flatten/grouping)
3. smaller PR to avoid conflict

XiaofeiCao:
First attempt would be create sample files for convenience methods(and client initialization).

Afterwards, fill in the files with method invocations.

XiaofeiCao:
Expected sample code after codegen finished:
https://github.com/Azure/azure-sdk-for-java/commit/a2ee043ffb48168141eed85d75845a035a5d6278

weidongxu-microsoft:
On next step, can you add some examples/json in our local test?

I don't have a good candidate. May be just the "list" op in "tsp/versioning.tsp".

weidongxu-microsoft:
Another pending one would be https://github.com/Azure/autorest.java/issues/2201

XiaofeiCao:
Start on working test generation. Similar to generated protocol tests, extend generated testBase and use example data as input and output verification.

PR #2129, opened 2023-05-08
tsp, generate convenience sample files

XiaofeiCao:
part of https://github.com/Azure/autorest.java/issues/2068

Generate sample file for convenience methods.

generate using mvad examples:
https://github.com/Azure/azure-sdk-for-java/commit/f58bdf49d611f82703a3df8545fc004242ebedae


javagen/src/main/java/com/azure/autorest/model/clientmodel/ConvenienceExample.java
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.azure.autorest.model.clientmodel;
+
+public class ConvenienceExample {
```
XiaofeiCao, line 6: Placeholder for `ConvenienceExample` for now. Will see if more parameters are needed.

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -327,13 +330,42 @@ public Client map(CodeModel codeModel) {
                 }
             };
 
-            syncClients.stream().filter(c -> c.getServiceClient() != null)
+            // protocol examples
+            syncClients.stream().filter(c -> c.getServiceClient() != null && CoreUtils.isNullOrEmpty(c.getConvenienceMethods()))
```
XiaofeiCao, line 334: Should we filter it in a per-method manner?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ConvenienceExample.java
weidongxu-microsoft: Maybe just MethodExample or ClientMethodExample?

We say protocol method example just because it is special. Convenience method is basically just client method in vanilla.

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
weidongxu-microsoft: Yeah, it could be a problem. We likely only want to generate example for client method with max overload. Previously it is done to filter for method with `Context`. But we didn't add this to convenience method.

XiaofeiCao: Got it. Maybe use the one with the longest parameters size?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ConvenienceExample.java
XiaofeiCao: Sounds good.

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -343,6 +318,102 @@ public Client map(CodeModel codeModel) {
...
+                    Set<String> convenienceProxyMethodNames = new HashSet<>();
+                    if (c.getConvenienceMethods() != null) {
+                        convenienceProxyMethodNames.addAll(c.getConvenienceMethods().stream()
+                                .map(convenienceMethod -> convenienceMethod
+                                        .getProtocolMethod().getProxyMethod().getBaseName())
+                                .collect(Collectors.toSet()));
+                    }
+                    c.getServiceClient().getClientMethods()
+                            .stream()
+                            .filter(m -> !convenienceProxyMethodNames.contains(m.getProxyMethod().getBaseName()))
```
XiaofeiCao, line 397: only generate protocol examples for protocol methods without convenience methods

javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -54,6 +55,8 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
+;
+
```
weidongxu-microsoft, line 59: ???

weidongxu-microsoft (commented):
what current output looks like?

XiaofeiCao:
> what current output looks like?

I used mvad's example files:
https://github.com/Azure/azure-rest-api-specs/tree/main/specification/cognitiveservices/AnomalyDetector/examples/v1.1

And generated samples:
https://github.com/Azure/azure-sdk-for-java/commit/f58bdf49d611f82703a3df8545fc004242ebedae

javagen/src/main/java/com/azure/autorest/Javagen.java
XiaofeiCao: removed...

XiaofeiCao:
Yes, I think the request model construction part can reuse that of fluent.

weidongxu-microsoft (approved):

PR #2133, opened 2023-05-10
convenience sample, write client initialization

XiaofeiCao:
part of https://github.com/Azure/autorest.java/issues/2068

Mainly refactor, extract client initialization code to a separate class.

generated samples:
https://github.com/Azure/azure-sdk-for-java/commit/c379b2c186d6c1e37232c0029bd812e66c603dc1

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
+++ b/javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
@@ -327,27 +327,28 @@ public class ClientMapper implements IMapper<CodeModel, Client> {
             ClientBuilder clientBuilder = c.getClientBuilder();
             if (clientBuilder != null && convenienceMethod.getProtocolMethod().getProxyMethod().getExamples() != null) {
                 // only generate sample for convenience methods with max overload parameters
-                convenienceMethod.getConvenienceMethods().stream().max((o1, o2) -> {
-                    int o1ParameterCount = o1.getOnlyRequiredParameters()
-                            ? o1.getMethodRequiredParameters().size()
-                            : o1.getMethodParameters().size();
-                    int o2ParameterCount = o2.getOnlyRequiredParameters()
-                            ? o2.getMethodRequiredParameters().size()
-                            : o2.getMethodParameters().size();
-                    return o1ParameterCount - o2ParameterCount;
-                }).ifPresent(clientMethod ->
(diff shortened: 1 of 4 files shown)


javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -327,27 +327,28 @@ private void addConvenienceExamples(Client.Builder builder, List<AsyncSyncClient
...
-                                    ClientMethodExample convenienceExample =
-                                            new ClientMethodExample(clientMethod, c, clientBuilder, filename, example);
-                                    convenienceExamples.add(convenienceExample);
-                                    convenienceExampleNameSet.add(filename);
-                                }
-                            }
-                        }));
+                convenienceMethod.getConvenienceMethods().stream()
+                        .filter(clientMethod -> Templates.getClientMethodSampleTemplate()
+                                .isExampleIncluded(clientMethod, convenienceMethod))
```
XiaofeiCao, line 332: bug fix.. filter out excluded convenience methods first

javagen/src/main/java/com/azure/autorest/template/example/ProtocolExampleWriter.java
```diff
@@ -203,64 +197,9 @@ public ProtocolExampleWriter(ProtocolExample protocolExample) {
...
-                            .defaultValueExpression(parameterValue.getObjectValue().toString());
-
-                    serviceClient.getProperties().stream().filter(p1 -> Objects.equals(p.getName(), p1.getName())).findFirst().ifPresent(serviceClientProperty -> {
-                        processedServiceClientProperties.add(serviceClientProperty);
-
-                        clientParameterLines.add(
-                                String.format(".%1$s(%2$s)", serviceClientProperty.getAccessorMethodSuffix(), clientValue));
-                    });
-                });
             }
```
XiaofeiCao, line 200: The biggest refactor here: 
Extracted handling of "proxy method parameters whose values come from client" out of the parameter value assignment loop. 

When the parameter is from client, I think we don't have to look into the method signatures to find it? The value should be set in client instead.

weidongxu-microsoft: Sounds correct.

weidongxu-microsoft (approved):

PR #2137, opened 2023-05-12
tsp, client sample generate client method invocation

XiaofeiCao:
Reused that of fluent.

Generated sample code:
https://github.com/Azure/azure-sdk-for-java/commit/87bbdf2735cd01d1f251913c5c33e132c24bba37


javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
```diff
@@ -230,6 +236,101 @@ public static Object getChildObjectValue(List<String> jsonPropertyNames, Object
...
+                    elements = Arrays.asList(value.split(Pattern.quote("|"), -1));
+                    break;
+                case TSV:
+                    elements = Arrays.asList(value.split(Pattern.quote("\t"), -1));
+                    break;
+                default:
+                    // TODO (weidxu): CollectionFormat.MULTI
+                    elements = Arrays.asList(value.split(Pattern.quote(","), -1));
+                    LOGGER.error("Parameter style '{}' is not supported, fallback to CSV", collectionFormat);
+                    break;
```
weidongxu-microsoft, line 322: Emm, why I say this?

OK, I assume I am not sure how service write the MULTI in example json. So current code assume it is "," separated values.

javagen/src/main/java/com/azure/autorest/template/example/ClientMethodExampleWriter.java
```diff
@@ -0,0 +1,71 @@
...
+                        ? String.format("%s %s = %s.%s(%s);",
+                        method.getReturnValue().getType(),
+                        "response",
+                        clientVarName,
+                        method.getName(),
+                        parameterInvocations)
+                        : String.format("%s.%s(%s);",
+                        clientVarName,
+                        method.getName(),
+                        parameterInvocations);
```
weidongxu-microsoft, line 55: This condition may be hard to read.

May be use `String.format("%s.%s(%s);"`, and decide whether to add the `%s %s = ` based on return type.

weidongxu-microsoft (approved):
Look good.

Maybe see if there is LRO case.
They have LRO but you mean need to copy the example json from their swagger folder
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/HealthInsights/healthinsights.oncophenotype/route.oncophenotype.tsp

javagen/src/main/java/com/azure/autorest/template/example/ClientMethodExampleWriter.java
XiaofeiCao: Changed.

weidongxu-microsoft:
may also take a try on https://github.com/Azure/azure-rest-api-specs/tree/main/specification/cognitiveservices/ContentSafety

XiaofeiCao:
generated sample code: https://github.com/Azure/azure-sdk-for-java/commit/be035de2d62dc698ea4ac520466085c64d63b41b

I had to make following change to the spec file to work: https://github.com/Azure/azure-rest-api-specs/commit/c89645fca0d8880e23e33650535e10b6d2f9e56e

Seems it's not LRO issue. It's their grouping of example body parameter `body` that's different from our client's parameter name: [beginInferCancerProfile](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/healthinsights/azure-health-insights-cancerprofiling/src/main/java/com/azure/health/insights/cancerprofiling/CancerProfilingClient.java#L175-L181)

I may need to fix this parameter mapping.

weidongxu-microsoft:
> generated sample code: [Azure/azure-sdk-for-java@be035de](https://github.com/Azure/azure-sdk-for-java/commit/be035de2d62dc698ea4ac520466085c64d63b41b)
> 
> I had to make following change to the spec file to work: [Azure/azure-rest-api-specs@c89645f](https://github.com/Azure/azure-rest-api-specs/commit/c89645fca0d8880e23e33650535e10b6d2f9e56e)
> 
> Seems it's not LRO issue. It's their grouping of example body parameter `body` that's different from our client's parameter name `oncoPhenotypeData`: [beginInferCancerProfile](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/healthinsights/azure-health-insights-cancerprofiling/src/main/java/com/azure/health/insights/cancerprofiling/CancerProfilingClient.java#L175-L181)
> 
> Same story with https://github.com/Azure/azure-rest-api-specs/tree/main/specification/cognitiveservices/ContentSafety. I may need to fix this parameter mapping.

No need to do that at present, I think it is agreed that the name in example JSON should be same of name of parameter (though I am not sure if service can control the parameter name in TSP case).

If all other is good, the PR should be good.

weidongxu-microsoft:
Maybe we can tried a hack in a follow-up PR to fallback the body parameter to `body`, if a prior attempt cannot find the e.g. `oncoPhenotypeData` property.

XiaofeiCao:
> Maybe we can tried a hack in a follow-up PR to fallback the body parameter to `body`, if a prior attempt cannot find the e.g. `oncoPhenotypeData` property.

Sure. 

Also ContentSafety has a case-insensitive operationId:
[`AddBlockItems` in example json](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/examples/2023-04-30-preview/AddBlockItems.json)
[`addBlockItems` in typespec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/routes.tsp#L68)

Guess we'll treat the example operation id as case-insensitive?

weidongxu-microsoft:
> > Maybe we can tried a hack in a follow-up PR to fallback the body parameter to `body`, if a prior attempt cannot find the e.g. `oncoPhenotypeData` property.
> 
> Sure.
> 
> Also ContentSafety has a case-insensitive operationId: [`AddBlockItems` in example json](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/examples/2023-04-30-preview/AddBlockItems.json) [`addBlockItems` in typespec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/routes.tsp#L68)
> 
> Guess we'll treat the example operation id as case-insensitive?

Maybe I forget to make the operationId Pascal case (the parsing of operationId currently is a bit different from typespec-autorest, because it uses openapi lib, we not use yet).

XiaofeiCao:
> Maybe I forget to make the operationId Pascal case (the parsing of operationId currently is a bit different from typespec-autorest, because it uses openapi lib, we not use yet).
> 
> Log an issue to me please. I guess I'd better just follow typespec-autorest.

typespec-autorest output is indeed the same with example file. Issue logged: https://github.com/Azure/autorest.java/issues/2145

XiaofeiCao:
local run `Generate.ps1` has no diff. Will skip CI.

PR #2138, opened 2023-05-15
client, extract duplicate code to prototypeAsLong and prototypeAsDouble

weidongxu-microsoft:
this is a refactor

there was some duplicated code when defining basic type or type based on basic type in PrimitiveType and ClassType.

- add a Builder to PrimitiveType
- add prototypeAsLong and prototypeAsDouble to both for type based on long and double
- fix issue about UnixTimeLong, it may still have bug since we not really have good test on it, may improve it if we see new test in tsp encode


javagen/src/main/java/com/azure/autorest/model/clientmodel/PrimitiveType.java
```diff
@@ -266,4 +297,92 @@ public java.lang.String xmlSerializationMethodCall(java.lang.String xmlWriterNam
     public String toString() {
         return getName();
     }
+
+    private static class Builder {
```
weidongxu-microsoft, line 301: maybe we should move clientType, convertTo/FromClient to builder as well. though they already in the class and not hard to find, so maybe later.

XiaofeiCao (approved):

PR #2142, opened 2023-05-16
tsp, upgrade cadl ranch

weidongxu-microsoft:


typespec-tests/src/test/java/com/encode/duration/EncodeDurationTests.java
```diff
@@ -0,0 +1,50 @@
...
+    @Test
+    public void testQuery() {
+        queryClient.defaultMethod(DAY40);
+
+        queryClient.floatSeconds(SECOND35);
+
+        queryClient.int32Seconds(SECOND36);
+
+        queryClient.iso8601(DAY40);
+    }
```
weidongxu-microsoft, line 35: we seems don't have it on query, will see if this is an easy fix.

typespec-extension/src/code-model-builder.ts
```diff
@@ -732,7 +732,7 @@ export class CodeModelBuilder {
         // utcDateTime in header maps to RFC 5322
         schema = this.processDateTimeSchema(param.param.type, param.param.name, true);
       } else {
-        schema = this.processSchema(param.param.type, param.param.name);
+        schema = this.processSchema(param.param, param.param.name);
```
weidongxu-microsoft, line 735: bug fix here, let it process ModelProperty

haolingdong-msft (approved):

Issue #2145, opened 2023-05-17
tsp, fix parsing of operationId to align with typespec-autorest


XiaofeiCao:
https://github.com/Azure/autorest.java/pull/2137#issuecomment-1550624322

[AddBlockItems in example json](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/examples/2023-04-30-preview/AddBlockItems.json)
[addBlockItems in typespec](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/ContentSafety/routes.tsp#L68)

Currently we parse the operationId as `TextBlocklists_addBlockItems`.
typespec-autorest output is `TextBlocklists_AddBlockItems`(same with example json).

PR #2146, opened 2023-05-17
tsp, use openapi, also use pascalCaseForOperationId

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2145

I think the cause is `pascalCaseForOperationId` vs `pascalCase`

### DIFF
--- a/typespec-extension/src/utils.ts
+++ b/typespec-extension/src/utils.ts
@@ -13,8 +13,6 @@ import {
   Type,
   Union,
   ignoreDiagnostics,
-  isGlobalNamespace,
-  isService,
   isTemplateDeclaration,
   isTemplateInstance,
   resolvePath,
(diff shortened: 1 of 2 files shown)


typespec-extension/src/utils.ts
```diff
@@ -147,21 +146,11 @@ export async function loadExamples(program: Program, options: EmitterOptions): P
   return operationExamplesMap;
 }
 
-export function resolveOperationId(program: Program, operation: Operation) {
-  // const explicitOperationId = getOperationId(program, operation);
-  // if (explicitOperationId) {
-  //   return explicitOperationId;
-  // }
```
weidongxu-microsoft, line 154: this would be handled by openapi lib, for `@operationId` though it is not recommended.

typespec-extension/src/utils.ts
```diff
@@ -137,7 +136,7 @@ export async function loadExamples(program: Program, options: EmitterOptions): P
     if (operationIdExamplesMap.size > 0) {
       const routes = service.operations;
       routes.forEach((it) => {
-        const operationId = pascalCase(resolveOperationId(program, it.operation));
+        const operationId = pascalCaseForOperationId(resolveOperationId(program, it.operation));
```
weidongxu-microsoft, line 139: This probably was the bug

XiaofeiCao (approved):

PR #2153, opened 2023-05-23
tsp, support "is Record<>"

weidongxu-microsoft:
tested locally with 

```ts
model ModelIsRecordUnknown is Record<unknown> {
  @doc("The name property")
  name: string;
}
```

treat is same as `model ModelIsRecordUnknown extends Record<unknown> { name }`

wait cadl-ranch for formal test https://github.com/Azure/cadl-ranch/pull/330


typespec-extension/src/code-model-builder.ts
```diff
@@ -1231,7 +1231,7 @@ export class CodeModelBuilder {
       case "Model":
         if (isArrayModelType(this.program, type)) {
           return this.processArraySchema(type, nameHint);
-        } else if (isRecordModelType(this.program, type)) {
+        } else if (isRecordModelType(this.program, type) && type.properties.size == 0) {
```
weidongxu-microsoft, line 1234: the "pure" Record...

typespec-extension/src/code-model-builder.ts
```diff
@@ -1598,7 +1598,14 @@ export class CodeModelBuilder {
             }
           });
         }
+      } else {
+        pushDistinct(objectSchema.parents.all, parentSchema);
       }
+    } else if (isRecordModelType(this.program, type)) {
```
haolingdong-msft, line 1604: Do we need pure Record check here?

weidongxu-microsoft: No, if it gets here (the `processObjectSchema`), it is already mixed.

pure one already handled above with `processObjectSchema`

weidongxu-microsoft:
Emm, cadl-ranch get updated. Refreshing.

typespec-tests/src/test/java/com/encode/duration/EncodeDurationTests.java
```diff
@@ -31,6 +33,8 @@ public void testQuery() {
         queryClient.int32Seconds(SECOND36);
 
         queryClient.iso8601(DAY40);
+
+//        queryClient.int32SecondsArray(Arrays.asList(SECOND36, Duration.ofSeconds(47)));
```
weidongxu-microsoft, line 37: there is bug on this one, will fix in a separate PR (as this one is not about duration).

Issue #1765, opened 2022-10-14
client, public EnvironmentsClientBuilder pipeline(HttpPipeline pipeline) {


weidongxu-microsoft:
https://github.com/Azure/azure-sdk-for-java/pull/31300#discussion_r994560745

for pipeline and httpClient we should update this to copy what the hand-written builders do with logging a debug message if the HttpPipeline, or HttpClient, is set to null after being configured as there is a chance that it is a bug in user code:

```
if (this.pipeline != null && pipeline == null) {
    logger.info("HttpPipeline is being set to 'null' when it was previously configured.");
}
```

PR #2155, opened 2023-05-23
client, log if null pipeline

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/1765


weidongxu-microsoft:
@alzimmermsft Let me know if the PR is OK.

PR #2156, opened 2023-05-24
tsp, support head boolean

weidongxu-microsoft:
Enhancement for response error.

"*" is regarded as error, model with `@error` is regarded as error e.g. https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/foundations.tsp#L113-L124
Else be normal response (non-error response).

Most for the case of HEAD with 200+404 response. Detail https://teams.microsoft.com/l/message/19:906c1efbbec54dc8949ac736633e6bdf@thread.skype/1684272172008?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=3e17dcb0-4257-4a30-b843-77f47f1d4121&parentMessageId=1684272172008&teamName=Azure%20SDK&channelName=TypeSpec%20(Cadl)%20Discussion%20%F0%9F%90%AE&createdTime=1684272172008

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/response/ResponseServiceVersion.java
+++ b/typespec-tests/src/main/java/com/cadl/response/ResponseServiceVersion.java
@@ -6,14 +6,14 @@ package com.cadl.response;
 
 import com.azure.core.util.ServiceVersion;
 
-/** Service version of CoreClient. */
-public enum CoreServiceVersion implements ServiceVersion {
+/** Service version of ResponseClient. */
+public enum ResponseServiceVersion implements ServiceVersion {
     /** Enum value 1.0.0. */
     V1_0_0("1.0.0");
 
     private final String version;
 
-    CoreServiceVersion(String version) {
+    ResponseServiceVersion(String version) {
         this.version = version;
     }
 
@@ -26,9 +26,9 @@ public enum CoreServiceVersion implements ServiceVersion {
     /**
      * Gets the latest service version supported by this client library.
      *
-     * @return The latest {@link CoreServiceVersion}.
+     * @return The latest {@link ResponseServiceVersion}.
      */
-    public static CoreServiceVersion getLatest() {
+    public static ResponseServiceVersion getLatest() {
         return V1_0_0;
     }
 }
--- a/typespec-tests/tsp/response.tsp
+++ b/typespec-tests/tsp/response.tsp
@@ -16,6 +16,20 @@ using Azure.ClientGenerator.Core;
 @useDependency(Azure.Core.Versions.v1_0_Preview_2)
 namespace Cadl.Response;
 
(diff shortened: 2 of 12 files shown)


typespec-tests/src/main/java/com/cadl/response/ResponseServiceVersion.java
weidongxu-microsoft, line 1: The rename here is caused by now ResponseClient also having api-version (previously on CoreClient has it), hence the name changed to the service name of ResponseServiceVersion

typespec-tests/tsp/response.tsp
```diff
@@ -92,4 +106,22 @@ interface ResponseOp {
...
+    {},
+    {
+      @doc("A response containing headers related to the Job Schedule, if it exists.")
+      @statusCode
+      code: "200";
+    },
+    {
+      @doc("A response containing headers related to the Job Schedule, if it exists.")
+      @statusCode
+      code: "404";
```
haolingdong-msft, line 122: what does default status code 404 mean?

weidongxu-microsoft: that 404 is expected response (not an error)

swagger https://github.com/Azure/azure-rest-api-specs/blob/main/specification/batch/data-plane/Microsoft.Batch/stable/2022-10-01.16.0/BatchService.json#L2829-L2855

typespec-tests/tsp/response.tsp
```diff
@@ -92,4 +106,22 @@ interface ResponseOp {
     @header("operation-location")
     operationLocation: ResourceLocation<Resource>;
   };
+
+  @route("/exists")
+  @head
```
haolingdong-msft, line 111: Found the response model is boolean, will @head operation always return boolean response? is this handled by compiler or our emitter?

typespec-tests/tsp/response.tsp
haolingdong-msft: ok, I see, thanks.

typespec-tests/tsp/response.tsp
weidongxu-microsoft: No, not always. Batch got some other HEAD op that does not do this.

haolingdong-msft: who will resolve response model type, compiler?

weidongxu-microsoft: codegen. that part already exists https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/util/SchemaUtil.java#L99-L102

as well as azure-core

haolingdong-msft: got it.

haolingdong-msft (approved):

PR #2159, opened 2023-05-24
Stream Style XML Code Generation Fixes

alzimmermsft:
Fixes issues found in stream-style XML code generation.

- Support for Swagger defined XML namespace prefixes.
- Incorrect ordering of parameters that didn't match the XML APIs.
- Add default constant for XML polymorphic discriminators.
- Correct tracking of XML namespace and XML element name for polymorphic models.
- Correct getting and setting of polymorphic properties in XML models.
- Updated to generate new `XmlSerializable` APIs.
- Correct handling for generating XML namespaces with prefixes.


lmolkova (approved):
Do we need any tests here or do we get enough coverage from SDK tests?

PR #2167, opened 2023-05-31
tsp, encode unixTimestamp

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2168

got confirmation that wire type of unixTimestamp is long.

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/builtin/BuiltinClient.java
+++ b/typespec-tests/src/main/java/com/cadl/builtin/BuiltinClient.java
@@ -96,14 +96,13 @@ public final class BuiltinClient {
      *         String: double (Required)
      *     }
      *     encoded (Required): {
-     *         timeInSeconds: long (Required)
-     *         timeInSecondsFraction: double (Required)
-     *         timeInSecondsOptional: Long (Optional)
-     *         dateTime: OffsetDateTime (Required)
-     *         dateTimeRfc7231: DateTimeRfc1123 (Required)
-     *         unixTimestamp: long (Required)
-     *         base64: byte[] (Required)
-     *         base64url: Base64Url (Required)
+     *         timeInSeconds: Long (Optional)
+     *         timeInSecondsFraction: Double (Optional)
+     *         dateTime: OffsetDateTime (Optional)
+     *         dateTimeRfc7231: DateTimeRfc1123 (Optional)
+     *         unixTimestamp: Long (Optional)
+     *         base64: byte[] (Optional)
+     *         base64url: Base64Url (Optional)
      *     }
      * }
      * }</pre>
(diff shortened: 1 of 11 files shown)


typespec-tests/src/main/java/com/cadl/builtin/BuiltinClient.java
```diff
@@ -101,6 +101,7 @@ public final class BuiltinClient {
      *         timeInSecondsOptional: Long (Optional)
      *         dateTime: OffsetDateTime (Required)
      *         dateTimeRfc7231: DateTimeRfc1123 (Required)
+     *         unixTimestamp: long (Required)
```
haolingdong-msft, line 104: We don't have convenient method for this?

weidongxu-microsoft: It's in Encoded class. No update to convenient API signature. Hence does not show in PR diff.

haolingdong-msft: I don't quite understand why `write()` method does not have convenient method?

haolingdong-msft: I see, convenient method  for `write` op is disabled in .tsp. But not sure why do we disable it.

weidongxu-microsoft: https://github.com/Azure/autorest.java/pull/1928/files#diff-192b6f227735483ceb77afa960ff7eae31d5ca5ec1e80f6f819ffe50a631148b

haolingdong-msft: I see, this `write` operation is a newly added operation to test `convenientApi(false)` previously when upgrading to 0.39. Now we've added encode to the model, do we want to add another operation to test encoded model as input?

weidongxu-microsoft: No, eventually we should get cadl-ranch tests on them.

For now, my current test on serialization should be enough. These are not exactly new features, all have their corresponding feature in Swagger.

haolingdong-msft (approved):

weidongxu-microsoft:
There is a problem on ctor when wire type is different from client type. Follow up on chat.

weidongxu-microsoft:
fluent ci fails on main https://dev.azure.com/azure-sdk/public/_build/results?buildId=2815218&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=1fb2445c-3c92-5f9b-bcad-710222d02bc3

seems some swagger changed

will solve in another PR

PR #2173, opened 2023-06-02
tsp, switch all LRO flow to getLroMetadata

weidongxu-microsoft:
`testRpcFinalOnLocation` case is to be removed (it does not work well with `getLroMetadata`).

---

will wait for Haoling's Generated get in first.


typespec-tests/src/main/java/com/_specs_/azure/core/lro/rpc/models/JobDataJobResultJobPollResultLongRunningFinalLocationJobResult.java
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) AutoRest Code Generator.
+
+package com._specs_.azure.core.lro.rpc.models;
+
+import com.azure.core.annotation.Immutable;
+
+/** The JobDataJobResultJobPollResultLongRunningFinalLocationJobResult model. */
+@Immutable
+public final class JobDataJobResultJobPollResultLongRunningFinalLocationJobResult {
```
haolingdong-msft, line 11: How do we decide the model name? The model name looks complex to me.

weidongxu-microsoft: It is probably from a template.

As mentioned in PR, this test case will be removed.

PR #2179, opened 2023-06-06
Remove required = true from Jackson property annotation

### LINKED ISSUE #2169: Do not set required = true in @JsonProperty for deserialization of properties
When deserializing a service response, we should not fail if a required property is not included in the response. This is a use case for OpenAI where the streaming response will not have some fields that are otherwise required for non-streaming responses.

srnagar:
Fixes #2169

### DIFF
--- a/typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/internal/models/PublicModel.java
+++ b/typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/internal/models/PublicModel.java
@@ -16,7 +16,7 @@ public final class PublicModel {
      * The name property.
      */
     @Generated
-    @JsonProperty(value = "name", required = true)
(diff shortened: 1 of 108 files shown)


typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/internal/models/PublicModel.java
```diff
@@ -16,7 +16,7 @@ public final class PublicModel {
      * The name property.
      */
     @Generated
-    @JsonProperty(value = "name", required = true)
+    @JsonProperty(value = "name")
```
weidongxu-microsoft, line 19: Technically we only need to remove the `required=true` in ctor. But I guess it is also fine here.

weidongxu-microsoft (approved):

weidongxu-microsoft (approved):

PR #2181, opened 2023-06-07
tsp, bump compiler 0.45

weidongxu-microsoft:
1st commit to bump dependencies to 0.45/0.31.

2nd commit to regen tests. It is all doc changes, though I am not sure whether the cause is the `@doc` change on Azure.Core, or behavior of handing `@doc` or doc on internal type from compiler.

### DIFF
--- a/typespec-extension/changelog.md
+++ b/typespec-extension/changelog.md
@@ -1,5 +1,9 @@
 # Release History
 
+## 0.8.0 (Unreleased)
+
+Compatible with compiler 0.45.
+
 ## 0.7.4 (2023-06-06)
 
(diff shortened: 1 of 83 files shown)


typespec-extension/changelog.md
```diff
@@ -1,5 +1,9 @@
 # Release History
 
+## 0.8.0 (Unreleased)
```
weidongxu-microsoft, line 3: we still uses this version range `>=0.45.0 <1.0.0"`.

not sure when Laurent's proposal get consensus.

hence probably delay the release a few days (unless asked for).

PR #2182, opened 2023-06-07
Change How Top-Level XML Wrapped Lists are Handled

alzimmermsft:
This PR changes how top-level XML wrapped lists are handled by javagen. Previously, when presented with the following Swagger segment:

```json
"EnqueuedMessageList": {
  "description": "The object returned when calling Put Message on a Queue",
  "type": "array",
  "items": {
    "$ref": "#/definitions/EnqueuedMessage",
    "xml": {
      "name": "QueueMessage"
    }
  },
  "xml": {
    "wrapped": true,
    "name": "QueueMessagesList"
  }
}
```

When this was used as the return model for REST APIs this would be treated as a list at the top-level, `Response<List<EnqueuedMessaged>>`, instead of an object with a list property, `Response<EnqueuedMessageList>`. This PR changes that by treating Swagger definitions like this specially by converting the array type returned by the code modeler to an ad hoc object type.

Additional changes in this PR are updates to stream style XML code generation. Handling for XML wrappers has changed to no longer generate a static inner class as stream style XML has more control over how serialization and deserialization for wrapped types than Jackson Dataformat XML does. Also, like how JSON works, if the XML model doesn't have constructor properties deserialization will happen in line with the object that will be returned.


javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -515,20 +516,18 @@ private void addProperties(ClientModel model, JavaClass classBlock, JavaSettings
...
-                    classBlock.staticFinalClass(JavaVisibility.PackagePrivate, wrapperClassDefinition, innerClass ->
-                        addXmlWrapperClass(innerClass, property, xmlWrapperClassName, settings));
+                    classBlock.staticFinalClass(JavaVisibility.PackagePrivate, xmlWrapperClassName,
+                        innerClass -> addXmlWrapperClass(innerClass, property, xmlWrapperClassName, settings));
 
                     fieldSignature = xmlWrapperClassName + " " + propertyName;
                 } else if (propertyType instanceof ListType) {
-                    fieldSignature = propertyType + " " + propertyName + " = new ArrayList<>()";
+                    if (settings.isStreamStyleSerialization()) {
+                        fieldSignature = propertyType + " " + propertyName + " = new LinkedList<>()";
```
weidongxu-microsoft, line 527: Any reason why LinkedList is used?

LinkedList may not be very cache friendly when having large number of items. (usually not a major concern though)

weidongxu-microsoft (approved):
I am not an expert on xml part.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
alzimmermsft: I'll be reverting this, for some reason I got `LinkedList` vs `ArrayList` confused with `LinkedHashMap` vs `HashMap` where the latter doesn't preserve deserialization order.

Issue #2186, opened 2023-06-09
tsp, test codegen for loadtesting


weidongxu-microsoft:
https://github.com/Azure/azure-rest-api-specs/tree/feature/loadtesting/specification/loadtestservice

---

I got failure on these to my local

```java
    /*
     * The aggregation function to be applied on the client metric. Allowed functions
     * - ‘percentage�? - for error metric , ‘avg�?, ‘p50�?, ‘p90�?, ‘p95�?, ‘p99�?, ‘min�?,
     * ‘max�? - for response_time_ms and latency metric, ‘avg�? - for requests_per_sec,
     * ‘count�? - for requests
     */
    @Generated
    @JsonProperty(value = "aggregate")
    private PfAgFunc aggregate;
```

source here https://github.com/Azure/azure-rest-api-specs/blob/feature/loadtesting/specification/loadtestservice/models.tsp#L209-L214

I am not sure about the cause.

haolingdong-msft:
I tried locally. The strange behavior is caused by the special characters used in the models,.tsp. Below are the trouble shooting process and fix solution.

**Trouble shooting**

1. use command `tsp compile` : generated comment is wrong like below. but the description in code-model.yaml is in correct.

java:
```java
    /*
     * The aggregation function to be applied on the client metric. Allowed functions
     * - ‘percentage�? - for error metric , ‘avg�?, ‘p50�?, ‘p90�?, ‘p95�?, ‘p99�?, ‘min�?,
     * ‘max�? - for response_time_ms and latency metric, ‘avg�? - for requests_per_sec,
     * ‘count�? - for requests
     */
```
code-model.yaml:
```yaml
 description: >-
    The aggregation function to be applied on the client metric.
    Allowed functions

    - ‘percentage’ - for error metric , ‘avg’, ‘p50’, ‘p90’, ‘p95’,
    ‘p99’, ‘min’,

    ‘max’ - for response_time_ms and latency metric, ‘avg’ - for
    requests_per_sec,

    ‘count’ - for requests
```
[...]

PR #2188, opened 2023-06-09
tsp, fix encoding issue for special chars

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/2186

Fix the wrong generated code for special chars.

got failure like below when generating loadingtesting:

```java
    /*
     * The aggregation function to be applied on the client metric. Allowed functions
     * - ‘percentage�? - for error metric , ‘avg�?, ‘p50�?, ‘p90�?, ‘p95�?, ‘p99�?, ‘min�?,
     * ‘max�? - for response_time_ms and latency metric, ‘avg�? - for requests_per_sec,
     * ‘count�? - for requests
     */
    @Generated
    @JsonProperty(value = "aggregate")
    private PfAgFunc aggregate;
source here https://github.com/Azure/azure-rest-api-specs/blob/feature/loadtesting/specification/loadtestservice/models.tsp#L209-L214
```


weidongxu-microsoft:
see https://github.com/Azure/autorest.java/issues/2186#issuecomment-1584164215

better not to rely on `file.encoding` env var.

haolingdong-msft:
> see [#2186 (comment)](https://github.com/Azure/autorest.java/issues/2186#issuecomment-1584164215)
> 
> better not to rely on `file.encoding` env var.

ok, I will try.

weidongxu-microsoft:
writer part may have bug as well.
```java
        try (FileWriter writer = new FileWriter(outputFile)) {
            writer.write(content);
        } catch (IOException e) {
```

you may need to write this
```java
        try (OutputStreamWriter writer = new OutputStreamWriter(Files.newOutputStream(outputFile.toPath()), StandardCharsets.UTF_8)) {
            writer.write(content);
        } catch (IOException e) {
```

typespec-extension/src/emitter.ts
```diff
@@ -100,6 +100,7 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
     try {
       const output = await promisify(execFile)("java", [
         `-DemitterOptions=${emitterOptions}`,
+        `-Dfile.encoding=UTF-8`,
```
weidongxu-microsoft, line 103: let's see if it will pass after remove this.

weidongxu-microsoft (approved):

weidongxu-microsoft (approved):

weidongxu-microsoft:
make sure you always use latest main when create branch

PR #2195, opened 2023-06-14
Tsp, fix special char encoding issue for partial update cases

haolingdong-msft:
Link https://github.com/Azure/autorest.java/pull/2188

We don't use encoding when reading existing file content in partial update. This will cause special chararacters displayed incorrectly in partial update case like below:

image

After fixing, the code looks good now.

```java

    /**
     * Get the aggregate property: The aggregation function to be applied on the client metric. Allowed functions -
     * ‘percentage’ - for error metric , ‘avg’, ‘p50’, ‘p90’, ‘p95’, ‘p99’, ‘min’, ‘max’ - for response_time_ms and
     * latency metric, ‘avg’ - for requests_per_sec, ‘count’ - for requests.
     *
     * @return the aggregate value.
     */
```


weidongxu-microsoft:
you probably need to fix the test code as well
https://github.com/Azure/autorest.java/blob/ecab1deaa47d563fb4739de6c9a4439623bc241f/partial-update/src/test/java/com/azure/autorest/partialupdate/util/PartialUpdateHandlerTest.java#L27-L28

weidongxu-microsoft (approved):

XiaofeiCao (approved):

Issue #2058, opened 2023-04-03
TypeSpec, group headers of SupportsConditionalRequests trait to RequestConditions


weidongxu-microsoft:
```
SupportsRepeatableRequests 
SupportsConditionalRequests 
SupportsClientRequestId
```
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-core/lib/traits.tsp

There 3 would involve lots of request headers
```
Repeatability-Request-ID
Repeatability-First-Sent

If-Match
If-None-Match
If-Unmodified-Since
If-Modified-Since

x-ms-client-request-id
```

Also a few in response
```
Repeatability-Result

ETag

x-ms-client-request-id
```

We might need certain design for client method, if one or several of them is used.

---

It requires SDK to provide auto-generated value for (random GUID and current date)
```
Repeatability-Request-ID
Repeatability-First-Sent

x-ms-client-request-id
```
if user does not provide these headers.
If user provides these headers, SDK should not generate (or equivalently, user input should overwrite SDK).

---

Currently SDK generate for
```
Repeatability-Request-ID
Repeatability-First-Sent
```
However user not able to overwrite.

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/2063 for SupportsRepeatableRequests 
https://github.com/Azure/autorest.java/issues/2064 for SupportsClientRequestId

For SupportsConditionalRequests, we could use https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/http/RequestConditions.java

weidongxu-microsoft:
A bit difficulty on convertion

```java
        RequestOptions requestOptions = new RequestOptions();
        String ifMatch = requestConditions.getIfMatch();
        String ifNoneMatch = requestConditions.getIfNoneMatch();
        OffsetDateTime ifUnmodifiedSince = requestConditions.getIfUnmodifiedSince();
        OffsetDateTime ifModifiedSince = requestConditions.getIfModifiedSince();
        if (ifMatch != null) {
            requestOptions.setHeader("If-Match", ifMatch);
        }
        if (ifNoneMatch != null) {
            requestOptions.setHeader("If-None-Match", ifNoneMatch);
        }
        if (ifUnmodifiedSince != null) {
            requestOptions.setHeader("If-Unmodified-Since", String.valueOf(ifUnmodifiedSince));
        }
        if (ifModifiedSince != null) {
[...]

PR #2196, opened 2023-06-14
tsp, group etag headers to RequestConditions

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2058

generated code
https://github.com/Azure/autorest.java/blob/89b479ed15e3b83b4f79ed903fe392a9b30cb164/typespec-tests/src/main/java/com/_specs_/azure/core/traits/TraitsClient.java#L141-L184


typespec-tests/src/main/java/com/cadl/specialheaders/SpecialHeadersClient.java
```diff
@@ -266,4 +443,196 @@ public Resource post(String name) {
...
+        if (ifMatch != null) {
+            requestOptions.setHeader("If-Match", ifMatch);
+        }
+        if (ifNoneMatch != null) {
+            requestOptions.setHeader("If-None-Match", ifNoneMatch);
+        }
+        return patchWithMatchHeadersWithResponse(name, BinaryData.fromObject(resource), requestOptions)
+                .getValue()
+                .toObject(Resource.class);
+    }
```
weidongxu-microsoft, line 540: Case of `if-match` and `if-none-match`.

typespec-tests/src/main/java/com/cadl/specialheaders/SpecialHeadersClient.java
```diff
@@ -266,4 +443,196 @@ public Resource post(String name) {
...
+            requestOptions.setHeader("If-None-Match", ifNoneMatch);
+        }
+        if (ifUnmodifiedSince != null) {
+            requestOptions.setHeader("If-Unmodified-Since", String.valueOf(new DateTimeRfc1123(ifUnmodifiedSince)));
+        }
+        if (ifModifiedSince != null) {
+            requestOptions.setHeader("If-Modified-Since", String.valueOf(new DateTimeRfc1123(ifModifiedSince)));
+        }
+        return putWithOptionalBodyWithResponse(format, requestOptions).getValue().toObject(Resource.class);
+    }
```
weidongxu-microsoft, line 617: Case of etag header with other optional query/header/body parameters. `RequestConditions` is put after the optional body parameter.

typespec-extension/src/code-model-builder.ts
```diff
@@ -852,7 +856,137 @@ export class CodeModelBuilder {
...
+      if (groupToRequestConditions || groupToMatchConditions) {
+        op.convenienceApi.requests = [];
+        const request = new Request();
+        request.parameters = [];
+        request.signatureParameters = [];
+        op.convenienceApi.requests.push(request);
+
+        for (const parameter of op.parameters) {
+          // copy all parameters to request
+          const clonedParameter = cloneOperationParameter(parameter);
```
haolingdong-msft, line 904: Could you please also paste the generated code-model.yaml? Because only looking at this logic is a bit hard to imagine the output code-model.yaml.

typespec-extension/src/code-model-builder.ts
```diff
@@ -852,7 +856,137 @@ export class CodeModelBuilder {
...
+        }
+      }
+
+      let groupToRequestConditions = false;
+      let groupToMatchConditions = false;
+
+      if (etagHeaders.length === 4) {
+        // all 4 headers available, use RequestConditions
+        groupToRequestConditions = true;
+      } else if (etagHeaders.length === 2) {
```
haolingdong-msft, line 887: is the length must be 2 or 4? Could the user only provide one header like `if-match`?

weidongxu-microsoft: I guess we'd ignore such case for now (1, or 3, or some mismatch of the 2 types).

Hope these won't happen. If we really see it, we may first check with REST API board on whether that definition is correct or not.

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Yeah, the grouping (and flattening) is hard to understand. It is originally from autorest m4.

I will upload the code-model.yaml, but that likely not understandable.

Guess I will need to do a write-up on it.

weidongxu-microsoft: [code-model.txt](https://github.com/Azure/autorest.java/files/11786370/code-model.txt)

rename that to yaml.

check the `groups` section.

weidongxu-microsoft: For the `requests` under `ConvenienceApi`, it is for handling of specific handling of method signature for convenientAPI.

If it is not set, codegen will assume it having same parameter as procotolAPI, just render differently (optional parameter etc.).

If it is set, usually this involve more handling, e.g. spread, or group (in etag case).

PR #2197, opened 2023-06-14
Remove the need for enable-xml

alzimmermsft:
This PR updates code generation to take into account `serializationFormats` produced by modeler four to determine which serialization formats a model or property uses. With this change `enable-xml` is no longer needed or used anywhere in code generation and can be removed at a later date.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientModel.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientModel.java
@@ -4,12 +4,11 @@
 package com.azure.autorest.model.clientmodel;
 
 import com.azure.autorest.extension.base.plugin.JavaSettings;
-import com.azure.autorest.util.ClientModelUtil;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * A model that is defined by the client.
@@ -80,9 +79,11 @@ public class ClientModel {
     private final boolean stronglyTypedHeader;
 
     private final ImplementationDetails implementationDetails;
+    private final boolean usedInXml;
 
     /**
      * Create a new ServiceModel with the provided properties.
+     *
      * @param name The name of this model.
      * @param imports The imports for this model.
      * @param description The description of this model.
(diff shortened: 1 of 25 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientModel.java
```diff
@@ -204,26 +207,35 @@ public ImplementationDetails getImplementationDetails() {
...
+            }
+        }
+
+        for (ClientModelPropertyReference clientModelPropertyReference : getPropertyReferences()) {
+            if (clientModelPropertyReference.isFromFlattenedProperty()) {
+                propertyAccesses.add(clientModelPropertyReference);
+            }
+        }
+
+        return propertyAccesses;
```
weidongxu-microsoft, line 233: nit, any concern over stream vs for-loop? Performance concern or what?

Usually, I view only it as personal preference?

javagen/src/main/java/com/azure/autorest/util/SchemaUtil.java
```diff
@@ -254,12 +254,13 @@ public static Optional<ClassType> mapExternalModel(ObjectSchema compositeType) {
...
-                            && compositeType.getLanguage().getJava().getNamespace().startsWith("com.azure.core")) {
+                    if ("PollResult".equals(name)
+                        && compositeType.getLanguage().getJava().getNamespace().startsWith("com.azure.core")) {
                         classType = new ClassType.Builder()
-                                .name(name)
-                                .packageName(compositeType.getLanguage().getJava().getNamespace())
-                                .build();
+                            .name(name)
+                            .packageName(compositeType.getLanguage().getJava().getNamespace())
+                            .usedInXml(treatAsXml(compositeType))
```
weidongxu-microsoft, line 262: We may not need this line. This [`PollResult`](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core-experimental/src/main/java/com/azure/core/experimental/models/PollResult.java) is only used for LRO poll result. I guess it won't be XML?  (emm, do we have LRO with XML?)

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/util/SchemaUtil.java
alzimmermsft: I don't think we have an XML-based `PollResult` so this could be removed.

Issue #2170, opened 2023-06-01
ctor for required properties has bug when client type differs from wire type


weidongxu-microsoft:
```java
public Encoded(@JsonProperty(value = "dateTimeRfc7231") OffsetDateTime dateTimeRfc7231);
public OffsetDateTime getDateTimeRfc1123();
private DateTimeRfc1123 dateTimeRfc7231;
```

In this class, the model property on wire is rfc1123 while client type is OffsetDateTime.

However Jackson also uses the ctor to de-serialize the JSON. At this ctor, since the parameter is OffsetDateTime, not DateTimeRfc1123, Jackson will fail on it.

We may able to solve it via generating a 2nd ctor, for Jackson only (but spotbugs may complain about this "unused" private ctor). E.g.
```java
private Encoded() {};
```
while having the public ctor for user only (i.e. with client type, and without `@JsonProperty`)
```java
public Encoded(OffsetDateTime dateTimeRfc7231);
```

---

And should we always do this? That is, do not create ctor with `@JsonProperty` on parameter.

weidongxu-microsoft:
https://github.com/weidongxu-microsoft/autorest.java/commit/15c0f2200ec21adf5fedf5636141f6ccf8ea9ac4

weidongxu-microsoft:
We would try to limit the possible (technically) breaking changes on annotations. So we will keep the code unchanged, if there is no mismatch.

There would be complication on inherit case, as superclass or subclass could have mismatch that might affect the other (even if they don't have mismatch).

weidongxu-microsoft:
Ignore corner case when mismatch happen to generic type.

```java
    @Generated
    public FloatSecondsDurationArrayProperty(List<Duration> value) {
        this.value =
                value.stream()
                        .map(el -> (double) el.toNanos() / 1000_000_000L)
                        .collect(java.util.stream.Collectors.toList());
    }

    @Generated
    @JsonCreator
    private FloatSecondsDurationArrayProperty(@JsonProperty(value = "value") List<Double> value) {
        this(
                value.stream()
                        .map(el -> Duration.ofNanos((long) (el * 1000_000_000L)))
                        .collect(java.util.stream.Collectors.toList()));
    }
```

PR #2202, opened 2023-06-20
client, fix Jackson de-serialize error when ctor has mismatched client type and wire type

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2170

this commit is a sample of the changes on generated code https://github.com/Azure/autorest.java/pull/2202/commits/f4612132421c618c6a25fc3a51b6fc0a5cbefe52


javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -762,6 +792,52 @@ private void addModelConstructor(ClientModel model, JavaVisibility constructorVi
...
+
+                addModelConstructorParameterAsWireType(property, constructorPropertiesAsWireType);
+
+                addParameterInvokePublicConstructor.accept(property);
+            }
+
+            classBlock.privateConstructor(model.getName() + "(" + constructorPropertiesAsWireType + ")", constructor -> {
+                constructor.line("this(" + constructorPropertiesInvokePublicConstructor + ");");
+            });
+        }
```
weidongxu-microsoft, line 840: Main logic here, to generate another private ctor with parameters as wire types.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -684,27 +685,55 @@ private void addModelConstructor(ClientModel model, JavaVisibility constructorVi
...
+                        return true;
+                    }
+                }
+            };
+            final boolean constructorParametersContainsMismatchWireType =
+                requiredProperties.stream().anyMatch(isWireTypeMismatch) || requiredParentProperties.stream().anyMatch(isWireTypeMismatch);
+
+            if (constructorParametersContainsMismatchWireType && !settings.isStreamStyleSerialization()) {
+                generatePrivateConstructorForJackson = true;
+            }
```
weidongxu-microsoft, line 726: Only do this, when there is mismatch, AND it is not stream-stype.

typespec-tests/src/main/java/com/encode/datetime/models/Rfc7231DatetimeProperty.java
```diff
@@ -27,11 +27,16 @@ public final class Rfc7231DatetimeProperty {
...
+    public Rfc7231DatetimeProperty(OffsetDateTime value) {
         this.value = new DateTimeRfc1123(value);
     }
 
+    @Generated
+    @JsonCreator
+    private Rfc7231DatetimeProperty(@JsonProperty(value = "value") DateTimeRfc1123 value) {
+        this(value.getDateTime());
+    }
+
```
weidongxu-microsoft, line 39: change in generated code like this

weidongxu-microsoft:
Latest cadl-ranch actually had a case on generic type (List) that we cannot handle even after this PR.

https://github.com/Azure/autorest.java/blob/9bfe7aa16a9129c57f7faab18230de954a321ff4/typespec-tests/src/main/java/com/encode/bytes/models/Base64UrlArrayBytesProperty.java#L29-L33
which would requires a ctor with `List<Base64Url>`.

In this case, the class does not have superclass or subclass. We can generate a empty private class for Jackson.
But this will not work in more complex case.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -684,27 +685,55 @@ private void addModelConstructor(ClientModel model, JavaVisibility constructorVi
...
         // If there are no required constructor properties this will simply be zero and result in a no-args constructor
         // being generated.
         StringBuilder constructorProperties =
-            new StringBuilder(128 * (requiredProperties.size() + requiredParentProperties.size()));
+            new StringBuilder(constructorPropertiesStringBuilderCapacity);
 
         StringBuilder superProperties = new StringBuilder(64 * requiredParentProperties.size());
 
         if (settings.isRequiredFieldsAsConstructorArgs()) {
+            final Predicate<ClientModelProperty> isWireTypeMismatch = p -> {
```
alzimmermsft, line 707: Could we turn this into a utility method? I imagine we have other places checking for `wireType != clientType`. Might want to design this utility method with an additional boolean parameter on whether to consider `GenericType`s as equalivalent.

alzimmermsft (approved):

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
weidongxu-microsoft: sure

PR #2203, opened 2023-06-20
tsp, upgrade cadl ranch

weidongxu-microsoft:
Also fix the conflict of local variable with client method parameter. https://github.com/Azure/autorest.java/commit/283e14412aa5c1348a4879bbf474d86cd8cca592

The cause is that client parameter has one named `value`, and generated code also uses `value` in lambda on mapping client type to wire type, e.g. `.map(value-> Base64Url.encode(value))`.

Here we rename the one in lambda to a more complex name, to reduce the chance of conflicting.


weidongxu-microsoft:
I merged it early, as cadl-ranch 0.16.0 got bug that won't pass test, and cadl-ranch 0.16.1 would produce the `value` parameter that get compilation fail.

Let me know if one think the name of the variable in lambda is not good. Happy to change.

PR #2204, opened 2023-06-21
Customization, debugging and sync stack

srnagar:
This PR includes 4 main changes:
- Enables sync stack by default and removes the option to configure it from TypeSpec codegen path (this a breaking change but very few services use this today and we can fix them all to not have to set this)
- Add support for customization from TypeSpec similar to the one supported with Swagger codegen
- Enable the Java process to log to console from TypeSpec (using either System.out.println or using logger that can be configured to log at different log levels)
- Enable debugging using IDE breakpoints to debug Java codegen code


customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageServerFacade.java
```diff
@@ -34,7 +34,7 @@ public EclipseLanguageServerFacade(String pathToLanguageServerPlugin, int port)
                 command += "-configuration ./config_linux";
             }
             server = Runtime.getRuntime().exec(command, new String[]{"CLIENT_PORT=" + port},
-                Paths.get(pathToLanguageServerPlugin, "jdt-language-server").toFile());
+                Paths.get(pathToLanguageServerPlugin, "../postprocessor/jdt-language-server").toFile());
```
weidongxu-microsoft, line 37: This probably won't work (or won't work alone) on released npm package.

I assume we need to modify the included `files` on npm package
https://github.com/Azure/autorest.java/blob/b1c855b0d641345ef4f3da5389b857aa81c1c1b1/typespec-extension/package.json#L41-L46

to include these as autorest npm package.
https://github.com/Azure/autorest.java/blob/b1c855b0d641345ef4f3da5389b857aa81c1c1b1/package.json#L40-L47

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -122,8 +124,7 @@ public String handlePartialUpdate(String filePath, String generatedContent) {
         SETTINGS_MAP.put("generate-sync-async-clients", true);
         SETTINGS_MAP.put("generate-builder-per-client", false);
         SETTINGS_MAP.put("sync-methods", "all");
-        // TODO(xiaofei) set to true when PagedIterable::mapPage is fixed in azure-core
-        SETTINGS_MAP.put("enable-sync-stack", false);
+        SETTINGS_MAP.put("enable-sync-stack", true);
```
weidongxu-microsoft, line 127: OK with it, but I remember the fix is merged but not released in 1.40 core.

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageServerFacade.java
weidongxu-microsoft: Also, include the language server would likely double the package size :-)

typespec-extension/src/emitter.ts
```diff
@@ -101,14 +104,19 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
...
+    javaArgs.push(codeModelFileName);
     try {
-      const output = await promisify(execFile)("java", [
-        `-DemitterOptions=${emitterOptions}`,
-        "-jar",
-        jarFileName,
-        codeModelFileName,
-      ]);
-      program.trace("typespec-java", output.stdout ? output.stdout : output.stderr);
+      spawn("java", javaArgs, {stdio: "inherit"});
```
weidongxu-microsoft, line 119: There seems to be some difference of async of the change.

After `promisify` I assume it be a `Promise`. Then `await` on it, as `$onEmit` is async function.

But `spawn` seems to not be a `Promise`.
https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

weidongxu-microsoft: The problem about it, is that `$onEmit` would finish (and tsp would continue), regardless of this java call is success or not.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
srnagar: Yeah, we will release azure-core this week. So, we should be able to make this the default and I will merge this PR after the next core release.

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageServerFacade.java
srnagar: Discussed this offline and moved the language server to resources to include this with the jar. Also, the size of the package will be bigger similar to autorest but it's needed to make customization work. We'll have to investigate an alternative for Eclipse Language Server as a separate task to improve the efficiency of customization and package size.

typespec-extension/src/emitter.ts
srnagar: Have updated the code to handle exit and error events and notify the emitter of any errors from running the java process.

typespec-extension/src/emitter.ts
```diff
@@ -101,27 +102,34 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
...
-        program.trace("typespec-java", msg);
-        program.reportDiagnostic({
-          code: "typespec-java",
-          severity: "error",
-          message: msg,
-          target: NoTarget,
-        });
-      } else {
+      const childProcess = spawn("java", javaArgs, {stdio: "inherit"});
+      childProcess.on('exit', (code, signal) => {
```
weidongxu-microsoft, line 118: I am still not sure about it, but guess we can merge and see.

This seems still calls `spawn`, register 2 callback function on it. The code flow still continue. Maybe the callback will be invoked, but then the main thread is not here.

typespec-extension/src/emitter.ts
```diff
@@ -7,8 +7,7 @@ import {
   createTypeSpecLibrary,
 } from "@typespec/compiler";
 import { dump } from "js-yaml";
-import { promisify } from "util";
-import { execFile } from "child_process";
+import {execFile, spawn, spawnSync} from "child_process";
```
weidongxu-microsoft, line 10: CI (lint) will fail if unused import.

typespec-extension/readme.md
```diff
@@ -72,6 +72,89 @@ A few exceptions are API of JSON Merge Patch, and API of long-running operation
...
+    service-name: "Authoring"
+    partial-update: false
+    service-versions:
+      - "2022-05-15-preview"
+    namer: false
+    generate-samples: true
+    generate-tests: true
+    examples-directory: "./examples"
+    custom-types-subpackage: "implementation.models"
+    custom-types: InternalModel1,InternalModel2
```
weidongxu-microsoft, line 97: nit, we can remove these, if we only want to emphasis customization options.

typespec-extension/src/emitter.ts
```diff
@@ -101,27 +102,34 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
...
+      childProcess.on('error', (err) => {
+        if (err.message.endsWith("ENOENT")) {
+          const msg = "'java' is not on PATH. Please install JDK 11 or above.";
+          throw new Error(msg + ": " + err.message);
+        }
         throw err;
-      }
+      });
+    } catch (err: any) {
+      throw err;
```
weidongxu-microsoft, line 132: I don't think throw in callback (lambda) would have effect on the catch here.

weidongxu-microsoft (approved):

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageServerFacade.java
```diff
@@ -40,6 +51,30 @@ public EclipseLanguageServerFacade(String pathToLanguageServerPlugin, int port)
         }
     }
 
+    private String unzipLanguageServer() throws IOException {
+        Path languageServerDirectory = Files.createTempDirectory("tmpjdt");
```
alzimmermsft, line 55: Thoughts on instead of using a temporary directory we use a well-known directory that has different version identifiers based on the Autorest Java version being used? Similar to what is done for Test Proxy. This way we can remove the entire need to bundle the Eclipse language server into Autorest Java and can instead use the download links provided by Eclipse to download the ZIP file if it doesn't exist, if it does exist we know all this work has already been done and can be skipped.

srnagar: Yep, that will certainly reduce the size of the npm package. I have created an issue to track this https://github.com/Azure/autorest.java/issues/2227

weidongxu-microsoft: Great idea. We can just upload the zip to github release artifacts, and code the URL and SHA to code. Code can verify the local zip with SHA, if pass, use local, if not pass, download from github.

PR #2206, opened 2023-06-21
Support XML namespaces in more locations in stream-style serialization

alzimmermsft:
Adds support for XML namespace in more locations in stream-style serialization. 

Internal methods for setting XML wrapped values have been removed as #2182 changed stream-style serialization to no longer need a wrapping class for wrapped XML elements as that is handled within serialization and deserialization in `toXml` and `fromXml`.

Checking the discriminator field for XML deserialization is also fixed for code generation that doesn't have required properties be part of the class constructor.

Also fixes a bug where `ClassCustomization.rename` could rename other classes that ended with the same string as the class being renamed, ex `RuleFilter` being renamed changing the name of `SqlRuleFilter` as both ended with `RuleFilter`. This now checks that the full class name is equal to what is being renamed.

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/streamstylexmlserialization/implementation/models/BananaWrapper.java
+++ b/vanilla-tests/src/main/java/fixtures/streamstylexmlserialization/implementation/models/BananaWrapper.java
@@ -12,6 +12,7 @@ import com.azure.xml.XmlWriter;
 import fixtures.streamstylexmlserialization.models.Banana;
 import java.util.ArrayList;
 import java.util.List;
+import javax.xml.namespace.QName;
 import javax.xml.stream.XMLStreamException;
 
 /** A wrapper around List&lt;Banana&gt; which provides top-level metadata for serialization. */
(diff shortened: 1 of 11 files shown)


vanilla-tests/src/main/java/fixtures/streamstylexmlserialization/implementation/models/BananaWrapper.java
```diff
@@ -65,9 +66,9 @@ public static BananaWrapper fromXml(XmlReader xmlReader, String rootElementName)
                     List<Banana> items = null;
 
                     while (reader.nextElement() != XmlToken.END_ELEMENT) {
-                        String elementName = reader.getElementName().getLocalPart();
+                        QName elementName = reader.getElementName();
 
-                        if ("banana".equals(elementName)) {
+                        if ("banana".equals(elementName.getLocalPart())) {
```
weidongxu-microsoft, line 71: nit, seems equivalent before/after change?

weidongxu-microsoft (approved):

vanilla-tests/src/main/java/fixtures/streamstylexmlserialization/implementation/models/BananaWrapper.java
alzimmermsft: Yeah, this is equivalent before and after but using `QName` allows for the namespace to be checked as well.

PR #2217, opened 2023-07-03
client, exclude serializationFormats=xml, if model only used in Exception

weidongxu-microsoft:
There is swagger that writes this way
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/web/resource-manager/Microsoft.Web/stable/2022-09-01/WebApps.json#L8295-L8348

It specifies `produces=application/xml` as response, and response is file. But that XML applies to the error response as well, hence `serializationFormats` be `json` and `xml`.

Without it, mgmt will have
```java
@Immutable
@JacksonXmlRootElement
public final class DefaultErrorResponseError extends ManagementError {
```

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/util/SchemaUtil.java
+++ b/javagen/src/main/java/com/azure/autorest/util/SchemaUtil.java
@@ -317,7 +317,8 @@ public class SchemaUtil {
     }
 
     public static boolean treatAsXml(Schema schema) {
-        return (schema.getSerializationFormats() != null && schema.getSerializationFormats().contains("xml"))
-            || (schema.getSerialization() != null && schema.getSerialization().getXml() != null);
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft:
`Error` in stream was XML https://github.com/Azure/autorest.java/blob/026ab5090fca611278267d4527997542ce531ada/vanilla-tests/src/main/java/fixtures/streamstylexmlserialization/models/Error.java

Should it be JSON instead?

weidongxu-microsoft:
@alzimmermsft 

Let me know your opinion.

If you'd prefer Error still stay XML, I can change to mgmt only.

weidongxu-microsoft:
I find in older tag, they were treated as XML
https://github.com/Azure/autorest.java/blob/v4.1.10/vanilla-tests/src/main/java/fixtures/streamstylexmlserialization/models/Error.java

So I guess I would modify the PR to make it only apply to mgmt (probably with a override in `ModelMapper`)

PR #2261, opened 2023-08-04
tsp, handle corner case for client type differs from wire type

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2170


typespec-tests/src/main/java/com/encode/bytes/models/Base64UrlArrayBytesProperty.java
```diff
@@ -27,13 +27,16 @@ public final class Base64UrlArrayBytesProperty {
...
         this.value = value.stream().map(el -> Base64Url.encode(el)).collect(java.util.stream.Collectors.toList());
     }
 
     @Generated
-    private Base64UrlArrayBytesProperty() {}
+    @JsonCreator
+    private static Base64UrlArrayBytesProperty fromJson(@JsonProperty(value = "value") List<Base64Url> value) {
+        return new Base64UrlArrayBytesProperty(
+                value.stream().map(el -> el.decodedBytes()).collect(java.util.stream.Collectors.toList()));
+    }
```
weidongxu-microsoft, line 39: Use factory method (the `static` method), so that we now can do "overload" on same `List<>` to the ctor.

Issue #2273, opened 2023-08-15
Code generation fails with latest Azure OpenAI 


jpalvarezl:
## Description

While bringing in the latest changes in this branch for the Azure OpenAI SDK rest API spec (expected to be merged in this [PR](https://github.com/Azure/azure-rest-api-specs/pull/25271/commits) ) into the Java Azure OpenAI SDK ([link](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/openai/azure-ai-openai)) running the code generation  steps:

- `\eng\common\scripts\TypeSpec-Project-Sync.ps1`
- `\eng\common\scripts\TypeSpec-Project-Generate.ps1`

The second script fails with the following error:

![image](https://github.com/Azure/autorest.java/assets/11056031/57e57a83-f866-4301-90c0-d95d9aa61dcf)

## Expected behaviour

The TSP project in isolation compiles fine with a typespec compiler version of `v0.47.0`

## Additional context

The source of the issue appears to be the usage of a `union` type between two models. See [here](https://github.com/Azure/azure-rest-api-specs/pull/25271/files#diff-dc5836bc2f2bc7d21322b10726611deb23591b7f0d60d65e09afdcaeb29bd224R74).

## Repro

Use the following `tsp-location.yaml` values for a repro of the error:

```yaml
commit: abde69c39810669c9a0d771bfdd8fb262b6c4de0
repo: trrwilson/azure-rest-api-specs
```

The latest commit in this branch has reverted the problematic changes. [Here](https://github.com/Azure/azure-rest-api-specs/pull/25271/commits/0d3d201639b9e7f3f448bbd08b1fe2fd461ecf49) you can see them in isolation.

weidongxu-microsoft:
acknowledged, will try make a 0.8.9 today

emitter-package.json update in SDK repo may take one more day.

weidongxu-microsoft:
The cause is actually this `EnumMember` as type
https://github.com/Azure/azure-rest-api-specs/blob/0d3d201639b9e7f3f448bbd08b1fe2fd461ecf49/specification/cognitiveservices/OpenAI.Inference/models/azure.cognitive.search.options.tsp#L67

[...]

PR #2274, opened 2023-08-16
tsp, support EnumMember as type

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/2273

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/enumservice/models/Operation.java
+++ b/typespec-tests/src/main/java/com/cadl/enumservice/models/Operation.java
@@ -47,6 +47,27 @@ public final class Operation {
     @JsonProperty(value = "color")
     private ColorModel color;
 
+    /*
+     * The priorityValue property.
+     */
+    @Generated
+    @JsonProperty(value = "priorityValue")
+    private Priority priorityValue = Priority.LOW;
+
+    /*
+     * The colorValue property.
+     */
+    @Generated
+    @JsonProperty(value = "colorValue")
+    private Color colorValue = Color.GREEN;
+
+    /*
+     * The colorModelValue property.
+     */
+    @Generated
+    @JsonProperty(value = "colorModelValue")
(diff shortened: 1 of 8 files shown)


typespec-tests/tsp/enum.tsp
```diff
@@ -45,6 +45,9 @@ model Operation {
   age: 50;
   priority: Priority;
   color: ColorModel;
+  priorityValue: Priority.Low;
+  colorValue: Color.Green;
+  colorModelValue: ColorModel.Blue;
```
weidongxu-microsoft, line 50: The cause of the issue is this `type: EnumMember`, instead of e.g. the usual `type: Enum`

typespec-tests/src/main/java/com/cadl/enumservice/models/Operation.java
```diff
@@ -47,6 +47,27 @@ public final class Operation {
...
+    @Generated
+    @JsonProperty(value = "colorValue")
+    private Color colorValue = Color.GREEN;
+
+    /*
+     * The colorModelValue property.
+     */
+    @Generated
+    @JsonProperty(value = "colorModelValue")
+    private ColorModel colorModelValue = ColorModel.BLUE;
```
weidongxu-microsoft, line 69: Generated code be this. Required type is treated as constant (which means it is read-only -- no setter).

XiaofeiCao (approved):

typespec-tests/tsp/enum.tsp
```diff
@@ -45,6 +45,9 @@ model Operation {
   age: 50;
   priority: Priority;
   color: ColorModel;
+  priorityValue: Priority.Low;
```
haolingdong-msft, line 48: What does it mean? define the default value of `priorityValue` as `Priority.Low`?

weidongxu-microsoft: yes

Issue #2149, opened 2023-05-19
mgmt, handle weird resource path in eventgrid 


XiaofeiCao:
(domain before eventSubscriptions): https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/preview/2023-06-01-preview/EventGrid.json#L2001

(topic before eventSubscriptions): https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/preview/2023-06-01-preview/EventGrid.json#L2067

Ideally they need to map to different resources. We can start with a workaround to enforce consistent mapping for `define` method, e.g. https://github.com/Azure/azure-sdk-for-java/pull/35026#pullrequestreview-1433721618

weidongxu-microsoft:
I am also need to handle this to support a case of e.g. map a resource to `null` to avoid breaking changes (previous stable the resource only happen to response, but in new stable it now can be createOrUpdate).

PR #2276, opened 2023-08-17
mgmt, resource collection association

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2149

eventgrid config:

```
resource-collection-association:
- resource: EventSubscription
  collection: TopicEventSubscriptions
```


XiaofeiCao (approved):

XiaofeiCao:
Should we map `EventSubscription` to `EventSubscription`? 
Current generated code seems wrong to me. It is calling `serviceClient().getTopicEventSubscriptions()` in its `create()` function. Instead, I believe it should call `serviceClient.getEventSubscriptions()`(it should have no `topics` information in `EventSubscriptions`)?

`create()` in `EventSubscriptionImpl`: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/eventgrid/azure-resourcemanager-eventgrid/src/main/java/com/azure/resourcemanager/eventgrid/implementation/EventSubscriptionImpl.java#L137-L138

swagger: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/stable/2022-06-15/EventGrid.json#L1204

weidongxu-microsoft:
From guide here seems right?

https://learn.microsoft.com/en-us/azure/event-grid/scripts/cli-subscribe-custom-topic#run-the-script

It would create a topic first, then create the event subscription.

Though not sure if there is other way to create event subscription (non custom topic?).

weidongxu-microsoft:
I remember that at first, `EventSubscription` follow this path https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/preview/2023-06-01-preview/EventGrid.json#L2736, "resource from scope"
Seems OK.

Then this was added https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/preview/2023-06-01-preview/EventGrid.json#L2067 and it moved to "subresource from resource group"
This seems the correct one. Better than the one above.

XiaofeiCao:
There is one without `topics`, only `domain`:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/preview/2023-06-01-preview/EventGrid.json#L3628

Though response is also `EventSubscription` and it always has `topics` property. 
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/eventgrid/resource-manager/Microsoft.EventGrid/preview/2023-06-01-preview/EventGrid.json#L13413-L13421

weidongxu-microsoft:
OK, maybe we can move it back to "scope as parent", it is be more generic? We should be able to add another field e.g. "type: SCOPE_AS_PARENT" to have it match the scope first.

XiaofeiCao:
Agree, `withScope` definitely seems more appropriate.

Issue #1616, opened 2022-08-11
Client, support PagedFlux.byPage(pageSize) for maxpagesize


weidongxu-microsoft:
pageable API would need to generate like this, if `maxpagesize` parameter in REST API.

```java
    public PagedFlux<Foo> listAsync() {
        return new PagedFlux<>(
                (pageSize) -> listSinglePageAsync(pageSize),
                (nextLink, pageSize) -> listNextSinglePageAsync(nextLink));
```

weidongxu-microsoft:
I will add a `enable-page-size` option.

weidongxu-microsoft:
Code may look like

```java
    public PagedIterable<BinaryData> list(RequestOptions requestOptions) {
        return new PagedIterable<>(
                (size) -> {
                    RequestOptions requestOptionsForSinglePage = requestOptions == null ? new RequestOptions() : requestOptions;
                    requestOptionsForSinglePage.addQueryParam("maxpagesize", String.valueOf(size));
                    return listSinglePage(requestOptions);
                },
                (nextLink, size) -> {
                    RequestOptions requestOptionsForNextPage = new RequestOptions();
                    requestOptionsForNextPage.setContext(
                            requestOptions != null && requestOptions.getContext() != null
                                    ? requestOptions.getContext()
                                    : Context.NONE);
                    requestOptionsForNextPage.addQueryParam("maxpagesize", String.valueOf(size));
                    return listNextSinglePage(nextLink, requestOptionsForNextPage);
                });
    }
```

PR #2277, opened 2023-08-18
dpg, support maxpagesize for Paged model

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/1616

This PR only work for DPG (protocol API and impl), not vanilla (nor fluent).


typespec-tests/tsp/protocol-api-convenient-api.tsp
```diff
@@ -97,6 +97,19 @@ model ResourceJ {
   type: string;
 }
 
+@trait
+model ListTrait {
+  @traitContext(TraitContext.List)
+  reqeust: {
```
XiaofeiCao, line 103: typo?

weidongxu-microsoft: yeah, typo. The name seems not have effect at the first place.

typespec-tests/tsp/protocol-api-convenient-api.tsp
```diff
@@ -97,6 +97,19 @@ model ResourceJ {
...
+    @traitLocation(TraitLocation.Parameters)
+    properties: {
+      @projectedName("client", "maxPageSize")
```
XiaofeiCao, line 106: Is `maxPageSize` SDK's conventional name for the pageSize? If service defines otherwise, we need to tell them to use `@projectedName` right?

XiaofeiCao (approved):

typespec-tests/tsp/protocol-api-convenient-api.tsp
weidongxu-microsoft: Yes. Ideally it should have serialziedName = maxpagesize. For those not define as such, we allow them to use projectedName to maxPageSize (as we cannot change their serialziedName)

weidongxu-microsoft:
test https://github.com/Azure/cadl-ranch/pull/410

PR #2278, opened 2023-08-18
tsp, support test generation

XiaofeiCao:
issue: https://github.com/Azure/autorest.java/issues/2068
Generated test example: https://github.com/Azure/autorest.java/pull/2278/files#diff-19f3c91dc765d44e978e894edfa31dd7112cd5a477a6c56b881b7293dca5de1e

Response assertion will be supported in following PR.


javagen/src/main/java/com/azure/autorest/model/clientmodel/TestContext.java
```diff
@@ -8,12 +8,12 @@
 import java.util.List;
 import java.util.Objects;
 
-public class TestContext {
+public class TestContext<Example> {
```
weidongxu-microsoft, line 11: nit, `TExample`?

weidongxu-microsoft (approved):
Not sure how to handle test. Maybe we disable generate-tests here, and later add a thing similar to https://github.com/Azure/autorest.java/tree/main/protocol-sdk-integration-tests (may use the same folder?)

XiaofeiCao:
> Not sure how to handle test. Maybe we disable generate-tests here, and later add a thing similar to https://github.com/Azure/autorest.java/tree/main/protocol-sdk-integration-tests (may use the same folder?)

Putting `@Disabled` on the class seems to do the trick, locally verified:
https://github.com/junit-team/junit5/issues/415#issuecomment-234776991

weidongxu-microsoft:
Alright, that should so do, at least in short term.

> Putting `@Disabled` on the class seems to do the trick, locally verified: [junit-team/junit5#415 (comment)](https://github.com/junit-team/junit5/issues/415#issuecomment-234776991)

javagen/src/main/java/com/azure/autorest/model/clientmodel/TestContext.java
XiaofeiCao: Fixed.

PR #2281, opened 2023-08-23
Handle read-only properties better in stream-style serialization

alzimmermsft:
Updates how Autorest generates JSON stream-style serialization to better handle read-only properties. In deserialization if the field is defined in the super type a package-private setter is now generated that takes the wire type value as the field is generated as private and stream-style serialization doesn't perform any reflective setting. In serialization read-only properties are no longer included in serialization.


weidongxu-microsoft:
There seems to be some class difference (between client and wire) in autorest.java CI.

weidongxu-microsoft (approved):
One may add some example that these changes have effect, to e.g. azure-dataplane-tests

PR #2285, opened 2023-08-29
tsp, support access and usage

weidongxu-microsoft:


typespec-tests/tsp/internal.tsp
```diff
@@ -26,35 +26,46 @@ model ResponseInternal {
...
+@access(Access.public)
 model Response {
   name: string;
 }
 
-// StandAloneData and StandAloneDataInner will be generated
-@include
+// StandAloneData will be generated as output
+@access(Access.public)
+@usage(Usage.output, "java")
```
haolingdong-msft, line 42: What is the difference between output-only model and default model?

weidongxu-microsoft: At present, no. output-only is currently the least possible model we can generate (it does not have a public ctor and setter).

But it may change, if there be difference of model of "none" and model of "output-only".

haolingdong-msft (approved):

PR #2294, opened 2023-09-06
Use KeyCredential instead of AzureKeyCredential

srnagar:
This PR is to update the generated builder classes to use `KeyCredential` instead for `AzureKeyCredential`. This is disabled by default for swagger-based generators and enabled for TypeSpec based generators. 

Also, centralizing the azure-core versioning in dependency management.


pom.xml
```diff
@@ -57,6 +57,12 @@
         <artifactId>jsonrpc4j</artifactId>
         <version>1.6.0</version>
       </dependency>
+
+      <dependency>
+        <groupId>com.azure</groupId>
+        <artifactId>azure-core</artifactId>
+        <version>1.43.0</version>
```
srnagar, line 64: This version is not yet available in Maven Central. It will be released by end of this week. This PR will be merged after the release and when the build is successful.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
```diff
@@ -5,12 +5,9 @@
 
 import com.azure.autorest.extension.base.plugin.JavaSettings;
 import com.azure.autorest.model.javamodel.JavaBlock;
-import com.azure.core.client.traits.AzureKeyCredentialTrait;
-import com.azure.core.client.traits.ConfigurationTrait;
-import com.azure.core.client.traits.EndpointTrait;
-import com.azure.core.client.traits.HttpTrait;
-import com.azure.core.client.traits.TokenCredentialTrait;
+import com.azure.core.client.traits.*;
```
weidongxu-microsoft, line 8: Do we want to avoid import *?

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -137,6 +137,8 @@ public String handlePartialUpdate(String filePath, String generatedContent) {
         SETTINGS_MAP.put("generic-response-type", true);
         SETTINGS_MAP.put("output-model-immutable", true);
         SETTINGS_MAP.put("disable-required-property-annotation", true);
+        // Defaulting to KeyCredential and not providing TypeSpec services to generate with AzureKeyCredential.
+        SETTINGS_MAP.put("use-key-credential", true);
```
weidongxu-microsoft, line 141: Would we set this as default for DPG as well? I can do in a follow-up PR.

For those already GAed, we can update their swagger/readme.md to still use AzureKeyCredential.

weidongxu-microsoft (approved):

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
srnagar: Ideally, we would want to do this for all libraries (including DPG). This will be a 2-step process:
- update autorest to make this the default and release
- update readme.md of all currently GA'd libraries 

We can start with just DPG in a separate PR.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
srnagar: Yes, I will fix this. We should have a checkstyle rule for this.

PR #2308, opened 2023-09-15
Handle wire type to client type conversions better

alzimmermsft:
Fixes an issue with stream-style serialization if the wire type to client type conversion requires the wire type value to be non-null.

Before the transformation would generate as

```java
OffsetDateTime.ofInstant(Instant.ofEpochSecond(reader.getNullable(JsonReader::getLong)), ZoneOffset.UTC));
```

which would result in a `NullPointerException` if the wire type long value was null.

It now generates as

```java
Long expires = reader.getNullable(JsonReader::getLong);
if (expires != null) {
    deserializedCertificateAttributes.setExpires(OffsetDateTime.ofInstant(Instant.ofEpochSecond(expires), ZoneOffset.UTC));
}
```

which null checks the nullable wire type for being null before converting to the client type.


srnagar (approved):
LGTM!

PR #2309, opened 2023-09-18
tsp, make API internal, for multipart/form-data

weidongxu-microsoft:
Treat it as internal API and no convenience API


typespec-tests/src/main/java/com/cadl/multipart/MultipartClient.java
```diff
@@ -0,0 +1,63 @@
...
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @return the {@link Response}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Response<Void> uploadWithResponse(BinaryData request, RequestOptions requestOptions) {
+        // Protocol API does not work, as operation 'upload' is 'multipart/form-data'
```
weidongxu-microsoft, line 60: Just say it won't work...

weidongxu-microsoft (commented):
May replace the local test with that from cadl-ranch (when available).

typespec-tests/src/main/java/com/cadl/multipart/MultipartClient.java
haolingdong-msft: If the protocol API does not work, why do we still generate it as `public`, probably make it package-private?

weidongxu-microsoft: make sense. change to internal as well.

haolingdong-msft (approved):

PR #2316, opened 2023-09-22
tsp, upgrade cadl-ranch; make all response body be BinaryData (include string/enum)

weidongxu-microsoft:
big fix for `byte[]` in body.
https://github.com/Azure/cadl-ranch/pull/424


typespec-extension/src/operation-utils.ts
```diff
@@ -17,15 +17,37 @@ import { EmitterOptions } from "./emitter.js";
...
+    "text/javascript",
+    "text/plain",
+    // not in azure-core
+    "application/merge-patch+json",
+  ]);
+
+  return contentTypes.some((it) => {
+    return supported_mime_types.has(it);
+  });
+}
```
weidongxu-microsoft, line 50: I am following java azure-core here. If core does not know how to serialize, we assume it be unknown.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1332,8 +1343,11 @@ export class CodeModelBuilder {
     let trackConvenienceApi = op.convenienceApi ?? false;
     if (resp.responses && resp.responses.length > 0 && resp.responses[0].body) {
       const responseBody = resp.responses[0].body;
+      const unknownResponseBody =
+        responseBody.contentTypes.length > 0 && !isKnownContentType(responseBody.contentTypes);
+
       bodyType = this.findResponseBody(responseBody.type);
-      if (bodyType.kind === "Scalar" && bodyType.name === "bytes") {
+      if (unknownResponseBody && bodyType.kind === "Scalar" && bodyType.name === "bytes") {
```
weidongxu-microsoft, line 1350: If content-type is unknown, and body is bytes, we'd take it means binary stream.

PR #2317, opened 2023-09-25
Return null by default for byte[] getters

alzimmermsft:
Changes the default behavior in `ArrayType` to return `null` instead of `new byte[0]` when the `defaultValueExpression` is null. To provide a way for backwards compatibility a new configuration `default-byte-array-returns-empty-array` has been added to `JavaSettings`. Additionally, the previous behavior has a minor improvement by reusing a static singleton for the empty byte array instead of creating a new instance on each get call.

### DIFF
--- a/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
+++ b/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
@@ -170,7 +170,7 @@ public class JavaSettings {
                 getBooleanValue(host, "disable-required-property-annotation", false),
                 getBooleanValue(host, "enable-page-size", false),
                 getBooleanValue(host, "use-key-credential", false),
-                getBooleanValue(host, "default-byte-array-returns-empty-array", false)
+                getBooleanValue(host, "null-byte-array-maps-to-empty-array", false)
             );
         }
         return instance;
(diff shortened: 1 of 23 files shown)


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -169,7 +169,8 @@ public static JavaSettings getInstance() {
                 // were generated with required = true set in JsonProperty annotation
                 getBooleanValue(host, "disable-required-property-annotation", false),
                 getBooleanValue(host, "enable-page-size", false),
-                getBooleanValue(host, "use-key-credential", false)
+                getBooleanValue(host, "use-key-credential", false),
+                getBooleanValue(host, "default-byte-array-returns-empty-array", false)
```
srnagar, line 173: `null-byte-array-maps-to-empty-array`

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -132,6 +132,15 @@ public final void write(ClientModel model, JavaFile javaFile) {
                 classBlock.privateStaticFinalVariable("Pattern KEY_ESCAPER = Pattern.compile(\"\\\\.\");");
             }
 
+            // If code is being generated with the behavior to return an empty byte array when the default value
+            // expression is null and the model has any array types that will need conversion within getter methods
+            // generate a static byte[] that will be returned instead of creating a new instance each get.
+            if (settings.isDefaultByteArrayReturnsEmptyArray()
+                && model.getProperties().stream().anyMatch(property -> property.getClientType() instanceof ArrayType
+                    && property.getWireType() != property.getClientType())) {
+                classBlock.privateStaticFinalVariable("byte[] EMPTY_BYTE_ARRAY = new byte[0]");
```
srnagar, line 141: Do we not have a constant in azure-core for this?

srnagar (approved):

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
alzimmermsft: Nothing that is public

weidongxu-microsoft:
null array could trigger some spotbug error

```java
    public byte[] data() {
        if (this.data == null) {
            return null;
        }
        return this.data.decodedBytes();
    }
```

see https://github.com/Azure/autorest.java/pull/2334

PR #2326, opened 2023-10-03
Adds partial update support for package-info.java

alzimmermsft:
Adds support for partial updates in code generated `package-info.java` files. In code generated with `partial-update: true` the Javadoc description for `package-info.java` files will have the generated Javadoc surrounded by `<!-- start generated doc -->` and `<!-- end generated doc -->`. When regenerating code, the contents between those tags will be replaced with the newly generated description, any Javadocs outside of those tags will be retained as-is. If the tags don't exist the existing Javadoc will be retains and the generated description will be discarded.


postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -287,8 +287,7 @@ private void handlePartialUpdate(Map<String, String> fileContents) {
                 if (Files.exists(existingFilePath)) {
                     try {
                         String existingFileContent = new String(Files.readAllBytes(existingFilePath), StandardCharsets.UTF_8);
-                        String updatedContent = PartialUpdateHandler.handlePartialUpdateForFile(generatedFileContent, existingFileContent);
-                        return updatedContent;
+                        return PartialUpdateHandler.handlePartialUpdateForFile(generatedFileContent, existingFileContent);
```
weidongxu-microsoft, line 290: nit, should we in future refactor it to use a class similar to `JavaFile` (i.e. have filename and content) as input parameter? So that it be much simpler to decide whether the file is package-info.java.

typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/access/implementation/models/package-info.java
```diff
@@ -2,5 +2,10 @@
 // Licensed under the MIT License.
 // Code generated by Microsoft (R) TypeSpec Code Generator.
 
-/** Package containing the data models for Access. Test for internal decorator. */
+/**
+ *
+ * <!-- start generated doc -->
+ * Package containing the data models for Access. Test for internal decorator.
+ * <!-- end generated doc -->
+ */
```
weidongxu-microsoft, line 10: Welcome advise from @srnagar and @haolingdong-msft (it seems not possible for `@Generated` approach here)

Also may want to consider whether we'd like to enable javadoc customization to APIs in Client as well.

weidongxu-microsoft:
For the conflict, `typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/internal` is removed.

partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -65,13 +70,20 @@ public static String handlePartialUpdateForFile(String generatedFileContent, Str
         // 2. If it's module-info.java file, then go to handlePartialUpdateForModuleInfoFile
         if (compilationUnitForExistingFile.getModule().isPresent() &&
                 compilationUnitForGeneratedFile.getModule().isPresent()) {
-            return handlePartialUpdateForModuleInfoFile(generatedFileContent, existingFileContent);
+            return handlePartialUpdateForModuleInfoFile(compilationUnitForGeneratedFile, compilationUnitForExistingFile);
         }
 
-        // 3. If it's class or interface file, handle partial update for class or interface file
+        // 3. If it's package-info.java file, then go to handlePartialUpdateForPackageInfoFile
```
haolingdong-msft, line 76: Please update the javadoc of this method, as well as the javadoc for the `PartialUpdateHandler`.

weidongxu-microsoft (approved):
unblock

PR #2338, opened 2023-10-11
tsp, anonymous model

weidongxu-microsoft:
it is under `namer` option, currently no service uses that option.

```
warning typespec-java: Rename anonymous TypeSpec model to 'parametersModel'
warning typespec-java: Rename anonymous TypeSpec model to 'getAnonymouseResponse'
```

we may later move it to another option, e.g. `unbranding`


typespec-tests/src/main/java/com/cadl/enumservice/EnumServiceAsyncClient.java
```diff
@@ -303,7 +303,7 @@ public Mono<Response<BinaryData>> setIntEnumArrayWithResponse(
      * String
      * }</pre>
      *
-     * @param stringArray Array of Response.
+     * @param stringArray Array of SetStringEnumArrayResponse.
```
weidongxu-microsoft, line 306: These strange thing is done by `prenamer` which we copied from autorest modules.

So far, no service actually uses this option, so we may eventually remove this from typespec-java.

Issue #1511, opened 2022-06-18
Client, set/replace api-version in LRO URL


weidongxu-microsoft:
Actually apply to all client.

Simple case is api-version is same across client.
A slight complex case is api-version per method (usually happen in mgmt, but also on synapse data-plane).

And we probably only handle for `api-version` in query. If service calls it another name, we probably not able to handle that.

It could be only addition if not available, or addition and replacement.

Might need support from azure-core.
```[tasklist]
### Tasks
- [ ] Use cadl-ranch azure core lro case to test if api version is correctly appended, add assertions in cadl-ranch as well
```

haolingdong-msft:
From Srikanta: We only should be able to set/replace API versions in LRO and paging APIs when it's a query param. The API version should be the same as the client version created by the customer. Do not modify api versions that are part of the path param

weidongxu-microsoft:
For api-version https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/clientmodel/ParameterSynthesizedOrigin.java#L26 (path parameter will have this too)

PR #2349, opened 2023-10-16
Client, set/replace api-version in LRO URL 

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/1511

Code changes:
1. `JavaSettings.java`: Add `.setServiceVersion({serviceVersion}))` to polling strategy by default
2. `ClientMethodTemplate.java`: If client method has `apiVersion` in query param and is DPG, then replace `{serviceVersion}` to `this.getServiceVersion().getVersion()`

Test:
1. cadl-ranch standard LRO case: add assertions to check lro polling url contains correct api version. https://github.com/Azure/cadl-ranch/pull/443
2. `vanilla-tests/swagger/lro.md`: only enable customized polling strategy for `LROs_put200Succeeded`, remaining operations will set `serviceVersion` to `null`.

Currently I use `this.getServiceVersion().getVersion()` to get client's service version, have not handle case where method's service version is different.


javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -1526,6 +1526,7 @@ private String getPollingStrategy(ClientMethod clientMethod, String contextParam
             .replace("{httpPipeline}", clientMethod.getClientReference() + ".getHttpPipeline()")
             .replace("{endpoint}", endpoint)
             .replace("{context}", contextParam)
+            .replace("{serviceVersion}", clientMethod.getClientReference() + ".getServiceVersion().getVersion()")
```
weidongxu-microsoft, line 1529: Please test the LRO case in vanilla
vanilla-tests/swagger/lro.md (remove the polling)

weidongxu-microsoft (commented):
Also update the cadl-ranch standard LRO to verify the api-version in poll GET

haolingdong-msft:
> Also update the cadl-ranch standard LRO to verify the api-version in poll GET

Yeah, wip on cadl-ranch's test.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
weidongxu-microsoft: you can modify the md to set polling to one or two API, that would leave others as test of default polling

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -1578,4 +1580,15 @@ protected void generateSendRequestSync(ClientMethod clientMethod, JavaType typeB
             function.methodReturn("this.sendRequestAsync(httpRequest).contextWrite(c -> c.putAll(FluxUtil.toReactorContext(context).readOnly())).block()");
         });
     }
+
+    private String getServiceVersionValue(ClientMethod clientMethod) {
```
weidongxu-microsoft, line 1584: can it be static?

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
```diff
@@ -1578,4 +1580,15 @@ protected void generateSendRequestSync(ClientMethod clientMethod, JavaType typeB
...
+        String serviceVersion = "null";
+        if (JavaSettings.getInstance().isDataPlaneClient() && clientMethod.getProxyMethod() != null && clientMethod.getProxyMethod().getParameters() != null) {
+            if (clientMethod.getProxyMethod().getParameters().stream()
+                    .anyMatch(p -> p.getRequestParameterLocation() == RequestParameterLocation.QUERY && "apiVersion".equals(p.getName()))) {
```
weidongxu-microsoft, line 1588: Can you use this? `ParameterSynthesizedOrigin.fromValue(parameter.getOrigin()) == ParameterSynthesizedOrigin.API_VERSION`

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplate.java
haolingdong-msft: It can be static. But I don't quite get the benefit of making it static? The method is `private` , so no instance should be instantiated to call it.

weidongxu-microsoft: `static` means it will never use any var in the class. So whenever method can be static, it better be.

haolingdong-msft: Make sense, I will make it static.

weidongxu-microsoft:
make sure you've tested it with cadl-ranch LRO

haolingdong-msft:
> make sure you've tested it with cadl-ranch LRO

Yes, tested, this is the log in UT, all urls contains api-version.
image

protocol-tests/src/main/java/fixtures/dpgcustomization/implementation/DpgClientImpl.java
```diff
@@ -677,7 +677,8 @@ public PollerFlux<BinaryData, BinaryData> beginLroAsync(String mode, RequestOpti
                                 .setContext(
                                         requestOptions != null && requestOptions.getContext() != null
                                                 ? requestOptions.getContext()
-                                                : Context.NONE)),
+                                                : Context.NONE)
+                                .setServiceVersion(null)),
```
srnagar, line 681: Can we skip setting this if the value is `null`? This looks a bit odd in generated code as this is a no-op.

haolingdong-msft: Hi Srikanta, thanks for your suggestion. I have the same concern with you previously. 
But I finally set to `null` because our current logic is to `setServiceVersion(this.getServiceVersion().getVersion())` when two conditions meet: 1. it's DPG 2. apiversion is in query param. And if we don't want to pass it, we need to update the [polling strategy template](https://github.com/Azure/autorest.java/blob/main/extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java#L948-L972) in `JavaSettings`, but in JavaSettings, we don't know if there is `api-version` in query param, So if it is DPG and `api-version` is not in query param, it still will be set to `null`. To make the logic for two conditions in one place, I add `setServiceVersion` to the default template and determine its value in `ClientMethodTemplate`.

PR #2353, opened 2023-10-16
verify new createSdkContext changes in tcgc

iscai-msft:


weidongxu-microsoft (approved):
Change looks good.

Do we expect to use the 0.36.0-dev.2 here for release (user might get a conflict on their end when running tsp compile), or we can wait for 0.36.0

iscai-msft:
@weidongxu-msft there won't be a conflict if you publish with the dev version, however this isn't a fix that needs to go out since it's all internal code.

weidongxu-microsoft:
I am doing typespec version bump in this PR

Issue #2345, opened 2023-10-13
Crash in Java emitter


dargilco:
I run this command at the root of the azure-rest-api-specs repo:
```
pwsh .\eng\scripts\TypeSpec-Generate-Sdk.ps1 -SdkRepoRootDirectory ..\sdk-repos\azure-sdk-for-java -TypeSpecProjectDirectory .\specification\cognitiveservices\Vision.ImageAnalysis -SdkLanguage java
```
Per the screen spew I see
```console
Emitter "@azure-tools/typespec-java" crashed! This is a bug.
Please file an issue at https://github.com/Azure/autorest.java/issues
```
so I'm filing this issue. Note that my TypeSpec files are not yet in Main branch of azure-rest-api-specs. They are in [draft PR](https://github.com/Azure/azure-rest-api-specs/pull/26146) in the topic branch [dargilco/image-analysis-typespec](https://github.com/Azure/azure-rest-api-specs/tree/dargilco/image-analysis-typespec)

Full screen spew

```console
e:\src\azure-rest-api-specs>pwsh .\eng\scripts\TypeSpec-Generate-Sdk.ps1 -SdkRepoRootDirectory ..\sdk-repos\azure-sdk-for-java -TypeSpecProjectDirectory .\specification\cognitiveservices\Vision.ImageAnalysis -SdkLanguage java
Generating java sdk code ...
Using module powershell-yaml with version 0.4.7.
WARNING: Parameter of Commithash or RepoUrl are not provided along with the local path of tspconfig.yaml, trying to re-generate sdk code based on the local type specs.
Generating sdk code based on local type specs at specRepoRoot: E:/src/azure-rest-api-specs.
WARNING: Failed to find tsp-location.yaml in 'E:\src\sdk-repos\azure-sdk-for-java\sdk\cognitiveservices\azure-ai-vision-imageanalysis'. Trying to create tsp-location.yaml using HEAD commit of the local spec repo then proceed the sdk generation based upon local typespecs at E:/src/azure-rest-api-specs. Alternatively, please make sure to provide CommitHash and RepoUrl parameters when running this script.
WARNING: Failed to find git origin repo of the local spec repo at specRepoRoot: E:/src/azure-rest-api-specs. Using default repo: Azure/azure-rest-api-specs
created package folder E:\src\sdk-repos\azure-sdk-for-java\sdk\cognitiveservices\azure-ai-vision-imageanalysis
[...]

PR #2355, opened 2023-10-17
fix sample generation failure

XiaofeiCao:
address https://github.com/Azure/autorest.java/issues/2345

generated sample/test: https://github.com/Azure/azure-sdk-for-java/commit/50576fa15adc575bbfc929cd23b0d9ec0910abe0#diff-63aed0c7ce618d000fd35a03c68985fe0f88bdf60ecc11313de43f2f55f13e3d

Code change: 
1. Add condition check `ClientModelUtil.isClientModel(type)`.
2. Remove unnecessary error throwing

Others are format change after removing `if(model != null)` check.


javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
```diff
@@ -106,74 +106,71 @@ public static ExampleNode parseNode(IType type, IType wireType, Object objectVal
             }
         } else if (type == ClassType.Object) {
             node = new ObjectNode(type, objectValue);
-        } else if (type instanceof ClassType && objectValue instanceof Map) {
+        } else if (type instanceof ClassType && ClientModelUtil.getClientModel(((ClassType) type).getName()) != null
+                && objectValue instanceof Map) {
```
weidongxu-microsoft, line 110: add some comment on which case this branch process?

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
```diff
@@ -106,74 +106,71 @@ public static ExampleNode parseNode(IType type, IType wireType, Object objectVal
...
             ClientModel model = ClientModelUtil.getClientModel(((ClassType) type).getName());
-            if (model != null) {
```
weidongxu-microsoft, line 111: Personally I still prefer call `ClientModelUtil.getClientModel` only once. This is at least `O(n)` op.

weidongxu-microsoft (commented):
Could you have some test case? I still didn't see how BinaryData is handled.

We had "multiple-content-types.tsp" that has binary input. If this is not easy to work with, feel free to add another case.

XiaofeiCao:
> Could you have some test case? From code, I still didn't see how BinaryData is handled.
> 
> We had "multiple-content-types.tsp" that has binary input. If this is not easy to work with, feel free to add another case.

Sure, will add test case.

By checking whether the given type is ClientModel type, BinaryData format(which fails the check) will be left in the default branch:
https://github.com/Azure/autorest.java/blob/7bfdfef0def76e379cacd954bb8444bef96240a8/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java#L180-L186

For BinaryData, it's `BinaryData.fromObject()`:
```java
BinaryData response =
                imageAnalysisClient.segmentFromStream(SegmentationMode.BACKGROUND_REMOVAL, BinaryData.fromObject("{}"));
```

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
XiaofeiCao: Changed.

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
XiaofeiCao: Added.

weidongxu-microsoft:
> Sure, will add test case.
> 
> By checking whether the given type is ClientModel type, BinaryData format(which fails the check) will be left in the default branch:
> 
> https://github.com/Azure/autorest.java/blob/7bfdfef0def76e379cacd954bb8444bef96240a8/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java#L180-L186
> 
> For BinaryData, it's `BinaryData.fromObject()`:
> 
> ```java
> BinaryData response =
>                 imageAnalysisClient.segmentFromStream(SegmentationMode.BACKGROUND_REMOVAL, BinaryData.fromObject("{}"));
> ```
> 
> I also have concern that there maybe other cases than BinaryData, which type is ClassType but not ClientModel, and value is Map. In this case, it may not be appropriate to throw exception. Probably garbage-in-garbage-out?

On one hand, if we expect `BinaryData`, it may actually helpful to handle it explicitly (e.g. as `} else if (type == ClassType.Object) {`), since `BinaryData` does behavior differently from the other "default" ones like string or int.

On the other hand, I am OK if we switch the "throw" to log error or lor warn,

weidongxu-microsoft:
Plan to release 4.1.22 today or tomorrow (there were some bug fixes, as well as graalvm).

Would be great if you can complete this PR before that. I am fine leave test case to a follow-up PR.

XiaofeiCao:
> On one hand, if we expect `BinaryData`, it may actually helpful to handle it explicitly (e.g. as `} else if (type == ClassType.Object) {`), since `BinaryData` does behavior differently from the other "default" ones like string or int.
> 
> On the other hand, I am OK if we switch the "throw" to log error or lor warn,

I've added test case for BinaryData with paramter value "{}" here: https://github.com/Azure/autorest.java/pull/2355/commits/2348d4b2d8e4b7f846168ee3ba81897eaac91a7d#diff-344f7a06bffc9df912dd893d7a59d6aaaa531c465be459846dffbe7130579602

Currently it behaves as `multiContentTypesClient.uploadImageForSingleContentType(BinaryData.fromObject("{}"));`

By special handling, do you mean we need to deal with byte array as paramter value(e.g. [1,2,3])?

weidongxu-microsoft:
May sure it also work if BinaryData provided with e.g "abc"

weidongxu-microsoft (approved):

XiaofeiCao:
For plain string`"abc"`, it'll generate:
```java
multiContentTypesClient.uploadImageForSingleContentType(BinaryData.fromObject("abc"));
```
`System.out.println(BinaryData.fromObject("abc"))` will print `"abc"`.

For int value `123`, will generate `multiContentTypesClient.uploadImageForSingleContentType(BinaryData.fromObject("123"))`, though I'm not sure `123` is a valid example value for BinaryData...

weidongxu-microsoft:
> For plain string`"abc"`, it'll generate:
> 
> ```java
> multiContentTypesClient.uploadImageForSingleContentType(BinaryData.fromObject("abc"));
> ```
> 
> `System.out.println(BinaryData.fromObject("abc"))` will print `"abc"`.
> 
> For int value `123`, will generate `multiContentTypesClient.uploadImageForSingleContentType(BinaryData.fromObject("123"))`, though I'm not sure `123` is a valid example value for BinaryData...

Yeah. Ideally it should use `BinaryData.fromBytes` or `BinaryData.fromString` in such case. I am OK if we adjust it later.

XiaofeiCao:
Sure. I'm not quite sure what types are valid for bytes input. Let me bring that up in tomorrow's meeting.

haolingdong-msft:
Hi @weidongxu-microsoft, I'm sorry I don't quite understand the use case where user defines `bytes` as data type and pass in integer or string as object value. Could you let me know more details about this kind of use case?

weidongxu-microsoft:
> Hi @weidongxu-microsoft, I'm sorry I don't quite understand the use case where user defines `bytes` as data type and pass in integer or string as object value. Could you let me know more details about this kind of use case?

It is just that we are not sure how service typically write the example JSON, for the case of e.g. upload a file.

PR #2358, opened 2023-10-18
Add crossLanguageDefinitionId to apiview-properties.json

srnagar:


typespec-extension/src/common/operation.ts
```diff
@@ -153,6 +155,10 @@ export class Operation extends Aspect implements Operation {
     (this.profile = this.profile || {})[profileName] = apiVersion;
     return this;
   }
+
+  setCrossLanguageDefinitionId(id: string) {
+    this.crossLanguageDefinitionId = id;
+  }
```
weidongxu-microsoft, line 161: I think typescript usually does not need a setter.

typespec-tests/main.tsp
weidongxu-microsoft, line 1: this file should not be committed?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft, line 1: please run `npm run format` under typespec-extension

typespec-extension/src/code-model-builder.ts
```diff
@@ -1730,37 +1733,37 @@ export class CodeModelBuilder {
...
+            namespace: namespace,
           },
-        }),
-      );
+          java: {
+            namespace: getJavaNamespace(namespace),
+          },
+        },
+      });
+      sealedChoiceSchema.crossLanguageDefinitionId = getCrossLanguageDefinitionId(type);
```
weidongxu-microsoft, line 1749: Totally perisomal opinion and likely very subjective: I kind of inclined to put `crossLanguageDefinitionId` into `language.default.crossLanguageDefinitionId` so it applies to any code model object (`crossLanguageDefinitionId` being a pretty universal identity, like `namespace`+`name` and `description` etc.).

No need to change, we can discuss and adjust in follow-up PR.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -48,6 +53,47 @@ public Client processClient(CodeModel codeModel) {
         // map to client model
         Client client = Mappers.getClientMapper().map(codeModel);
 
+        client.getAsyncClients()
+                .forEach(asyncClient -> crossLanguageDefinitionsMap
+                        .put(asyncClient.getPackageName() + "." + asyncClient.getClassName(), asyncClient.getCrossLanguageDefinitionId()));
```
weidongxu-microsoft, line 58: Is it possible that some of these `.getCrossLanguageDefinitionId()` returns `null`?

If we just output `null` in `apiview_properties.json`, then we are good.

srnagar: Yes, the json will have null. Not a big issue as apiview is able to handle it.

weidongxu-microsoft (approved):
@XiaofeiCao Please also take a look on whether this may impact mgmt.

Issue #2354, opened 2023-10-17
tsp, able to easily switch core package/namespace


weidongxu-microsoft:
If possible, switch as one line.

e.g. we'd like
``` java
    public static final ClassType RestProxy = new ClassType.Builder(false)
        .knownClass(com.azure.core.http.rest.RestProxy.class)
        .build();
```

be in `com.azure.core.http.rest` if azure-core, be in say `com.generic.core.http.rest` if generic core.

We may modify `.knownClass` API, if the mapping of azure to generic is predictable.

weidongxu-microsoft:
We also need to handle the pom.xml, use different package in different case.

PR #2360, opened 2023-10-19
tsp, unbranding, package name in ClassType and Annocation

weidongxu-microsoft:
part of https://github.com/Azure/autorest.java/issues/2354

Besides namespace, PR also solve some intricate problem of naming anonymous model from openai https://github.com/bterlson/openai-in-typespec

move to https://github.com/Azure/autorest.java/pull/2370


javagen/src/main/java/com/azure/autorest/model/clientmodel/Annotation.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public Builder knownClass(Class<?> clazz) {
+            this.packageName(clazz.getPackage().getName())
+                    .name(clazz.getSimpleName());
+
+            if (!JavaSettings.getInstance().isBranding()) {
+                this.packageName(clazz.getPackage().getName()
+                        .replace("com.azure.core", "com.generic.core")
+                        .replace("com.azure.json", "com.generic.json"));
```
weidongxu-microsoft, line 112: I will probably abstract these package name to class e.g. ExternalPackage

javagen/src/main/java/com/azure/autorest/model/clientmodel/Package.java
```diff
@@ -0,0 +1,93 @@
...
+            .packageName(GENERIC_CORE_PACKAGE_NAME)
+            .groupId("com.generic")
+            .artifactId("generic-core")
+            .build();
+
+    public static final Package JSON = new Builder()
+            .packageName(GENERIC_JSON_PACKAGE_NAME)
+            .groupId("com.generic")
+            .artifactId("generic-json")
+            .build();
```
weidongxu-microsoft, line 27: Currently, we use this for module-info and pom.xml

The value will be changed to azure-core lib if branding=true

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
```diff
@@ -696,8 +738,48 @@ public Builder prototypeAsDouble() {
...
+                    this.packageName("com.generic.core.http.pipeline");
+                } else if (Objects.equals(clazz, com.azure.core.http.policy.HttpLogOptions.class)
+                        || Objects.equals(clazz, com.azure.core.http.policy.HttpLoggingPolicy.class)) {
+                    this.packageName("com.generic.core.http.policy.logging");
+                } else {
+                    this.packageName(clazz.getPackage().getName()
+                            .replace(Package.AZURE_CORE_PACKAGE_NAME, Package.GENERIC_CORE_PACKAGE_NAME)
+                            .replace(Package.AZURE_JSON_PACKAGE_NAME, Package.GENERIC_JSON_PACKAGE_NAME));
+                }
+            }
```
weidongxu-microsoft, line 780: Though we have most generic/azure package/name here, this does not look good.

We might want to be able to write e.g. `Package.CORE.HttpPipeline` (`Package.CORE` would map it to generic or azure) instead.

Though would keep it temporary here, before generic-core get stable.

javagen/src/main/java/com/azure/autorest/util/ReturnTypeDescriptionAssembler.java
```diff
@@ -135,12 +139,13 @@ private boolean isGenericTypeClassSubclassOf(IType type, Class<?>... parentClass
...
-        }
+        return TYPE_CLASS_MAP.computeIfAbsent(className, key -> {
+            try {
+                return Optional.of(Class.forName(key));
+            } catch (ClassNotFoundException e) {
+                logger.warn(String.format("class %s not found!", key), e);
+                return Optional.empty();
+            }
+        }).orElse(null);
     }
```
weidongxu-microsoft, line 150: A cache here.

We need future work on this class, as generic-core class e.g. `Response` not be able to be loaded from runtime.

typespec-extension/src/emitter.ts
```diff
@@ -17,6 +17,8 @@ export interface EmitterOptions {
   "namespace"?: string;
   "output-dir"?: string;
 
+  "branding"?: boolean;
```
weidongxu-microsoft, line 20: Add a `branding` option for now. May change in future.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1530,7 +1531,8 @@ export class CodeModelBuilder {
           // use it for extensible enum
           schema = this.processChoiceSchema(knownValues, this.getName(knownValues), false);
         } else {
-          schema = this.processSchema(type.type, nameHint + "Model");
+          const schemaNameHint = pascalCase(getModelNameForProperty(type)) + pascalCase(nameHint);
+          schema = this.processSchema(type.type, schemaNameHint);
```
weidongxu-microsoft, line 1535: This part is pretty tricky. In future, we'd switch to name from TCGC.

For now, make it works in java emitter, and output name similar to .NET

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -591,7 +594,11 @@ private static ModuleInfo getModuleInfo(List<String> modelsPackages) {
             }
 
             // open models package to azure-core and jaskson
-            List<String> openToModules = Arrays.asList("com.azure.core", "com.fasterxml.jackson.databind");
+            List<String> openToModules = new ArrayList<>();
+            openToModules.add(Package.CORE.getPackageName());
+            if (!settings.isStreamStyleSerialization()) {
```
haolingdong-msft, line 599: Do we have any documents on StreamStyleSerialization? I found I'm lack of knowledge of StreamStyleSerialization, and want to learn mot. Sorry it may not really related to this pr.

haolingdong-msft:
Do we have tests on how non-azure look like?

haolingdong-msft:
classtype name and annotation look good to me.

haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
weidongxu-microsoft: Not sure if we had good doc on it. You can ping Alan in chat.

It mostly about model. It would use `JsonSerializable` and the class do serialization, not Jackson via reflect.
https://github.com/Azure/autorest.java/tree/main/vanilla-tests/src/main/java/fixtures/streamstyleserialization/models

It actually have tight relation to your Patch work.

weidongxu-microsoft:
> Do we have tests on how non-azure sdk code look like?

No, we don't. As we don't have anything concret on generic-core. So generated code probably going to change a lot in following months.

It looks not very different from azure-core, just it does not use `com.azure.*`. Mostly swtich to `com.generic.*`, include pom etc.

javagen/src/main/java/com/azure/autorest/util/ReturnTypeDescriptionAssembler.java
XiaofeiCao: I wonder why. Does this mean codegen won't depend on generic-core?

weidongxu-microsoft: I don't know.

Strickly speaking, codegen does not need to depends on azure-core either. We just have it here for convenience, not necessity.

XiaofeiCao (approved):

weidongxu-microsoft:
[tsp-output.zip](https://github.com/Azure/autorest.java/files/13162727/tsp-output.zip)

current output from https://github.com/bterlson/openai-in-typespec

weidongxu-microsoft:
We got some conflict with Srikanta's PR https://github.com/Azure/autorest.java/pull/2365

I am breaking this PR to small ones.

weidongxu-microsoft:
move to https://github.com/Azure/autorest.java/pull/2370

PR #2362, opened 2023-10-20
tsp, fix binarydata example

XiaofeiCao:
For BinaryData as ClientType, generate sample code:
```java
BinaryData.fromBytes(exampleValue.toString().getBytes());
```

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
+++ b/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
@@ -12,9 +12,10 @@ import com.azure.autorest.model.clientmodel.ClientModelProperty;
 import com.azure.autorest.model.clientmodel.IType;
 import com.azure.autorest.model.clientmodel.ModelProperty;
 import com.azure.autorest.model.clientmodel.PrimitiveType;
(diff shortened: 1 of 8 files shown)


javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
```diff
@@ -309,9 +313,15 @@ public String accept(ExampleNode node) {
...
+            } else if (node instanceof BinaryDataNode) {
+                this.imports.add(com.azure.core.util.BinaryData.class.getName());
+                return binaryDataNodeExpression((BinaryDataNode) node);
             }
             return null;
         }
     }
 
+    private static String binaryDataNodeExpression(BinaryDataNode binaryDataNode) {
+        return String.format("BinaryData.fromBytes(\"%s\".getBytes())", binaryDataNode.getExampleValue());
```
weidongxu-microsoft, line 325: You probably need some escaping. E.g. what if example value be `"example"` with quote?

XiaofeiCao: Thanks, added escape and test case.

weidongxu-microsoft (approved):

PR #2363, opened 2023-10-23
tsp, example getbytes utf8

XiaofeiCao:
Change `getBytes()` to `getBytes(StandardCharset.UTF_8)` in examples.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
+++ b/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
@@ -146,6 +146,7 @@ public class ModelExampleWriter {
                 }
             } else if (node instanceof BinaryDataNode) {
                 this.imports.add(com.azure.core.util.BinaryData.class.getName());
+                this.imports.add(java.nio.charset.StandardCharsets.class.getName());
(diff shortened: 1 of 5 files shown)


weidongxu-microsoft:
could you put 2 to another PR?

XiaofeiCao:
> could you put 2 to another PR?

Sure.

javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
```diff
@@ -146,6 +146,7 @@ public void accept(ExampleNode node, String getterCode) {
                 }
             } else if (node instanceof BinaryDataNode) {
                 this.imports.add(com.azure.core.util.BinaryData.class.getName());
+                this.imports.add(java.nio.charset.StandardCharsets.class.getName());
                 addEqualsAssertion(binaryDataNodeExpression((BinaryDataNode) node), getterCode);
```
weidongxu-microsoft, line 150: Are you sure this is correct?

There is no `equals` in `BinaryData`?

Just assertNotNull?

XiaofeiCao: I see... Just notNull then...

weidongxu-microsoft (approved):

Issue #2321, opened 2023-09-27
HTML tags getting replaced/removed when replacing file contents through customization


vcolin7:
I'm using AutoRest to customize the contents of a `package-info.java` file in my Azure SDK library and noticed that the string I used to replace the file's contents is not being copied and pasted exactly as I formatted it. For example:

1. The first `<p>` tag of a Javadoc comment was removed.
2. Every instance of `</p>` was removed as well.
3. Every instance of `<br/>` was replaced by `<br>` and moved to the end of the last line with text, when it was originally in a line of its own.
4. An empty line was added after every paragraph that did not already have an empty line follow it.
5. A new line is getting added after every line that starts with `<!--src_embed` and before every line that starts with `<!--end`. This might not be specific to these two types of line but all HTML comments altogether.
6. Comments were reformatted to try to maximize characters in each line, instead of leaving them specifically as I wrote them. Here's an example:
    ```
     * <p>In order to interact with the Azure Key Vault service, you will need to create an instance of the
     * {@link com.azure.security.keyvault.administration.KeyVaultAccessControlAsyncClient} class, a vault url and a
     * credential object.</p>
    ```
    Was turned into

    ```
     * <p>In order to interact with the Azure Key Vault service, you will need to create an instance of the {@link
     * com.azure.security.keyvault.administration.KeyVaultAccessControlAsyncClient} class, a vault url and a credential
     * object.
    ```
    By moving `{@link` from the beginning of the second line to the end of the first one.

I don't know exactly what's being done to process HTML for customization, but I would expect that when replacing a file's contents with my own the source material is not modified at all.

[...]

PR #2364, opened 2023-10-24
Use Maven Spotless plugin in postprocessor to handle all code formatting

alzimmermsft:
This PR removes usage of Google format in favor of having the postprocessor using Maven Spotless plugin with an Eclipse format configuration perform formatting. This removes our managed dependency and uses a format configuration that could be shared so that all formatting is the same.


fluentgen/pom.xml
```diff
@@ -40,12 +40,6 @@
       <version>1.0.0-beta.1</version>
     </dependency>
 
-    <dependency>
-      <groupId>com.google.googlejavaformat</groupId>
-      <artifactId>google-java-format</artifactId>
-      <version>1.17.0-beta.1</version>
-    </dependency>
```
weidongxu-microsoft, line 47: fluent does not use postprocess nor Eclipse language server.

Maybe we first keep it here, and let @XiaofeiCao do the migration later.

I am a bit worried about the extra time when invoking Eclipse language server (it's pretty slow).

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -118,7 +99,7 @@ public boolean processInternal() {
             } else if (className.startsWith("src") && className.endsWith(".java")) {
                 customizationClass = loadCustomizationClassFromJavaCode(className, getBaseDirectory(), logger);
             } else {
-                customizationClass = loadCustomizationClassFromReadme(className, readme);
+                throw new RuntimeException("Unable to load customization class.");
```
weidongxu-microsoft, line 103: why this change about readme?

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
```diff
@@ -321,54 +286,28 @@ private static void attemptMavenInstall(Path pomPath, Logger logger) {
...
-            logger.info("Unable to determine Java version to verify if Java 11+ is being used, which is the "
-                + "requirement to run Autorest code customizations.");
-            return -1;
+        try {
+            Process process = new ProcessBuilder(command)
+                .redirectOutput(ProcessBuilder.Redirect.DISCARD)
+                .redirectError(ProcessBuilder.Redirect.DISCARD)
+                .start();
+            process.waitFor(30, TimeUnit.SECONDS);
+            process.exitValue();
```
weidongxu-microsoft, line 306: why return value ignored?

typespec-extension/src/main/java/com/azure/typespec/Main.java
weidongxu-microsoft, line 1: typespec-java does not pass postprocess like autorest does.

Currently it would call here https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java#L144-L149, which seems won't call the spotless code.

weidongxu-microsoft:
For fluent CI, I think it fails due to https://github.com/Azure/autorest.java/pull/2364#discussion_r1371032869

For typespec CI, you first need to move them to Java 11 as well https://github.com/alzimmermsft/autorest.java/blob/UseSpotlessInPostProcessor/eng/pipelines/ci-typespec-java.yaml#L30
(nightly-build CI as well)
Then this issue https://github.com/Azure/autorest.java/pull/2364#discussion_r1371063630

haolingdong-msft:
@alzimmermsft  May I know the benefit of this change, is it due to https://github.com/Azure/autorest.java/issues/2321#issuecomment-1743477724 the javadoc issue? Is there any other benefits?

fluentgen/pom.xml
alzimmermsft: Fortunately, Spotless doesn't require the Eclipse language server to perform Eclipse code formatting. Don't know the exact design but I know from investigating the removal of Google Format that Eclipse JDT Core has the ability to run code formatting without use of the language server and is just as fast as Google Format. Though, using Spotless requires code to be on disk rather than in memory, which will be slower but shouldn't be by much.

Based on that, thoughts on copying the code I added into postprocessor to fluent, and leaving the Google Format dependency out.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
alzimmermsft: This is removing the ability to configure code customizations using a Java code block in the README. This is pushing the requirement to use a proper Java source code file or precompiled JAR for code customizations.

fluentgen/pom.xml
weidongxu-microsoft: Yeah, we can just copy the code to fluentgen.

Maybe another solution is to put the spotless related code to a module that can be used independently from postprocess (like current "partial-update" module). Anyway if we really want it we can always refactor later.

postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java
weidongxu-microsoft: Got it. This feature is likely rarely used.

weidongxu-microsoft:
> @alzimmermsft May I know the benefit of this change, is it due to [#2321 (comment)](https://github.com/Azure/autorest.java/issues/2321#issuecomment-1743477724) the javadoc issue? Is there any other benefits?

One thing is about the google java format. It is built from our own fork (and we actually built 2 of them :-)) and never released to maven. Using package outside of maven could be a security risk.

fluentgen/pom.xml
alzimmermsft: I would save the larger refactor for later as there is other functionality copied across the plugins which would be useful to centralize.

alzimmermsft:
Closing in favor of #2369 as it shows the code changes plus on regeneration change to make the review easier.

PR #2369, opened 2023-10-26
Use Spotless instead of Google format

alzimmermsft:


weidongxu-microsoft (approved):
LGTM. For fluent CI, it appears to be working as expected.

If possible, you might want to test one data-plane SDK code output from spotless code format and test a draft PR to azure-sdk-for-java repo, to see if it miss some configure.
<-- OK, I remember we already included an CI with checkstyle and spotbugs. Then if CI pass,  you are good.
https://github.com/Azure/autorest.java/tree/main/protocol-sdk-integration-tests

If no one have objection, you can regen everything and merge immediately after CI pass (to avoid being conflicted from other PR).

@XiaofeiCao @haolingdong-msft

alzimmermsft:
I've added a regenerated example.

And Fluent CI does check Checkstyle and Spotbugs which passed, so I believe we're in a good spot.

XiaofeiCao (approved):

weidongxu-microsoft:
I am good with it (again, as long as CI passes).

PS: I've merged https://github.com/Azure/autorest.java/pull/2371

PR #2373, opened 2023-10-30
mpg codegen

XiaofeiCao:
Refactor current typespec-extension to allow for MPG.
Share the same code-model with DPG, different codegen path. 
This PR does not affect existing DPG codegen output.

Generated code: https://github.com/Azure/azure-sdk-for-java/commit/7572c1343becf92023fcb8da5c5221253dbb6352

Known issues: 
1. LRO not generated
2. Operation group id null
3. Paged model name is wrong


fluentgen/src/main/java/com/azure/autorest/fluent/FluentGen.java
```diff
@@ -201,20 +200,22 @@ Client handleMap(CodeModel codeModel) {
...
-            javaPackage
-                    .addServiceClientInterface(interfacePackage, client.getServiceClient().getInterfaceName(), client.getServiceClient());
+        if (CoreUtils.isNullOrEmpty(client.getServiceClients())) {
+            ServiceClient serviceClient = client.getServiceClient();
+            addServiceClient(javaSettings, javaPackage, interfacePackage, serviceClient);
+        } else {
+            // multi-client from TypeSpec
+            for (ServiceClient serviceClient : client.getServiceClients()) {
+                addServiceClient(javaSettings, javaPackage, interfacePackage, serviceClient);
+            }
```
weidongxu-microsoft, line 218: mgmt likely won't need multi-client.

typespec-extension/pom.xml
```diff
@@ -61,7 +61,7 @@
     </dependency>
     <dependency>
       <groupId>com.azure.tools</groupId>
-      <artifactId>azure-autorest-preprocessor</artifactId>
+      <artifactId>azure-autorest-fluentgen</artifactId>
       <version>1.0.0-beta.1</version>
       <exclusions>
         <exclusion>
```
weidongxu-microsoft, line 67: Should we replace preprocessor+javagen with fluentgen?

typespec-extension/src/main/java/com/azure/typespec/util/FileUtil.java
```diff
@@ -0,0 +1,35 @@
...
+package com.azure.typespec.util;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class FileUtil {
```
weidongxu-microsoft, line 13: nit, private ctor?

typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java
```diff
@@ -0,0 +1,153 @@
...
+
+    static {
+        SETTINGS_MAP.put("data-plane", false);
+
+        SETTINGS_MAP.put("sdk-integration", true);
+        SETTINGS_MAP.put("regenerate-pom", true);
+
+        SETTINGS_MAP.put("license-header", "MICROSOFT_MIT_SMALL_TYPESPEC");
+        SETTINGS_MAP.put("enable-sync-stack", false);
+        SETTINGS_MAP.put("enable-page-size", true);
```
weidongxu-microsoft, line 107: probably still false.

I think mostly follow https://github.com/Azure/autorest.java/blob/main/fluentnamer/readme.md?

typespec-extension/pom.xml
XiaofeiCao: OK, I'll remove dependency from javagen.

typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentNamer.java
```diff
@@ -0,0 +1,21 @@
...
+public class TypeSpecFluentNamer extends FluentNamer {
+    public TypeSpecFluentNamer() {
+        super(new TypeSpecPlugin.MockConnection(), "dummy", "dummy");
+    }
+
+    @Override
+    public <T> T getValue(Type type, String key) {
+        // Logger uses getValue to get log level, null will use default
+        return null;
+    }
```
weidongxu-microsoft, line 20: We had some JavaSettings check in FluentNamer e.g. `JavaSettings.getInstance().isGenerateClientInterfaces()`

Will it safer if we have the same MAP here from `TypeSpecFluentPlugin`?

XiaofeiCao: OK, there's FluentJavaSettings instance in `transform` method. We should use the same map here.

typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+
+        JavaSettingsAccessor.setHost(this);
+        LOGGER.info("Output folder: {}", emitterOptions.getOutputDir());
+        LOGGER.info("Namespace: {}", JavaSettings.getInstance().getPackage());
+    }
+
+    public Client processClient(CodeModel codeModel) {
+        // transform code model
+        FluentNamer fluentNamer = new TypeSpecFluentNamer();
+        codeModel = fluentNamer.transform(codeModel);
```
weidongxu-microsoft, line 62: Personally I'd prefer Plugin sets the SETTINGS_MAP into FluentNamer, in ctor or other place, instead of FluentNamer pulling it out from Plugin.

weidongxu-microsoft (approved):

PR #2374, opened 2023-10-30
Use Maven to compile code customization

alzimmermsft:
Removes usage of Eclipse Language Server to compile the code customization Java file and instead use Maven compiler. Maven compiler has a lower overhead and should reduce time taken to run code customizations by about 10-20 seconds per each run and reduce the number of times Eclipse Language Server fails to start.


postprocessor/pom.xml
```diff
@@ -21,8 +21,20 @@
...
+                <configuration>
+                    <source>11</source>
+                    <target>11</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 
-  <dependencies>
+    <dependencies>
```
weidongxu-microsoft, line 37: nit, make it consistent? 2 space or 4 space for the whole file.

weidongxu-microsoft (approved):

PR #2376, opened 2023-10-31
Remove synchronization and semaphore

alzimmermsft:
Remove the synchronization and semaphore block in `Connection` and rewrote `write` to write the entire request at once. Writing the request in two write statements was likely the reason for this synchronized behavior as it allowed for race conditions in writing and parallelization.


extension-base/src/main/java/com/azure/autorest/extension/base/jsonrpc/Connection.java
```diff
@@ -295,28 +294,20 @@ protected void close() throws IOException {
...
-            write(("Content-Length: " + buffer.length + "\r\n\r\n").getBytes(StandardCharsets.US_ASCII));
             write(buffer);
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
-
-        synchronized (streamReady) {
-            streamReady.release();
-        }
     }
```
weidongxu-microsoft, line 311: I am a bit uncertain about this part. It is ancient code and I am not expert on jsonrpc and communication with autorest.

Though if CI is good (we should have lots of autorest call there), it likely be fine.

alzimmermsft: From reading through documentation, JSON RPC communication with Autorest is done using standard in and out which synchronize read and write calls, making this safe to remove.

That said, I'm not really sure this made anything more performant as it's likely the request handling isn't completed fast enough to make the parallelization worthwhile.

I was thinking, what if we had the plugins use the local disk in the same way that TypeSpec does and we can remove the read and write file usage of Autorest.

weidongxu-microsoft: I remember the design principal of the autorest is that it would handle the communication between plugins. Hence stdin stdout for communication, include write files. (so, ideally it would work without file system)

At first (and most of the time), autorest.java adhere to this.

Later, we add feature to generate auxiliary files like pom.xml and readme.md and other files, which require we "read" something from file system e.g. sdk repo structure and version_client.txt.

And partial-update also involve "read" the previous version of the code.

Then finally we had spotless that "write" the file.

alzimmermsft:
Closing this as other changes were made which makes this less important. May revisit in the future in a new PR with the potential to add batch support if Autorest supports it.

PR #2378, opened 2023-11-02
tsp, unbranded, branch for Trait and addQueryParam

weidongxu-microsoft:
remove use of `KeyCredentialTrait`

also avoid 3rd param in `RequestOptions.addQueryParam`

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
@@ -402,10 +402,7 @@ abstract class ConvenienceMethodTemplateBase {
             IType elementType = ((IterableType) parameter.getClientMethodParameter().getWireType()).getElementType();
             String elementTypeExpression = expressionConvertToString("paramItemValue", elementType, parameter.getProxyMethodParameter());
             writeLine = javaBlock -> {
-                String addQueryParamLine = String.format("requestOptions.addQueryParam(%1$s, %2$s, %3$s);",
-                        ClassType.String.defaultValueExpression(parameter.getSerializedName()),
-                        elementTypeExpression,
-                        parameter.getProxyMethodParameter().getAlreadyEncoded());
+                String addQueryParamLine = getAddQueryParamExpression(parameter, elementTypeExpression);
 
                 javaBlock.line(String.format("for (%1$s paramItemValue : %2$s) {", elementType, parameter.getName()));
                 javaBlock.indent(() -> {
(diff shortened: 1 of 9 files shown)


javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -419,10 +416,8 @@ private static void writeQueryParam(MethodParameter parameter, JavaBlock methodB
             };
         } else {
             writeLine = javaBlock -> javaBlock.line(
-                    String.format("requestOptions.addQueryParam(%1$s, %2$s, %3$s);",
-                            ClassType.String.defaultValueExpression(parameter.getSerializedName()),
-                            expressionConvertToString(parameter.getName(), parameter.getClientMethodParameter().getWireType(), parameter.getProxyMethodParameter()),
-                            parameter.getProxyMethodParameter().getAlreadyEncoded()));
+                    getAddQueryParamExpression(parameter,
```
haolingdong-msft, line 419: I did not see generated code get updated, but from my understanding, this change will impact generated code.

weidongxu-microsoft: It is in non-azure code branch, so not affect Azure.

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -434,6 +429,20 @@ private static void writeQueryParam(MethodParameter parameter, JavaBlock methodB
         }
     }
 
+    private static String getAddQueryParamExpression(MethodParameter parameter, String variable) {
+        // TODO: generic not having 3rd parameter "encoded"
+        if (JavaSettings.getInstance().isBranded()) {
```
weidongxu-microsoft, line 434: here

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
haolingdong-msft: ok, I see. BTW, may I know why non-azure wants to remove this parameter?

weidongxu-microsoft: I don't know. Maybe they do not want to support the case that parameter already encoded.

But anyway they are not finalized, so they could add it back later. But I had to generate code that compiles with generic-core, righ now...

weidongxu-microsoft: Same for the Trait. I think they will eventually have Trait.

haolingdong-msft: I see, thanks.

PR #2379, opened 2023-11-03
tsp, fix issue related to Enum/Union with float number

weidongxu-microsoft:
Test https://github.com/Azure/autorest.java/commit/da961584d055b37603480a3fd8d964e1a209951e

There is more test on enum from Union of Literals from cadl-ranch, but we cannot upgrade yet, due to single client issue.
Something like `unit: 1 | 1000 | 0.001` or `size: 0.8 | 1 | 1.2 | float32`


typespec-extension/src/code-model-builder.ts
```diff
@@ -443,6 +444,27 @@ export class CodeModelBuilder {
...
+          nameCount.set(name, count + 1);
+          schema.language.default.name = name + count;
+        }
+      }
+    };
+    this.codeModel.schemas.objects?.forEach((it) => deduplicateName(it));
+    this.codeModel.schemas.groups?.forEach((it) => deduplicateName(it)); // it has RequestConditions
+    this.codeModel.schemas.choices?.forEach((it) => deduplicateName(it));
+    this.codeModel.schemas.sealedChoices?.forEach((it) => deduplicateName(it));
+    this.codeModel.schemas.ors?.forEach((it) => deduplicateName(it));
```
weidongxu-microsoft, line 467: Will remove after move to TCGC.

Currently use it to unblock the cadl-ranch case, so we can test other issues.

typespec-extension/src/type-utils.ts
```diff
@@ -263,6 +263,10 @@ export function getAccess(type: Model | Operation | Enum): string | undefined {
   });
 }
 
+export function isAllValueInteger(values: number[]): boolean {
+  return values.every((it) => Number.isInteger(it));
+}
```
weidongxu-microsoft, line 268: Not a good way to decide the Enum is integer or float. But typespec does not tell the type.

haolingdong-msft (approved):

PR #2381, opened 2023-11-06
Update versions for release

alzimmermsft:
Updates versions of Autorest libraries in preparation for release.


weidongxu-microsoft (approved):

eng/pipelines/fluent_integration.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
       jdkArchitectureOption: 'x64'
       publishJUnitResults: true
       testResultsFiles: '**/TEST-*.xml'
-      goals: 'verify'
+      goals: 'clean install'
```
weidongxu-microsoft, line 45: Thanks. Previous fluent does not depends on azure-autorest-parent hence it does not need a installed lib of it. Now it does.

PR #2387, opened 2023-11-08
tsp, fix script and add override

weidongxu-microsoft:
The override on "@typespec/openapi" is for "@azure-tools/typespec-azure-resource-manager", which depends on openapi

@XiaofeiCao Do we really need "@azure-tools/typespec-azure-resource-manager" at present?

### DIFF
--- a/typespec-tests/Setup.ps1
+++ b/typespec-tests/Setup.ps1
@@ -4,7 +4,7 @@ param (
 )
 
 if ($RebuildJar) {
-    mvn clean install package -f ../pom.xml -Plocal -Ptsp -DskipTests
+    mvn clean install -f ../pom.xml -P local,tsp -DskipTests "-Djacoco.skip"
     if ($LASTEXITCODE -ne 0) {
         exit $LASTEXITCODE
     }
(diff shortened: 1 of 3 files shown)


typespec-tests/Setup.ps1
```diff
@@ -4,7 +4,7 @@ param (
 )
 
 if ($RebuildJar) {
-    mvn clean install package -f ../pom.xml -Plocal -Ptsp -DskipTests
+    mvn clean install -f ../pom.xml -P local,tsp -DskipTests "-Djacoco.skip"
```
weidongxu-microsoft, line 7: if both install and package, it fails

weidongxu-microsoft:
put it into https://github.com/Azure/autorest.java/pull/2353/

XiaofeiCao:
> The override on "@typespec/openapi" is for "@azure-tools/typespec-azure-resource-manager", which depends on openapi
> 
> @XiaofeiCao Do we really need "@azure-tools/typespec-azure-resource-manager" at present?

We don't need that in our emitter/codegen. We can safely delete the dependency.

Issue #2384, opened 2023-11-07
tsp, handle Union as BinaryData


weidongxu-microsoft:
## Design

Intend to align with C#. Also this would allow the class that refers Union be public.

Though it probably not easy for customer to direct use that `BinaryData`. We can add javadoc ref (or `@see`) to refer to the union variants (in future), but it usually not enough.

Hence it likely still better to be customized. We allow partial-update on model class.

---

If Union is `Cat | Dog`, if it is public, `Cat` class and `Dog` class will also be public.

weidongxu-microsoft:
quote from Michael

> BinaryData is the only real safe "fallback" solution but we intend to have smarter options for certain kinds of unions.  Such as a union of all literals of the same type will become a fixed enum, a union of all literl types plus the type of the literals will become an extensible enum, etc.

> We do generate them but in some cases they get removed because they aren't referenced anywhere.  We plan to fix this because we are writing some summary docs which enumerates each type this BinaryData.

weidongxu-microsoft:
got this :-)

```
        } else {
            // TODO (alzimmer): Resolve this as deserialization logic generation needs to handle all cases.
            throw new RuntimeException("Unknown wire type " + wireType + " in serialization. Need to add support for it.");
        }
```

PR #2390, opened 2023-11-08
tsp, TypeSpec Union as BinaryData

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2384

The PR also added flag of `stream-style-serialization` to emitter, and enabled some test with it.
generic-core must use this flag `stream-style-serialization`, as it won't have Jackson (that is the reason I need to test Union under stream-style).


javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -291,7 +291,9 @@ private static void serializeJsonProperty(JavaBlock methodBlock, ClientModelProp
                 methodBlock.line(fieldSerializationMethod + ";");
             }
         } else if (wireType == ClassType.Object) {
-            methodBlock.line("jsonWriter.writeUntyped(" + propertyValueGetter + ");");
+            methodBlock.line("jsonWriter.writeUntypedField(\"" + serializedName + "\", " + propertyValueGetter + ");");
+        } else if (wireType == ClassType.BinaryData) {
+            methodBlock.line("jsonWriter.writeUntypedField(\"" + serializedName + "\", " + propertyValueGetter + ".toObject(Object.class));");
```
weidongxu-microsoft, line 296: I've changed the `Object` path as well. I think previous `writeUntyped` may be a bug.

typespec-tests/src/main/java/com/type/union/UnionClient.java
```diff
@@ -42,7 +46,7 @@ public final class UnionClient {
      * </p>
      * <pre>{@code
      * {
-     *     simpleUnion: ModelWithSimpleUnionPropertySimpleUnionModelBase (Required)
+     *     simpleUnion: BinaryData (Required)
```
weidongxu-microsoft, line 49: javadoc will be addressed in future

We hope it would let use know which type can be assigned to this `BinaryData`

typespec-tests/src/main/java/com/type/union/UnionClient.java
```diff
@@ -235,7 +232,158 @@ public Response<BinaryData> receiveFirstNamedUnionValueWithResponse(RequestOptio
...
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public void sendInt(ModelWithSimpleUnionProperty input) {
+        // Generated convenience method for sendIntWithResponse
+        RequestOptions requestOptions = new RequestOptions();
+        sendIntWithResponse(BinaryData.fromObject(input), requestOptions).getValue();
+    }
```
weidongxu-microsoft, line 255: public method now

typespec-tests/src/test/java/com/type/union/UnionsClientTest.java
weidongxu-microsoft, line 1: Update test cases

If we expose `BinaryData` to customer, we expect them use it like this.

typespec-tests/Generate.ps1
```diff
@@ -45,6 +45,8 @@ function Generate($tspFile) {
     $tspOptions = "--options=""@azure-tools/typespec-java.customization-class=../customization/src/main/java/CustomizationTest.java"""
   } elseif ($tspFile -match "encode[\\/]bytes[\\/]main.tsp") {
     $tspOptions = "--options=""@azure-tools/typespec-java.customization-class=../customization/src/main/java/CustomizationEncodeBytes.java"""
+  } elseif ($tspFile -match "type[\\/]union[\\/]main.tsp") {
+    $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
```
weidongxu-microsoft, line 49: I only enable stream style in test for this union case.

There is still too many case that compile failure when switched to stream.

azure-dataplane-tests/src/main/java/com/azure/containers/containerregistry/implementation/models/AcrErrorInfo.java
```diff
@@ -102,7 +102,7 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
         jsonWriter.writeStringField("code", this.code);
         jsonWriter.writeStringField("message", this.message);
-        jsonWriter.writeUntyped(this.detail);
+        jsonWriter.writeUntypedField("detail", this.detail);
```
weidongxu-microsoft, line 105: This be the result of change on `Object` serialization. I think this be the correct form?

typespec-extension/src/code-model-builder.ts
```diff
@@ -389,15 +387,15 @@ export class CodeModelBuilder {
...
-        // check it does not contain Union
-        const union = unionReferredByType(this.program, model, this.typeUnionRefCache);
-        if (union) {
-          const errorMsg = `Model '${getTypeName(
-            model,
-            this.typeNameOptions,
-          )}' cannot be set as access=public, as it refers Union '${getUnionDescription(union, this.typeNameOptions)}'`;
-          throw new Error(errorMsg);
-        }
+        // // check it does not contain Union
```
mssfang, line 390:
```suggestion
        // check it does not contain Union
```

typespec-tests/src/test/java/com/type/union/UnionsClientTest.java
```diff
@@ -3,56 +3,64 @@
...
+import java.util.List;
+
 class UnionsClientTest {
 
     private final UnionClient client = new UnionClientBuilder().buildClient();
 
     @Test
     void sendInt() {
-        client.sendIntWithResponse(BinaryData.fromString("{ \"simpleUnion\": 1 }"), null);
+        client.sendInt(new ModelWithSimpleUnionProperty(BinaryData.fromObject(1)));
```
weidongxu-microsoft, line 25: Here again, an age-old argument of `BinaryData` vs. `Object`.

In setter, `Object` is simpler, one can write `client.sendInt(new ModelWithSimpleUnionProperty(1));` here. No need for `BinaryData.from##`.

But in response, `BinaryData` give user the ability to deserialize to the exact class.

haolingdong-msft (approved):

PR #2391, opened 2023-11-08
mpg, support lro

XiaofeiCao:
1. When @extension("x-ms-long-running-operation") is present on operation, add x-ms-long-running-operation to operation code model.
2. Switch `fluent` to `arm` in `EmitterOptions`.
3. Remove `typespec-azure-resource-manager` and `typespec-providerhub` from dependencies.

generated LRO method: 
```java
    private Mono<TrackedResourceAssociationPropertiesInner> createOrUpdateAsync(String subscriptionId,
        String resourceGroupName, String trafficControllerName, String associationName, AssociationInner resource) {
        return beginCreateOrUpdateAsync(subscriptionId, resourceGroupName, trafficControllerName, associationName,
            resource).last().flatMap(this::getLroFinalResultOrError);
    }

```

### DIFF
--- a/typespec-extension/src/common/code-model.ts
+++ b/typespec-extension/src/common/code-model.ts
@@ -20,7 +20,7 @@ export interface CodeModel extends Metadata {
 
   clients: Array<Client>;
 
-  arm: boolean;
+  arm?: boolean;
 }
 
 export class CodeModel extends Metadata implements CodeModel {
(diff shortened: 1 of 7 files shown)


typespec-extension/src/common/code-model.ts
```diff
@@ -19,6 +19,8 @@ export interface CodeModel extends Metadata {
   security: Security;
 
   clients: Array<Client>;
+
+  arm: boolean;
```
weidongxu-microsoft, line 23: Is it required or optional?

weidongxu-microsoft (approved):

PR #2395, opened 2023-11-09
tsp, drop PollResult from core-experimental, use PollOperationDetails in core

weidongxu-microsoft:
`PollResult` moved to azure-core and get renamed to `PollOperationDetails`.


haolingdong-msft (approved):

haolingdong-msft:
Is there any difference between `PollResult` and `PollOperationDetails`? Only name difference?

weidongxu-microsoft:
> Is there any difference between `PollResult` and `PollOperationDetails`? Only name difference?

Only name difference (besides moved from core-experimental to core module).

PR #2397, opened 2023-11-10
Add `Next` to generated post pageable method when conflicting with operation name itself

XiaofeiCao:
https://github.com/Azure/azure-rest-api-specs/blob/6301be1289cf6b8cf44074f0e4229c2adf822991/specification/paloaltonetworks/resource-manager/PaloAltoNetworks.Cloudngfw/stable/2023-09-01/PaloAltoNetworks.Cloudngfw.json#L3621-L3624

Currently we don't throw exceptions on this situation. Instead, we won't generate any methods in service clients for this operation.
This PR adds `Next` to the nextPage operation name when such conflict happens, hence client methods are successfully generated.

This will cause breaking changes to the original methods(without `x-ms-pageable`, previous operation will not be considered pageable).

Generated code: https://github.com/Azure/azure-sdk-for-java/commit/4f4dc3679f26f266239405e22a970f680b34509f#diff-5f826fe15f6fbd3cd098c64250563b7c1d21627ebc4cf10cf4524b1d6f639500

### DIFF
--- a/azure-tests/src/main/java/fixtures/custombaseuri/paging/Pagings.java
+++ b/azure-tests/src/main/java/fixtures/custombaseuri/paging/Pagings.java
@@ -84,6 +84,12 @@ public final class Pagings {
             @PathParam(value = "nextLink", encoded = true) String nextLink,
             @HostParam("accountName") String accountName, @HostParam("host") String host,
             @HeaderParam("Accept") String accept, Context context);
+
+        @Get("{nextLink}")
+        @ExpectedResponses({ 200 })
+        @UnexpectedResponseExceptionType(HttpResponseException.class)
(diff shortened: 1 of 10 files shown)


weidongxu-microsoft (commented):
Is all these changes in test caused by the code change? Are they expected (aka are them all as cases that having nextPage operationName referring to itself?)

XiaofeiCao:
> Is all these changes in test caused by the code change? Are they expected (aka are them all as cases that having nextPage operationName referring to itself?)

Sadly yes. Afaik, these are caused by authorization RP, which seems to have m4 generate nextLink operations with names identical to `operationName` defined in `x-ms-pageable` extension. 

Before I figure that out, I'll wait for service swagger fix.

PR #2400, opened 2023-11-10
Fix running customization using Java 17+

alzimmermsft:
Fixes running customizations on Java 17+ by reverting the change to use a newer version of Eclipse language server when Java 17+ is running code generation. The Java 17+ version of Eclipse language server ended up deadlocking based on its logging which needs further investigation before it can be enabled again.

This PR also removes the usage of sockets to communicate with the Eclipse language server and instead relies on standard input and output of the Java process running the Eclipse language server. This should help reduce random port errors when running code customization.


customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageClient.java
```diff
@@ -71,33 +65,8 @@ public EclipseLanguageClient(String workspaceDir) {
...
-        try {
-            serverSocket = new ServerSocket(0);
-            int port = serverSocket.getLocalPort();
-            Thread thread = new Thread(() -> {
-                try {
-                    clientSocket.set(serverSocket.accept());
-                } catch (IOException e) {
-                    throw new RuntimeException(e);
-                }
-            });
```
weidongxu-microsoft, line 83: I am not sure why we remove this?

alzimmermsft: The PR changed it so we don't use sockets anymore to connect to the language server. Now it's using the standard in and out from the JVM process running the language server to communicate, effectively the same that Autorest does with autorest.java!

alzimmermsft:
Replaced by #2405

PR #2401, opened 2023-11-10
[DO NOT MERGE] Showing code changes needed to make the demo work.

vcolin7:
I wrote [a sample app](https://github.com/vcolin7/openai-sample-app) that we can use to demonstrate how our generic-core allows it to communicate with the OpenAI service. There's plenty to do, like making the OpenAI client take models as parameters instead of BinaryData ([like a better version of this one](https://github.com/vcolin7/openai-sample-app/blob/main/src/main/java/org/example/ChatCompletionParameters.java)), but things work as is.

### DIFF
--- a/openai-in-typespec/openai-sdk/src/main/java/com/openai/OpenAIClientBuilder.java
+++ b/openai-in-typespec/openai-sdk/src/main/java/com/openai/OpenAIClientBuilder.java
@@ -62,7 +62,7 @@ public final class OpenAIClientBuilder {
         HttpPipelineBuilder httpPipelineBuilder = new HttpPipelineBuilder();
         List<HttpPipelinePolicy> policies = new ArrayList<>();
         if (keyCredential != null) {
-            policies.add(new KeyCredentialPolicy("Authorization", keyCredential, "Bearer"));
+            policies.add(new KeyCredentialPolicy("authorization", keyCredential, "Bearer"));
         }
         httpPipelineBuilder.policies(policies.toArray(new HttpPipelinePolicy[0]));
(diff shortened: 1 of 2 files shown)


weidongxu-microsoft:
Thanks Victor.

I will fix the "bearer" one ASAP. (currently it follows the TSP scheme)

Let me see if the RestProxy overload is already there. If so, I will fix that one as well.

weidongxu-microsoft:
For "making the OpenAI client take models as parameters instead of BinaryData". It should be there very soon.

I've have PR merged https://github.com/Azure/autorest.java/pull/2390. I just need to test it on the Azure OpenAI.

weidongxu-microsoft:
`x-ms-error-code` seems not consistent. Some API has it and some not.

This relates to your service behavior. It is recommended to have it for all APIs.

weidongxu-microsoft:
@vcolin7 

I've update the "openai" branch to latest, and regenerate the code.

It should include what you have here, as well as the model classes.

Plan to release a 0.11.0 with all these contained.

vcolin7:
> Thanks Victor.
> 
> I will fix the "Bearer" one ASAP. (currently it follows the TSP scheme) I assume "Authentication" (as header name) is case insensitive? Does OpenAI having problem with this as well? https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive
> 
> Let me see if the RestProxy overload is already there. If so, I will fix that one as well.

You're right, the only thing we needed to update was Bearer started with an uppercase "B", the header name can be all lowercase and still work. @samvaity just added the `RestProxy` overload, so we should be good on that department as well.

PR #2403, opened 2023-11-14
tsp, fix NPE when using BinaryData

weidongxu-microsoft:
bug fix for https://github.com/Azure/autorest.java/issues/2384

### DIFF
--- a/typespec-tests/Generate.ps1
+++ b/typespec-tests/Generate.ps1
@@ -47,6 +47,8 @@ function Generate($tspFile) {
     $tspOptions = "--options=""@azure-tools/typespec-java.customization-class=../customization/src/main/java/CustomizationEncodeBytes.java"""
   } elseif ($tspFile -match "type[\\/]union[\\/]main.tsp") {
     $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
+  } elseif ($tspFile -match "tsp[\\/]union[\\/]main.tsp") {
+    $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
   }
 
   $tspTrace = "--trace import-resolution --trace projection --trace typespec-java"
--- a/typespec-tests/Generate.ps1
+++ b/typespec-tests/Generate.ps1
@@ -47,7 +47,7 @@ function Generate($tspFile) {
     $tspOptions = "--options=""@azure-tools/typespec-java.customization-class=../customization/src/main/java/CustomizationEncodeBytes.java"""
   } elseif ($tspFile -match "type[\\/]union[\\/]main.tsp") {
     $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
-  } elseif ($tspFile -match "tsp[\\/]union[\\/]main.tsp") {
+  } elseif ($tspFile -match "tsp[\\/]union.tsp") {
     $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
   }
 
--- a/typespec-tests/src/main/java/com/cadl/union/models/Result.java
+++ b/typespec-tests/src/main/java/com/cadl/union/models/Result.java
@@ -7,34 +7,36 @@ package com.cadl.union.models;
 import com.azure.core.annotation.Fluent;
 import com.azure.core.annotation.Generated;
 import com.azure.core.util.BinaryData;
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.azure.json.JsonReader;
+import com.azure.json.JsonSerializable;
+import com.azure.json.JsonToken;
+import com.azure.json.JsonWriter;
+import java.io.IOException;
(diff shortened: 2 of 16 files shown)


typespec-tests/src/main/java/com/cadl/union/models/SubResult.java
```diff
@@ -0,0 +1,173 @@
...
+                    dataFound = true;
+                } else if ("result".equals(fieldName)) {
+                    result = Result.fromJson(reader);
+                } else if ("text".equals(fieldName)) {
+                    text = reader.getString();
+                } else if ("arrayData".equals(fieldName)) {
+                    Object arrayDataAsObject = reader.readUntyped();
+                    if (arrayDataAsObject != null) {
+                        arrayData = BinaryData.fromObject(arrayDataAsObject);
+                    }
```
weidongxu-microsoft, line 148: null protection

typespec-tests/src/main/java/com/cadl/union/models/SubResult.java
```diff
@@ -0,0 +1,173 @@
...
+    @Override
+    public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
+        jsonWriter.writeStartObject();
+        jsonWriter.writeStringField("name", getName());
+        jsonWriter.writeUntypedField("data", getData().toObject(Object.class));
+        jsonWriter.writeJsonField("result", getResult());
+        jsonWriter.writeStringField("text", this.text);
+        if (this.arrayData != null) {
+            jsonWriter.writeUntypedField("arrayData", this.arrayData.toObject(Object.class));
+        }
```
weidongxu-microsoft, line 108: null protection

typespec-tests/src/main/java/com/cadl/union/models/Result.java
```diff
@@ -78,4 +91,64 @@ public Result getResult() {
     public BinaryData getData() {
         return this.data;
     }
+
+    @Override
+    public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
+        jsonWriter.writeStartObject();
+        jsonWriter.writeStringField("name", this.name);
+        jsonWriter.writeUntypedField("data", this.data.toObject(Object.class));
```
weidongxu-microsoft, line 99: no null protection, when property is required.

typespec-tests/Generate.ps1
```diff
@@ -47,6 +47,8 @@ function Generate($tspFile) {
     $tspOptions = "--options=""@azure-tools/typespec-java.customization-class=../customization/src/main/java/CustomizationEncodeBytes.java"""
   } elseif ($tspFile -match "type[\\/]union[\\/]main.tsp") {
     $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
+  } elseif ($tspFile -match "tsp[\\/]union.tsp") {
+    $tspOptions = "--options=""@azure-tools/typespec-java.stream-style-serialization=true"""
```
weidongxu-microsoft, line 51: Use stream-style for this Union test case

typespec-tests/src/test/java/com/cadl/union/UnionTests.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Collections;
+
+public class UnionTests {
+
+    private static final byte[] bytes = new byte[] { 'a' };
+
+    @Test
+    public void testSerialization() {
+        // test non-null Union (required property)
```
haolingdong-msft, line 20: Do we have javadoc in client method to show user how to pass in `BinaryData` (union parameter)? It could be hard for user  to create `BinaryData` object.

weidongxu-microsoft: Yes, we know. No time to do this yet.

weidongxu-microsoft: For Azure lib, we do not expect to expose this to user.

haolingdong-msft: sure, I'm fine with adding the javadoc later.

typespec-tests/src/test/java/com/cadl/union/UnionTests.java
```diff
@@ -0,0 +1,40 @@
...
+        Result result = new Result("name", BinaryData.fromObject(bytes));
+        BinaryData json = BinaryData.fromObject(result);
```
haolingdong-msft, line 22: nit: Putting the test here is a bit strange to me as we are testing the `BinaryData` serialization/deserialization logic, seems it has nothing to do with tsp. But seems we don't have test suites to test the stream-style serialization's serialization and desefialization cases.. So I'm OK with it for now.

haolingdong-msft (approved):

typespec-tests/src/test/java/com/cadl/union/UnionTests.java
weidongxu-microsoft: BinaryData is just a wrapper for the serialization and de-serialization of JSON <-> class.

PR #2410, opened 2023-11-16
Fix stream style serialization bugs

### LINKED ISSUE #2396: Enable stream-style-serialization for TypeSpec
PR: https://github.com/Azure/autorest.java/pull/2398
CI : https://github.com/Azure/autorest.java/pull/2398/checks?check_run_id=18553581122

```[tasklist]
### Tasks
- [x] Support Enum contains long/double deserialization
- [ ] `ResponseError` deserialization
- [x] `DateTimeRfc7231` deserialization
- [x] `Base64Url` array deserialization
- [x] `Base64Url` bytes deserialization
- [x] `Double` to `Duration` deserialization
```

alzimmermsft:
Fixes #2396

- Adds better support for non-String based enums
- Uses a different temporary holder for nullable values that need to be converted from wire type to client type
- Handles required polymorphic discriminators correctly
- Adds support for `ResponseError` while `azure-core` hasn't shipped the model implementing `JsonSerializable`
- Cleans up returning models with only constructor properties
- Cleans up error throwing for models with only one required property


haolingdong-msft:
There are still several CI failures, we may need to regen for both typespec and non-typespec tests.

typespec-tests/src/main/java/com/cadl/partialupdate/partialupdate/PartialUpdateAsyncClient.java
weidongxu-microsoft, line 1: This folder (and files within) seems should not exist. Script error?

typespec-tests/customization/src/main/java/CustomizationEncodeBytes.java
```diff
@@ -22,6 +22,8 @@ public void customize(LibraryCustomization customization, Logger logger) {
         PackageCustomization packageCustomization = customization.getPackage("com.encode.bytes.models");
         ClassCustomization classCustomization = packageCustomization.getClass("Base64UrlArrayBytesProperty");
 
+        classCustomization.getProperty("value").setModifier(Modifier.PRIVATE);
```
weidongxu-microsoft, line 25: Is it affect serialization? Or it is just a test on customization.

alzimmermsft: The customization was wrong for stream-style serialization. Since stream-style has knowledge of the model being deserialized it can better generate models using `final` variable (constructor only, read only, constants) whereas Jackson based deserialization the properties are left non-final to allow Jackson to reflectively set them after creating an instance. So, this customization was adding another constructor to the model which had constructor only properties that were `final` and this is just an update to the customization to also remove the final from the property to make it valid code.

typespec-tests/src/main/java/com/cadl/partialupdate/partialupdate/PartialUpdateAsyncClient.java
alzimmermsft: Yeah, I'm not sure what happened here, looking into it. Will also be fixing the CI errors.

typespec-tests/src/main/java/com/cadl/enumservice/models/Operation.java
```diff
@@ -202,4 +192,56 @@ public ColorModel getColorModelValue() {
...
+    @Override
+    public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
+        jsonWriter.writeStartObject();
+        jsonWriter.writeStringField("name", this.name == null ? null : this.name.toString());
+        jsonWriter.writeBooleanField("best", this.best);
+        jsonWriter.writeLongField("age", this.age);
+        jsonWriter.writeLongField("priority", this.priority == null ? null : this.priority.toLong());
+        jsonWriter.writeStringField("color", this.color == null ? null : this.color.toString());
+        jsonWriter.writeDoubleField("unit", this.unit == null ? null : this.unit.toDouble());
+        jsonWriter.writeLongField("priorityValue", this.priorityValue == null ? null : this.priorityValue.toLong());
```
haolingdong-msft, line 205:
```ts
model Operation {
  name: "Read" | "Write";
  best: true;
  age: 50;
  priority: Priority;
  color: ColorModel;
  unit: Unit;
  priorityValue: Priority.Low;
  colorValue: Color.Green;
  unitValue: Unit.Milligram;
}
```
I wonder if we need the `null` check here. I think for this model, `priorityValue` should only be `Priority.Low`, cannot be `null`. If the field in tsp is `priorityValue: Priority.Low | null`, then it can be `null`.

typespec-tests/src/main/java/com/cadl/naming/models/RequestType.java
```diff
@@ -66,7 +66,7 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
 
     /**
      * Reads a RequestType from the JSON stream.
-     * &lt;p&gt;.
+     * <p>
```
haolingdong-msft, line 69: will need to add `@Generated` annotation to `fromJson` and `toJson`. Can be done in following pr.

javagen/src/main/resources/CoreToCodegenBridgeUtils.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * Converts a {@link Duration} to a string in ISO-8601 format with support for a day component.
+     *
+     * @param duration The {@link Duration} to convert.
+     * @return The {@link Duration} as a string in ISO-8601 format with support for a day component, or null if the
+     * provided {@link Duration} was null.
+     */
+    public static String durationToStringWithDays(Duration duration) {
```
weidongxu-microsoft, line 100: I assume the case of `P24H` is not exactly an error, it can be `P24H` or `P1D`. Here we just make the behavior better?

javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
```diff
@@ -547,4 +549,33 @@ public static boolean isWireTypeMismatch(ClientModelProperty clientModelProperty
...
+        }
+
+        boolean ret = model.getProperties().stream()
+            .anyMatch(p -> p.getClientType() == ClassType.ResponseError || p.getClientType() == ClassType.Duration);
+
+        // flatten properties
+        if (!ret && settings.getClientFlattenAnnotationTarget() == JavaSettings.ClientFlattenAnnotationTarget.NONE) {
+            ret = model.getPropertyReferences().stream()
+                .filter(ClientModelPropertyReference::isFromFlattenedProperty)
+                .anyMatch(p -> p.getClientType() == ClassType.ResponseError || p.getClientType() == ClassType.Duration);
```
weidongxu-microsoft, line 576: Is there possible other cases? E.g. `List<Duration>` or `Map<>`?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
```diff
@@ -419,6 +438,7 @@ private static ClassType.Builder getClassTypeBuilder(Class<?> classKey) {
 
     public static final ClassType JsonPatchDocument = new ClassType.Builder(false)
         .knownClass(com.azure.core.models.JsonPatchDocument.class)
+        .jsonToken("JsonToken.START_OBJECT")
         .build();
 
     public static final ClassType BinaryData = getClassTypeBuilder(com.azure.core.util.BinaryData.class)
```
weidongxu-microsoft, line 444: `BinaryData` may also need a `jsonToken`, likely same as `Object`.

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
```diff
@@ -173,8 +218,27 @@ private void writeEnum(EnumType enumType, JavaFile javaFile, JavaSettings settin
...
+                String serializationCall = enumType.getElementType().jsonSerializationMethodCall("jsonWriter", null, "value");
+                enumBlock.annotation("Override");
+                enumBlock.publicMethod("JsonWriter toJson(JsonWriter jsonWriter) throws IOException",
+                    function -> function.methodReturn(serializationCall));
+
+                String deserializationCall = enumType.getElementType().jsonDeserializationMethod("jsonReader");
+                enumBlock.javadocComment(javadoc -> {
+                    javadoc.description("Reads a " + enumName + " from the JSON stream.");
+                    javadoc.line("<p>");
+                    javadoc.description("The passed JsonReader must be positioned at a " + enumType.getElementType().jsonToken() + " value.");
```
weidongxu-microsoft, line 234: This seems the only use of `jsonToken()` I found in PR.

Does it have other usages?

weidongxu-microsoft (approved):
To unblock.

There are some pending comments/questions. If we got bug, we can fix in follow-up PR.

typespec-tests/src/main/java/com/cadl/enumservice/models/Operation.java
alzimmermsft: The `null` check isn't needed but I'd rather get this changed in a future PR as this doesn't do anything and will likely be handled by JIT.

javagen/src/main/resources/CoreToCodegenBridgeUtils.java
alzimmermsft: I added this to align with what Jackson serialization was doing. `P24H` and `P1D` should be treated that same as a Duration.

We could decide to remove this and change the test logic to directly compare the duration value and not the String value.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java
alzimmermsft: Yeah, wasn't sure what to do for `BinaryData` as the underlying content could be `START_OBJECT` (fromObject), `START_ARRAY` (fromBytes), or `STRING` (fromString).

javagen/src/main/java/com/azure/autorest/template/EnumTemplate.java
alzimmermsft: This was the only usage I have for now. There may be more uses for this in the future but for now just this.

javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
alzimmermsft: Yeah, it could be used in `List` and `Map` but I didn't see any cases at this time.

typespec-tests/src/main/java/com/cadl/naming/models/RequestType.java
alzimmermsft: Will add this in a follow up PR

PR #2413, opened 2023-11-17
eng, ignore "generated" folder in sdk sync, if that folder not in sdk repo

weidongxu-microsoft:


eng/sdk/sync_sdk.py
```diff
@@ -44,26 +47,45 @@ def update_emitter(version: str):
...
             os.path.join(sdk_root, 'eng/common/scripts/TypeSpec-Project-Sync.ps1'),
-            artifact_path
+            tsp_location_path
         ]
         subprocess.check_call(cmd, cwd=sdk_root)
 
         cmd[1] = os.path.join(sdk_root, 'eng/common/scripts/TypeSpec-Project-Generate.ps1')
         subprocess.check_call(cmd, cwd=sdk_root)
 
+        if not generated_samples_exists:
```
haolingdong-msft, line 84: What is the case where `generated folder` not in sdk repo?

weidongxu-microsoft: service can generate it, copy some, then delete the "generated" folder.

in some case, some string in there would trigger cspell.

weidongxu-microsoft: But when service generate next api-version, regen these test/sample still helpful.

haolingdong-msft: I see. Thanks

PR #2415, opened 2023-11-20
Parallelize Generate scripts

alzimmermsft:
Parallelizes `/Generate.ps1` and `/typespec-tests/Generate.ps1`.

### DIFF
--- a/Generate.ps1
+++ b/Generate.ps1
@@ -17,10 +17,14 @@ $SWAGGER_PATH = "node_modules/@microsoft.azure/autorest.testserver/swagger"
 $AZURE_DATAPLANE_ARGUMENTS = "--use=./ --output-folder=./azure-dataplane-tests"
 $AZURE_DATAPLANE_PATH = "azure-dataplane-tests/swagger"
 $AZURE_SDK_FOR_JAVA = "https://github.com/Azure/azure-sdk-for-java/blob/main/sdk"
-$PARALLELIZATION = 4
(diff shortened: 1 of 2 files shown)


Generate.ps1
```diff
@@ -17,11 +17,15 @@ $SWAGGER_PATH = "node_modules/@microsoft.azure/autorest.testserver/swagger"
...
 $AZURE_DATAPLANE_PATH = "azure-dataplane-tests/swagger"
 $AZURE_SDK_FOR_JAVA = "https://github.com/Azure/azure-sdk-for-java/blob/main/sdk"
-$PARALLELIZATION = 4
-if ($IsWindows) {
-    $PARALLELIZATION = (Get-CIMInstance -Class 'CIM_Processor').NumberOfCores - 1
+$PARALLELIZATION = [Environment]::ProcessorCount - 1
+if ($PARALLELIZATION -lt 1) {
+  $PARALLELIZATION = 1
+} elseif ($PARALLELIZATION -gt 12) {
+  $PARALLELIZATION = 12
```
srnagar, line 24: Why is this hard coded to 12?

srnagar (approved):

Generate.ps1
alzimmermsft: No major reason other than when I allowed it to be purely `thread count - 1` it used too many resources on my machine. I can remove these if we think that'd be better.

PR #2416, opened 2023-11-20
Add version tooling and CI caching

alzimmermsft:
Adds version tooling and DevOps caching logic from `azure-sdk-for-java`.


eng/versioning/external_dependencies.txt
```diff
@@ -1,45 +1,68 @@
...
-org.apache.maven.plugins:maven-jxr-plugin
-org.apache.maven.plugins:maven-plugin-plugin
-org.apache.maven.plugins:maven-project-info-reports-plugin
-org.apache.maven.plugins:maven-release-plugin
-org.apache.maven.plugins:maven-resources-plugin
-org.apache.maven.plugins:maven-shade-plugin
-org.apache.maven.plugins:maven-site-plugin
-org.apache.maven.plugins:maven-source-plugin
-org.apache.maven.plugins:maven-surefire-plugin
+# Format;
```
srnagar, line 1: I wonder if we need this for autorest as we don't ship these libraries to maven central. We could just use Maven's `<dependencyManagement>` to maintain the versioning in the root pom and the children could just add the group/artifact in `<dependencies>`. Kind of similar to BOM but for multi-module project.

eng/versioning/__pycache__/utils.cpython-311.pyc
weidongxu-microsoft, line 1: should not be here?

weidongxu-microsoft (commented):
I still think https://github.com/Azure/autorest.java/blob/299c098887d57417f28e18ea921314455b9ff168/eng/versioning/sync_versions.py is much light-weighted and it also update version in code.

eng/versioning/external_dependencies.txt
weidongxu-microsoft: Not sure whether we'd like to maintain another file like this.

Should the script just copy the version_client and external_dependencies from SDK repo, and concat them?

fluentgen/src/main/resources/eclipse-format-azure-sdk-for-java.xml
haolingdong-msft, line 1: why delete this?

PR #2418, opened 2023-11-21
Limit code generation parallelization to local builds

alzimmermsft:


Generate.ps1
```diff
@@ -20,6 +20,13 @@ $AZURE_SDK_FOR_JAVA = "https://github.com/Azure/azure-sdk-for-java/blob/main/sdk
 $PARALLELIZATION = [Environment]::ProcessorCount - 1
 if ($PARALLELIZATION -lt 1) {
   $PARALLELIZATION = 1
+} elseif ($PARALLELIZATION -gt 8) {
+    $PARALLELIZATION = 8
+}
```
srnagar, line 25: remove this hard coded value.

srnagar (approved):

Generate.ps1
alzimmermsft: Made this overridable

Issue #1530, opened 2022-06-22
Patch operation supoprt, Generate models with nullable properties for patch operations


srnagar:
Use `Option<T>` in azure-core to generate models that support nullable properties in patch operations.

https://github.com/anuchandy/azure-sdk-for-java/blob/patch-model-example/sdk/metricsadvisor/azure-ai-metricsadvisor/src/main/java/com/azure/ai/metricsadvisor/patch/models/AnomalyDetectionConfigurationPatch.java

Generated sdk design:
https://github.com/srnagar/azure-sdk-for-java/tree/loadtesting-patch

Codegen tasks listed below:
```[tasklist]
### Tasks
- [x] Set stream-style-serialization=true for all models and check gaps
- [ ] https://github.com/Azure/autorest.java/issues/2396
- [x] Emitter: Add json-merge-patch into model usage
- [x] JsonMergePatchHelper: Add Helper class to get/set accessors
- [x] Model: Add jsonMergePatch flag and related static block to set flag
- [x] Model: Add updatedProperties map and the setters
- [x] Model: serialization and deserialization, `toJson()` and `toJsonMergePatch()`
- [x] Client: client method change on calling Helper
- [x] Client: support patch on optional body
- [x] Emitter: remove the code to disable convenient method for json-merge-patch, when convenientAPI=false, no need to add json-merge-patch handling
- [x] Client: use BinaryData.toBytes() to fire serialization
- [x] Emitter: turnoff convenient method for json-merge-patch if stream-style-serialization is disabled
- [x] cadl-ranch tests
- [ ] https://github.com/Azure/autorest.java/issues/2468
- [ ] https://github.com/Azure/autorest.java/issues/2471
```

anuchandy:
Hi @srnagar, as discussed, posting the related gist here (its a bit old but probably useful for this work) https://gist.github.com/anuchandy/ca914f4a8dfaa49ea8e082de7e619c8f

anuchandy:
Also a PPT on merge-patch+json and json-patch+json
[JsonMergePatch.pptx](https://github.com/Azure/autorest.java/files/8960834/JsonMergePatch.pptx)

weidongxu-microsoft:
5. All the model that used by the PATCH model would be also a PATCH model.

weidongxu-microsoft:
PS: the method (vanilla or convenience) is not in the scope of the DPG.

weidongxu-microsoft:
still waiting for arch design

[...]

PR #2419, opened 2023-11-22
tsp, json-merge-patch, add `JsonMergePatchHelper`

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/1530

Code Changes:
1. Add `JsonMergePatchTemplate.java` and related facotry code.
2. `TypeSpecPlugin.java`: generate JsonMergePatchHelper only when there is model used in json-merge-patch operation and needs to generate model.
3. `TemplateUtil.java`: add JsonMergePatchHelper class's name into it.

Generated `JsonMergePatchHelper` example: typespec-tests/src/main/java/com/_specs_/azure/core/basic/implementation/JsonMergePatchHelper.java

PS: Name the helper class `JsonMergePatchHelper` rather than `XXXHelper` (XXX is service name), because later we will need to use the helper in model, but `ModelTemplate` does not have the context of service name. Discussed with @weidongxu-microsoft  for simplicity, we will call it `JsonMergePatchHelper` for now.

image


extension-base/src/main/java/com/azure/autorest/extension/base/plugin/JavaSettings.java
```diff
@@ -1230,4 +1230,11 @@ private static void loadStringOrArraySettingAsArray(String settingName, Consumer
             action.accept(settingValues);
         }
     }
+
+    private final String jsonMergePatchHelperClassName = "JsonMergePatchHelper";
+
+    public String getJsonMergePatchHelperClassName() {
+        return jsonMergePatchHelperClassName;
+    }
```
weidongxu-microsoft, line 1238: just put it in some util class.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -0,0 +1,114 @@
...
+import com.azure.autorest.model.javamodel.JavaClass;
+import com.azure.autorest.model.javamodel.JavaFile;
+import com.azure.autorest.model.javamodel.JavaVisibility;
+import com.azure.autorest.util.ClientModelUtil;
+import com.azure.autorest.util.CodeNamer;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class JsonMergePatchHelperTemplate implements IJavaTemplate<Client, JavaFile>{
```
weidongxu-microsoft, line 17: Does it really need the whole `Client`?

If not, provide what is necessary, e.g. `List<ClientModel>`

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -0,0 +1,114 @@
...
+    public static JsonMergePatchHelperTemplate getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public void write(Client client, JavaFile javaFile) {
+        // imports
+        JavaSettings settings = JavaSettings.getInstance();
+        Set<String> imports = new HashSet<>();
+        addImports(imports, client, settings);
```
weidongxu-microsoft, line 33: same, why import on `Client`? Does the code has anything to do with `Client`?

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -57,6 +58,15 @@ protected void writeClientModels(Client client, JavaPackage javaPackage, JavaSet
...
                 .forEach(model -> javaPackage.addModel(model.getPackage(), model.getName(), model));
 
+        // JsonMergePatchHelper
+        boolean hasJsonMergePatchModelToGenerate = client.getModels().stream()
+                .filter(ModelUtil::isGeneratingModel)
+                .filter(ClientModelUtil::isJsonMergePatchModel)
+                .findFirst().isPresent();
+        if (hasJsonMergePatchModelToGenerate) {
+            javaPackage.addJsonMergePatchHelper(client);
+        }
```
weidongxu-microsoft, line 68: Adjust this accordingly.

weidongxu-microsoft (commented):
Before merge, make sure to turn off one link of the flow so no additional file will be generated till the whole thing is ready to work.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -57,6 +58,15 @@ protected void writeClientModels(Client client, JavaPackage javaPackage, JavaSet
...
                 .forEach(model -> javaPackage.addModel(model.getPackage(), model.getName(), model));
 
+        // JsonMergePatchHelper
+        boolean hasJsonMergePatchModelToGenerate = client.getModels().stream()
+                .filter(ModelUtil::isGeneratingModel)
+                .filter(ClientModelUtil::isJsonMergePatchModel)
+                .findFirst().isPresent();
+//        if (hasJsonMergePatchModelToGenerate) {
+//            javaPackage.addJsonMergePatchHelper(client.getModels());
+//        }
```
weidongxu-microsoft, line 68: recommend to filter the models (as you already done here), and only send these to the template.

and template no longer need to do the filter again.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+
+    /**
+     * Add getter and setter for each model that is used in json-merge-patch.
+     *
+     * @param models
+     * @param javaClass
+     */
+    private void addGettersAndSetters(List<ClientModel> models, JavaClass javaClass) {
```
weidongxu-microsoft, line 101: use static method whenever possible.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -0,0 +1,118 @@
...
+        if (models != null && !models.isEmpty()) {
+            models.forEach(model -> {
+                if (ClientModelUtil.isJsonMergePatchModel(model)) {
+                    // setters
+                    javaClass.publicStaticMethod(String.format("void set%sAccessor(%sAccessor accessor)", model.getName(), model.getName()),methodBlock -> {
+                        methodBlock.line(String.format("%sAccessor = accessor;", CodeNamer.getModelNamer().modelPropertySetterName(model.getName())));
+                    });
+                    // getters
+                    javaClass.publicStaticMethod(String.format("%sAccessor get%sAccessor()", model.getName(), model.getName()), methodBlock -> {
+                        methodBlock.line(String.format("return %sAccessor;",CodeNamer.getModelNamer().modelPropertySetterName(model.getName())));
```
weidongxu-microsoft, line 111: it should have a `methodReturn` method.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -0,0 +1,118 @@
...
+    private void addGettersAndSetters(List<ClientModel> models, JavaClass javaClass) {
+        if (models != null && !models.isEmpty()) {
+            models.forEach(model -> {
+                if (ClientModelUtil.isJsonMergePatchModel(model)) {
+                    // setters
+                    javaClass.publicStaticMethod(String.format("void set%sAccessor(%sAccessor accessor)", model.getName(), model.getName()),methodBlock -> {
+                        methodBlock.line(String.format("%sAccessor = accessor;", CodeNamer.getModelNamer().modelPropertySetterName(model.getName())));
+                    });
+                    // getters
+                    javaClass.publicStaticMethod(String.format("%sAccessor get%sAccessor()", model.getName(), model.getName()), methodBlock -> {
```
weidongxu-microsoft, line 110: Use e.g. `%1$s` whenever you need to use same string more than once in format.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
haolingdong-msft: sure, will pass in the filtered models to addJsonMergePatchHelper(), but 'm not sure if we want to delete the checking logic in template, as template should not assume all the passed-in models are json merge patch model, it's safer to add  the check in template as well? What do you think?

weidongxu-microsoft: Of course don't delete the check logic. It just becomes "if list is not empty"

haolingdong-msft: I mean  the logic on checking whether model is json merge patch model. Do we want to keep it for safety?

weidongxu-microsoft: No. For safety you should do null check on every single string usage, do you?

haolingdong-msft: ok, fine.

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
```diff
@@ -59,6 +59,8 @@ public class TemplateUtil {
 
     public static final String DATE_UTC = "date-utc";
 
+    public static final String JSON_MERGE_PATCH_HELPER_CLASS_NAME = "JsonMergePatchHelper";
+
```
weidongxu-microsoft, line 63: move to `TemplateHelper` or `ClientModelUtil`

at present `TemplateUtil` is for the replacement of placeholder strings.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaPackage.java
```diff
@@ -321,4 +322,11 @@ protected void checkDuplicateFile(String filePath) {
             logger.warn(String.format("Name conflict for output file '%1$s'.", filePath));
         }
     }
+
+    public void addJsonMergePatchHelper(List<ClientModel> models) {
+        JavaFile javaFile = javaFileFactory.createSourceFile(settings.getPackage("implementation"), ClientModelUtil.JSON_MERGE_PATCH_HELPER_CLASS_NAME);
```
weidongxu-microsoft, line 327:
```suggestion
        JavaFile javaFile = javaFileFactory.createSourceFile(settings.getPackage(settings.getImplementationSubpackage()), ClientModelUtil.JSON_MERGE_PATCH_HELPER_CLASS_NAME);
```

weidongxu-microsoft (approved):

PR #2421, opened 2023-11-24
tsp, support simple multipart/form-data

weidongxu-microsoft:
Tested on https://github.com/Azure/cadl-ranch/pull/452

Use `InputStream` as content of `BinaryData`, because sync pipeline will throw if it find `Flux`, but async pipeline won't throw on `InputStream` (though it may have performance penalty).

Ideally we would like to be able to concat `BinaryData` directly and have `replayable` feature maintained with it.

Field of content-type "application/json" is not supported.

At current state (`BinaryData` or `InputStream`), it probably won't able to support retry, if first REST call fails.
If possible, we'd wish we can cancat on `BinaryData` directly, e.g. `BinaryData.concat`


javagen/src/main/java/com/azure/autorest/model/javamodel/JavaPackage.java
```diff
@@ -213,6 +218,20 @@ public final void addPom(String name, Pom pom) {
         xmlFiles.add(xmlFile);
     }
 
+    public final void addJavaFromResources(String packageName, String name) {
```
weidongxu-microsoft, line 221: Add a convenience method to load Java file resources and write out, since Alan got such use case as well.

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -503,6 +519,52 @@ private static String expressionConvertToType(String name, MethodParameter conve
...
+                        ClassType.String.defaultValueExpression(property.getSerializedName()),
+                        name,
+                        property.getGetterName()
+                ));
+            }
+            // TODO (weidxu): application/json
+        }
+        builder.append(".end().getRequestBody()");
+        return builder.toString();
+    }
```
weidongxu-microsoft, line 566: serialize a model of multipart/form-data to `BinaryData`, with the help of `MultipartFormDataHelper`

javagen/src/main/resources/MultipartFormDataHelper.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.SequenceInputStream;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+
+// DO NOT modify this helper class
+
+public final class MultipartFormDataHelper {
```
weidongxu-microsoft, line 14: The helper class

typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
```diff
@@ -0,0 +1,154 @@
...
+     * The image property.
+     */
+    @Generated
+    private final BinaryData image;
+
+    /*
+     * The filename for image
+     */
+    @Generated
+    private String imageFilename = "image";
```
weidongxu-microsoft, line 39: this "filename" property is generated by codegen

javagen/src/main/resources/MultipartFormDataHelper.java
```diff
@@ -0,0 +1,123 @@
...
+                    + CRLF;
+            byte[] data = serialized.getBytes(encoderCharset);
+            appendBytes(data);
+        }
+        return this;
+    }
+
+    // application/octet-stream
+    public MultipartFormDataHelper serializeField(String fieldName, BinaryData file, String filename) {
+        if (file != null) {
```
weidongxu-microsoft, line 86: It allows optional field. If value is null, this field is skipped.

typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
haolingdong-msft: Where is default file name defined?

weidongxu-microsoft: Currently using the field name, aka `image` in this case.

If user does not provide the filename, current assumption is that the filename does not matter. So `image` or random UUID should both work (as long as there is no duplicate).

haolingdong-msft (approved):

typespec-tests/Generate.ps1
```diff
@@ -55,6 +55,8 @@ $generateScript = {
     $tspOptions += " --option ""@azure-tools/typespec-java.stream-style-serialization=true"""
   } elseif ($tspFile -match "tsp[\\/]union.tsp") {
     $tspOptions += " --option ""@azure-tools/typespec-java.stream-style-serialization=true"""
+  } elseif ($tspFile -match "tsp[\\/]multipart.tsp") {
+    $tspOptions += " --option ""@azure-tools/typespec-java.stream-style-serialization=true"""
```
haolingdong-msft, line 59: so currently it's user to set stream-style-serialization=true in tspconfig.yaml, instead of us adding the flag automatically if it's multipart?

weidongxu-microsoft: multipart does not depends on stream-style. None of JSON applies in this case.

I did this just to make Alan's life easier (as the test code generated is stream-style, which matches the default on his PR).

haolingdong-msft: I see.

javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -333,6 +335,23 @@ protected void writeClientModels(Client client, JavaPackage javaPackage, JavaSet
         }
     }
 
+    protected void writeHelperClasses(Client client, JavaPackage javaPackage, JavaSettings settings) {
+        // While azure-core's ResponseError hasn't shipped implementing JsonSerializable add a utility method that
+        // will serialize and deserialize ResponseError.
```
srnagar, line 340: Is there an issue tracking this to add support in core?

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaPackage.java
srnagar: Why do we need this helper class to be loaded from resources?

javagen/src/main/resources/MultipartFormDataHelper.java
```diff
@@ -0,0 +1,123 @@
...
+    private BinaryData requestBody;
+
+    /**
+     * Default constructor used in the code. The boundary is a random value.
+     */
+    public MultipartFormDataHelper(RequestOptions requestOptions) {
+        this(requestOptions, UUID.randomUUID().toString().substring(0, 16));
+    }
+
+    private MultipartFormDataHelper(RequestOptions requestOptions, String boundary) {
```
srnagar, line 53: Since this is private, we don't really need this overload. We can move the logic to the public ctor.

srnagar: If we want to make `boundary` configurable, then we can make this ctor public.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaPackage.java
weidongxu-microsoft: It is simipler to maintain the helper class this way, when the whole class is to be outputted as generated code.

E.g. [MultipartFormDataHelper](https://github.com/Azure/autorest.java/blob/b49056c5e03488dd61ee8884597014179df10a3a/javagen/src/main/resources/MultipartFormDataHelper.java) from codegen resources is outputted as [code](https://github.com/Azure/autorest.java/blob/b49056c5e03488dd61ee8884597014179df10a3a/typespec-tests/src/main/java/com/cadl/multipart/implementation/MultipartFormDataHelper.java) in test case (the only diff is package name and header).

javagen/src/main/java/com/azure/autorest/Javagen.java
weidongxu-microsoft: I think Alan already merged the code in azure-core https://github.com/Azure/azure-sdk-for-java/pull/37046

@alzimmermsft After new azure-core get released, are we able to direct depends on azure-core to handle the de-serialization?

javagen/src/main/resources/MultipartFormDataHelper.java
weidongxu-microsoft: Sure. Merged into the public ctor.

weidongxu-microsoft:
test case here (after cadl-ranch upgrade)
https://github.com/Azure/autorest.java/pull/2421/commits/3fd3f792603c2ab92c41a7454ec887d44b891b90#diff-0b261ad9261dffea162985054910fe6cebac468aa441d619f5ac1ddb60d50dde

javagen/src/main/java/com/azure/autorest/Javagen.java
alzimmermsft: Yeah, I've already merged in a change to `azure-core` where `ResponseError` now implements JsonSerializable so this won't be an issue any longer once that ships.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientModelProperty.java
```diff
@@ -344,6 +344,40 @@ protected void addJsonFlattenAnnotationImport(Set<String> imports) {
         imports.add("com.azure.core.annotation.JsonFlatten");
     }
 
+    /**
+     * Creates a builder that is initialized with all the builder properties set to current values of this instance.
+     * @return A new builder instance initialized with properties values of this instance.
+     */
+    public Builder toNewBuilder() {
```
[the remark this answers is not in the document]
weidongxu-microsoft: Used this name as it got prior :-)
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodParameter.java#L202

I may change it to `newBuilder` (seems clearer in "new" phrase that it creates a Builder).

Yeah, having the code in Builder itself should lower the possibility that a property added but forget to update this method.

PR #2424, opened 2023-11-27
tsp, json-merge-patch, model and client part code change

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/1530

WIP cadl-ranch test case. Will do e2e test with cadl-ranch test case.

Would recommend to look at the test first. 
  **tsp**: https://github.com/Azure/autorest.java/pull/2424/files#diff-28f987a95a4c3ed97b17b24b3dfd3013742f18513542f0d0449eada2afadc54f
  **generated code**: typespec-tests/src/main/java/com/cadl/patch/
  **unit test on generated code**:
  https://github.com/Azure/autorest.java/pull/2424/files#diff-e44d5197c14ca4f7a2c3a9163c54c95f18c5ca77e9a229b547cf1aa076685b28

**Unit test includes:**
1. testSerializationForNumbers
2. testSerializationForString: **content safety** service has similar case.
3. testSerializationForNestedModelProperty
4. testSerializationForMapProperty: **loadtesting** service has similar case. 
FYI - Loadtesting Java [PR](https://github.com/Azure/azure-sdk-for-java/pull/37656) to generate with latest spec. [TestRunAppComponents](https://github.com/Azure/azure-sdk-for-java/blob/4ce24e6a1c5dd8a077c9ba3fd62dd6b8fe8e2674/sdk/loadtesting/azure-developer-loadtesting/src/main/java/com/azure/developer/loadtesting/models/TestRunAppComponents.java) is the model that uses `Record` property in patch operation.
6. testSerializationForEnum
7. testSerializationForHierarchicalModel (aka. discriminator)

**Code change includes:**
1. Add jsonMergePatch flag and related static block to set flag
2. Add updatedProperties map and the setters
3. Serialization and deserialization, toJson() and toJsonMergePatch()
4. Enable generating JsonMergePatchHelper and model generation by turn off disabling convenient method for patch in emitter
5. Support serialization when map/array entry value is `null`. 
  logic: 45fb608c99afaef81eb12c03942b4fe5e45925e6
  sdk: https://github.com/Azure/autorest.java/blob/2a91d5e90cc80736cb59a00bb3c511180582920e/typespec-tests/src/main/java/com/cadl/patch/models/Resource.java#L266-L275
6. Add logic in convenient method to enable json merge patch for the required body model.
7. Support patch on optional body: 5d0d8f4a407ed7dd809c3a48d8d5faba980ed3b2

**Not Supported case:**
1. Patch on required property. e.g. this case in cadl-ranch: https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/property/nullable/main.tsp#L95-L101 
   Issue tracked: https://github.com/Azure/autorest.java/issues/2471


javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -263,13 +294,41 @@ private static void serializeJsonProperty(JavaBlock methodBlock, ClientModelProp
             return;
         }
 
+        if (isJsonMergePatch) {
+            methodBlock.ifBlock(String.format("%s !=null", property.getName()), codeBlock -> {
```
weidongxu-microsoft, line 298: fix "%s !=null"

What if it is int or double? Plan to have test on such case.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -263,13 +294,41 @@ private static void serializeJsonProperty(JavaBlock methodBlock, ClientModelProp
...
+                                              String serializedName, boolean fromSuperType) {
         IType clientType = property.getClientType();
         IType wireType = property.getWireType();
         String propertyValueGetter;
         if (fromSuperType) {
             propertyValueGetter = (clientType != wireType)
-                ? wireType.convertFromClientType(property.getGetterName() + "()")
-                : property.getGetterName() + "()";
+                    ? wireType.convertFromClientType(property.getGetterName() + "()")
+                    : property.getGetterName() + "()";
```
weidongxu-microsoft, line 331: avoid this kind of unnecessary formatting change

haolingdong-msft: I will revert back those spaces. 
BTW, Do we have command to do code format, like ts have `npm run format`, so there will not have unnecessary formatting change.

weidongxu-microsoft: No. checkstyle is all we need for now.

Just give other dev the freedom to write their own code, as long as it does not against checkstyle.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
weidongxu-microsoft: What's changed code (and what get generated)?

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -98,7 +98,7 @@ private static void addGettersAndSetters(List<ClientModel> models, JavaClass jav
             models.forEach(model -> {
                 // setters
                 javaClass.publicStaticMethod(String.format("void set%1$sAccessor(%1$sAccessor accessor)", model.getName()),methodBlock -> {
-                    methodBlock.line(String.format("%sAccessor = accessor;", CodeNamer.getModelNamer().modelPropertySetterName(model.getName())));
+                    methodBlock.line(String.format("%sAccessor = accessor;", CodeNamer.toCamelCase(model.getName())));
```
weidongxu-microsoft, line 101: Why not `model.getName()`? It shouldbe the class name.

typespec-tests/tsp/patch.tsp
```diff
@@ -23,16 +26,66 @@ model Resource {
...
-    @header("content-type") contentType: "application/merge-patch+json",
-    @path name: string,
-    @body body: Resource,
-  ): Resource | ErrorResponse;
+  @route("/resource")
+  createOrUpdateResource is ResourceOperations.ResourceCreateOrUpdate<Resource>;
+
+  @patch
+  @route("/fish")
+  createOrUpdateFish is ResourceOperations.ResourceCreateOrUpdate<Fish>;
```
weidongxu-microsoft, line 90: nit, Do we have to use core?

`@header("content-type") contentType: "application/merge-patch+json"` would be a valid case for non-azure.

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
```diff
@@ -0,0 +1,355 @@
...
+            return jsonWriter.writeEndObject();
+        }
+    }
+
+    public JsonWriter toJsonMergePatch(JsonWriter jsonWriter) throws IOException {
+        jsonWriter.writeStartObject();
+        if (this.map != null) {
+            jsonWriter.writeMapField("map", this.map, (writer, element) -> {
+                element.serializeAsJsonMergePatch(true);
+                writer.writeJson(element);
```
weidongxu-microsoft, line 266: Set serializeAsJsonMergePatch back to `false` after write to json.

weidongxu-microsoft: Here you need null check and write null value. You can do it in follow-up PR.

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
```diff
@@ -0,0 +1,355 @@
...
+            });
+        } else if (updatedProperties.contains("map")) {
+            jsonWriter.writeNullField("map");
+        }
+        if (this.description != null) {
+            jsonWriter.writeStringField("description", this.description);
+        } else if (updatedProperties.contains("description")) {
+            jsonWriter.writeNullField("description");
+        }
```
weidongxu-microsoft, line 275: I am unsure, so just a discussion. Should we check `updatedProperties` before write any value (even if it is not null)?

typespec-tests/src/test/java/com/cadl/patch/PatchClientTest.java
```diff
@@ -0,0 +1,104 @@
...
+public class PatchClientTest {
+
+    @Test
+    public void testSerializationForNumbers() {
+        // patch not enabled
+        Resource resource = new Resource(new HashMap<>());
+        resource.setIntValue(null);
+        resource.setLongValue(null);
+        String json = BinaryData.fromObject(resource).toString();
+        Assertions.assertEquals("{\"map\":{}}", json);
```
weidongxu-microsoft, line 26: nit, don't do this string equal on testing JSON. Use e.g. `JsonNode node = mapper.readTree(json);` and validate the JsonNode.

typespec-tests/tsp/patch.tsp
```diff
@@ -23,16 +26,66 @@ model Resource {
...
+  longValue?: int64;
+  intValue?: int32;
+  enumValue?: "a" | "b" | "c";
+
+  @projectedName("json", "wireNameForInnerModelProperty")
+  innerModelProperty?: InnerModel;
+}
+
+model InnerModel {
+  name: string;
```
weidongxu-microsoft, line 39: Add an optional property to InnerModel.

Then you are able to test `{"map": {"key": {"optional": null}}}`

typespec-tests/tsp/patch.tsp
```diff
@@ -23,16 +26,66 @@ model Resource {
...
+
+@doc("The third level model SawShark in polymorphic multiple levels inheritance.")
+model SawShark extends Shark {
+  sharktype: "saw";
+}
+
+@doc("The third level model GoblinShark in polymorphic multiple levels inheritance.")
+model GoblinShark extends Shark {
+  sharktype: "goblin";
 }
```
weidongxu-microsoft, line 80: question: Do we support nested polymorphic (kind -> sharktype) in stream-style (I remember it is not supported in Jackson)?

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -1159,4 +1180,29 @@ public boolean add(String s) {
...
+        classBlock.packagePrivateMethod("void serializeAsJsonMergePatch(boolean jsonMergePatch)", method -> {
+            method.line("this.jsonMergePatch = jsonMergePatch;");
+        });
+
+        // static code block to access jsonMergePatch setter
+        classBlock.staticBlock(staticBlock -> {
+            staticBlock.text(String.format("JsonMergePatchHelper.set%sAccessor((model, jsonMergePatchEnabled) -> {\n" +
+                    "            model.serializeAsJsonMergePatch(jsonMergePatchEnabled);\n" +
+                    "            return model;\n" +
+                    "        });", model.getName()));
```
weidongxu-microsoft, line 1205: do not do extra indent

typespec-tests/tsp/patch.tsp
haolingdong-msft: I asked what should tsp look like before because I'm not sure how user (azure, non-azure) should model it. I learned from the chat to provide code for the core case. So I add the test case using core. I can also add a case not using the template.
image

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -282,18 +321,43 @@ private static void serializeJsonProperty(JavaBlock methodBlock, ClientModelProp
...
-        String propertyValueGetter;
-        if (fromSuperType) {
-            propertyValueGetter = (clientType != wireType)
-                ? wireType.convertFromClientType(property.getGetterName() + "()")
-                : property.getGetterName() + "()";
-        } else if (property.isPolymorphicDiscriminator()) {
-            propertyValueGetter = property.getDefaultValue();
+        if (isJsonMergePatch) {
+            if (property.getClientType().isNullable()) {
+                methodBlock.ifBlock(String.format("%s !=null", getPropertyGetterStatement(property, fromSuperType)), codeBlock -> {
```
weidongxu-microsoft, line 326: details "%s !=null" -> "%s != null"

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -1159,4 +1180,29 @@ public boolean add(String s) {
...
+
+    /**
+     * Add json-merge-patch related flag and accessors.
+     */
+    private void addJsonMergePatchRelatedPropertyAndAccessors(JavaClass classBlock, ClientModel model) {
+        // properties
+        addGeneratedAnnotation(classBlock);
+        classBlock.privateMemberVariable("boolean jsonMergePatch");
+        addGeneratedAnnotation(classBlock);
+        classBlock.privateFinalMemberVariable("Set<String> updatedProperties = new HashSet<>()");
```
weidongxu-microsoft, line 1192: Suggestion to doc comment here that `updatedProperties` stores the java property name (not serialized name).

typespec-tests/tsp/patch.tsp
weidongxu-microsoft: You can add the Azure case to cadl-ranch.

I would still prefer this simple patch.tsp covers the non-azure (most cases here is non-azure).

typespec-extension/src/main/java/com/azure/typespec/model/EmitterOptions.java
```diff
@@ -44,7 +44,7 @@ public class EmitterOptions {
     private Boolean enableSyncStack = true;
 
     @JsonProperty(value = "stream-style-serialization")
-    private Boolean streamStyleSerialization = false;
+    private Boolean streamStyleSerialization = true;
```
weidongxu-microsoft, line 47: revert this

weidongxu-microsoft (commented):
Overall LGTM

typespec-tests/tsp/patch.tsp
haolingdong-msft: Sure, will update the case.

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
haolingdong-msft: `writer.writeJson(element);` will call `element.toJson()`, and in `toJson()` will call `toJsonMergePatch()` which will handle the null values.
Test case: `testSerializationForNestedModelProperty`

weidongxu-microsoft: I mean `element` be null.

haolingdong-msft: oh, I see. This is already fixed in this pr. output code: https://github.com/Azure/autorest.java/pull/2424/files#diff-0318904499f974c90db93a7e3663c59db25ffcb359ea2f7081a2d1b5e3b748d7

typespec-tests/src/main/java/com/cadl/specialheaders/models/Resource.java
```diff
@@ -41,6 +44,27 @@ public final class Resource implements JsonSerializable<Resource> {
...
+     */
+    @Generated
+    private final Set<String> updatedProperties = new HashSet<>();
+
+    @Generated
+    void serializeAsJsonMergePatch(boolean jsonMergePatch) {
+        this.jsonMergePatch = jsonMergePatch;
+    }
+
+    static {
```
haolingdong-msft, line 61: Currently we cannot add `@Generated` annotation to static code block, so partial updatemay treat this static code block as customized code. We may need to provide a whitelist for partial update supported elements: member properties, member methods, interface implemented types. Other elements change will be ignored by partial-update. I've tracked this in this issue: https://github.com/Azure/autorest.java/issues/2468.

typespec-tests/tsp/patch.tsp
haolingdong-msft: We need to pass in `sharktype` in constructor currently, and it will do correct serialization. May not be the expected 'support'.
I have removed the nested hierarchy case. Only keep simple inheritance.

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
haolingdong-msft: I think both are ok. Don't have preference actually.The current way seems straightforward to me.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
```diff
@@ -106,9 +131,36 @@ public String getEtag() {
...
+
+    public JsonWriter toJsonMergePatch(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
-        jsonWriter.writeStringField("name", this.name);
-        jsonWriter.writeArrayField("orders", this.orders, (writer, element) -> writer.writeJson(element));
+        if (this.name != null) {
+            jsonWriter.writeStringField("name", this.name);
+        } else if (updatedProperties.contains("name")) {
+            jsonWriter.writeNullField("name");
+        }
```
alzimmermsft, line 150: Maybe I'm getting JSON merge and JSON merge patch confused here.

Should there be a check first to see if `name` is in `updatedProperties`, if it's not then nothing is serialized. Then check if the `name` property is non-null and serialize that. Finally serialize null as the value was removed.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
```diff
@@ -42,6 +45,27 @@ public final class User implements JsonSerializable<User> {
...
     private String etag;
 
+    @Generated
+    private boolean jsonMergePatch;
+
+    /**
+     * Stores updated model property, the value is property name, not serialized name.
+     */
+    @Generated
+    private final Set<String> updatedProperties = new HashSet<>();
```
alzimmermsft, line 55: If we plan to use this a lot it may be better long-term to look into a design that doesn't use a `Set`. Something I've been thinking about doing for stream-style serialization in deserialization instead of having all the `boolean *Found` properties is to use an `int` or `long` and bit manipulation. Ex,

_Current_

```java
String value = null;
boolean valueFound = false;
Double nullableDecimal = null;
boolean nullableDecimalFound = false;
....
if ("value".equals(fieldName)) {
  valueFound = true;
} else if ("nullableDecimal".equals(fieldName)) {
  nullableDecimalFound = true;
}
....
if (!valueFound) {
  // Add value as missing property
} else if (!nullableDecimalFound) {
  // Add nullableDecimal as missing property
} 
....
```

You can do something based on property ordering:

```java
int foundTracker = 0;
String value = null;
Double nullableDecimal = null;
....
if ("value".equals(fieldName)) {
  foundTracker |= 1;
} else if ("nullableDecimal".equals(fieldName)) {
  foundTracker |= 2;
}
....
if (foundTracker & 1 == 0) {
  // Add value as missing property
} else if (foundTracker & 2 == 0) {
  // Add nullableDecimal as missing property
}
....
```

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
```diff
@@ -42,6 +45,27 @@ public final class User implements JsonSerializable<User> {
     @Generated
     private String etag;
 
+    @Generated
+    private boolean jsonMergePatch;
```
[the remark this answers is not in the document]
srnagar: So, the toJsonMergePatch() method will set this to true, serialize the field, then set this back to false. So, the customer's serialization should not be impacted except for a brief amount of time when the serialization is happening. We don't guarantee thread safety for models. In any case, the serialized model is still correct and will just have `null` values if they were modified, nothing else should be impacted and it's not incorrect json just have more fields that are just `null`.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -340,6 +349,13 @@ private void addImports(Set<String> imports, ClientModel model, JavaSettings set
         addGeneratedImport(imports);
 
         model.addImportsTo(imports, settings);
+
+        // add Json merge patch related imports
+        if (ClientModelUtil.isJsonMergePatchModel(model) && settings.isStreamStyleSerialization()) {
```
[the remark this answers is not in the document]
haolingdong-msft: Yes, because we need to handle json merge patch serialization specifically. (i.e. to serialize `null`)

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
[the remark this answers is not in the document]
haolingdong-msft: In client, we will generate the code  to set the flag, user does not need to set the flag. The code may look like below:
```
    @Generated
    public Resource createOrUpdateResource(Resource resource) {
        // Generated convenience method for createOrUpdateResourceWithResponse
        RequestOptions requestOptions = new RequestOptions();
        JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);
        BinaryData data = BinaryData.fromObject(resource);
        JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, false);
        return createOrUpdateResourceWithResponse(data, requestOptions).getValue()
            .toObject(Resource.class);
    }
```

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
haolingdong-msft: The current logic in `toJsonMergePatch()` is: if the property is not null, the serialization is as usual, if the property is null and the property is in `updatedProperties`(which means the user updates the property to `null` explicitly), we will serialization the `null` value.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
```diff
@@ -106,9 +131,36 @@ public String getEtag() {
...
+            jsonWriter.writeStringField("name", this.name);
+        } else if (updatedProperties.contains("name")) {
+            jsonWriter.writeNullField("name");
+        }
+        if (this.orders != null) {
+            jsonWriter.writeArrayField("orders", this.orders, (writer, element) -> {
+                if (element != null) {
+                    element.serializeAsJsonMergePatch(true);
+                    writer.writeJson(element);
+                    element.serializeAsJsonMergePatch(false);
```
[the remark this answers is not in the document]
haolingdong-msft: Yes, I agree it would be better, but I'm not sure if this will the code a bit complex. Because, the user will not call set the flag themselves (that's why JsonMergePatchHelper is in `impl` package), it will be called in client method through `       JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);`. So I guess simply set to false here is fine.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
haolingdong-msft: I will monitor the performance, we can refactor this part if we found the performance is bad.

srnagar: Any reason for not using `Set`? JMP is not very used a lot (at least in data-plane libraries). Also, bit manipulation makes the code less readable and hard to maintain especially when we rely on the order of the properties because the TypeSpec authors can change the order and add new properties in any order. If this is for perf optimization, I would recommend not using bit manipulation prematurely until we have data to show that `Set` is causing perf bottleneck.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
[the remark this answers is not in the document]
srnagar: Yes, we decided not to support the non-stream style serialization for JMP as that's anyway going away soon.

typespec-extension/src/code-model-builder.ts
```diff
@@ -663,9 +663,9 @@ export class CodeModelBuilder {
       // check if the convenience API need to be disabled for some special cases
       if (operationIsJsonMergePatch(op)) {
         // do not generate convenience method for JSON Merge Patch
-        generateConvenienceApi = false;
-        apiComment = `Convenience API is not generated, as operation '${op.operation.name}' is 'application/merge-patch+json'`;
-        this.logWarning(apiComment);
+        // generateConvenienceApi = false;
+        // apiComment = `Convenience API is not generated, as operation '${op.operation.name}' is 'application/merge-patch+json'`;
+        // this.logWarning(apiComment);
```
srnagar, line 668: Why are these lines commented out?

typespec-tests/src/main/java/com/_specs_/azure/core/basic/implementation/JsonMergePatchHelper.java
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+// Code generated by Microsoft (R) TypeSpec Code Generator.
+
+package com._specs_.azure.core.basic.implementation;
+
+import com._specs_.azure.core.basic.models.User;
+import com._specs_.azure.core.basic.models.UserOrder;
+
+public class JsonMergePatchHelper {
```
srnagar, line 10: Add javadoc to describe what this helper does.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
srnagar: This is to support roundtrippable models. We write all properties that are not null in the given model and also write null properties if they are marked as updated.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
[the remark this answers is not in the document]
srnagar: Yes, resetting this to `false` is sufficient as this shouldn't be called by external code and should only be used in JMP codepath.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
haolingdong-msft: I agree with @srnagar. `Set` here is more readable and we can refactor if we have perf issue later.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: This is to **turn off disabling convenient method for patch in emitter**, so that models will be generated. Otherwise, if convenient methods are diabled, models will not be generated.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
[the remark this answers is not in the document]
weidongxu-microsoft: Setting `true`, run the code, then setting `false` could have problem when there is concurrency of serialization of these models. The chance is likely very low (first or all, user need to call different API e.g. PUT and PATCH, with a same model, concurrently).

Since all the code is not exposed to customer, we can e.g. try switch to better concurrency code (e.g. `JsonMergePatchHelper.getUserOrderAccessor().writeJsonAsJsonMergePatch(element)`) if we really see user do such concurrency.

weidongxu-microsoft:
Please consider the case of shared model between PUT and Json-Merge-Patch e.g.

```yaml
model SharedModel {
  requiredProp1: string
  requiredProp2: int32
  optionalProp3: ...
}
```

And how we'd like to do the class. Think e.g. what user need to do on the patch, when they only want to modify `requiredProp1`

haolingdong-msft:
> Please consider the case of shared model between PUT and Json-Merge-Patch e.g.
> 
> ```yaml
> model SharedModel {
>   requiredProp1: string
>   requiredProp2: int32
>   optionalProp3?: ...
> }
> ```
> 
> And how we'd like the class be. Think e.g. what user need to do on the patch, when they only want to modify `requiredProp1`
> 
> Not needed in this PR. But please consider and discuss with Srikanta, if you are unsure about the design.

Thanks Weidong, I created an issue to track this: https://github.com/Azure/autorest.java/issues/2471.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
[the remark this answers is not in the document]
weidongxu-microsoft: nit, when you really do the code, use a complex var name `BinaryData data = BinaryData.fromObject(resource);`

We do not use `data` which is pretty likely to conflict with some parameter or other variables. Maybe `requestBodyInBinaryData` etc.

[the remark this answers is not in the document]
haolingdong-msft: Yes, I will, there could be multiple parameters.

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -154,11 +154,21 @@ protected void writeMethodImplementation(
...
                 .map(p -> {
-                    String expression = parameterExpressionsMap.get(p.getName());
-                    if (expression == null) {
-                        expression = p.getName();
+                    String parameterName = p.getName();
+                    String expression = parameterExpressionsMap.get(parameterName);
+                    IType parameterClientType = p.getClientType();
+                    IType parameterRawType = p.getRawType();
+
+                    if (ClientModelUtil.isClientModel(parameterRawType) && ClientModelUtil.isJsonMergePatchModel(ClientModelUtil.getClientModel(((ClassType) parameterRawType).getName()))) {
```
weidongxu-microsoft, line 162: nit, also add a check for `RequestParameterLocation.BODY`?

If possible, add a test case when this request body is optional.

haolingdong-msft: I will add the request body optional case into cadl-ranch.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/implementation/JsonMergePatchHelper.java
haolingdong-msft: updated

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
haolingdong-msft: Seems not much code change, I updated together in this pr.

weidongxu-microsoft: OK, then it be great.

I wasn't expect these old code could handle that without any modification outside of this block, but appears it does.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -369,17 +463,35 @@ private static void serializeJsonContainerProperty(JavaBlock methodBlock, String
...
+                                }
+                                ifBlock.line(valueSerializationMethod + ";");
+                                if (elementType instanceof ClassType && ((ClassType) elementType).isSwaggerType()) {
+                                    codeBlock.line(elementName + ".serializeAsJsonMergePatch(false);");
+                                }
+                            }).elseBlock(elseBlock -> elseBlock.line(lambdaWriterName + ".writeNull();"));
+                        });
+                    } else {
+                        methodBlock.line(valueSerializationMethod);
+                    }
```
weidongxu-microsoft, line 481: Is this for the Map?

If yes, all Map element in json-merge-patch should be treated as nullable. So I don't think we should test that (as isNullable here is the state of property in shared model).

Even if element is primitive integer, it should still test for null, as the type is `Map<String, Integer>` and user can still set it to null to remove a key-value.

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: You now can just delete these code now.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/BasicAsyncClient.java
```diff
@@ -469,6 +469,33 @@ public Mono<Response<BinaryData>> exportWithResponse(int id, String format, Requ
...
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return details about a user on successful completion of {@link Mono}.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Mono<User> createOrUpdate(int id, User resource) {
```
weidongxu-microsoft, line 489: I think we had a test calling protocol API. You can change it now to call this one.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/BasicAsyncClient.java
```diff
@@ -489,8 +516,11 @@ public Mono<Response<BinaryData>> exportWithResponse(int id, String format, Requ
     public Mono<User> createOrReplace(int id, User resource) {
         // Generated convenience method for createOrReplaceWithResponse
         RequestOptions requestOptions = new RequestOptions();
-        return createOrReplaceWithResponse(id, BinaryData.fromObject(resource), requestOptions)
-            .flatMap(FluxUtil::toMono).map(protocolMethodData -> protocolMethodData.toObject(User.class));
+        JsonMergePatchHelper.getUserAccessor().prepareModelForJsonMergePatch(resource, true);
+        BinaryData resourceInBinaryData = BinaryData.fromObject(resource);
+        JsonMergePatchHelper.getUserAccessor().prepareModelForJsonMergePatch(resource, false);
```
weidongxu-microsoft, line 521: Emm, this is wrong. It is PUT. We probably need to additionally check the content-type in the ConvenienceMethodTemplateBase

haolingdong-msft: Good catch, emm, yes, we probably need to check on operation level as well, because the model can be used in both json-merage-patch operation or non json-merge-patch operation..though I'm not sure if this is a common case.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
haolingdong-msft: I add it mainly for the Map case. But this is for serializing container case, e.g. `Map`, `Array`, `List`. I'm thinking if it is `int[]` array, the parameter can be `int`. So I added this for safety reason.

weidongxu-microsoft: So far, codegen don't have `int[]`. `ArrayType` is only for `byte[]`.

weidongxu-microsoft: Or you may target Map only?

I don't really sure what's
`List['a', 'b', null]` means for json-merge-patch. That probably not valid.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -304,6 +368,9 @@ private static void serializeJsonProperty(JavaBlock methodBlock, ClientModelProp
             methodBlock.line("jsonWriter.writeFieldName(\"" + serializedName + "\");");
             methodBlock.line(ClientModelUtil.CORE_TO_CODEGEN_BRIDGE_UTILS_CLASS_NAME + ".responseErrorToJson(jsonWriter, " + propertyValueGetter + ");");
         } else if (fieldSerializationMethod != null) {
+            if (isJsonMergePatch && wireType instanceof ClassType && ((ClassType) wireType).isSwaggerType()) {
```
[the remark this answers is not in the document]
haolingdong-msft: From my understanding, this is to see if the type is defined in swagger. In tsp, it's type defined in tsp. Add @alzimmermsft to comment as well.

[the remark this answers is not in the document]
weidongxu-microsoft: Ok, so this appears to say it is a model, not a string/number something like `String` `Duration` `BigDecimal`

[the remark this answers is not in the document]
haolingdong-msft: Yes

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
haolingdong-msft: ok, if that's the case, I might simply remove the null check.

weidongxu-microsoft: I mean Map should have it in any case (this is a json-merge-patch to clean a value -- Map is just a JSON object, same as your Class model). List/Array may not need any special handling in json-merge-patch (I think `null` here does not have special meaning, but safer to double check RFC).

haolingdong-msft: I see, we may only need to do special handling for json-mergpatch+Map case.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
alzimmermsft: Okay, sounds like I may have been getting things confused. What I was thinking is if the model was return with

`{ "value" : "aValue" }`

And the user never edited it, do we need to send back

`{ "value" : "aValue" }`?

Or would the choice be to send `{ }` as nothing was updated.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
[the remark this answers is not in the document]
alzimmermsft: Yeah, `isSwaggerType` is meant to be something not in the base language or a well-known type in `azure-core`, basically anything defined by Swagger or TypeSpec.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/BasicAsyncClient.java
weidongxu-microsoft: Azure may ask new service only do PATCH, but it is pretty common for non-azure service to have both POST+PATCH or PUT+PATCH for create/update.

I'm in the progress of adding `mimeType` into the template (we would need it for e.g. `text/plain` anyway)
https://github.com/Azure/autorest.java/pull/2476

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
haolingdong-msft: Updated code, add a test case for array. If we remove nullable check here, when we set one element of array to `null`, it will not be serialized.
```
@Test
    public void testSerializationForArrayProperty() throws JsonProcessingException {
        Resource resource = new Resource(new HashMap<>());
        resource.setArray(Arrays.asList(new InnerModel("value1"), new InnerModel("value2")));
        JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);
        resource.getMap().put("key", null);
        resource.getArray().set(0, null);
        String json = BinaryData.fromObject(resource).toString(); // {"map":{"key":null},"array":[{"name":"value2"}]}
        JsonNode node = OBJECT_MAPPER.readTree(json);
        Assertions.assertEquals(JsonNodeType.NULL, node.get("map").get("key").getNodeType());
        Assertions.assertEquals(1, node.get("array").size());
        Assertions.assertEquals("value2", node.get("array").get(0).get("name").asText());
    }
```

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -113,6 +113,7 @@ protected void writeMethodImplementation(
 
         writeValidationForVersioning(convenienceMethod, parametersMap.keySet(), methodBlock);
 
+        boolean isJsonMergePatchOperation = protocolMethod != null && protocolMethod.getProxyMethod() != null && protocolMethod.getProxyMethod().getRequestContentType().contains("application/merge-patch+json");
```
weidongxu-microsoft, line 119: RequestContentType should have only 1 item?

Please do an case-insensitive equal.

weidongxu-microsoft (approved):
Most LGTM

Please fix https://github.com/Azure/autorest.java/pull/2424/commits/895c6e99a6e967377fc44053b72d153c473ccd4f#r1432675198

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
haolingdong-msft: This is for roundtrippable models. So I think this one: `{ "value" : "aValue" }`. 
Use case is like below, `resource` will be used as output(create case) and input(update case):
```
Resource resource = client.createOrUpdate(new Resource("aValue")); // this is to create the resource
resource.setValue(null);
client.createOrUpdate(resource); // this is to update the resource, write all properties include the `null` values.
```
/cc @srnagar

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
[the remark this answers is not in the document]
haolingdong-msft: I found there is a bug with this code,  while working on cadl-ranch e2e tests.
```java
    @Generated
    public Resource updateResoure(Resource body) {
        // Generated convenience method for updateResoureWithResponse
        RequestOptions requestOptions = new RequestOptions();
        JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(body, true);
        BinaryData bodyInBinaryData = BinaryData.fromObject(body);
        JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(body, false);
        return updateResoureWithResponse(bodyInBinaryData, requestOptions).getValue().toObject(Resource.class);
    }
```
Serialization (i.e. call `toJsonMergePatch()`) will not actually happen during `BinaryData.fromObject()`, it happens `return updateResoureWithResponse(bodyInBinaryData, requestOptions).getValue().toObject(Resource.class);`. But since we  disable json merge patch before it, json-merge-patch serialization will not be available. I am checking how to fix it. May need to update the code in `writeInvocationAndConversion()`.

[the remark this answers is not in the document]
haolingdong-msft: Fixed using Solution 2 now.

haolingdong-msft:
https://github.com/Azure/autorest.java/pull/2424#discussion_r1437341155

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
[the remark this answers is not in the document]
weidongxu-microsoft: OK for now.

Got it, here the impl of `BianryData` gets invovled. The serialization does not happen when creating the `BinaryData` (here the impl only wraps the object). It happens when the data of the `BinaryData` get used.

PR #2426, opened 2023-11-28
Fix constant names

alzimmermsft:
Fixes constant names to use `UPPERCASING_WITH_UNDERSCORES`


weidongxu-microsoft (approved):

weidongxu-microsoft:
Both dpg and mgmt CI seems reporting about `getOdataNextLink` to `getOdatanextLink`.

```
[ERROR] /D:/a/_work/1/s/fluent-tests/src/main/java/com/azure/mgmtlitetest/mediaservices/implementation/StreamingLocatorsClientImpl.java:[184,94] cannot find symbol
  symbol:   method odatanextLink()
```

There is too many naming changes, so I could not see where in your PR would affect this.

PR #2431, opened 2023-11-29
Fix constant names

alzimmermsft:
Replaces #2426 as it had too many merge conflicts.

Fixes constant names to use `UPPERCASING_WITH_UNDERSCORES`.


javagen/pom.xml
```diff
@@ -68,6 +68,12 @@
       <version>4.13.2</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>io.projectreactor.netty</groupId>
+          <artifactId>reactor-netty-http</artifactId>
+          <version>1.1.7</version>
+          <scope>compile</scope>
+      </dependency>
```
srnagar, line 76: Why is this dependency needed?

srnagar (approved):

javagen/pom.xml
alzimmermsft: I think IntelliJ added that when I was fixing a problem, will remove it.

PR #2434, opened 2023-11-30
Add preview pipeline for regenerating clients in azure-sdk-for-java

hallipr:
- Add preview.yml for use by the `autorest.java - preview`
- Add powershell scripts used by the common preview template
- Update package.json to pin floating peer dependencies using a devDependency version
- Commit package-lock.json files to the repo for deterministic CI builds


hallipr:
This pipeline does not yet:
- run code generation tests in autorest.java prior to running sdk generation in azure-sdk-for-java
- use the newly packaged autorest generator to regenerate autorest based sdks

typespec-extension/package.json
```diff
@@ -70,6 +71,13 @@
...
     "@typescript-eslint/eslint-plugin": "~6.12.0",
-    "eslint-plugin-deprecation": "~2.0.0"
+    "eslint-plugin-deprecation": "~2.0.0",
+    "@azure-tools/typespec-azure-core": "0.36.0",
+    "@azure-tools/typespec-client-generator-core": "0.36.0",
+    "@typespec/compiler": "0.50.0",
+    "@typespec/http": "0.50.0",
+    "@typespec/openapi": "0.50.0",
+    "@typespec/rest": "0.50.0",
+    "@typespec/versioning": "0.50.0"
```
weidongxu-microsoft, line 81: why do we need this? they should already be defined in `peerDependencies`?

typespec-extension/package-lock.json
weidongxu-microsoft, line 1: Does the preview job need the package-lock? Generally, dev want to always take the latest (non-dev), to catch anything new.

eng/pipelines/preview.yml
```diff
@@ -0,0 +1,41 @@
+schedules:
+- cron: '0 0 * * *'
+  displayName: Nightly dev build
```
weidongxu-microsoft, line 3: Is it a daily job?

eng/scripts/Initialize-Repository.ps1
```diff
@@ -0,0 +1,35 @@
...
+    invoke "npm ci"
+    
+    Set-Location "$RepoRoot/typespec-extension"
+    invoke "npm ci"
+  }
+  
+  $env:path = $env:path = "$RepoRoot/node_modules/.bin;$env:path"
+}
+finally {
+    Pop-Location
```
weidongxu-microsoft, line 34: nit, inconsistent indent?

eng/scripts/Build-Packages.ps1
```diff
@@ -0,0 +1,85 @@
...
+    
+    Write-Host "Dependencies of typespec-java"
+    invoke "npm ls --all" -ExecutePath "$RepoRoot/typespec-extension"
+}
+finally {
+    Pop-Location
+}
+
+if ($PublishInternal) {
+  $feedUrl = "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry"
```
weidongxu-microsoft, line 70: why it is js-test-autorest?

eng/scripts/Build-Packages.ps1
```diff
@@ -0,0 +1,85 @@
...
+
+  $overrides = @{
+      "@azure-tools/typespec-java" = "$feedUrl/@azure-tools/typespec-java/-/typespec-java-$emitterVersion.tgz"
```
weidongxu-microsoft, line 73: Anywhere uses eth package?

weidongxu-microsoft:
It would be quick fix in typespec-java CI to ignore the diff on package-lock.json. Though I'd like to know whether package-lock.json is required for the PR.

I understand usually npm module would contain package-lock.json. It is the current design not to have them.

typespec-extension/package-lock.json
hallipr: To reduce supply chain vulnerabilities and support deterministic builds, all CI builds should use lock files provided by the package manager.

Local devs can use `npm ci` to delete and recreate their node_module folder, `npm install` to keep the folder and add missing modules, or `npm update` to update to newer versions that match the version specs in package.json.

eng/pipelines/preview.yml
hallipr: it is

eng/scripts/Build-Packages.ps1
hallipr: This is the common internal feed used for npm modules.

eng/scripts/Build-Packages.ps1
hallipr: The package is used by the sdk repo during preview generation via emitter-package.json:
https://github.com/Azure/azure-sdk-for-java/pull/37850/files#diff-bac355930627a26bbb83d1f84d110b1e3ec82614739953f538daffc255e31d73
``` json
  "dependencies": {
    "@azure-tools/typespec-java": "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/@azure-tools/typespec-java/-/typespec-java-0.11.2-alpha.20231128.3.tgz"
  },
```

We should also use the generator package from the same feed via URL to support validation of autorest generation in the sdk repo.

typespec-extension/package.json
hallipr: There are 2 primary motivations for listing the package in both devDependencies and peerDependencies.

1. To construct an installable emitter-package.json when the emitter depends on a dev version of typespec, we have to declare our dependency on the dev version of all of the peer dependencies:
    https://github.com/Azure/azure-sdk-for-java/pull/37850/files#diff-bac355930627a26bbb83d1f84d110b1e3ec82614739953f538daffc255e31d73
    ``` json
      "dependencies": {
        "@azure-tools/typespec-java": "https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-js-test-autorest/npm/registry/@azure-tools/typespec-java/-/typespec-java-0.11.2-alpha.20231128.3.tgz"
      },
      "devDependencies": {
        "@typespec/openapi": "0.51.0-dev.1",
        "@typespec/compiler": "0.51.0-dev.9",
        "@typespec/versioning": "0.51.0-dev.6",
        "@typespec/rest": "0.51.0-dev.2",
        "@typespec/http": "0.51.0-dev.3",
        "@azure-tools/typespec-azure-core": "0.37.0-dev.6",
        "@azure-tools/typespec-client-generator-core": "0.37.0-dev.6"
      }
    ```

    The devDependency list in emitter-package json is constructed by comparing the peerDependency and devDependency list of the emitter's package.json file and including any package that's in both.

2. The specific version specified in devDependencies tells devs which version within the typically broad peerDependency range we're actually working with and is known to work with the module.

eng/scripts/Initialize-Repository.ps1
hallipr: thanks

hallipr:
> It would be quick fix in typespec-java CI to ignore the diff on package-lock.json.

I think the better solution would be for the ci pipeline to use `npm ci` rather than `npm install` once the repo contains a package-lock.json

weidongxu-microsoft:
> I think the better solution would be for the ci pipeline to use `npm ci` rather than `npm install` once the repo contains a package-lock.json

As mentioned, we'd like CI to always use latest, so if there were incompatibility, it would fail on next CI (I am glad to have this kind of CI failure when I prepare a release), instead of fail on customer.

Though if you insist, it is fine to override the checkenforcer and merge. I will follow up to update the script to use "npm ci".
Likely this 2 files
https://github.com/Azure/autorest.java/blob/5a6a29a4767b4b5160a4af23ba3b0823c28275fd/Generate-TypeSpec.ps1
https://github.com/Azure/autorest.java/blob/5a6a29a4767b4b5160a4af23ba3b0823c28275fd/typespec-tests/Setup.ps1

weidongxu-microsoft (approved):
Approve to unblock.

But see comment https://github.com/Azure/autorest.java/pull/2434#discussion_r1413312651

I am not very comfortable to fix the devDependencies in package.json for all situations, where it would only be used in preview release.

typespec-extension/package.json
weidongxu-microsoft: One thought on generating the `devDependencies` on "emitter-package.json" in SDK repo is always use latest dev version (ignore the "devDependencies" here).

If both typespec-java and SDK preview is daily, these dev version should match.

Just a though. Feel free to disagree :-)

hallipr:
> Approve to unblock.

Thanks Weidong.

We're in the process of updating all of the autorest repos to use lock files, `npm ci` and `npx --no-install` to avoid supply chain vulnerabilities in CI builds.   All of that to say, we don't want CI builds to install any module versions that aren't explicitly specified in the repo. I'm adding changes to use `npm ci` for the remaining scripts used in CI that call `npm install`.  This will mean that all automated runs will use the lock file.

weidongxu-microsoft:
> > Approve to unblock.
> 
> Thanks Weidong.
> 
> We're in the process of updating all of the autorest repos to use lock files, `npm ci` and `npx --no-install` to avoid supply chain vulnerabilities in CI builds. All of that to say, we don't want CI builds to install any module versions that aren't explicitly specified in the repo. I'm adding changes to use `npm ci` for the remaining scripts used in CI that call `npm install`. This will mean that all automated runs will use the lock file.

Alright.

There been some update on package.json. I am trying to see whether getting them to latest would solve the CI error https://github.com/Azure/autorest.java/pull/2452

weidongxu-microsoft:
https://github.com/Azure/autorest.java/pull/2452 merges latest main, and made very minor change upon this PR

The only thing it worth noticing is that currently it still use `npm install` on local test module.

We may later see if we can put the lock there as well. That would require us to separate the script when CI runs the test, and when dev runs the test (dev always want latest typespec-java built on their local).

PR #2437, opened 2023-12-01
tsp, allow access(public) to API in pollingOperation

weidongxu-microsoft:
quote Michael

> I agree with the suggestion here that we should always generate all operations regardless of whether they are marked with polling operation.  We can use `@access` for now to hide them if we don’t want them and `@scope` once that decorator comes online.

---

Update: still in discussion.

Now in this PR, behavior is not changed, but we allow `@access(public)` to unhide the operation.


typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -260,7 +260,6 @@ public TypeSpecPlugin(EmitterOptions options, boolean sdkIntegration) {
             SETTINGS_MAP.put("license-header", "SMALL_TYPESPEC");
 
             SETTINGS_MAP.put("sync-methods", "sync-only");
-            SETTINGS_MAP.put("stream-style-serialization", true);
```
weidongxu-microsoft, line 263: default is here
https://github.com/Azure/autorest.java/blob/a118c84ba4c6f6b80fea925c2d664911abbc5ad3/typespec-extension/src/main/java/com/azure/typespec/model/EmitterOptions.java#L46-L47

weidongxu-microsoft:
Hold on as there is more discussion in email and chat.

Issue #2436, opened 2023-12-01
tsp, cadl-ranch


weidongxu-microsoft:
dashboard https://azure.github.io/cadl-ranch/

![image](https://github.com/Azure/autorest.java/assets/53292327/a75c87ab-c74c-4d6f-87cf-a44b130a3e0a)

Add for "Type/Property/Optional".

weidongxu-microsoft:
click the case on dashboard for details (expected input, expected output that need assert, etc.)

do not call "WithResponse" API, unless no alternative.

v-hongli1:
relate to [pr#2438](https://github.com/Azure/autorest.java/pull/2438)

PR #2438, opened 2023-12-01
Add test cases for tsp, cadl-ranch Type/Property/Optional

v-hongli1:
1. Add test cases for fix [issues#2436](https://github.com/Azure/autorest.java/issues/2436).
2. Now there are four test cases that are not working due to `NullPointer` exception occurs. 
   The detail info following as : 
image
3. Disabled not working test cases, the tag is `NullPointer Cannot invoke "java.lang.Long.longValue()"`.

### DIFF
--- a/typespec-tests/src/test/java/com/type/property/optional/BooleanLiteralClientTests.java
+++ b/typespec-tests/src/test/java/com/type/property/optional/BooleanLiteralClientTests.java
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.type.property.optional;
+
+import com.type.property.optional.models.BooleanLiteralProperty;
+import com.type.property.optional.models.BooleanLiteralProperty1;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class BooleanLiteralClientTests {
+    BooleanLiteralClient client = new OptionalClientBuilder().buildBooleanLiteralClient();
+
+    @Test
+    void getAll() {
+        BooleanLiteralProperty booleanLiteralProperty = client.getAll();
+        Assertions.assertEquals(BooleanLiteralProperty1.TRUE, booleanLiteralProperty.getProperty());
+    }
+
+    @Test
+    void getDefault() {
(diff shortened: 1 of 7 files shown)


typespec-tests/src/test/java/com/type/property/optional/BooleanLiteralClientTests.java
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT License.
+
+package com.type.property.optional;
+
+import com.type.property.optional.models.BooleanLiteralProperty;
+import com.type.property.optional.models.BooleanLiteralProperty1;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class BooleanLiteralClientTests {
```
weidongxu-microsoft, line 11: nit, make all class and method `public`.

typespec-tests/src/test/java/com/type/property/optional/BooleanLiteralClientTests.java
```diff
@@ -0,0 +1,38 @@
...
+    BooleanLiteralClient client = new OptionalClientBuilder().buildBooleanLiteralClient();
```
weidongxu-microsoft, line 12: nit, `private final`

v-hongli1:
> cause is likely
> 
> ```java
>     @Override
>     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
>         jsonWriter.writeStartObject();
>         jsonWriter.writeDoubleField("property", this.property == null ? null : this.property.toDouble());
>         return jsonWriter.writeEndObject();
>     }
> ```
> 
> `writeDoubleField` takes `double`. Hence should not pass `null`. The code should be inside a null protection.

Yes, it is caused by this problem.

haolingdong-msft (approved):
LGTM

typespec-tests/src/test/java/com/type/property/optional/BooleanLiteralClientTests.java
v-hongli1: Fixed in the new version.

weidongxu-microsoft:
failed tests fixed via https://github.com/Azure/autorest.java/pull/2444

PR #2439, opened 2023-12-01
Add test cases for for tsp, cadl-ranch Projection.ProjectedName

v-hongli1:
Add test cases for for tsp, cadl-ranch Projection.ProjectedName


weidongxu-microsoft (approved):
nit, make test class/method public

PR #2441, opened 2023-12-01
Use slf4j-simple instead of logback

alzimmermsft:
Replaces logback with slf4j-simple while retaining the ability to pass a different log level configuration when running code generation.


joshfree (approved):

srnagar (commented):
LGTM

PR #2442, opened 2023-12-01
Don't generate enums with JsonSerializable

alzimmermsft:
#2410 added functionality which generates `enum`s and `ExpandableStringEnum`s implementing `JsonSerializable` to resolve failures in APIs using `BinaryData.fromObject(enum/expandableStringEnum)` or `Map<String, Object>` which is later passed to `BinaryData.fromObject`. This did fix the issues being seen but after generating a few more libraries outside of TypeSpec, and notably an XML-based library, I realized this wasn't the best solution for a few reasons:

- Ties `enum`s and `ExpandableStringEnum`s logically to JSON given the `JsonSerializable` implementation. These types shouldn't be logically tied to a specific serialization format.
- This doesn't fix the problem for XML-based libraries which cannot use `JsonSerializable`. Unlike JSON, XML doesn't have a clear way to handle this anyways as the `enum` may be used as an attribute or XML element value.

This PR instead changes how code generation uses `enum`s and `ExpandableStringEnum`s when generating code where `BinaryData.fromObject` handles the payload to use the backing value of the enum. For example, String-based enums will call `toString` before passing to either the `BinaryData` factory method or `Map.put`, long-based `toLong`, double-based `toDouble`, etc.

This removes the need to modify the API generated, any potential confusion or errors with `enum`s implementing `JsonSerializable`, and better aligns with the documentation generated for these APIs which mention the value needs to be a `String` or `long` or `double`, etc. And it gives more time for stream-style serialization to create a better end-to-end story when `enum`s are being used outside of generated code.


weidongxu-microsoft:
Got it.

I guess previously toObject and fromObject works because in the background they are done by Jackson?

And if we purely rely on azure-json, the `toObject` code on extensible enum would be written as `Enum.fromString(binaryData.toObject(String.class))`?

One consideration is that customer may actually use the simple form of e.g. `binaryData.toObject(Enum.class)` or `BinaryData.fromObject(enum)`. We seems not able to disallow them from doing this.

weidongxu-microsoft:
And does it (to and from BinaryData) apply to other object like `OffsetDateTime`, `Duration`, `BigDecimal` etc.?

PR #2454, opened 2023-12-11
tsp, do not skip op with pollingOperation decorator

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/pull/2454/commits/560e6deead6fdd0aadbfce9813fb77d362e5fedc

there is existing cadl-ranch case with "get" op in `@service` namespace (lro/rpc-legacy).

test case in local that "get" op is not in `@service` namespace: https://github.com/Azure/autorest.java/pull/2454/commits/e701c10d77084a66f17691911e226e46d06e0674
(later will see if we do a case in cadl-ranch -- probably together with other legacy cases like "importUpdate" and "importDevices" in deviceupdate)


typespec-tests/tsp/lro.tsp
```diff
@@ -0,0 +1,117 @@
...
+  version: "1.0.0",
+})
+@useDependency(Azure.Core.Versions.v1_0_Preview_2)
+@route("/long-running")
+namespace LongRunning {
+  // case of LRO having pollingOperation in non-service namespace
+  @pollingOperation(NonService.poll)
+  @post
+  @route("/post")
+  op longRunning(): AcceptedResponse;
```
weidongxu-microsoft, line 20: The case of expected way to write typespec when service do not want the op in `@pollingOperation` get generated.

weidongxu-microsoft: There could be a disagreement on whether `PollResponse` model should be in namespace here.

haolingdong-msft: I'm also thinking about the namespace of `PollResponse`, as it will be used in the `Client`, so may be better putting it in the @service namespace.

haolingdong-msft (approved):
emitter code and generated code look good to me, just where to put `PollResponse` is a bit ambiguous.

typespec-tests/tsp/lro.tsp
weidongxu-microsoft: OK. Moved PollResponse to service namespace.

weidongxu-microsoft: Both should be same in current processing. However, when switched to TCGC `getAllModels`, this could have a difference.

Also the necessity of the model could be different across language. E.g. Java in `SyncPoller<PollResult, Void>` would need the model, while .NET in `Operation` probably not.

PR #2458, opened 2023-12-12
tsp, adapt finalResult from LroMetadata

weidongxu-microsoft:
typespec-azure-core https://github.com/Azure/typespec-azure-pr/pull/3982

Even if LroMetadata give `finalResult="void"`, we still use `##Poller<BinaryData, BinaryData>` for protocol API (so that it does not break DPG 1.0 SDK).

code https://github.com/Azure/autorest.java/pull/2458/commits/b35b757e44fb245e2e6bc333f8723ff0b5056311


typespec-tests/src/main/java/com/_specs_/azure/core/lro/rpc/legacy/LegacyClient.java
```diff
@@ -172,7 +172,7 @@ public JobResult getJob(String jobId) {
      */
     @Generated
     @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
-    public SyncPoller<JobResult, JobResult> beginCreateJob(JobData jobData) {
+    public SyncPoller<JobResult, Void> beginCreateJob(JobData jobData) {
```
haolingdong-msft, line 175: Does `Void` mean this `create` operation does not have final result? Wonder how to define in tsp if we want to have final result.

weidongxu-microsoft: Yes, if we follow the info given by LroMetadata, it is.

haolingdong-msft: I see. Thanks.

haolingdong-msft (approved):

Issue #2461, opened 2023-12-13
tsp, bug on projectedName and spread/flatten


weidongxu-microsoft:
case

```ts
  @route("send-projected-name")
  @post
  sendProjectedName(@query id: string, @projectedName("json", "file_id") fileId: string): OkResponse;
```

PR #2462, opened 2023-12-13
tsp, fix projectedName json on spread/flatten

weidongxu-microsoft:
case in openai https://github.com/Azure/azure-rest-api-specs-pr/blob/1647cabca96bb39ee8470246ebb52ebbcfce1797/specification/ai/OpenAI.Assistants/assistants/routes.tsp#L87-L89

fix https://github.com/Azure/autorest.java/issues/2461

case
```ts
  @route("send-projected-name")
  @post
  sendProjectedName(
    @query id: string,
    @projectedName("json", "file_id") @projectedName("client", "fileIdentifier") fileId: string,
  ): OkResponse;
```

tested case for discriminator (not sure whether to add to cadl-ranch, as we haven't seen this usage)
```ts
@discriminator("kind")
model Data {
  @projectedName("json", "kind_id")
  @projectedName("client", "type")
  kind: string;
}

model BytesData extends Data {
  kind: "bytes";
  data: bytes;
}
```


typespec-extension/src/code-model-builder.ts
```diff
@@ -1987,10 +1988,30 @@ export class CodeModelBuilder {
     const discriminator = getDiscriminator(this.program, type);
     if (discriminator) {
       discriminatorPropertyName = discriminator.propertyName;
```
weidongxu-microsoft, line 1990: `discriminator.propertyName` is the name

typespec-extension/src/code-model-builder.ts
```diff
@@ -1987,10 +1988,30 @@ export class CodeModelBuilder {
...
+          discriminatorProperty = Array.from(deriveModel.properties.values()).find(
+            (it) => it.name === discriminatorPropertyName,
+          );
+          if (discriminatorProperty) {
+            break;
+          }
+        }
+      }
+      const discriminatorPropertySerializedName = discriminatorProperty
+        ? this.getSerializedName(discriminatorProperty)
```
weidongxu-microsoft, line 2008: `this.getSerializedName(discriminatorProperty)` would give the serialized name (it differs in the case of `projectedName("json",)`)

typespec-extension/src/code-model-builder.ts
```diff
@@ -1273,8 +1273,9 @@ export class CodeModelBuilder {
         request.parameters = [];
         op.convenienceApi.requests.push(request);
 
-        for (const [key, _] of parameters.properties) {
-          const existParameter = op.parameters.find((it) => it.language.default.serializedName === key);
+        for (const [_, opParameter] of parameters.properties) {
+          const serializedName = this.getSerializedName(opParameter);
```
weidongxu-microsoft, line 1277: Same here, for the case that name and serializedName differs.

haolingdong-msft:
I don't quite get what is the relationship with spread? The case seems not have spread(i.e. `...`)

XiaofeiCao (commented):
Are there tests for flatten/grouping behavior?

XiaofeiCao (approved):
Unblock

weidongxu-microsoft:
> Are there tests for flatten/grouping behavior?

Case already added to PR and written in description.

weidongxu-microsoft:
> I don't quite get what is the relationship with spread? The case seems not have spread(i.e. `...`)

In typespec, it is equivalent to https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/parameters/spread/main.tsp#L45-L66

You can write `alias Whatever; op send(...Whatever)`, or just write `op send(param1, param2)`. They are same.

weidongxu-microsoft:
> > Are there tests for flatten/grouping behavior?
> 
> Case already added to PR and written in PR description.
> 
> This issue is not yet related to grouping so I didn't test the projectedName in such case (as long as all follows the "serializedName", it should be good).

Added a case for grouping https://github.com/Azure/autorest.java/pull/2465/files

PR #2464, opened 2023-12-13
stream-style-serialization, support nullable value map

haolingdong-msft:
For @alzimmermsft


weidongxu-microsoft:
could you use a better one to demonstrate?

e.g. `Record<InnerModel>`
https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/type/dictionary/models/InnerModel.java#L79

I assume the general problem is that if `null` value in map, the serialization (at current code) would fail?

PR #2467, opened 2023-12-14
Fixed the not working test case for PR#2445

v-hongli1:
1. The JsonTokes in the record from `ExtendsModelArrayClient.getWithResponse` are following as:
`START_OBJECT`,`FIELD_NAME`,`START_ARRAY`,`START_OBJECT`,`FIELD_NAME`,`STRING`,`END_OBJECT`,`END_ARRAY`,`END_OBJECT`
According to the logic of `ExtendsModelArrayAdditionalProperties#fromJson`
```
                 reader.nextToken();

                     additionalProperties = new LinkedHashMap<>();
                 }

                 additionalProperties.put(fieldName, null);
             }
```
The values of `reader.nextToken()` in the while statement evaluates expression are `FIELD_NAME`,`START_OBJECT`,`STRING`and `END_ARRAY`, so the while loop becomes an infinite loop.

2. `ExtendsModelArrayAdditionalProperties#fromJson` also does not complete the encapsulation of the `Array` section.

3. Please replace 
```
                 additionalProperties.put(fieldName, null);
```` 
with
```
                additionalProperties.put(fieldName, reader.readArray(ModelForRecord::fromJson));
```

### DIFF
--- a/typespec-tests/src/main/java/com/type/property/additionalproperties/models/ExtendsModelArrayAdditionalProperties.java
+++ b/typespec-tests/src/main/java/com/type/property/additionalproperties/models/ExtendsModelArrayAdditionalProperties.java
@@ -11,7 +11,6 @@ import com.azure.json.JsonSerializable;
 import com.azure.json.JsonToken;
 import com.azure.json.JsonWriter;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: typespec-tests/src/main/java/com/type/property/additionalproperties/models/ExtendsModelArrayAdditionalProperties.java [hunk 1]
```diff
@@ -90,9 +91,14 @@ public static ExtendsModelArrayAdditionalProperties fromJson(JsonReader jsonRead
                 }
 
                 additionalProperties.put(fieldName, null);
+                if (reader.currentToken() == JsonToken.START_ARRAY) {
+                    additionalProperties.replace(fieldName, new ArrayList<>());
+                    while (reader.nextToken() != JsonToken.END_ARRAY) {
+                        additionalProperties.get(fieldName).add(ModelForRecord.fromJson(reader));
+                    }
+                }
```
weidongxu-microsoft, line 99: Would it work if the code here be

`additionalProperties.put(fieldName, reader.readArray(reader1 -> ModelForRecord.fromJson(reader1)));`

v-hongli1: It can work, now fixed in the new version.

weidongxu-microsoft:
Please see https://github.com/Azure/autorest.java/pull/2470/files#diff-be5494678a67acad019c812c9ada00037515c9586128fdd84577800b75d5a2e1 on how to debug the codegen Java code in IDE.

Code change likely here
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java#L1135-L1142

It is a difficult task. Try it when you have time (but not high priority).

PR #2466, opened 2023-12-14
tsp, simplify BinaryData serialization

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/commit/367bb657bf5e7a45fecc8ea2a02033a75db49805
We had existing serialization tests like https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/test/java/com/cadl/union/UnionTests.java and cadl-ranch tests

also fixed a minor issue about MapType of javadoc https://github.com/Azure/autorest.java/commit/95f8fa3dfea482af14cb4711f6f8049f357dacff

PR #2469, opened 2023-12-14
tsp, fix regression on BinaryData serialization

weidongxu-microsoft:
fix a regression introduced in https://github.com/Azure/autorest.java/pull/2466

test case https://github.com/Azure/autorest.java/commit/8a26a51ada8ce256a718d15a7581622612d52901

remembered the reason I wrote the null protection logic -- the non-azure OpenAI fails the chat completion request upon "function: null"...

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
+++ b/typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
@@ -151,7 +151,7 @@ public final class FormData implements JsonSerializable<FormData> {
         jsonWriter.writeIntField("resolution", this.resolution);
         jsonWriter.writeStringField("type", this.type == null ? null : this.type.toString());
         jsonWriter.writeJsonField("size", this.size);
-        jsonWriter.writeUntypedField("image", this.image == null ? null : this.image.toObject(Object.class));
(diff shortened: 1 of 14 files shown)


XiaofeiCao (approved):

haolingdong-msft (approved):

typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
```diff
@@ -151,7 +151,7 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeIntField("resolution", this.resolution);
         jsonWriter.writeStringField("type", this.type == null ? null : this.type.toString());
         jsonWriter.writeJsonField("size", this.size);
-        jsonWriter.writeUntypedField("image", this.image == null ? null : this.image.toObject(Object.class));
```
haolingdong-msft, line 154: will we have issue if we remove null protection?

weidongxu-microsoft: The `image` property is required in tsp, so it should not be `null`. User will get a null exception if they pass a null, but I guess that is better than silent ignore.

PR #2475, opened 2023-12-19
Allow setting the working directory

alzimmermsft:
Adds an environment configuration `codegen.java.temp.directory` which allows for configuring the directory used when creating temporary folders. This allows for a few things:

- Clearer location and management of temporary files generated by Autorest Java. Allows to use a non-standard temporary directory that can more clearly show working files of Autorest Java.
- Potential optimizations of the temporary directory, such as using Windows Dev Drive.

### DIFF
--- a/extension-base/src/main/java/com/azure/autorest/extension/base/util/FileUtils.java
+++ b/extension-base/src/main/java/com/azure/autorest/extension/base/util/FileUtils.java
@@ -17,12 +17,12 @@ public final class FileUtils {
     /**
      * Creates a temporary directory.
      * <p>
-     * If the environment setting {@code autorest.java.temp.directory} is set, the directory will be created under the
+     * If the environment setting {@code codegen.java.temp.directory} is set, the directory will be created under the
      * specified path. Otherwise, the directory will be created under the system default temporary directory.
      * <p>
      * {@link System#getProperty(String)} is checked before {@link System#getenv(String)}.
      * <p>
-     * If {@code autorest.java.temp.directory} is set to a non-existent path, the directory will be created under the
+     * If {@code codegen.java.temp.directory} is set to a non-existent path, the directory will be created under the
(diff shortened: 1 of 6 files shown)


weidongxu-microsoft (approved):

extension-base/src/main/java/com/azure/autorest/extension/base/util/FileUtils.java
```diff
@@ -0,0 +1,47 @@
...
+        String tempDirectory = System.getProperty("codegen.java.temp.directory");
+        if (tempDirectory == null) {
+            tempDirectory = System.getenv("codegen.java.temp.directory");
+        }
+
+        if (tempDirectory != null) {
+            Path tempDirectoryPath = Paths.get(tempDirectory);
+            if (Files.exists(tempDirectoryPath)) {
+                return Files.createTempDirectory(tempDirectoryPath, prefix);
+            }
```
srnagar, line 42: Since this package already depends on azure-core, we can consider using `Configuration` to get system property or env var with a single method call.

alzimmermsft: I initially left this out as I don't know how tightly we want to couple code generation to azure-core as it becomes a bit of a circular dependency at times.

srnagar: Autorest/codegen will always take a dependency on azure-core package as we need to use the types defined in azure-core. So, this dependency is expected but the opposite direction for core to depend on autorest shouldn't be allowed and we should not have circular dependency.

weidongxu-microsoft:
@alzimmermsft 

Should we avoid dot in env var name? `codegen.java.temp.directory`

Maybe underscore? I need to google about how to set it in bash...

PR #2476, opened 2023-12-20
tsp, support text request/response body

weidongxu-microsoft:
cadl-ranch test https://github.com/Azure/autorest.java/pull/2476/commits/7d6004d8853e0c3091afe946dc76d74469c82886


typespec-tests/package.json
```diff
@@ -9,7 +9,7 @@
     "testserver-run": "npx cadl-ranch serve ./node_modules/@azure-tools/cadl-ranch-specs/http --coverageFile ./cadl-ranch-coverage-java.json"
   },
   "dependencies": {
-    "@azure-tools/cadl-ranch-specs": "0.28.0",
+    "@azure-tools/cadl-ranch-specs": "0.28.1",
```
weidongxu-microsoft, line 12: bump cadl-ranch

most new files is from this

javagen/src/main/java/com/azure/autorest/template/ConvenienceAsyncMethodTemplate.java
```diff
@@ -143,28 +154,45 @@ private String expressionConvertFromBinaryData(IType responseBodyType, IType raw
...
-            mapExpression = String.format("protocolMethodData -> protocolMethodData.toObject(%1$s.class)", responseBodyType.asNullable());
-        } else if (responseBodyType == ArrayType.BYTE_ARRAY) {
-            // byte[]
-            if (rawType == ClassType.BASE_64_URL) {
-                return "protocolMethodData -> protocolMethodData.toObject(Base64Url.class).decodedBytes()";
-            } else {
-                return "protocolMethodData -> protocolMethodData.toObject(byte[].class)";
-            }
+        SupportedMimeType mimeType = getResponseKnownMimeType(mediaTypes);
+        // TODO (weidxu): support XML etc.
```
weidongxu-microsoft, line 162: Currently I seems not able to get a XML type ObjectSerializer from azure-core, nor from azure-xml.

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -387,18 +392,65 @@ protected boolean isModelOrBuiltin(IType type) {
...
+    protected enum SupportedMimeType {
+        TEXT,
+        XML,
+        MULTIPART,
+        BINARY,
+        JSON
+    }
+
+    // azure-core SerializerEncoding.SUPPORTED_MIME_TYPES
+    private static final Map<String, SupportedMimeType> SUPPORTED_MIME_TYPES = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
haolingdong-msft, line 404: why do we use `TreeMap`?

haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
weidongxu-microsoft: azure-core uses TreeMap and I copied them https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/core/azure-core/src/main/java/com/azure/core/util/serializer/SerializerEncoding.java#L41

I don't really see the reason why TreeMap is used there as well. It has some difference in ordering when iterating, but their use is also a simple `Map.get`.

weidongxu-microsoft: OK, I get it. It is for ctor with `String.CASE_INSENSITIVE_ORDER`, which is not available in HashMap.

weidongxu-microsoft: Make sure you also do case-insensitive compare with "application/json-merge-patch"

PR #2479, opened 2023-12-26
Use in-memory code formatting

alzimmermsft:
Replaces Spotless with in-memory formatting which is effectively equivalent to what Spotless was doing. This adds dependencies on Google Java Format and Eclipse JDT Core to handle unused import removal and Eclipse-based code formatting.

Code formatting is now much quicker than it was before with Spotless which required multiple file reads and writes.

Longer term, it should be checked if Google Java Format can be replaced with Checkstyle's unused import check.


haolingdong-msft:
I'm a bit confused on the formatting part. Summarize my understanding for your pr, pleaase correct me if I'm wrong. 
- we are using in-memory formatting to replace spotless now. We do not use spotless not.
- Google format is in-memory formatting's dependency, we only use it to remove unused imports now.

alzimmermsft:
> I might lose some context on formatting part. Summarize my understanding for your pr, pleaase correct me if I'm wrong.
> 
> * we are using in-memory formatting to replace spotless now. We do not use spotless not.
> * Google format is in-memory formatting's dependency, we only use it to remove unused imports now.

Correct, Spotless is a wrapper around many different code formatters and it works really well. But for Autorest it requires in-memory files to be written out to disks, Spotless ran (which loads from disk and process), and Autorest must load from disk again to then write out the file to the final location. This results in a lot of overhead to run formatting.

We only use Eclipse code formatting and Google Java format, both which have in-memory equivalents. This PR replaces usage of Spotless with those and when running the generate scripts used in testing this is shown to reduce runtime by about 30%. With initiatives to regenerate all libraries in `azure-sdk-for-java` when changes are made here, this will have a large impact.

There is one issue to resolve in Google Java format making an API call to JVM internals which isn't allowed in later versions. I've been looking for ways to replace it with something else to handle removing unused imports (which oddly isn't something Eclipse code format can do).

eng/pipelines/ci.yaml
```diff
@@ -63,7 +63,6 @@ jobs:
         inputs:
           pwsh: true
           filePath: Generate.ps1
-          arguments: '-Parallelization 1'
```
weidongxu-microsoft, line 66: Are we sure the parallelization is stable enough for CI.

I do prefer longer time in CI, than CI not stable.

postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
weidongxu-microsoft, line 1: I assume here we kind of break down the spotless process logic (as I remember it uses google format to remove unused imports) to code?

typespec-tests/src/main/java/com/cadl/enumservice/EnumServiceAsyncClient.java
```diff
@@ -756,7 +756,7 @@ public Mono<String> setStringEnumArray(List<ColorModel> colorArray, List<ColorMo
         }
         return setStringEnumArrayWithResponse(colorArray.stream()
             .map(paramItemValue -> Objects.toString(paramItemValue, "")).collect(Collectors.toList()), requestOptions)
-                .flatMap(FluxUtil::toMono).map(protocolMethodData -> protocolMethodData.toObject(String.class));
+            .flatMap(FluxUtil::toMono).map(protocolMethodData -> protocolMethodData.toObject(String.class));
```
weidongxu-microsoft, line 759: There is some small diff. Not sure what's the cause. Maybe some version diff of dependency lib?

We can merge and hope SDK repo checkstyle still happy with it.

Report here should give me an early warning.
https://github.com/Azure/autorest.java/wiki/Fluent-Codegen-Report

weidongxu-microsoft (approved):

eng/pipelines/ci.yaml
alzimmermsft: This change should get us both faster and more stable CI 😄 

The reason it became unstable with Spotless being ran in parallel is that there could be multiple threads trying to download the Maven packages for Spotless and they could end up locking files against one another.

Now, all the code needed to run code formatting is bundled into the JAR, making it thread-safe.

postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
alzimmermsft: Yeah, so behind the scenes the Spotless configuration we were using ran Google Java Formatter to remove unused imports then ran Eclipse Code Formatter to format our code in the configured Eclipse style. This PR moves that functionality to being in-code directly within the Autorest packages rather than an out of process, file-based Spotless run.

typespec-tests/src/main/java/com/cadl/enumservice/EnumServiceAsyncClient.java
alzimmermsft: Yeah, this small formatting is something I'll keep an eye on whether we see skew between this change and running with Spotless once I begin Spotless integration in azure-sdk-for-java

Issue #2422, opened 2023-11-24
mpg, initial functional version


XiaofeiCao:
ApiView diff: https://apiview.dev/Assemblies/Review/a5478a8310b64355af6237d9e06c63c1?revisionId=56db5905a50e4d69b80d25d184d528bb&diffOnly=False&doc=False&diffRevisionId=47b359ec7aa94def8e7b89a06234bfda

```[tasklist]
### Tasks
- [ ] https://github.com/Azure/autorest.java/issues/2520
- [ ] https://github.com/Azure/autorest.java/issues/2518
- [ ] https://github.com/Azure/autorest.java/issues/2519
- [ ] https://github.com/Azure/autorest.java/issues/2487
- [ ] https://github.com/Azure/autorest.java/issues/2505
- [ ] when resource's update method is non-LRO, `apply` method will have wrong implementation
- [ ] https://github.com/Azure/autorest.java/issues/2484
- [ ] https://github.com/Azure/autorest.java/issues/2503
- [ ] https://github.com/Azure/autorest.java/issues/2481
- [x] `listBySubscription` and `get` need to be renamed to `list` and `getByResourceGroup`.
- [x] `listNext` is generated in resource collections(hardcoded)
- [x] Identify `subscriptionId` parameter in emitter. (hardcoded)
- [x] `TrackedResource(Base)`, `ProxyResource(Base)`, `ArmResource` shouldn't be generated. (hardcoded)
- [x] xxResourceProperties seems to be modeled as resource.
- [x] ParentResource not correctly identified, e.g. Association should be child of TrafficController
- [x] define/list/get/deleteById is removed from `AssociationInterfaces`(resource collection).
```

weidongxu-microsoft:
Could you prioritize the 1st one? I could not imagine how this could happen. Resource must have id+type+name.

XiaofeiCao:
Sure. 

I saw in generated code, that these resources like `AssociationInner` now extends newly generated model `TrackedResourceBase`, instead of `Resource`. This is not correct.

weidongxu-microsoft:
There is lots of processing on schema (and its parent schema) in https://github.com/Azure/autorest.java/blob/main/fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/ResourceTypeNormalization.java

Maybe this be the cause (as class hierarch is now different; previously it should only be subclass of `TrackedResource`)

XiaofeiCao:
[...]

PR #2480, opened 2023-12-29
mpg, fix compilation errors found in generating Azure Sphere and removed workaround code

XiaofeiCao:
issue: https://github.com/Azure/autorest.java/issues/2422
1. fixed hardcoded `TrackedResource(Base)`, `ProxyResource(Base)`, `ArmResource` resource identification.
2. fixed `listBySubscription` and `get` not renamed to `list` and `getByResourceGroup`.
3. fixed `listNext` is generated in resource collections
4. fixed hardcoded `subscriptionId` parameter in emitter.
5. fixed `systemData` in https://github.com/Azure/autorest.java/pull/2480/commits/eb344d3b64083bfba9b2a151bd1885ec0ed5e9a6, issue: https://github.com/Azure/autorest.java/issues/2481

generated code: https://github.com/Azure/azure-sdk-for-java/compare/main...XiaofeiCao:azure-sdk-for-java:mpg_fluentgen


javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
```diff
@@ -1703,15 +1703,19 @@ private static IType createTypeFromModelName(String modelName, JavaSettings sett
...
-            return new MethodNamer(proxyMethod.getName());
+            return resolveMethodNamer(proxyMethod);
+        }
+    }
+
+    protected final MethodNamer resolveMethodNamer(ProxyMethod proxyMethod) {
+        if (proxyMethod.isSync()) {
+            return new MethodNamer(proxyMethod.getBaseName());
         }
+        return new MethodNamer(proxyMethod.getName());
```
XiaofeiCao, line 1718: Fluent doesn't rely on convenienceApi.

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
```diff
@@ -310,86 +313,97 @@ private void addPagingNextOperation(Client client, OperationGroup operationGroup
...
+                  .forEach(param -> {
+                    nextOperationLocal.getRequests().get(0).getSignatureParameters().add(param);
+                  });
+        }
+        operation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
+        nextOperation.getExtensions().getXmsPageable().setNextOperation(nextOperation);
+        operationNextOperationMap.put(operationSignature, nextOperation);
+      } else {
+        // Same pageable operation should always share the same next link operation
+        nextOperation = operationNextOperationMap.get(operationSignature);
```
XiaofeiCao, line 405: Change here. 

Current implementation shares `operation.extension` across all generated NextOperation for the same operation. When the operation is processed multiple times(in different operation groups), previous generated NextOperation's `nextOperation` will be set to the newly generated one, causing it to be considered normal pageable operation. Hence `listNext` is generated in service client.

typespec-extension/package.json
```diff
@@ -72,7 +73,7 @@
     "@typescript-eslint/parser": "~6.13.2",
     "@typescript-eslint/eslint-plugin": "~6.13.2",
     "eslint-plugin-deprecation": "~2.0.0",
-    "@azure-tools/typespec-azure-core": "0.37.0",
+    "@azure-tools/typespec-azure-core": "0.37.2",
```
XiaofeiCao, line 76: need to set to 0.37.2 to work, haven't digged into the cause yet..

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
weidongxu-microsoft: I remember rename on operation is done in `OperationNameNormalization`.

Do we rename it anywhere else?

typespec-extension/src/operation-utils.ts
```diff
@@ -26,6 +26,7 @@ export const SPECIAL_HEADER_NAMES = new Set([
 ]);
 
 export const ORIGIN_API_VERSION = "modelerfour:synthesized/api-version";
+export const ORIGIN_SUBSCRIPTION_ID = "modelerfour:synthesized/subscription-id";
```
weidongxu-microsoft, line 29: m4 actually didn't have `modelerfour:synthesized/subscription-id`.

Maybe we should avoid adding this, till we really need it in Java code.

typespec-extension/src/emitter.ts
```diff
@@ -49,6 +49,7 @@ export interface DevOptions {
   "support-versioning"?: boolean;
   "debug"?: boolean;
   "loglevel"?: "off" | "debug" | "info" | "warn" | "error";
+  "java-temp-dir"?: string; // working directory for java codegen, e.g. transformed code-model file
```
weidongxu-microsoft, line 52: Does user or dev need to use this config?

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
```diff
@@ -235,11 +235,11 @@ private static boolean nonNullNextLink(Operation operation) {
     return operation.getExtensions().getXmsPageable().getNextLinkName() != null && !operation.getExtensions().getXmsPageable().getNextLinkName().isEmpty();
   }
 
-  private static class PagingNextOperationSignature {
+  private static class OperationSignature {
```
weidongxu-microsoft, line 238: Is it just a rename? If this is only for "next operation", the origin name seems more approparate?

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
```diff
@@ -258,7 +258,10 @@ public int hashCode() {
     }
   }
 
-  private final Map<PagingNextOperationSignature, Schema> pagingNextOperationResponseSchemaMap = new HashMap<>();
+  private final Map<OperationSignature, Schema> pagingNextOperationResponseSchemaMap = new HashMap<>();
+
+  // Operation -> NextOperation
```
weidongxu-microsoft, line 263: nit, Operation -> next page Operation

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
XiaofeiCao: here: https://github.com/Azure/autorest.java/blob/8c72ed64834e22cfebcf49fab81b29d90bf9d5ee/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java#L754-L761

It happens after preprocessor(`OperationNameNormalization`).

typespec-extension/src/operation-utils.ts
XiaofeiCao: Removed.

typespec-extension/src/emitter.ts
XiaofeiCao: Current codegen puts preprocessed code-model under `java.io.tmpdir`, which is very tricky to find for linux-based OS.
https://github.com/Azure/autorest.java/blob/ecf3d350fd1cd59459965d435d2ce30945cf5ab4/fluentnamer/src/main/java/com/azure/autorest/fluentnamer/FluentNamer.java#L63

 It also provides a way to override the default tmp dir:
https://github.com/Azure/autorest.java/blob/ecf3d350fd1cd59459965d435d2ce30945cf5ab4/extension-base/src/main/java/com/azure/autorest/extension/base/util/FileUtils.java#L33

This `java-temp-dir` is equivalent to the `codegen.java.temp.directory` in codegen.

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
XiaofeiCao: Yeah, just a rename. I'm more towards the new name. There are two places that use this class:
1. `Map<OperationSignature, Schema> pagingNextOperationResponseSchemaMap` which maps the next page operation to the paging operation's response schema.
2. `Map<OperationSignature, Operation> operationNextPageOperationMap`, added in this PR, maps paging operation to its next page operation.

In the first usage, it's next page operation, while in the second it's not(rather paging operation itself). Guess `OperationSignature` is more general?

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
XiaofeiCao: updated

typespec-extension/src/code-model-builder.ts
```diff
@@ -904,7 +903,11 @@ export class CodeModelBuilder {
 
   private processParameter(op: CodeModelOperation, param: HttpOperationParameter, clientContext: ClientContext) {
     function isSubscriptionId(sdkContext: SdkContext, param: HttpOperationParameter): boolean {
-      return "subscriptionId".toLocaleLowerCase() === param?.name?.toLocaleLowerCase() && param.param && isArmCommonType(param.param);
+      return (
+        "subscriptionId".toLocaleLowerCase() === param?.name?.toLocaleLowerCase() &&
+        param.param &&
```
weidongxu-microsoft, line 908: nit, maybe also verify that it is path param.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2146,14 +2154,39 @@ export class CodeModelBuilder {
 
   private dummyResourceSchema(type: Model, name?: string): ObjectSchema {
```
weidongxu-microsoft, line 2155: nit, see if you can move this func (and related ones) out of this ts (it is already too big), e.g. type-utils or external-schemas

typespec-extension/src/code-model-builder.ts
```diff
@@ -2589,24 +2622,18 @@ export class CodeModelBuilder {
...
-        origin: ORIGIN_SUBSCRIPTION_ID,
-        required: true,
-        protocol: {
-          http: new HttpParameter(ParameterLocation.Path),
-        },
-        language: {
-          default: {
-            serializedName: "subscriptionId",
-          },
+    return new Parameter("subscriptionId", "subscription ID", this.stringSchema, {
```
weidongxu-microsoft, line 2625: nit, "Subscription ID" as description?

or their description https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/parameters.tsp#L35-L48

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
weidongxu-microsoft: OK. Seems you've enlarged the usage of this class.

Previously it only created under `operation.getExtensions().getXmsPageable().getOperationName() != null` condition, and it only for the next page operation.

Here in `operationNextPageOperationMap` you seems using it for normal operation as well?

typespec-extension/package.json
weidongxu-microsoft: when you are about ready, you need to run `npm install` to update the package-lock file.

typespec-extension/src/code-model-builder.ts
```diff
@@ -548,7 +549,7 @@ export class CodeModelBuilder {
       );
       clientContext.preProcessOperations(this.sdkContext, client);
 
-      const operationGroups = listOperationGroups(this.sdkContext, client, true);
+      const operationGroups = listOperationGroups(this.sdkContext, client);
```
weidongxu-microsoft, line 552: why this change?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1975,6 +1987,14 @@ export class CodeModelBuilder {
 
   private processObjectSchema(type: Model, name: string): ObjectSchema {
     const namespace = getNamespace(type);
+    if (this.isArm() && isArmCommonType(type) && name?.includes("Resource")) {
+      const objectSchema = this.dummyResourceSchema(type, name);
+      this.codeModel.schemas.add(objectSchema);
+  
+      // cache this now before we accidentally recurse on this type
```
weidongxu-microsoft, line 1994: comment seems not valid?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: OK, probably not possible.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2123,13 +2143,52 @@ export class CodeModelBuilder {
 
     if (type.kind === "Model") {
       const effective = getEffectiveModelType(program, type, isSchemaProperty);
-      if (effective.name) {
+      if (this.isArm() && getNamespace(effective as Model)?.startsWith("Azure.ResourceManager")) {
+        return type;
+      } else if (effective.name) {
         return effective;
       }
```
weidongxu-microsoft, line 2150: why this branch?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Better to comment here about what you are doing.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: Wrongly modified.. Reverted, thanks...

typespec-extension/src/code-model-builder.ts
XiaofeiCao: Seems removing the deduplication logic here still works... Looking into the cause..

typespec-extension/src/code-model-builder.ts
XiaofeiCao: For this particular model:
```typescript
model Catalog is TrackedResource<CatalogProperties> {
  @doc("Name of catalog")
  @pattern("^[A-Za-z0-9_-]{1,50}$")
  @key("catalogName")
  @path
  @segment("catalogs")
  name: string;
}
```

`getEffectiveModelType` will return `TrackedResource` instead of `Catalog`. Resulting in `TrackedResourceCatalogProperties` been generated as the response body type.

Not sure what the expected behavior is. Is this a bug for `getEffectiveModelType`?

weidongxu-microsoft: OK, if it makes it wrong, then keep this.

`getEffectiveModelType` would find a model that be same of the `type` (excluding those `@path` `@header` `@query`).
It may find an unexpected model, if they are same otherwise.

PS: maybe we can see if we call `getEffectiveModelType` only if `type.name` is empty. No need to do this here.

XiaofeiCao: Yeah. After adding all the properties to the dummy model, it's now considered effective model...

preprocessor/src/main/java/com/azure/autorest/preprocessor/tranformer/Transformer.java
XiaofeiCao: Added comment.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: updated

typespec-extension/src/code-model-builder.ts
XiaofeiCao: added

typespec-extension/src/code-model-builder.ts
XiaofeiCao: We won't recurse on this dummy schema, so not needed.

typespec-extension/package.json
XiaofeiCao: added

typespec-extension/src/code-model-builder.ts
```diff
@@ -902,7 +902,7 @@ export class CodeModelBuilder {
   }
 
   private processParameter(op: CodeModelOperation, param: HttpOperationParameter, clientContext: ClientContext) {
-    if (clientContext.apiVersions && isApiVersion(this.sdkContext, param)) {
+    if ((clientContext.apiVersions || this.isArm()) && isApiVersion(this.sdkContext, param)) {
```
XiaofeiCao, line 905: mgmt typespec hasn't used `@versioned` yet:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/sphere/Sphere.Management/main.tsp#L25

typespec-tests/tsp/arm/tspconfig.yaml
weidongxu-microsoft, line 1: Why do we need a different tspconfig (and a separate folder)?

TypeSpecPlugin for mgmt should already sets the options required for mgmt?

typespec-tests/tsp/arm/arm.tsp
```diff
@@ -0,0 +1,100 @@
...
+import "@azure-tools/typespec-azure-resource-manager";
+
+@armProviderNamespace
+@service({
+  title: "ArmResource",
+  version: "2023-11-01",
+})
+@doc("Arm Resource Provider management API.")
+@useDependency(Azure.ResourceManager.Versions.v1_0_Preview_1)
+@useDependency(Azure.Core.Versions.v1_0_Preview_1)
```
weidongxu-microsoft, line 14: use latest of them

typespec-tests/tsp/arm/arm.tsp
```diff
@@ -0,0 +1,100 @@
...
+namespace Cadl.ArmResource;
```
weidongxu-microsoft, line 15: nit, `ArmResourceProvider`?

typespec-tests/tsp/arm/arm.tsp
```diff
@@ -0,0 +1,100 @@
...
+@lroStatus
+enum ProvisioningState {
+  Provisioning,
+  Updating,
+  Deleting,
+  Accepted,
+  Succeeded,
+  Failed,
+  Canceled,
+}
```
weidongxu-microsoft, line 79: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/models.tsp#L144

typespec-tests/tsp/arm/arm.tsp
```diff
@@ -0,0 +1,100 @@
...
+
+@doc("Resource ID definition used by parent to reference child resources.")
+model ResourceId {
+  @doc("Resource ID of child resource.")
+  id: string;
+}
```
weidongxu-microsoft, line 85: SubResource?

Or https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/models.tsp#L108 ?

Basically I didn't see a reason this kind of basic model need to be defined...

typespec-tests/tsp/arm/tspconfig.yaml
XiaofeiCao: mgmt tests will generate random mock values so I disabled `generate-tests` in the new tspconfig. 

For this particular case, an alternative is to add .gitignore to ignore the generated tests. I'm not sure in the future whether mgmt will need to test configs of its own(service-name, title, etc), so I added it in a separate folder. 

Any suggestions?

typespec-tests/tsp/arm/arm.tsp
weidongxu-microsoft: Maybe check with Chenjie about whether there is `SubResource` in resourcemanager lib.

typespec-tests/tsp/arm/tspconfig.yaml
weidongxu-microsoft: 2 options come to mind:
1. turn it off in script https://github.com/Azure/autorest.java/blob/main/typespec-tests/Generate.ps1#L52-L54
2. use a `tsp-arm` folder? at least this feels clearer.

weidongxu-microsoft: And do we need `/typespec-tests/tsp/arm/tsp-output` in .gitignore?

XiaofeiCao: > And do we need `/typespec-tests/tsp/arm/tsp-output` in .gitignore?

If we use a separate folder and put new `tspconfig.yaml` there, it's needed. Seems `project-root` is considered where the `tspconfig.yaml` is.

XiaofeiCao: For now I'll turn if off in script.

javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java
weidongxu-microsoft, line 1: All the changes here about `resolveMethodNamer` seems only because we didn't rename name in `convenienceApi` in `OperationNameNormalization`.

Should we do this there instead?

typespec-tests/tsp/arm/arm.tsp
XiaofeiCao: Chenjie is aware of ResourceIdentifier:
https://github.com/Azure/autorest/blob/faf5c1168232ba8a1e8fe02fbc28667c00db8c96/packages/extensions/openapi-to-typespec/test/arm-networkanalytics/tsp-output/models.tsp#L226

I found Fleet has such use case:
https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/containerservice/Fleet.Management/update/run.tsp#L42-L47

swagger equivalent:
https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/containerservice/resource-manager/Microsoft.ContainerService/fleet/preview/2023-08-15-preview/fleets.json#L2182

.Net treat it as ResourceIdentifier:
https://github.com/Azure/azure-sdk-for-net/blob/b9469adaa8f5cb0076ffcb9b71c667519a300b1d/sdk/fleet/Azure.ResourceManager.ContainerServiceFleet/src/Generated/ContainerServiceFleetUpdateRunData.cs#L83

Not SubResource(in VM):
https://github.com/Azure/azure-sdk-for-net/blob/b9469adaa8f5cb0076ffcb9b71c667519a300b1d/sdk/compute/Azure.ResourceManager.Compute/src/Generated/VirtualMachineData.cs#L163

Though he's not sure how to represent `SubResource` in tsp either. I'll remove the reference property here till we reach agreement on how to define it.

.gitignore
```diff
@@ -89,6 +89,7 @@ javagen/*.yaml
 /typespec-tests/cadl-ranch-coverage.json
 /typespec-tests/cadl-ranch-coverage-java.json
 /typespec-tests/tsp-src/
+/typespec-tests/tsp/arm/tsp-output
```
weidongxu-microsoft, line 92: Seems no longer need this?

weidongxu-microsoft (approved):

.gitignore
XiaofeiCao: removed

PR #2488, opened 2024-01-02
tsp, fix handling of client api-version (Azure-styled api-version + typespec versioned)

weidongxu-microsoft:
major change is here
https://github.com/Azure/autorest.java/pull/2488#discussion_r1440009390
https://github.com/Azure/autorest.java/pull/2488#discussion_r1440009515


typespec-extension/src/code-model-builder.ts
```diff
@@ -896,7 +896,8 @@ export class CodeModelBuilder {
   }
 
   private processParameter(op: CodeModelOperation, param: HttpOperationParameter, clientContext: ClientContext) {
-    if (isApiVersion(this.sdkContext, param)) {
+    if (clientContext.apiVersions && isApiVersion(this.sdkContext, param)) {
+      // pre-condition for "isApiVersion": the client supports ApiVersions
```
weidongxu-microsoft, line 900: Major change is here.

If the client is not versioned, do not take the result of "isApiVersion". Make it a normal parameter.

typespec-extension/src/code-model-builder.ts
```diff
@@ -519,18 +518,19 @@ export class CodeModelBuilder {
...
+      //   const service = getService(this.program, client.service);
+      //   if (service?.version) {
+      //     codeModelClient.apiVersions = [];
+      //     const apiVersion = new ApiVersion();
+      //     apiVersion.version = service.version;
+      //     codeModelClient.apiVersions.push(apiVersion);
+      //   } else {
+      //     throw new Error(`API version not available for client ${client.name}.`);
+      //   }
+      // }
```
weidongxu-microsoft, line 533: Do not use `@service.version` as api-version

weidongxu-microsoft: However, we may reconsider this

From Timothee

- if the service provides no versioning: then this is kinda an error if we were explicit about the param. Should just be a regular param
- if the service provide a version via `@service version` it technically is a constant param
- if the service provide `@versioned` then it should allow any of those values 

And I kind of agree. `@service` without `version` seems best suit to say that the service does not have any version support.

typespec-tests/tsp/builtin.tsp
```diff
@@ -6,7 +6,6 @@ using Azure.ClientGenerator.Core;
 
 @service({
   title: "Builtin",
-  version: "1.0.0",
 })
 namespace Cadl.Builtin;
```
weidongxu-microsoft, line 10: Ideally, this is the way to say that this service does not support `version`.

typespec-tests/src/main/java/com/server/versions/notversioned/NotVersionedAsyncClient.java
```diff
@@ -114,15 +118,17 @@ public Mono<Void> withoutApiVersion() {
      */
     @Generated
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public Mono<Void> withQueryApiVersion() {
+    public Mono<Void> withQueryApiVersion(String apiVersion) {
```
haolingdong-msft, line 121: One question, for not-versioned, in tsp definition, the `api-version` query param can be any name, it's just like a normal query param?

weidongxu-microsoft: Yeah, for that test of "not-versioned", it intends to say that the service is not versioned, so no parameter (even it is called `api-version`) mean anything related to version, and hence just a normal parameter like others.

But the cadl-ranch test may not be the best case for that because it has "version" in `@service`. (discussed https://github.com/Azure/autorest.java/pull/2488#discussion_r1440009515 -- but anyway I assume Azure service would not do this)

haolingdong-msft (approved):
LGTM

Issue #2485, opened 2024-01-02
tsp, add test for cadl-ranch Type/Model and Type/Property


weidongxu-microsoft:
![image](https://github.com/Azure/autorest.java/assets/53292327/c59e7eaf-fb35-4362-9c0a-2d919756e620)

PR #2489, opened 2024-01-03
tsp, add test for cadl-ranch Type/Model and Type/Property

v-hongli1:
Add test cases for [issue#2485](https://github.com/Azure/autorest.java/issues/2485)

### DIFF
--- a/typespec-tests/src/main/java/com/type/model/inheritance/nesteddiscriminator/models/Salmon.java
+++ b/typespec-tests/src/main/java/com/type/model/inheritance/nesteddiscriminator/models/Salmon.java
@@ -116,11 +116,11 @@ public final class Salmon extends Fish {
     @Override
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
-        jsonWriter.writeIntField("age", getAge());
         jsonWriter.writeStringField("kind", "salmon");
-        jsonWriter.writeJsonField("partner", this.partner);
+        jsonWriter.writeIntField("age", getAge());
         jsonWriter.writeArrayField("friends", this.friends, (writer, element) -> writer.writeJson(element));
         jsonWriter.writeMapField("hate", this.hate, (writer, element) -> writer.writeJson(element));
+        jsonWriter.writeJsonField("partner", this.partner);
         return jsonWriter.writeEndObject();
     }
 
--- a/typespec-tests/src/main/java/com/type/model/inheritance/nesteddiscriminator/models/Salmon.java
+++ b/typespec-tests/src/main/java/com/type/model/inheritance/nesteddiscriminator/models/Salmon.java
@@ -116,11 +116,11 @@ public final class Salmon extends Fish {
     @Override
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
-        jsonWriter.writeStringField("kind", "salmon");
         jsonWriter.writeIntField("age", getAge());
+        jsonWriter.writeStringField("kind", "salmon");
+        jsonWriter.writeJsonField("partner", this.partner);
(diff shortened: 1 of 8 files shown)


typespec-tests/src/main/java/com/type/model/inheritance/nesteddiscriminator/models/Salmon.java
```diff
@@ -116,11 +116,11 @@ public Salmon setPartner(Fish partner) {
     @Override
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
-        jsonWriter.writeStringField("kind", "salmon");
         jsonWriter.writeIntField("age", getAge());
+        jsonWriter.writeStringField("kind", "salmon");
```
haolingdong-msft, line 120: This is directly generated, or manual change?

v-hongli1: This is directly generated, so removed now.

typespec-tests/src/test/java/com/type/property/additionalproperties/ExtendsTests.java
```diff
@@ -104,4 +108,43 @@ public void testExtendsUnknownClient() {
         Assertions.assertEquals("ExtendsUnknownAdditionalProperties", properties.getName());
         Assertions.assertIterableEquals(propertyMap.entrySet(), properties.getAdditionalProperties().entrySet());
     }
+
+    @Test
+    @Disabled("Polymorphic deserialization doesn't support multiple levels of inheritance in Jackson, https://github.com/FasterXML/jackson-databind/issues/1188")
```
weidongxu-microsoft, line 113: I don't think this is the reason?

If you are not sure about it, just keep it empty.

typespec-tests/src/test/java/com/type/model/inheritance/NestedDiscriminatorTests.java
weidongxu-microsoft, line 1: This case would be lowest priority. We haven't decided to support the "multiple levels of inheritance" case.

typespec-tests/src/test/java/com/type/property/additionalproperties/ExtendsTests.java
v-hongli1: The reason for disable has been fixed in the new version.

typespec-tests/src/test/java/com/type/model/inheritance/NestedDiscriminatorTests.java
v-hongli1: Test cases `putModel` and `putRecursiveModel` have been set to disabled.

typespec-tests/src/test/java/com/type/model/inheritance/NestedDiscriminatorTests.java
```diff
@@ -39,10 +45,35 @@ void getRecursiveModel() {
     }
 
 
-    @Disabled("Polymorphic deserialization doesn't support multiple levels of inheritance in Jackson, https://github.com/FasterXML/jackson-databind/issues/1188")
     @Test
+    @Disabled("1. The item `kind` is missing in the generated json file by the method `toJson` of the class `Shark`. " +
+            "2. The output order of each item in the generated json file by the method `toJson` of the class `Salmon` is incorrect.")
```
weidongxu-microsoft, line 50: Why order affect the test? JSON object does not care about order, isn't it?

v-hongli1: Fixed in the new version.

weidongxu-microsoft (approved):

weidongxu-microsoft:
https://github.com/Azure/autorest.java/issues/2500

weidongxu-microsoft:
some of them being fixed in https://github.com/Azure/autorest.java/pull/2504

Issue #2483, opened 2024-01-02
mgmt, add byte-buddy and byte-buddy-agent to dependency


weidongxu-microsoft:
depends on https://github.com/Azure/azure-sdk-for-java/pull/38144

PR #2490, opened 2024-01-04
client, add bytebuddy to test dependency in pom

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2483

it (mockito as test dependency) takes effect when `--generate-tests` is enabled.

output pom
```diff
       <version>4.11.0</version> <!-- {x-version-update;org.mockito:mockito-core;external_dependency} -->
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>net.bytebuddy</groupId>
+      <artifactId>byte-buddy</artifactId>
+      <version>1.12.23</version> <!-- {x-version-update;net.bytebuddy:byte-buddy;external_dependency} -->
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>net.bytebuddy</groupId>
+      <artifactId>byte-buddy-agent</artifactId>
+      <version>1.14.8</version> <!-- {x-version-update;net.bytebuddy:byte-buddy-agent;external_dependency} -->
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
```


javagen/src/main/java/com/azure/autorest/mapper/PomMapper.java
```diff
@@ -53,6 +53,10 @@ private Pom createAzurePom(Project project) {
                 Project.Dependency.JUNIT_JUPITER_ENGINE, true);
         addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes,
                 Project.Dependency.MOCKITO_CORE, true);
+        addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes,
+                Project.Dependency.BYTE_BUDDY, true);
+        addDependencyIdentifier(dependencyIdentifiers, addedDependencyPrefixes,
+                Project.Dependency.BYTE_BUDDY_AGENT, true);
```
weidongxu-microsoft, line 59: I actually didn't remember why we add `mockito-core` by default to DPG (under `generate-tests` option)

eng/versioning/external_dependencies.txt
```diff
@@ -41,6 +41,8 @@ org.apache.maven.plugins:maven-source-plugin
 org.apache.maven.plugins:maven-surefire-plugin
 org.codehaus.mojo:exec-maven-plugin
 
-org.mockito:mockito-core;4.11.0
```
haolingdong-msft, line 44: why remove `4.11.0`?

weidongxu-microsoft: When syncing, we always take version from SDK repo e.g. https://github.com/Azure/azure-sdk-for-java/blob/main/eng/versioning/external_dependencies.txt

So we don't have version here in this repo.

XiaofeiCao (approved):
nit, maybe a comment on the reason we include the dependency, like this here?
https://github.com/Azure/azure-sdk-for-java/pull/38144/files#diff-35f1d760f519daf49663c7a20a9a25e9a509c63377455c505c1af1beb7cffa41R78

haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/mapper/PomMapper.java
weidongxu-microsoft: Emm, actually those in DPG is not under `generate-tests` option.

weidongxu-microsoft: @alzimmermsft @srnagar 
Any preference about whether we should include all the test dependencies when generating pom.xml for DPG?

Some like `azure-core-test` and `azure-identity` should actually be almost always needed by dev in follow up work, when adding test cases (also when we generate test cases under `generate-tests` option).

But dev may not use `mockito` as it would only for mock test, while mostly they do record/playback test.

Should I remove `mockito` in DPG pom?

alzimmermsft: If the generated code isn't using Mockito I'd remove it. While Mockito helps out a lot in testing it's one of those dependencies that has a lot of gotchas around support in newer Java versions or gotchas during usage which could be replaced by slightly more verbose testing code (creating mock classes locally, re-writing tests to not use mocking).

weidongxu-microsoft: Got you. Let me remove it from DPG.

weidongxu-microsoft:
need a fix, find that they uses `testdep_` prefix...

```
testdep_net.bytebuddy:byte-buddy;1.14.8
testdep_net.bytebuddy:byte-buddy-agent;1.14.8
```

Issue #2473, opened 2023-12-19
tsp, follow up on multipart/form-data


weidongxu-microsoft:
A model only used by multipart/form-data better not to have Jackson annotation nor extend JsonSerialization.

PR #2494, opened 2024-01-08
tsp, refactor, move multipart in code-model to serializationFormats

weidongxu-microsoft:
Moved from `usage` to `serializationFormats`.

```yaml
      usage:
        - input
        - public
      serializationFormats:
        - multipart
```

When we supports `xml`, it would also be in `serializationFormats`. (`xml` would have further data in `serialization` property)

This would be compatible with code-model from m4.

---

One existing problem here (same for m4) is that `json` is not in `serializationFormats`.
So the assumption is kind of -- if it is not specified otherwise (`xml` or `multipart`), it is `json`

prepare for https://github.com/Azure/autorest.java/issues/2473


typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -159,7 +158,7 @@ protected void writeHelperClasses(Client client, JavaPackage javaPackage, JavaSe
         // MultipartFormDataHelper
         final boolean generateMultipartFormDataHelper = client.getModels().stream()
                 .filter(ModelUtil::isGeneratingModel)
-                .anyMatch(m -> m.getImplementationDetails() != null && m.getImplementationDetails().getUsages().contains(ImplementationDetails.Usage.MULTIPART_FORM_DATA));
+                .anyMatch(m -> m.getSerializationFormats().contains("multipart"));
```
haolingdong-msft, line 161: "multipart" -> use constant?

weidongxu-microsoft: fixed via `KnownMediaType.MULTIPART.value()`

haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientModel.java
```diff
@@ -670,6 +683,17 @@ public Builder crossLanguageDefinitionId(String crossLanguageDefinitionId) {
             return this;
         }
 
+        /**
+         * Sets the cross language definition id for the model.
```
XiaofeiCao, line 687: nit, comment is wrong

weidongxu-microsoft: Thanks. Fixed.

PR #2496, opened 2024-01-08
tsp, remove serialization interface/code from multipart model

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2473

Add this for case that the model does not need any serialization code.
```java
    private static boolean modelRequireSerialization(ClientModel model) {
        // TODO (weidxu): any other case? "binary"?
        return !model.getSerializationFormats().contains("multipart");
    }
```

`model.getSerializationFormats()` is due to https://github.com/Azure/autorest.java/pull/2494 (if any recommendation for this, I can improve here as well)

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
+++ b/typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
@@ -7,17 +7,12 @@ package com.cadl.multipart.models;
 import com.azure.core.annotation.Fluent;
 import com.azure.core.annotation.Generated;
 import com.azure.core.util.BinaryData;
-import com.azure.json.JsonReader;
-import com.azure.json.JsonSerializable;
-import com.azure.json.JsonToken;
-import com.azure.json.JsonWriter;
-import java.io.IOException;
 
 /**
  * The FormData model.
  */
 @Fluent
-public final class FormData implements JsonSerializable<FormData> {
+public final class FormData {
     /*
      * The name property.
      */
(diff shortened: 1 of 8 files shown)


typespec-tests/src/main/java/com/cadl/multipart/models/FormData.java
```diff
@@ -7,17 +7,12 @@
...
-import com.azure.json.JsonToken;
-import com.azure.json.JsonWriter;
-import java.io.IOException;
 
 /**
  * The FormData model.
  */
 @Fluent
-public final class FormData implements JsonSerializable<FormData> {
+public final class FormData {
```
weidongxu-microsoft, line 15: Most for this.

In Jackson it is annotations (so changing them would not count as breaking changes).

But in stream-style, it would be in class signature. Here we'd better to avoid having serialization code when not needed.

haolingdong-msft: Can a model be used both for `form/multipart` and other content types? Do we add the stream-style serialization code into the model if a model can is shared.

weidongxu-microsoft: It can. But probably not going to happen (having same model for multipart and json seems not meaningful).

The short-coming is partly m4. As mentioned in https://github.com/Azure/autorest.java/pull/2494, `json` is not included in `serializationFormats`.

haolingdong-msft: ok, I'm fine with it for now.

haolingdong-msft (approved):

PR #2499, opened 2024-01-09
tsp, fix subclass of class with additionalProperties - part1

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/pull/2499/commits/7833aa3b1ffb0ef46f5034af8c1d0d7ab484c532
Jackson should be good.

pending part2 on fix to stream-style https://github.com/Azure/autorest.java/issues/2500


typespec-extension/src/code-model-builder.ts
```diff
@@ -2062,8 +2062,16 @@ export class CodeModelBuilder {
       }
     } else if (isRecordModelType(this.program, type)) {
       // "pure" Record processed elsewhere
+
       // "mixed" Record that have properties, treat the model as "additionalProperties"
-      const parentSchema = this.processDictionarySchema(type, this.getName(type));
```
weidongxu-microsoft, line 2066: Bug was that previously this `type` get cached as a `DictionarySchema` in this line.

It should be cached as an `ObjectSchema`, while its sourceModel be potentially `DictionarySchema`

weidongxu-microsoft:
As typespec, the `additionalProperty` is not a "meta concept".

It is currently done like this
```ts
model SuperClass is Record<unknown> {
  prop1InSuperClass: string;
}

model Class extends SuperClass {
  prop1InClass: string;
}
```

So it is a basic "extends" syntax.

PS: typespec is discussing writing this as
```ts
model SuperClass {
  ...Record<unknown>;
  prop1InSuperClass: string;
}

model Class extends SuperClass {
  prop1InClass: string;
}
```

typespec-tests/src/main/java/com/type/property/additionalproperties/models/IsFloatAdditionalProperties.java
```diff
@@ -26,7 +26,7 @@ public final class IsFloatAdditionalProperties implements JsonSerializable<IsFlo
     private final double id;
 
     /*
-     * The model is from Record<float32> type.
+     * Dictionary of ExtendsUnknownAdditionalPropertiesDerivedAge
      */
```
haolingdong-msft, line 30: A bit confused, the dictionay is `Map<String, Double>`, why is this 'Dictionary of `ExtendsUnknownAdditionalPropertiesDerivedAge`', not dictionay of `Double`? Is it saying it's the `age` property of `ExtendsUnknownAdditionalPropertiesDerived` model?

weidongxu-microsoft: Yeah, this is a bit strange. (seems from the doc of the Record<>)

I will take a look.

weidongxu-microsoft: The cause is the `float` got names as `ExtendsUnknownAdditionalPropertiesDerivedAge`. This alone won't affect anything, as we don't generate a float class/struct.

But the `namer` module would find that there is no `@doc` on the Record, and use the name above to coin a description, hence the strange name.

weidongxu-microsoft: fix in https://github.com/Azure/autorest.java/pull/2499/commits/51f4ea5cd0c3869882c1c41f42217d39cd312dcc and https://github.com/Azure/autorest.java/pull/2499/commits/7f5c7434f7dc22cba074a521aeee00df2dc938ee

also disabled `namer` in test, and hide it in readme.

haolingdong-msft (approved):

XiaofeiCao (approved):

Issue #2500, opened 2024-01-09
tsp, bug, subclass of class with additionalProperty


weidongxu-microsoft:
Works in Jackson, but not work in stream-style. See https://github.com/Azure/autorest.java/pull/2499

---

[`IsUnknownAdditionalPropertiesDerived`](https://github.com/Azure/autorest.java/blob/2bcc238e3edefdc09a2960dbc3ea1239ee85bf64/typespec-tests/src/main/java/com/type/property/additionalproperties/models/IsUnknownAdditionalPropertiesDerived.java) is a subclass of [`IsUnknownAdditionalProperties`](https://github.com/Azure/autorest.java/blob/2bcc238e3edefdc09a2960dbc3ea1239ee85bf64/typespec-tests/src/main/java/com/type/property/additionalproperties/models/IsUnknownAdditionalProperties.java) which has `additionalProperties`.

Therefore the serialization code of `IsUnknownAdditionalPropertiesDerived` would need to process `additionalProperties` as well.

Plan to generate code like this in subclass.

```diff
    @Override
    public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
        jsonWriter.writeStartObject();
        jsonWriter.writeStringField("name", getName());
        jsonWriter.writeIntField("index", this.index);
        jsonWriter.writeNumberField("age", this.age);
+        if (this.getAdditionalProperties() != null) {
+            for (Map.Entry<String, Object> additionalProperty : this.getAdditionalProperties().entrySet()) {
+                jsonWriter.writeUntypedField(additionalProperty.getKey(), additionalProperty.getValue());
+            }
+        }
        return jsonWriter.writeEndObject();
    }

    public static IsUnknownAdditionalPropertiesDerived fromJson(JsonReader jsonReader) throws IOException {
        return jsonReader.readObject(reader -> {
            String name = null;
            int index = 0;
            Double age = null;
+            Map<String, Object> additionalProperties = null;
            while (reader.nextToken() != JsonToken.END_OBJECT) {
                String fieldName = reader.getFieldName();
                reader.nextToken();

                if ("name".equals(fieldName)) {
                    name = reader.getString();
                } else if ("index".equals(fieldName)) {
                    index = reader.getInt();
                } else if ("age".equals(fieldName)) {
                    age = reader.getNullable(JsonReader::getDouble);
                }  else {
-                    reader.skipChildren();
+                    if (additionalProperties == null) {
[...]

PR #2504, opened 2024-01-11
tsp, fix additionalProperties from superclass

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2500

discuss https://github.com/Azure/autorest.java/pull/2499#issuecomment-1884058050


javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
```diff
@@ -52,6 +52,7 @@ public final class ClientModelPropertiesManager {
     private final List<ClientModelProperty> superRequiredProperties;
     private final List<ClientModelProperty> superSetterProperties;
     private final List<ClientModelProperty> superReadOnlyProperties;
+    private final ClientModelProperty superAdditionalProperties;
```
weidongxu-microsoft, line 55: Should we call it `superAdditionalPropertiesProperty`?

Currently follow naming of `private final ClientModelProperty additionalProperties;`

XiaofeiCao (approved):

haolingdong-msft (approved):

Issue #2468, opened 2023-12-14
partial-update, not treat static block as customized code even it does not have `@Generated` annotation


haolingdong-msft:

weidongxu-microsoft:
What if we do this `// @Generated <tag>`, and base on keep

```
    // @Generated setResourceAccessor
    static {
        JsonMergePatchHelper.setResourceAccessor((model, jsonMergePatchEnabled) -> {
            model.serializeAsJsonMergePatch(jsonMergePatchEnabled);
            return model;
        });
    }
```

`// @Generated <tag>` block in generated code would overwrite `// @Generated <tag>` block in existing source.

Block without `// @Generated` in existing source will not be overwritten. (with solves https://github.com/Azure/autorest.java/issues/2493)

This approach maybe unnecessary complex. Though it seems won't hurt if we start with generate the block with `// @Generated <tag>`. So that in future, at least we know the block is generated.

haolingdong-msft:
Thanks @weidongxu-microsoft, I like the approach to identify the code block is generated for JsonMergePatch(or something else). During partial update, we would be able to identify the corresponding newly generated code and replace the code block in existing file. And we can support partial-update for static code block as well.

I have thought about a solution similar as you previously, but did not add to the issue's design options as I thought it is a bit complex and I'm not sure if we want user to customize static code block. I added it as option3 in the issue's description now for reference. Difference is my solution needs to update the `@Generated` annotation, while yours does not need to.

srnagar:
Static blocks are used only for the accessors and nothing else today. So, we should treat this as a generated block and always overwrite. We don't need partial update support for this scenario (yet).

PR #2506, opened 2024-01-11
Enable stream-style-serialization by default

alzimmermsft:
Enables `stream-style-serialization` by default for Swagger and TypeSpec code generation. Also removes `enable-xml` as it isn't used anymore.


weidongxu-microsoft:
One reminding issue before we can do the switch is the missing `@Generated` annotation on "toJson" "fromJson" method
https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java#L132-L197

@haolingdong-msft

weidongxu-microsoft:
It could depends on the decision in https://github.com/Azure/autorest.java/issues/2468 as well.
(json-merge-patch in stream-style model would generate code with static block, and we need to decide on how partial-update handles that)

haolingdong-msft:
> One reminding issue before we can do the switch is the missing `@Generated` annotation on "toJson" "fromJson" method https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java#L132-L197
> 
> @haolingdong-msft
> 
> PS: we probably need to let openai still use Jackason due to the "missing 'kind'" discussion (or we may need to customize the "fromJson" for them to handle the case).

Yes, Hi @alzimmermsft, will you fix this, or you want me to fix this? (I have not fixed it currently as you said you will add this) https://github.com/Azure/autorest.java/pull/2410#discussion_r1408712850 Thanks!

alzimmermsft:
Thanks @haolingdong-msft!

It appears this PR needs some more work as the types used in Management plane generation don't extend JsonSerializable (azure-core-management types). So, I'll mark those as `--stream-style-serialization false` to prevent those libraries from failing build. Additional to that, `azure-xml` is still in beta so I may add a switch specific for that, or we can just say it's an all or nothing for now.

weidongxu-microsoft:
I've set `stream-style-serialization` to always `false` on mgmt flow.
- https://github.com/Azure/autorest.java/pull/2506/commits/921cbb206cde4abaefbcf588affea5bfe9a5a966
- https://github.com/Azure/autorest.java/pull/2506/commits/e4b9f381049d3c416686a76f7ee0fe849ad4f1dc

It is not tested there, so we don't want to switch it on there, at present.

srnagar (approved):

typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/access/implementation/models/AbstractModel.java
```diff
@@ -47,6 +75,7 @@ public String getName() {
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
```
weidongxu-microsoft, line 75: @alzimmermsft @haolingdong-msft @srnagar 

We need `@Generated` on this method.

Else, partial-update would be in a mess.

typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/access/implementation/models/AbstractModel.java
```diff
@@ -56,8 +85,7 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
      * @param jsonReader The JsonReader being read.
      * @return An instance of AbstractModel if the JsonReader was pointing to an instance of it, or null if it was
      * pointing to JSON null.
-     * @throws IllegalStateException If the deserialized JSON object was missing any required properties or the
-     * polymorphic discriminator.
+     * @throws IllegalStateException If the deserialized JSON object was missing any required properties.
      * @throws IOException If an error occurs while reading the AbstractModel.
      */
     public static AbstractModel fromJson(JsonReader jsonReader) throws IOException {
```
weidongxu-microsoft, line 91: and this

typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/access/implementation/models/AbstractModel.java
```diff
@@ -88,17 +116,23 @@ public static AbstractModel fromJson(JsonReader jsonReader) throws IOException {
     static AbstractModel fromJsonKnownDiscriminator(JsonReader jsonReader) throws IOException {
```
weidongxu-microsoft, line 116: and this

typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/access/implementation/models/AbstractModel.java
```diff
@@ -30,9 +36,31 @@ public class AbstractModel implements JsonSerializable<AbstractModel> {
...
+        return this.kind;
+    }
+
+    /**
+     * Set the kind property: The kind property.
+     * 
+     * @param kind the kind value to set.
+     * @return the AbstractModel object itself.
+     */
+    AbstractModel setKind(String kind) {
```
weidongxu-microsoft, line 59: and this

PR #2509, opened 2024-01-15
tsp, support multiple files (List<BinaryData>) in multipart/form-data

weidongxu-microsoft:
basically for the `files: bytes[]` case in multipart.

code https://github.com/Azure/autorest.java/pull/2509/commits/37c0584927eba15b1fd7dbed5f36862f5f4b1014


javagen/src/main/resources/MultipartFormDataHelper.java
```diff
@@ -156,8 +176,30 @@ public MultipartFormDataHelper end() {
...
+    }
+
     private void appendBytes(byte[] bytes) {
         requestLength += bytes.length;
         requestDataStream = new SequenceInputStream(requestDataStream, new ByteArrayInputStream(bytes));
     }
+
+    private String normalizeAscii(String text) {
+        return Normalizer.normalize(text, Normalizer.Form.NFD).replaceAll("[^\\x00-\\x7F]", "");
+    }
```
weidongxu-microsoft, line 204: This is for better safety of the request. Generally, RFC ask the MIME be US-ASCII.

typespec-tests/src/main/java/com/payload/multipart/MultiPartAsyncClient.java
```diff
@@ -151,7 +154,10 @@ Mono<Response<Void>> jsonPartWithResponse(BinaryData body, RequestOptions reques
      * {
      *     id: String (Required)
      *     pictures (Required): [
-     *         byte[] (Required)
+     *         BinaryData (Required)
+     *     ]
+     *     pictures (Optional): [
```
haolingdong-msft, line 159: The two properties are all called `pictures`?

weidongxu-microsoft: The other one if "picturesFilenames" we appended during codegen.

This method will not get exposed, so user should not able to see this javadoc.

weidongxu-microsoft: But this reminds me of the "additionalProperties" case, that the Javadoc is exposed and be wrong...
https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/type/property/additionalproperties/IsFloatClient.java#L44-L51

I will had a follow up task to say e.g. this ModelProperty is not via JSON etc.

haolingdong-msft: ok, I see.

haolingdong-msft (approved):

typespec-tests/src/test/java/com/payload/multipart/MultipartTests.java
```diff
@@ -36,19 +43,60 @@ public void testBasic() {
...
+                Arrays.asList(BinaryData.fromFile(PNG_FILE), BinaryData.fromFile(PNG_FILE))));
+
+        // provide 3 filenames, when there are 2 files
+        asyncClient.complex(new ComplexPartsRequest(
+                "123",
+                new Address("X"),
+                BinaryData.fromFile(FILE),
+                Arrays.asList(new Address("Y"), new Address("Z")),
+                Arrays.asList(BinaryData.fromFile(PNG_FILE), BinaryData.fromFile(PNG_FILE)))
+                .setPicturesFilenames(Arrays.asList("picture1", "picture2", "picture3"))).block();
```
haolingdong-msft, line 100: cadl-ranch seems not assert for file name? do we need to verify the serialized result?

weidongxu-microsoft: Yeah. Currently I've verified the file name by human (i.e. me) reading the http log :-)

Verify automatically would require we do a more advanced verifier than https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/test/java/com/parameters/bodyoptionality/BodyTests.java#L36-L44
Let me see if I want to do this...

typespec-tests/src/main/java/com/payload/multipart/MultiPartAsyncClient.java
weidongxu-microsoft: Another "fix" for avoiding the confusing multipart schema is to just "do not generate Javadoc if the method is not public or protected". It could save some file space as we have this kind of very long Javadoc. But that likely be another round of discussion.

haolingdong-msft: How about first remove the javadoc for the package-private method for multipart only? Because the schema for multipart is meaningless

typespec-tests/src/test/java/com/payload/multipart/MultipartTests.java
weidongxu-microsoft: validation added

typespec-tests/src/main/java/com/payload/multipart/MultiPartAsyncClient.java
weidongxu-microsoft: OK, I've skip the body request schema, if it is multipart.

Issue #2503, opened 2024-01-10
mpg, api-version not present in service client(operation's api-version is empty in code-model)


XiaofeiCao:

weidongxu-microsoft:
This still exist?

It means the generated SDK cannot actually work, right?

XiaofeiCao:
Yeah, it still exists. I'm looking into the issue.

weidongxu-microsoft:
I guess it probably due to the difference to api-version parameter. In Swagger it is a constant with the version of that Swagger.

But in TypeSpec there be multiple versions, so we didn't do this to the api-version parameter.

XiaofeiCao:
Current observation is, we do have a global api-version parameter with constant schema. Though each operation has its own api-version schema definition(which doesn't reference to the global parameter) and the schema is empty.

We may need to update the operation's api-version parameter's schema to refer to the global one like swagger does.

weidongxu-microsoft:
We can refer to api-version in client (which can be more than one). But not api-version in code-model (which is only one).

TypeSpec can have different api-version on different namespace, which maps to client.

weidongxu-microsoft:
And autorest also support mixing Swagger of different api-version (this is not same as the case above). I just mean even autorest does not always have a global api-version.

XiaofeiCao:
For single api-version, m4 will put `api-version` in `globalParameters`. 
![image](https://github.com/Azure/autorest.java/assets/92354331/a433175f-1cb5-48b8-abcc-f2039da17330)
![image](https://github.com/Azure/autorest.java/assets/92354331/7a270e93-2f44-40f1-9584-8b990d55a907)

For mixing api-versions, m4 will put `api-version`'s constants into `constants`, and inline api-version parameter definition in `parameters` of `operation`.
![image](https://github.com/Azure/autorest.java/assets/92354331/31ed729a-82e8-41e4-9999-a4c684f435c1)
![image](https://github.com/Azure/autorest.java/assets/92354331/0525bdd6-4724-470b-8e05-e633a0095848)

Will align with this behavior.

weidongxu-microsoft:
Maybe we first handle the former case.

weidongxu-microsoft:
Is it now possible to send and receive a e.g. List from service (using tsp generated sdk)?

XiaofeiCao:
> Is it now possible to send and receive a e.g. List from service (using tsp generated sdk)?

Yes, live test can pass for list:

```log
[...]

PR #2511, opened 2024-01-16
mpg, fix api-version

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2503

Currently we don't know how to define mixed api-versions in tsp, so I assume there's only one client and one api-version. 
We don't support multiple client for mpg anyways.

code-model: 
image

generated code:
image
image


weidongxu-microsoft (approved):
add a TODO in your comment

Issue #2510, opened 2024-01-15
DPG, partial-update, merge javadoc with user content


weidongxu-microsoft:
ask https://github.com/Azure/azure-sdk-for-java/pull/38307

> Let me see if I can modify the codegen to let it "merge" the Javadoc via codegen and the "<!-- src_embed ... <!-- end" section via dev/codesnippet-maven-plugin. <-- it may not be easy; and even after we merge it, there could be some format difference after the formatter that still trigger the diff

weidongxu-microsoft:
Another is where do we put the `start` and `end`.

E.g. for a full Javadoc, should we do this
```
    /**
     * <!-- start generated doc -->
     * The read operation.
     * <!-- end generated doc -->
     * 
     * @param filter A sequence of textual characters.
     * @throws IllegalArgumentException thrown if parameters fail the validation.
     * @throws HttpResponseException thrown if the request is rejected by server.
     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
     * @return the response.
     */
```

or this
```
    /**
     * <!-- start generated doc -->
     * The read operation.
     * 
     * @param filter A sequence of textual characters.
     * @throws IllegalArgumentException thrown if parameters fail the validation.
     * @throws HttpResponseException thrown if the request is rejected by server.
     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
     * @return the response.
     * <!-- end generated doc -->
     */
```

haolingdong-msft:
Option 1:
Add tags for generated docs. We will have a single block for generated comments. We will overwrite the javadoc within <start> and <end> wrapper. If the user wants to add in the middle of <start> and <end> wrapper, they will need to maintain the javadoc by themselves by removing the <start> and <end> tags. 
How to do backfill? Have script to go through the repo to update the method's javadoc.
Option 2：
Add customized tags. <customized start><customized end>. Only allow customization at the top of javadoc description.

[...]

Issue #2502, opened 2024-01-10
partial-update, support partial update class/interface's Javadoc


haolingdong-msft:

weidongxu-microsoft:
LGTM

If service does not want any generated doc, they would do this?
```java
/**
 * Some other custom Javadocs which shouldn't be overwritten as they aren't contained in the generated Javadoc block.
 * <p>
 * <!-- start generated doc -->
 * <!-- end generated doc -->
 */
```

haolingdong-msft:
`<!-- start generated doc -->` and `<!-- end generated doc -->` is added by codegen. If there is no generated javadoc, we will not add it.

weidongxu-microsoft:
We can release a version with start/end but without update in partial-update logic, to refresh existing libs. Then another version with partial-update logic.

haolingdong-msft:
Yes, I plan to do so as well.

> We can release a version with start/end but without update in partial-update logic, to refresh existing libs. Then another version with partial-update logic.

srnagar:
Do we plan to support multiple start and end tags?

For example, can this be supported?

```java
/**
 * Some other custom Javadocs which shouldn't be overwritten as they aren't contained in the generated Javadoc block.
 * <p>
 * <!-- start generated doc -->
 * <!-- end generated doc -->
 * 
 *  Some more custom docs added in the middle.
 * 
  * <!-- start generated doc -->
 * <!-- end generated doc -->
 */
```

srnagar:
> We will need to support partial update for class level javadoc.

We should also support partial update for all javadocs. We will have codesnippets added to javadocs of client methods too. So, we should support this wherever we have javadocs.

haolingdong-msft:
currently if service team remove the `@Generated` annotation on client method, and update javadoc, we will keep the javadoc they updated. But at class level, it does not have `@Generated` annotation, so we don't do the same thing.

weidongxu-microsoft:
> > We will need to support partial update for class level javadoc.
> 
> We should also support partial update for all javadocs. We will have codesnippets added to javadocs of client methods too. So, we should support this wherever we have javadocs.

client method issue here https://github.com/Azure/autorest.java/issues/2510

or you prefer we do them together?

PR #2512, opened 2024-01-16
partial-update,  add generated wrapper for class level javadoc

haolingdong-msft:
On hold this pr as we are having furthur discussions. https://github.com/Azure/autorest.java/issues/2510#issuecomment-1897697164

Link https://github.com/Azure/autorest.java/issues/2502

This PR adds start and end wrapper to class level generated javadoc. 
Currently for those classes have `@Generated` annotation, we will add the the start and end wrapper. The classes include:
1. DPG/TypeSpec: all models, clients and builders classes.
2. vanilla: builders

Logic mostly in this commit: https://github.com/Azure/autorest.java/pull/2512/commits/64a2277ca9743ae8ed5cedef05ef906fbe24ba37


typespec-tests/src/main/java/com/cadl/armresourceprovider/fluent/ArmResourceProviderClient.java
```diff
@@ -8,7 +8,9 @@
 import java.time.Duration;
 
 /**
+ * <!-- start generated doc -->
```
haolingdong-msft, line 11: mgmt plane impacted part

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaJavadocComment.java
```diff
@@ -37,6 +38,16 @@ public final void description(String description) {
         line(processedText);
     }
 
+    public final void description(String description, boolean withStartAndEndGeneratedComment) {
```
weidongxu-microsoft, line 41: Personally, prefer this to be in a util class. This not exactly belong to `javamodel`.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -95,7 +95,7 @@ public final void write(ClientModel model, JavaFile javaFile) {
         javaFile.declareImport(imports);
 
         javaFile.javadocComment(settings.getMaximumJavadocCommentWidth(),
-            comment -> comment.description(model.getDescription()));
+            comment -> comment.description(model.getDescription(), JavaSettings.getInstance().isDataPlaneClient()));
```
weidongxu-microsoft, line 98: Thinking of a protected `boolean supportPartialUpdate()` method in the class.

This way, it could control all related features about partial-update, including this and `@Generated`.
One reason is that `<!-- start generated doc -->` without `@Generated` is probably useless (I assume we still check the existance of `@Generated` annotation for whether the class supports partial-update during processing).

weidongxu-microsoft (commented):
If the change involves vanilla, do loop in Alan.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
haolingdong-msft: I think this is a good way to remove the generated javadoc wrapper if some classes do not want. I will update.

typespec-tests/src/main/java/com/cadl/armresourceprovider/fluent/ArmResourceProviderClient.java
haolingdong-msft: removed the change on mgmt

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaJavadocComment.java
haolingdong-msft: Yes, it does not have meaning in Java. But I did not find a util class very suitable for putting this code. Do you have recommendation. 
@srnagar would also like to hear your thoughts on this before doing the change.

haolingdong-msft:
Hi @alzimmermsft, this change will add start and end wrapper  for class level javadoc. Could you please also take a look as it will add the wrapper to vanilla builders and clients as well.

androidgen/src/main/java/com/azure/autorest/android/template/AndroidServiceAsyncClientTemplate.java
```diff
@@ -24,4 +24,9 @@ protected void addServiceClientAnnotationImports(java.util.Set<String> imports)
     @Override
     protected void addGeneratedAnnotation(JavaContext classBlock) {
     }
+
+    @Override
+    protected boolean isPartialUpdateSupported() {
```
alzimmermsft, line 29: Given Android generates targeting Java as well, do we want to just enable this? At the end of the day it's just a set of XML comments. Plus, given the limited usage of the Android generator it would simplify code in Javagen.

Comment applies to the other spots in Androidgen as well.

fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentModelTemplate.java
```diff
@@ -143,6 +143,11 @@ protected void addGeneratedImport(Set<String> imports) {
     protected void addGeneratedAnnotation(JavaContext classBlock) {
     }
 
+    @Override
+    protected boolean isPartialUpdateSupported() {
+        return false;
```
alzimmermsft, line 148: Similar question as in Androidgen, do we even want to disable this? It is a bit more overhead during generation but will simplify logic between generators.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -95,7 +95,7 @@ public final void write(ClientModel model, JavaFile javaFile) {
         javaFile.declareImport(imports);
 
         javaFile.javadocComment(settings.getMaximumJavadocCommentWidth(),
-            comment -> comment.description(model.getDescription()));
+            comment -> comment.description(model.getDescription(), isPartialUpdateSupported()));
```
alzimmermsft, line 98: As we keep adding more one-offs for partial update support, do we just want to enable it always and for everyone? Or at least in documentation, enable it for everyone. It's an extremely useful feature to allow extending documentation without complicated Swagger/TypeSpec transforms or code customizations.

javagen/src/main/java/com/azure/autorest/template/ServiceAsyncClientTemplate.java
```diff
@@ -144,6 +144,10 @@ protected void addGeneratedAnnotation(JavaContext classBlock) {
     classBlock.annotation(Annotation.GENERATED.getName());
   }
 
+  protected boolean isPartialUpdateSupported() {
+    return true;
```
alzimmermsft, line 148: Any reasons this just always returns true while other cases are based on the Autorest setting for enabling partial update? To be clear, I like this better where it's always true but was just wondering.

javagen/src/main/java/com/azure/autorest/template/ServiceClientInterfaceTemplate.java
```diff
@@ -37,7 +37,7 @@ public final void write(ServiceClient serviceClient, JavaFile javaFile) {
 
         javaFile.javadocComment(comment ->
         {
-            comment.description(String.format("The interface for %1$s class.", serviceClient.getInterfaceName()));
+            comment.description(String.format("The interface for %1$s class.", serviceClient.getInterfaceName()), true);
```
alzimmermsft, line 40: Another small inconsistency, this doesn't even have the ability to override the setting. And again, I'll say I like this always on with no ability to disable

androidgen/src/main/java/com/azure/autorest/android/template/AndroidServiceAsyncClientTemplate.java
weidongxu-microsoft: From what I remember, Android turns it off because their azure-core does not have `Generated` annotation.

Not sure if it's changed though.

fluentgen/src/main/java/com/azure/autorest/fluent/template/FluentModelTemplate.java
weidongxu-microsoft: The reason is that MGMT doing lots of interface as fluent code. That is not supported by our current partial-update logic.
E.g. https://github.com/Azure/autorest.java/blob/eba2861394f41f3b16f22ea18521960f35287f72/typespec-tests/src/main/java/com/cadl/armresourceprovider/models/TopLevelArmResource.java

So, at present, MGMT is a full overwrite (with all generated code deleted, then regen).

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaJavadocComment.java
weidongxu-microsoft: `TemplateUtil.writeJavadocDescriptionWithGenerated` seems good enough.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
weidongxu-microsoft: Enable for all javadoc could be fine, but we may need to check a few more situations.

E.g. for a full Javadoc, should we do this
```
    /**
     * <!-- start generated doc -->
     * The read operation.
     * <!-- end generated doc -->
     * 
     * @param filter A sequence of textual characters.
     * @return the response.
     */
```

or this
```
    /**
     * <!-- start generated doc -->
     * The read operation.
     * 
     * @param filter A sequence of textual characters.
     * @return the response.
     * <!-- end generated doc -->
     */
```

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaJavadocComment.java
srnagar: Yeah, I am fine with moving this to a util method.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
weidongxu-microsoft: Further discussion on Javadoc
https://github.com/Azure/autorest.java/issues/2510#issuecomment-1897697164

PR #2513, opened 2024-01-17
tsp, bump cadl-ranch

weidongxu-microsoft:
the global `service.version` is removed from `@scenarioService`

### DIFF
--- a/typespec-extension/package.json
+++ b/typespec-extension/package.json
@@ -51,8 +51,7 @@
     "@typespec/openapi": ">=0.51.0 <1.0.0",
     "@typespec/versioning": ">=0.51.0 <1.0.0",
     "@azure-tools/typespec-azure-core": ">=0.37.2 <1.0.0",
-    "@azure-tools/typespec-client-generator-core": ">=0.37.0 <1.0.0",
-    "@azure-tools/typespec-azure-resource-manager": ">=0.37.1 <1.0.0"
+    "@azure-tools/typespec-client-generator-core": ">=0.37.0 <1.0.0"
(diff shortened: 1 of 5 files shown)


typespec-extension/src/code-model-builder.ts
```diff
@@ -527,8 +527,7 @@ export class CodeModelBuilder {
           apiVersion.version = version.value;
           codeModelClient.apiVersions.push(apiVersion);
         }
-      } else if (this.isArm()) {
-        // todo: there's ongoing discussion of whether to apply it to DPG as well
+      } else {
```
weidongxu-microsoft, line 530: From current discussion, `@service.version` is a valid version.

XiaofeiCao (approved):

typespec-extension/package.json
```diff
@@ -51,8 +51,7 @@
     "@typespec/openapi": ">=0.51.0 <1.0.0",
     "@typespec/versioning": ">=0.51.0 <1.0.0",
     "@azure-tools/typespec-azure-core": ">=0.37.2 <1.0.0",
-    "@azure-tools/typespec-client-generator-core": ">=0.37.0 <1.0.0",
-    "@azure-tools/typespec-azure-resource-manager": ">=0.37.1 <1.0.0"
+    "@azure-tools/typespec-client-generator-core": ">=0.37.0 <1.0.0"
```
XiaofeiCao, line 54: Removed dependency on `typespec-azure-resource-manager`. It has dependency on `typespec-autorest` that we don't want.

PR #2525, opened 2024-01-23
tsp, multipart with ##FileDetails class

weidongxu-microsoft:
fix https://github.com/Azure/azure-sdk-for-java/issues/38408

code https://github.com/Azure/autorest.java/pull/2525/commits/aad9029478fe756c75cea7339315081bbba37e5d
maybe easier to see test about user code change https://github.com/Azure/autorest.java/pull/2525/commits/1d5a8ce587c3b2937f7be4fc7cdf268880932564

---

For OpenAI, I would need tspconfig.yaml to move `FileDetails` to impl
https://github.com/Azure/azure-sdk-for-java/pull/38460/files#diff-d40d4b7eb514d86d1ccdde99636b8882a33bc56362830e7c7fe6e82d072f4dc5

---

test with unreleased cadl-ranch https://github.com/weidongxu-microsoft/autorest.java/pull/11


javagen/src/main/java/com/azure/autorest/mapper/ModelMapper.java
```diff
@@ -586,45 +586,32 @@ private static String disambiguatePropertyNameOfFlattenedSchema(Set<String> prop
...
-                iterator.add(property.newBuilder()
-                        .name(property.getName() + ClientModelUtil.FILENAME_SUFFIX)
-                        .defaultValue(ClassType.STRING.defaultValueExpression(property.getSerializedName()))
-                        .description("The filename for " + property.getName())
-                        .wireType(ClassType.STRING)
-                        .clientType(ClassType.STRING)
-                        .required(false)
+                        .wireType(fileDetailsModelType)
+                        .clientType(fileDetailsModelType)
                         .build());
```
weidongxu-microsoft, line 607: Good news, got rid of the "filename" property. Bad news, get a new class...

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -132,13 +133,16 @@ public Client map(CodeModel codeModel) {
...
-        final List<ClientModel> clientModels = autoRestModelTypes
+        List<ClientModel> clientModelsFromCodeModel = autoRestModelTypes
             .distinct()
             .map(autoRestCompositeType -> Mappers.getModelMapper().map(autoRestCompositeType))
             .filter(Objects::nonNull)
             .distinct()
             .collect(Collectors.toList());
-
+        // append some models not from CodeModel
+        final List<ClientModel> clientModels = Stream.concat(clientModelsFromCodeModel.stream(), ClientModels.getInstance().getModels().stream())
```
haolingdong-msft, line 143: Confirm: This will add `MultipartFileDetails` model only currently?

weidongxu-microsoft: It could be multiple this kind of "FileDetails".

If the field is `image`, it be `ImageFileDetails`. e.g. https://github.com/Azure/autorest.java/tree/015cbeecb103ce84152d93cb4ce5827b772f26c7/typespec-tests/src/main/java/com/payload/multipart/models

Srikanta wants different name, to avoid different SDK conflict on this model name.

It should be same class, if moved to azure-core.

haolingdong-msft: I see, will this add other classes unrelated to multipart?

weidongxu-microsoft: As far as I remember, we didn't have "other classes", till this.

And test does not have new class.

weidongxu-microsoft: And I can delete this, if FileDetails class get to azure-core

haolingdong-msft: ok, I see.

haolingdong-msft (approved):

typespec-extension/src/code-model-builder.ts
```diff
@@ -1288,7 +1288,14 @@ export class CodeModelBuilder {
       this.trackSchemaUsage(schema, { serializationFormats: [KnownMediaType.Multipart] });
     }
 
-    if (!schema.language.default.name && schema instanceof ObjectSchema) {
+    if (schema instanceof ObjectSchema && schema.serializationFormats?.includes(KnownMediaType.Multipart)) {
+      // TODO: anonymous model for multipart is not supported
```
haolingdong-msft: I don't quite get why is this commit's code change related with 'avoid flatten'?

weidongxu-microsoft: I got a bug on this one. Correct code should be within the `!schema.language.default.name` check.
https://github.com/Azure/autorest.java/blob/05d287353282540d4c4d95b624f1217e0618289f/typespec-extension/src/code-model-builder.ts#L1291-L1306

weidongxu-microsoft: BTW, I probably need to update the anonymous model / flatten logic.

We were using the `Map<String, Object>` to handle that. But this won't work as expected in stream-serialization, as `Map` or `List` would trigger the Jackson flow in core.

javagen/src/main/java/com/azure/autorest/util/ClientModelUtil.java
```diff
@@ -714,4 +714,83 @@ public static Set<String> getExternalPackageNamesUsedInClient(List<ClientModel>
...
+     * Gets or creates a new FileDetails model for a multipart/form-data request
+     *
+     * @param compositeType the object schema of the multipart/form-data request model.
+     * @param filePropertyName the property name of the file in the multipart/form-data request model.
+     * @return the ##FileDetails model
+     */
+    public static IType getMultipartFileDetailsModel(
+            ObjectSchema compositeType,
+            String filePropertyName) {
+        // TODO (weidxu): this ##FileDetails model may get renamed and moved to azure-core
```
XiaofeiCao, line 728: Wouldn't it cause breaking change if moved to azure-core?

weidongxu-microsoft: Yes, it would.

The issue is that DocumentTranslation need it this week. And one won't have azure-core thing designed and merged that soon.

So, at present, a model in SDK. The decision about azure-core would be later, but should be before any of them get GA.

XiaofeiCao: I see.

PR #2526, opened 2024-01-23
mpg, remove lro workaround, use getLroMetadata

XiaofeiCao:
Seems getLroMetadata is working in typespec-azure-core 0.37.2.

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -171,7 +171,6 @@ import {
 } from "./operation-utils.js";
 import { isArmCommonType } from "./type-utils.js";
 import pkg from "lodash";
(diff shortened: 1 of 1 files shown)


weidongxu-microsoft (approved):

weidongxu-microsoft:
I assume we had an LRO in arm.tsp?

XiaofeiCao:
> I assume we had an LRO in arm.tsp?

Yeah: https://github.com/Azure/autorest.java/blob/main/typespec-tests/tsp/arm.tsp#L82

https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/cadl/armresourceprovider/implementation/TopLevelArmResourceInterfacesClientImpl.java#L351

PR #2527, opened 2024-01-25
tsp, upgrade compiler 0.52

weidongxu-microsoft:
also include the support for `@clientName`

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/naming/models/BytesData.java
+++ b/typespec-tests/src/main/java/com/cadl/naming/models/BytesData.java
@@ -18,36 +18,36 @@ import java.io.IOException;
 @Immutable
 public final class BytesData extends Data {
     /*
-     * The data property.
+     * The data_bytes property.
      */
     @Generated
-    private final byte[] data;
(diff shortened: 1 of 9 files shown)


typespec-tests/tsp/naming.tsp
```diff
@@ -52,6 +54,10 @@ model Data {
 
 model BytesData extends Data {
   kind: "bytes";
+
+  @clientName("dataAsBytes")
```
weidongxu-microsoft, line 58: test

typespec-tests/src/main/java/com/cadl/naming/models/BytesData.java
```diff
@@ -18,36 +18,36 @@
 @Immutable
 public final class BytesData extends Data {
     /*
-     * The data property.
+     * The data_bytes property.
      */
     @Generated
-    private final byte[] data;
+    private final byte[] dataAsBytes;
```
weidongxu-microsoft, line 24: result

haolingdong-msft (approved):

PR #2529, opened 2024-01-25
mgmt, support flattenProperty

weidongxu-microsoft:
However, saw a critical problem when trying to write a test. The `properties` property is in lib, not in service tsp...
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/models.tsp#L23

Locally, I've tested on other property. Not checked in. We should got a cadl-ranch case soon.

### DIFF
--- a/typespec-extension/changelog.md
+++ b/typespec-extension/changelog.md
@@ -5,6 +5,7 @@
 Compatible with compiler 0.52.
 
 - Supported `@clientName` from "@azure-tools/typespec-client-generator-core".
+- Supported `@flattenProperty` from "@azure-tools/typespec-client-generator-core".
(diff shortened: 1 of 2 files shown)


XiaofeiCao (approved):

weidongxu-microsoft:
I can write this

```
@@flattenProperty(TopLevelArmResource.properties);
@@flattenProperty(ChildResource.properties);
@@doc(ChildResource.name, "aaaaa");
@@doc(ChildResource.properties, "bbbbb");
```

But seems the one on `ChildResource.properties` have no effect. Not sure if its my parsing of ModelProperty.

Issue #2484, opened 2024-01-02
mpg, support `endpoint`/`host` parameter in service client


XiaofeiCao:
Currently there's no `hostParameters` generated from `getServers` in `typespec/http`. 
<img width="531" alt="Screenshot 2024-01-25 at 14 24 03" src="https://github.com/Azure/autorest.java/assets/92354331/a4f1fb01-ba40-4847-b622-a2ab4125c82d">

It's added by `@armProviderNamespace`:
https://github.com/Azure/typespec-azure/blob/8b8d7c05f168d9305a09691c4fedcb88f4a57652/packages/typespec-azure-resource-manager/src/namespace.ts#L121-L128

We took the `url` as the `baseURL` for every operation.

XiaofeiCao:
We could do:
1. In typespec-java, add one synthesized hostParameter if isArm() and noHostParameter()
2. Use the default namespace, though user will lack the ability to set to different Azure environment
3. Update `typespec-azure-resource-manager`, though not sure if it will affect other languages and how

Probably will go with 1 first.

XiaofeiCao:
fixed by https://github.com/Azure/autorest.java/pull/2532

PR #2532, opened 2024-01-25
mpg, add host parameter if `server` is synthesized by `@armProviderNamespace`

XiaofeiCao:
- issue: https://github.com/Azure/autorest.java/issues/2484

Not sure if we want to modify the synthesized server in typespec-azure-resource-manager. Add one in our emitter first.

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -173,7 +173,6 @@ import {
 } from "./operation-utils.js";
 import { isArmCommonType } from "./type-utils.js";
 import pkg from "lodash";
-import { Server } from "http";
 const { isEqual } = pkg;
 
 export class CodeModelBuilder {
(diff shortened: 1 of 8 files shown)


typespec-extension/src/code-model-builder.ts
```diff
@@ -542,7 +542,7 @@ export class CodeModelBuilder {
       // server
       let baseUri = "{endpoint}";
       const servers = getServers(this.program, client.service);
-      if (servers && servers.length === 1) {
+      if (servers && servers.length === 1 && !this.isArmSynthesizedServer(servers[0])) {
         baseUri = servers[0].url;
       }
```
weidongxu-microsoft, line 547: This part was a bit strange (why code conditioned on `servers.length`). Thought? (I know I wrote this...)

weidongxu-microsoft: OK, probably it would only has 1 server anyway

weidongxu-microsoft (approved):

typespec-extension/src/code-model-builder.ts
XiaofeiCao: Yeah, we seem to deal with one server only: https://github.com/Azure/autorest.java/blob/95268e17df0d1728e1db71fed63e9a248d2c0df3/typespec-extension/src/code-model-builder.ts#L548

PR #2533, opened 2024-01-25
Ensure polymorphic models close reader and support Java 21 Eclipse language server

alzimmermsft:
- Fix to JSON stream-style serialization to ensure the buffered `JsonReader` used to handle polymorphic types is closed.
- Support for Java 21-based Eclipse language server and more logging in that space for debugability.


customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageServerFacade.java
```diff
@@ -93,22 +99,31 @@ private static Path getLanguageServerDirectory() throws IOException {
...
+            downloadUrl = URI.create(DOWNLOAD_BASE_URL + "1.31.0/jdt-language-server-1.31.0-202401111522.tar.gz")
+                .toURL();
+            languageServerPath = autorestLanguageServer.resolve("1.31.0");
         }
 
         Path languageServer = languageServerPath.resolve("jdt-language-server");
         if (!Files.exists(languageServerPath) || !Files.exists(languageServer)) {
             Files.createDirectories(languageServerPath);
             Path zipPath = languageServerPath.resolve("jdt-language-server.tar.gz");
+            logger.info("Downloading Eclipse JDT language server from {} to {}", downloadUrl, zipPath);
```
srnagar, line 122: It's also worth adding a log when we skip the download because we found a downloaded local copy already. Log the path to the local copy in case we want to just delete and start from scratch.

srnagar (approved):

weidongxu-microsoft (approved):

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageServerFacade.java
alzimmermsft: this is already being done on line 78

`logger.info("Starting Eclipse JDT language server at {}", languageServerPath);`

PR #2534, opened 2024-01-26
Move Fluentgen to same pattern used in Javagen and TypeSpec

alzimmermsft:
Updates `fluent-tests` and `fluent-integration` CI to match the patterns used in Javagen and TypeSpec where the following pattern is used:

- Build JARs to run code generation
- Run code generation
- Use generated code for testing

This PR removes the old bat script being used and replaces it with a PowerShell script to enable Linux and macOs users to run the script more easily. The PowerShell script is also running code generation in parallel.

Overall, within CI this reduces the time for `fluent-integration` from about 18-22 minutes down to about 7-10 minutes.


eng/pipelines/fluent_integration.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
     inputs:
       mavenPomFile: pom.xml
       goals: 'clean install'
-      options: '$(DefaultOptions) -P local,testFluent -T 1C -DskipTests'
+      options: '$(DefaultOptions) -P local -T 1C -DskipTests'
```
weidongxu-microsoft, line 47: Why do we remove this? I think ci.yaml for javagen still have `testVanilla` and `testAzure`

fluent-tests/Initialize-Tests.ps1
weidongxu-microsoft, line 1: fluent seems still uses the old bat script.
https://github.com/Azure/autorest.java/blob/main/fluent-tests/pom.xml#L38

If the ps1 script works for you, you can merge it and I would try to make the switch.

weidongxu-microsoft (approved):

eng/pipelines/fluent_integration.yaml
alzimmermsft: This was moved further down the call chain. Fluent integration is now running like Javagen and TypeSpec where it runs the following pattern in CI

- Build JARs to generate code
- Generate code
- Run tests with generated code

fluent-tests/Initialize-Tests.ps1
alzimmermsft: Updated the PowerShell script with all cases in the bat script. Deleted the bat script as we shouldn't use those anymore given their OS limitations.

PR #2537, opened 2024-01-30
tsp, support encodedName for JSON

weidongxu-microsoft:


typespec-extension/src/code-model-builder.ts
```diff
@@ -2520,12 +2522,14 @@ export class CodeModelBuilder {
...
-    // First get projected name, if not found, return target.name
+    // First get projected name, if not found, call resolveEncodedName
+    // TODO: deprecate getProjectedName
     const jsonProjectedName = getProjectedName(this.program, target, "json");
     if (jsonProjectedName) {
       return jsonProjectedName;
     }
-    return target.name;
+    // TODO: the MIME better be from SchemaUsage.serializationFormats. However, at this step, the values is not propagated to inner models.
+    return resolveEncodedName(this.program, target, "application/json");
```
weidongxu-microsoft, line 2532: I had to put it after `projectedName`, as this `resolveEncodedName` won't tell us whether it resolved it from `@encodedName`, or fallback to target.name.

XiaofeiCao (approved):

haolingdong-msft:
One question: Currently `encodedName`'s key can only be `application-json`? I don't know if it has other key name, and what is the meaning.

weidongxu-microsoft:
From example they also have xml. But we don't have that yet.
https://typespec.io/docs/standard-library/built-in-decorators#@encodedName

json-merge-patch would get to this same application/json.

haolingdong-msft (approved):

PR #2541, opened 2024-01-31
mgmt, fix fluent operations with header response

XiaofeiCao:
This PR did:
1. Set `generic-response-type` to `false` for mpg, aligning with mgmt plane.
2. Fixed the scenario where `generic-response-type` is `true`.

Issue is, when service is using their custom ARM template(custom update), current mpg is having compilation issue. 
```typespec
// Custom update operations from sphere, similar issue with containerstorage
// This is not current recommended way of defining LRO update, but it's a real world example.
@armResourceOperations
interface CustomUpdateOperations {
  @autoRoute
  @doc("Update a {name}", TResource)
  @extension("x-ms-long-running-operation", true)
  @extension(
    "x-ms-long-running-operation-options",
    {
      `final-state-via`: "location",
    }
  )
  @armResourceUpdate(TResource)
  @patch
  ArmCustomPatchAsyncWithLocation<
    TResource extends ArmResource,
    TPatchModel extends TagsUpdateModel<TResource>,
    TBaseParameters = BaseParameters<TResource>
  >(
    ...ResourceInstanceParameters<TResource, TBaseParameters>,

    @doc("The resource properties to be updated.")
    @body
    properties: TPatchModel,
  ): ArmResponse<TResource> | ArmAcceptedResponse<"Resource update request accepted."> | ErrorResponse;
}
```

Generated client method(since the LRO update template is not standard, codegen did not recognize it as LRO, but that's another issue...):
```java
ResponseBase<DevicesUpdateHeaders, DeviceInner> updateWithResponse(String resourceGroupName, String catalogName,
        String productName, String deviceGroupName, String deviceName, DeviceUpdate properties, Context context);
``` 

Generated resource `apply` method implementation has compilation issue:
```java
public Device apply() {
    this.innerObject = serviceManager.serviceClient().getDevices().updateWithResponse(resourceGroupName,
        catalogName, productName, deviceGroupName, deviceName, updateProperties, Context.NONE);
    return this;
}
```

Instead, there should be a `.getValue()`:
```java
public Device apply() {
    this.innerObject = serviceManager.serviceClient().getDevices().updateWithResponse(resourceGroupName,
        catalogName, productName, deviceGroupName, deviceName, updateProperties, Context.NONE).getValue();
    return this;
}
```

The cause is, current mpg from TypeSpec sets `generic-response-type` to `true`(same with dpg). FluentGen logic did not handle this since it seems to have assumed the setting false. Setting `generic-response-type` to `false`(which should be done for mpg) will resolve this. 
However, since this setting is not explicitly tied to mgmt plane, we probably want to fix it.


typespec-tests/src/main/java/com/cadl/armresourceprovider/implementation/CustomTemplateResourceImpl.java
```diff
@@ -0,0 +1,147 @@
...
+            .updateWithResponse(resourceGroupName, customTemplateResourceName, this.innerModel(), Context.NONE)
+            .getValue();
+        return this;
+    }
+
+    public CustomTemplateResource apply(Context context) {
+        this.innerObject = serviceManager.serviceClient().getCustomTemplateResourceInterfaces()
+            .updateWithResponse(resourceGroupName, customTemplateResourceName, this.innerModel(), context).getValue();
+        return this;
+    }
```
XiaofeiCao, line 117: Fixed generated code with `generic-response-type` set to `true`.

typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java
```diff
@@ -99,7 +99,7 @@ protected FluentMapper getFluentMapper() {
 
         SETTINGS_MAP.put("license-header", "MICROSOFT_MIT_SMALL_TYPESPEC");
 
-        SETTINGS_MAP.put("generic-response-type", true);
+        SETTINGS_MAP.put("generic-response-type", false);
```
weidongxu-microsoft, line 102: Haven't read the details yet, but yeah I think mgmt from swagger have it false.

XiaofeiCao: Yeah. This PR did:
1. Set generic-response-type to false for mpg, aligning with mgmt plane.
2. Fixed the scenario where generic-response-type is true.

fluentgen/src/main/java/com/azure/autorest/fluent/util/FluentUtils.java
```diff
@@ -345,6 +353,18 @@ public static IType getValueTypeFromResponseType(IType clientType) {
...
+    private static IType getValueTypeFromResponseTypeSubType(GenericType type) {
+        IType bodyType = null;
+        if (ResponseBase.class.getSimpleName().equals(type.getName())) {
+            bodyType = type.getTypeArguments()[1];
+        } else if (SimpleResponse.class.getSimpleName().equals(type.getName())) {
+            bodyType = type.getTypeArguments()[0];
+        } else if (StreamResponse.class.getSimpleName().equals(type.getName())) {
+            bodyType = GenericType.FLUX_BYTE_BUFFER;
+        }
+        return bodyType;
```
weidongxu-microsoft, line 365: Wondering whether fallback to `type.getTypeArguments()[1]`?

weidongxu-microsoft (approved):

fluentgen/src/main/java/com/azure/autorest/fluent/util/FluentUtils.java
XiaofeiCao: Sure. I'm not sure if there's case where there's no type arguments(like StreamResponse). Better to throw exception when such case happens anyway.

PR #2543, opened 2024-02-01
tsp, generate ServiceVersion regardless API uses it or not

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2539

If `@versioned` (or `@service.version` -- this may change) defined on client/namespace, generate ServiceVersion on that client.

code change https://github.com/Azure/autorest.java/pull/2543/commits/07e74f54964f161d77f4fb8123658d48f1c225ec

---

last 3 commit is cherry-picked from multipart filename PR


typespec-tests/src/main/java/com/cadl/multipleapiversion/NoApiVersionClient.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public void action(String param1) {
+        // Generated convenience method for actionWithResponse
+        RequestOptions requestOptions = new RequestOptions();
+        if (!Arrays.asList("2022-12-01-preview").contains(serviceClient.getServiceVersion().getVersion())) {
+            throw LOGGER.logExceptionAsError(
+                new IllegalArgumentException("Parameter param1 is only available in api-version 2022-12-01-preview."));
+        }
```
weidongxu-microsoft, line 93: We actually still need to use the ServiceVersion when doing versioning resiliency.

typespec-tests/src/main/java/com/client/structure/service/ServiceVersion.java
weidongxu-microsoft, line 1: This because cadl-ranch service.tsp has `@service.version` in it. https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/client/structure/common/service.tsp#L32

protocol-tests/src/main/java/fixtures/bodyfile/AutoRestSwaggerBatFileServiceClientBuilder.java
```diff
@@ -193,6 +193,25 @@ public AutoRestSwaggerBatFileServiceClientBuilder host(String host) {
...
+     * @param serviceVersion the serviceVersion value.
+     * @return the AutoRestSwaggerBatFileServiceClientBuilder.
+     */
+    @Generated
+    public AutoRestSwaggerBatFileServiceClientBuilder
+        serviceVersion(AutoRestSwaggerBatFileServiceVersion serviceVersion) {
+        this.serviceVersion = serviceVersion;
+        return this;
+    }
+
```
weidongxu-microsoft, line 214: This get to DPG as well.

typespec-tests/src/main/java/com/client/structure/service/implementation/BazFoosImpl.java
```diff
@@ -48,6 +49,15 @@ public final class BazFoosImpl {
         this.client = client;
     }
 
+    /**
+     * Gets Service version.
+     * 
+     * @return the serviceVersion value.
+     */
+    public ServiceVersion getServiceVersion() {
```
haolingdong-msft, line 57: where is this method `getServiceVersion()` used?

weidongxu-microsoft: It is not used. (that was the reason why ServiceVersion is not generated)

typespec-tests/src/main/java/com/cadl/versioning/VersioningServiceVersion.java
```diff
@@ -4,12 +4,10 @@
 
 package com.cadl.versioning;
 
-import com.azure.core.util.ServiceVersion;
-
 /**
  * Service version of VersioningClient.
  */
-public enum VersioningServiceVersion implements ServiceVersion {
+public enum VersioningServiceVersion implements com.azure.core.util.ServiceVersion {
```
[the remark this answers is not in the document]
weidongxu-microsoft: Because one of the test case result in this class named `ServiceVersion`.

I saw no reason to block this... So use the full namespace for the `com.azure.core.util.ServiceVersion`

XiaofeiCao (approved):
Wonder the context. Why do we want to generate api-version class if it's not used by any APIs?

weidongxu-microsoft:
> Reference in

One potential case is version resiliency https://github.com/Azure/autorest.java/pull/2543#discussion_r1473775652

javagen/src/main/resources/MultipartFormDataHelper.java
```diff
@@ -215,7 +216,9 @@ private void appendBytes(byte[] bytes) {
         requestDataStream = new SequenceInputStream(requestDataStream, new ByteArrayInputStream(bytes));
     }
 
+    private static final Pattern REDACT_FILENAME = Pattern.compile("[^\\x20-\\x7E]|\"");
```
srnagar, line 219: Can you add a comment why this is needed? Why is the filename redacted?

typespec-tests/src/main/java/com/_specs_/azure/core/lro/rpc/legacy/LegacyServiceVersion.java
```diff
@@ -4,12 +4,10 @@
 
 package com._specs_.azure.core.lro.rpc.legacy;
 
-import com.azure.core.util.ServiceVersion;
-
 /**
  * Service version of LegacyClient.
  */
-public enum LegacyServiceVersion implements ServiceVersion {
+public enum LegacyServiceVersion implements com.azure.core.util.ServiceVersion {
```
srnagar, line 10: Why do we now have fully qualified package name instead of importing it?

javagen/src/main/java/com/azure/autorest/template/ServiceVersionTemplate.java
```diff
@@ -22,19 +20,14 @@ public static ServiceVersionTemplate getInstance() {
...
-
         javaFile.javadocComment(comment -> {
             comment.description("Service version of " + serviceVersion.getServiceName());
         });
 
         String className = serviceVersion.getClassName();
         List<String> serviceVersions = serviceVersion.getServiceVersions();
 
-        javaFile.publicEnum(className + " implements ServiceVersion", classBlock -> {
+        javaFile.publicEnum(className + " implements com.azure.core.util.ServiceVersion", classBlock -> {
```
srnagar, line 30: We should still import the ServiceVersion enum and not have the fully qualified name here in `implements` section.

typespec-tests/src/main/java/com/_specs_/azure/core/lro/rpc/legacy/LegacyServiceVersion.java
weidongxu-microsoft: Will revert.

Reason is one test case got name=`Service` and that result to `ServiceVersion`.

javagen/src/main/resources/MultipartFormDataHelper.java
weidongxu-microsoft: The header allows ASCII, while the filename may include much more.
Client is allowed to change the filename (somehow).

[rfc8187](https://www.ietf.org/rfc/rfc8187.html#section-3.2.3) & [rfc6266](https://www.ietf.org/rfc/rfc6266.html#section-4.3) recommend "filename" be ASCII, and a "filename*" with encoded UTF8.

At present, we hadn't get to the "filename*" yet.

PR #2545, opened 2024-02-02
tsp. use internal model for anonymous model

weidongxu-microsoft:
Improve model serialization on flattened (anonymous) model.

Generic types like Map/List is not well handled by stream-serialization (to a lesser extent, Jackson as well).

code https://github.com/Azure/autorest.java/pull/2545/commits/204dbc2e113d272e43a381154c8e1c3a4f5ab1cd

---

Support flatten for multipart: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/OpenAI.Assistants/files/routes.tsp#L45

code https://github.com/Azure/autorest.java/pull/2545/commits/816691aa95794cf08b5f439447f439a4b19aca46 (there is some hack)

test https://github.com/Azure/azure-sdk-for-java/pull/38595


XiaofeiCao (approved):
unblock

PR #2547, opened 2024-02-04
tsp, bug fix on 1. ServiceVersion, 2. package for FileDetails in anonymous model

weidongxu-microsoft:
Main code change https://github.com/Azure/autorest.java/pull/2547/commits/8b664c6e4a564c6dcb0c1cee9739b28223e6daf2

Considering a future PR to move FileDetails model to code-model (currently it only in client-model, as a different interpretation of `bytes` in multipart serialization -- from conceptual level, client-model only seems make sense, as this be more "interpretation based on language and implementation"; e.g. another implementation may use `java.nio.file.Path` instead of `BinaryData`+`filename`).
1. it would be better aligned with future TCGC model
2. anonymous model and flatten should be easier to implement in Java
3. propagation on `usages` would be better handled in code-model


typespec-tests/src/main/java/com/client/structure/service/implementation/RenamedOperationClientImpl.java
```diff
@@ -131,7 +131,7 @@ public GroupsImpl getGroups() {
      * @param client Need to be set as 'default', 'multi-client', 'renamed-operation', 'two-operation-group' in client.
      * @param serviceVersion Service version.
      */
-    public RenamedOperationClientImpl(String endpoint, String client, ServiceVersion serviceVersion) {
+    public RenamedOperationClientImpl(String endpoint, String client, ServiceServiceVersion serviceVersion) {
```
haolingdong-msft, line 134: why changed to `ServiceServiceVersion`?

weidongxu-microsoft: Name conflict with ServiceVersion.

And Srikanta ask me not to use namespace...

haolingdong-msft (approved):

PR #2553, opened 2024-02-07
eng, integrate GitHub Release to devop pipeline

weidongxu-microsoft:
For now, just put `- Bug fix` into release note.


haolingdong-msft (approved):

haolingdong-msft:
> For now, just put `- Bug fix` into release note.

I don't quite get what does this mean? Does it mean we just put a dummy release note when trigger pipeline and later update the release notes in Github manually?

weidongxu-microsoft:
It means the pipeline will create the GitHub release (which at present is created by the github action). So after this PR, one does not need to run the github action (and they are deleted).

Release note may get improved later. At present, if nothing major, `- Bug fix` is good enough. Not sure any dev (other than we) actually read those...

weidongxu-microsoft:
Or put as this, now step 2 can be removed (after I verified they works as expected)
image

haolingdong-msft:
Thanks. I see. But if there is major features, we may still need to manually update the release note in github releases.

weidongxu-microsoft:
Yes, if something major, we could just edit that release.

In future, if we got time, we can improve it to similar of SDK repo, that e.g. extract the latest item in [changelog](https://github.com/Azure/autorest.java/blob/main/typespec-extension/changelog.md) and put it to release note.
Just not feeling the work have much benefit right now.

PR #2554, opened 2024-02-09
Remove Google Java formatter

alzimmermsft:
Removes the dependency on Google Java formatter due to the need to use `add-exports` in Java 17+. It is replaced with a simple process using JavaParser to clean up import usage.


postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -96,8 +106,83 @@ private static Map<String, String> loadEclipseSettings() throws Exception {
...
+                return true;
+            }
+
+            String importType = importDeclaration.getName().getIdentifier();
+            return !types.contains(importType);
+        }).collect(Collectors.toMap(importDeclaration -> importDeclaration.getRange().get().begin.line,
+            importDeclaration -> importDeclaration));
+
+        // Get the list of duplicate imports.
+        imports.stream().collect(Collectors.groupingBy(ImportDeclaration::getNameAsString)).entrySet().stream()
```
srnagar, line 158: nit: it would be nice to format all functional method calls in separate lines:

```java
imports.stream()
       .collect(..)
       .entrySet()
       .stream()
       ...
```

PR #2556, opened 2024-02-12
Investigate type spec nullable long date time

alzimmermsft:
Investigation into handling of `type: utcDateTime | null` vs `type?: utcDateTime`

### DIFF
--- a/typespec-tests/src/main/java/com/cadl/optional/OptionalAsyncClient.java
+++ b/typespec-tests/src/main/java/com/cadl/optional/OptionalAsyncClient.java
@@ -112,6 +112,8 @@ public final class OptionalAsyncClient {
      *     double: Double (Optional)
      *     duration: Duration (Optional)
      *     dateTime: OffsetDateTime (Optional)
+     *     epochDateTimeRequiredNullable: OffsetDateTime (Required)
+     *     epochDateTimeNullable: Long (Optional)
      *     stringList (Optional): [
      *         String (Optional)
      *     ]
@@ -140,6 +142,8 @@ public final class OptionalAsyncClient {
      *     double: Double (Optional)
      *     duration: Duration (Optional)
      *     dateTime: OffsetDateTime (Optional)
+     *     epochDateTimeRequiredNullable: OffsetDateTime (Optional)
+     *     epochDateTimeNullable: Long (Optional)
      *     stringList (Optional): [
      *         String (Optional)
      *     ]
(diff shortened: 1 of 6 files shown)


weidongxu-microsoft:
From my memory current emitter would output code-model as optional property, on `T | null` type.

Not sure if it adds `x-ms-nullable` though. Currently `T | null` would affect Dict/Map as `nullableItems = true/false`. This affect serialization/de-serialization when using Jackson (not sure about stream-serialization).

weidongxu-microsoft:
OK. I guess the issue is from here
https://github.com/Azure/azure-rest-api-specs/blob/ecc0170a2005f5f38231ae4dbba40594d3c00a04/specification/ai/OpenAI.Assistants/runs/models.tsp#L58-L94

I kind of assume codegen should map them to `ClassType.UNIX_TIME_LONG`, but seems not so...
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/openai/azure-ai-openai-assistants/src/main/java/com/azure/ai/openai/assistants/models/ThreadRun.java#L99-L139

weidongxu-microsoft:
> OK. I guess the issue is from here https://github.com/Azure/azure-rest-api-specs/blob/ecc0170a2005f5f38231ae4dbba40594d3c00a04/specification/ai/OpenAI.Assistants/runs/models.tsp#L58-L94
> 
> I kind of assume codegen should map them to `ClassType.UNIX_TIME_LONG` (hence a `Long` in class variable), but seems not so... https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/openai/azure-ai-openai-assistants/src/main/java/com/azure/ai/openai/assistants/models/ThreadRun.java#L99-L139

OK. TypeScript part actually separate the handling of "optional" and "nullable"
https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L2389-L2390

Think why we did this...

However, the output of typespec-autorest seems also not correct
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/data-plane/OpenAI.Assistants/OpenApiV2/preview/2024-02-15-preview/assistants_generated.json#L3411-L3423
`expires_at` property is "x-nullable" and "required", which is expected. But it should also be `type=integer, format=unixtime` as the `created_at` property.

weidongxu-microsoft:
I think it is due to incorrect typespec
https://github.com/Azure/autorest.java/pull/2643

weidongxu-microsoft:
I should solved it in https://github.com/Azure/autorest.java/pull/2686

To the "required nullable", current solution is to add a null protection in ctor (there were existing code for null protection in setter and getter).

PR #2557, opened 2024-02-19
tsp, in test, use union as extensible enum

weidongxu-microsoft:
First update test to use union as "extensible enum".

For now, enum is still as "extensible enum". We cannot break those on SDK repo.
We can change that behavior (switch to treat enum as "fixed enum") only after all lib and spec is updated, and the commit ID on SDK repo is also updated.

### DIFF
--- a/typespec-tests/tsp/enum.tsp
+++ b/typespec-tests/tsp/enum.tsp
@@ -19,10 +19,11 @@ enum Color {
   Green,
 }
 
-enum ColorModel {
-  Red,
-  Blue,
-  Green,
+union ColorModel {
+  string,
+  "Red",
+  "Blue",
+  "Green",
 }
 
 #suppress "@azure-tools/typespec-azure-core/use-extensible-enum" "For testing"
--- a/typespec-tests/tsp/enum.tsp
+++ b/typespec-tests/tsp/enum.tsp
@@ -72,10 +72,11 @@ interface EnumOp {
   @route("/color")
   getColor(): Color;
 
-  // TODO: the response as Union is ambiguous
-  // @get
-  // @route("/colormodel")
-  // getColorModel(): ColorModel;
+  @get
+  @route("/colormodel")
+  getColorModel(): {
+    @body body: ColorModel;
+  };
 
   @post
   @route("operation/colormodel")
--- a/typespec-tests/tsp/enum.tsp
+++ b/typespec-tests/tsp/enum.tsp
@@ -21,9 +21,9 @@ enum Color {
 
 union ColorModel {
   string,
-  "Red",
-  "Blue",
-  "Green",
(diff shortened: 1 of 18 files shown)


typespec-tests/tsp/enum.tsp
```diff
@@ -19,10 +19,11 @@ enum Color {
...
 }
 
-enum ColorModel {
-  Red,
-  Blue,
-  Green,
+union ColorModel {
+  string,
+  Red: "Red",
+  Blue: "Blue",
```
haolingdong-msft, line 25: why is this key value pairs? i.e. `Red: "Red"`, but others are not?

weidongxu-microsoft: There is a line ref to that UnionVariant
```
```

so it needs a name.

typespec-tests/tsp/enum.tsp
```diff
@@ -71,9 +72,10 @@ interface EnumOp {
   @route("/color")
   getColor(): Color;
 
-  @get
-  @route("/colormodel")
-  getColorModel(): ColorModel;
+  // TODO: the response as Union is ambiguous
+  // @get
+  // @route("/colormodel")
+  // getColorModel(): ColorModel;
```
weidongxu-microsoft, line 78: Disabled for now.

This op now equivalent to
```
getColorModel(): string | "Red" | "Blue" | "Green";
```
this union in response looks a bit strange (e.g. if compare to `getColorModel(): OkResponse | NoContentResponse`)

typespec-tests/tsp/enum.tsp
haolingdong-msft: I see.

typespec-tests/tsp/enum.tsp
weidongxu-microsoft: I guess I still need to solve this now... cadl-ranch has such test case.

weidongxu-microsoft: Changed to explicit `@body` on Timothee's advice.

typespec-tests/tsp/lro.tsp
```diff
@@ -46,11 +46,12 @@ namespace Cadl.LongRunning {
...
   @lroStatus
-  enum JobStatus {
+  union JobStatus {
+    string,
     NotStarted: "notStarted",
     Running: "running",
-    Succeeded: "succeeded",
-    Failed: "failed",
+    Succeeded: "Succeeded",
+    Failed: "Failed",
```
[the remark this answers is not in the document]
haolingdong-msft: yeah, a bit strange to be case sensitive.

haolingdong-msft (approved):

Issue #2548, opened 2024-02-06
MPG, migrate fluentgen modelerfour(/identity) naming settings to typespec-extension


XiaofeiCao:
https://github.com/Azure/autorest.java/blob/5acb73c3e8bfa56271247d3616cb1f5e942a5848/fluentgen/readme.md

#### IPRules
**tsp**: https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/networkanalytics/NetworkAnalytics.Management/main.tsp#L77-L83
**swagger**: https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/networkanalytics/resource-manager/Microsoft.NetworkAnalytics/stable/2023-11-15/networkanalytics.json#L1686

-----------------------------------------------

#### userName
**tsp**: https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/networkanalytics/NetworkAnalytics.Management/main.tsp#L434
**swagger**: https://github.com/Azure/azure-rest-api-specs/blob/f776434f63fb6505926273db8d4f9a93b75ee4a1/specification/networkanalytics/resource-manager/Microsoft.NetworkAnalytics/stable/2023-11-15/networkanalytics.json#L2455
```[tasklist]
### Tasks
- [x] Naming override from FluentGen
- [ ] Naming override from m4
```

weidongxu-microsoft:
Maybe not high priority?

XiaofeiCao:
From the difference of tsp/swagger generated SDK for networkanalytics, there seems to be quite a few related to this, 
e.g. IpRules -> IPRules, timestamp -> timeStamp, etc..
https://apiview.dev/Assemblies/Review/d42c19683b6c4d17ab34625bc9f2b750?revisionId=2554ec62ec9d4bc18c34c77d8ac88bfe&diffOnly=False&doc=False&diffRevisionId=b222576086f949fea4540b6ea88f3822

weidongxu-microsoft:
I see...

A bit unsure whether we should have emitter to handle it for them, or have service do `@clientName`.

Could you check e.g. `userName` still triggers a spotbug error on SDK repo?

XiaofeiCao:
Currently not needed for greenfield. Will see whether `@clientName` can be used for brownfield.

PR #2558, opened 2024-02-19
mpg, migrate fluent naming override from FluentGen

XiaofeiCao:
phase 1 of https://github.com/Azure/autorest.java/issues/2548


fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/SchemaNameNormalization.java
```diff
@@ -371,21 +371,29 @@ private String overrideName(String name) {
...
+    private boolean allUppercase(String name) {
+        return name.chars().allMatch(Character::isUpperCase);
+    }
+
+    // A partial match is when the configured keyword is only part of the whole word in the name.
+    // E.g. "lower": "Lower", and the actual name is "flower". We won't replace it to be "fLower".
+    private boolean partialMatch(String name, int index, int endIndex) {
+        return (index > 0 && isSameCase(name.charAt(index - 1), name.charAt(index)))
+            || (endIndex < name.length() && isSameCase(name.charAt(endIndex - 1), name.charAt(endIndex)));
+    }
```
XiaofeiCao, line 395: Refactor, extract method.

fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/SchemaNameNormalization.java
```diff
@@ -371,21 +371,29 @@ private String overrideName(String name) {
...
-                    replace = false;
-                } else if (endIndex < newName.length() && isSameCase(newName.charAt(endIndex - 1), newName.charAt(endIndex))) {
-                    replace = false;
-                }
-
-                if (replace) {
+                if (!partialMatch(newName, index, endIndex)
+                    // hack, mgmt only allow `preserve-uppercase-max-length` set to 2, so there can't be a partial match
+                    // for all uppercase, e.g. IP. Otherwise, the max uppercase length is 3, which is larger than 2.
+                    // TODO(xiaofei) Remove when we done migrating m4's `preserve-uppercase-max-length`
```
weidongxu-microsoft, line 377: I may not get what's here. One example?

XiaofeiCao: What I thought was, with `preserve-uppercase-max-length=2`, m4 won't allow `IPRules`(which has uppercase length= 3). We don't have the m4 restriction yet. Thus we can safely assume, when a name has "IP" in it, it's the max uppercase length and the whole word, and we will replace it.

Though through double check, `preserve-uppercase-max-length` seems not working as I assumed.. May need revisit this hack myself..

weidongxu-microsoft: My guess is m4 first breaks `IPRules` to "ip" + "rules", then check max-uppercase. So it be 2, not 3. Not sure.

XiaofeiCao: Got it.. Then it seems m4 has different naming override strategy with us..

weidongxu-microsoft: I think so. m4 logic is more complete. My logic here is not intended for uppercase or lowercase and probably not work for short words.

I guess you can ignore ID IP for now.

XiaofeiCao: Sure, will revisit it later.

weidongxu-microsoft (approved):

PR #2559, opened 2024-02-19
json-merge-patch, add test case for map's key is `null`

haolingdong-msft:
When a map's key is `null`, it should throw error when doing serialization both in `toJson()` and `toJsonMergePatch()`.
We will throw `NullPointorException` when map's key value is `null`. This PR add a test case for this.

### DIFF
--- a/typespec-tests/src/test/java/com/cadl/patch/PatchClientTest.java
+++ b/typespec-tests/src/test/java/com/cadl/patch/PatchClientTest.java
@@ -74,6 +74,21 @@ public class PatchClientTest {
     }
 
     @Test
+    public void testSerializationForMapNullKeyProperty() {
+        Exception exception = Assertions.assertThrows(NullPointerException.class, () -> {
+            Resource resource = new Resource(new HashMap<>());
+            JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);
+            resource.getMap().put(null, new InnerModel("value1"));
+            BinaryData.fromObject(resource).toString();
+        });
+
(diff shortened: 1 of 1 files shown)


typespec-tests/src/test/java/com/cadl/patch/PatchClientTest.java
```diff
@@ -73,6 +73,21 @@ public void testSerializationForMapProperty() throws JsonProcessingException {
...
+            Resource resource = new Resource(new HashMap<>());
+            JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);
+            resource.getMap().put(null, new InnerModel("value1"));
+            BinaryData.fromObject(resource).toString();
+        });
+
+        String expectedMessage = "'fieldName' cannot be null.";
+        String actualMessage = exception.getMessage();
+
+        Assertions.assertTrue(actualMessage.contains(expectedMessage));
```
weidongxu-microsoft, line 88: Not sure if we want to verify the message. `fieldName` a bit strange to me?

weidongxu-microsoft (approved):

typespec-tests/src/test/java/com/cadl/patch/PatchClientTest.java
haolingdong-msft: I add this because I want to verify it's the same error message with non-json-merge-patch cases. But I do agree the message is a bit strange.

weidongxu-microsoft: Fine for me. It should be from core, hope they won't change that.

PR #2561, opened 2024-02-20
partial-update related, code refactor to seperate javadoc description and tags calls

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/2510
Link https://github.com/Azure/autorest.java/issues/2502

This PR is before adding the markers for javadoc description and javadoc tags seperately. It will refactor existing code, add the helper APIs and implementations to allow user to seperately call `javadocComment()` and pass in `javadocCommentDescAction` and `javadocCommentTagsAction`.

Code changes:
1. `JavaType.java`: add interface method to write method level javadoc
2. `JavaClass.java` and `JavaInterface.java`: add override implementations to write method level javadoc
3. `JavaFile.java`: add public methods to write class/interface level javadoc
4. `util/TemplateUtil.java`: add the javadoc marker constants
5. `ClientMethodTemplate.java`, `ClientMethodTemplateBase.java`, `ConvenienceMethodTemplateBase.java`, `WrapperClientMethodTemplate.java`, `ServiceSyncClientTemplate.java`, `ServiceAsyncClientTemplate.java`: refactor for `Client`'s javadoc.
6. `ServiceClientBuilderTemplate.java`: refactor for `Builder`'s javadoc
7. `ModelTemplate`: refactor for model's javadoc


javagen/src/main/java/com/azure/autorest/model/javamodel/JavaFileContents.java
```diff
@@ -261,6 +261,16 @@ public void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment> comme
...
 
+    public void javadocCommentStart() {
+        line("/**");
+        addToPrefix(" * ");
+    }
+
+    public void javadocCommentEnd() {
+        removeFromPrefix(" * ");
+        line(" */");
+    }
```
weidongxu-microsoft, line 272: package private? -- assume we don't want it called out of javamodel

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
```diff
@@ -199,6 +204,64 @@ public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment>
         contents.javadocComment(wordWrapWidth, commentAction);
     }
 
+    @Override
+    public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment> commentDescriptionAction, Consumer<JavaJavadocComment> commentTagsAction, boolean withGeneratedMarker) {
```
weidongxu-microsoft, line 208: Can we make `Consumer<JavaJavadocComment> commentDescriptionAction, Consumer<JavaJavadocComment> commentTagsAction` to `Consumer<JavadocDescription> commentDescriptionAction, Consumer<JavadocTags> commentTagsAction`?

Then no one need to remember which parameter is for which. The type tells everything.

And `JavadocDescription` would only take the description, while `JavadocTags` takes parameter, return, throw etc. And one cannot add description on tags block, or tags to description block.

Also prefer to have overload of `javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment> commentDescriptionAction, boolean withGeneratedMarker)` for cases that no tag is needed.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
```diff
@@ -199,6 +204,64 @@ public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment>
         contents.javadocComment(wordWrapWidth, commentAction);
```
weidongxu-microsoft, line 204: Should we remove other old `javadocComment` overload, to avoid anyone calling the unwanted method?

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
```diff
@@ -199,6 +204,64 @@ public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment>
...
+            if (withWordWrap) {
+                contents.withWordWrap(wordWrapWidth, () -> commentDescriptionAction.accept(new JavaJavadocComment(contents)));
+            } else {
+                commentDescriptionAction.accept(new JavaJavadocComment(contents));
+            }
+            if (withGeneratedMarker) {
+                contents.line(GENERATED_JAVADOC_DESC_END_MARKER);
+            }
+            if (commentTagsAction == null) { // if no tags action, then end the javadoc comment
+                contents.javadocCommentEnd();
```
weidongxu-microsoft, line 241: Should it add GENERATED_JAVADOC_TAG_START_MARKER and GENERATED_JAVADOC_TAG_END_MARKER?

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaFileContents.java
```diff
@@ -69,7 +69,7 @@ public final void setWordWrapWidth(Integer wordWrapWidth) {
         this.wordWrapWidth = wordWrapWidth;
     }
 
-    private void withWordWrap(int wordWrapWidth, Runnable action) {
+    public void withWordWrap(int wordWrapWidth, Runnable action) {
```
weidongxu-microsoft, line 72: package private?

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
weidongxu-microsoft: BTW, not sure if we really need the `boolean withWordWrap` or `int wordWrapWidth` param anymore.

Maybe we can clean them up in the new APIs?

PS: I have no problem drop `boolean withGeneratedMarker` as well. Just make it always true is OK for me.

haolingdong-msft: I'm fine to drop `withWordWrap`, as long as everyone agrees, as there is some code diff if we remove this.
`withGeneratedMarker `: I will keep it first because currently it is always `false`, and we can merge the pr first and generated code will be the same.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
haolingdong-msft: updated, and it will add tags marker even it's `commentTagsAction` is null.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
srnagar: We should also drop `wordWrapWidth` as we eventually will use Spotless codeformatter to make it consistent with the monorepo.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaFileContents.java
```diff
@@ -261,6 +261,16 @@ public void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment> comme
                 commentAction.accept(new JavaJavadocComment(this))));
     }
 
+    void javadocCommentStart() {
+        line("/**");
+        addToPrefix(" * ");
+    }
+
+    void javadocCommentEnd() {
```
srnagar, line 269: Why are these package private? Can we make them public if it's used outside this class?

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
```diff
@@ -199,6 +204,59 @@ public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment>
...
+
+    /**
+     * Helper method to add a javadoc comment to the contents.
+     * @param commentDescriptionAction
+     * @param commentTagsAction
+     * @param withGeneratedMarker
+     * @param withWordWrap
+     * @param wordWrapWidth
+     */
+    private void javadocComment(Consumer<JavaJavadocComment> commentDescriptionAction, Consumer<JavaJavadocComment> commentTagsAction, boolean withGeneratedMarker, boolean withWordWrap, int wordWrapWidth) {
```
srnagar, line 225: This method seems to be repeated in a few places. We should move this to a common helper method and remove duplication.

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -36,7 +36,19 @@
 
 public abstract class ClientMethodTemplateBase implements IJavaTemplate<ClientMethod, JavaType> {
 
+//    protected static void generateProtocolMethodJavadoc(ClientMethod clientMethod, JavaType typeBlock) {
+//        typeBlock.javadocComment(
+//                comment -> generateProtocolMethodJavadocDescription(clientMethod, comment),
+//                comment -> generateProtocolMethodJavadocTags(clientMethod, comment),
+//                false);
+//    }
+
```
srnagar, line 45: Delete this commented code block.

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
```diff
@@ -59,6 +59,14 @@ public class TemplateUtil {
 
     public static final String DATE_UTC = "date-utc";
 
+    public static final String GENERATED_JAVADOC_DESC_START_MARKER = "<!-- start generated doc description -->";
+
+    public static final String GENERATED_JAVADOC_DESC_END_MARKER = "<!-- end generated doc description-->";
+
+    public static final String GENERATED_JAVADOC_TAG_START_MARKER = "<!-- start generated doc tags -->";
+
+    public static final String GENERATED_JAVADOC_TAG_END_MARKER = "<!-- end generated doc tags-->";
```
srnagar, line 68: TemplateUtil is being used for all utilities and will make it hard to find things. So, I suggest we have a JavadocUtil or something similar to group utilities that are more specific.

partial-update-tests/generated/src/main/java/fixtures/bodystring/AutoRestSwaggerBatServiceClientBuilder.java
```diff
@@ -186,7 +186,7 @@ public AutoRestSwaggerBatServiceClientBuilder configuration(Configuration config
 
     /**
      * Sets server parameter.
-     *
+     *
```
srnagar, line 189: This seems like an unintentional space added here. We should revert this to reduce noise in the diff.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaFileContents.java
weidongxu-microsoft: I was intended for these as internal usage within javamodel. We do not want dev writing template to call `javadocCommentStart` and `javadocCommentEnd` explicitly, as it is possible to mismatch the start/end call.

Though no strong opinion on this. Anyway only our dev uses it.

javagen/src/main/java/com/azure/autorest/model/javamodel/JavaClass.java
```diff
@@ -199,6 +204,59 @@ public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment>
         contents.javadocComment(wordWrapWidth, commentAction);
     }
 
+    @Override
+    public final void javadocComment(int wordWrapWidth, Consumer<JavaJavadocComment> commentDescriptionAction, Consumer<JavaJavadocComment> commentTagsAction, boolean withGeneratedMarker) {
+        javadocComment(commentDescriptionAction, commentTagsAction, withGeneratedMarker, true, wordWrapWidth);
+    }
+
+    @Override
+    public final void javadocComment(Consumer<JavaJavadocComment> commentDescriptionAction, Consumer<JavaJavadocComment> commentTagsAction, boolean withGeneratedMarker) {
```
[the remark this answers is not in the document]
haolingdong-msft: If having one consumer, we cannot easily distinguish description section and tags section and add the markers seperately.

weidongxu-microsoft:
Find we sometimes call `comment.line` as well.

If we keep 2 part, 1st part be description, allow `description` and `line` (maybe `inheritDoc` as well). 2nd part be tags, allow `param`, `methodReturns`, `methodThrows`, `deprecated`.

Issue #2411, opened 2023-11-17
client, model serialized to JSON does not have discriminator property, in option "pass-discriminator-to-child-deserialization"


weidongxu-microsoft:
See https://github.com/Azure/autorest.java/pull/2408

Current generated code seems get confused around

- JsonTypeId https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonTypeId.html
- JsonTypeInfo.As.EXISTING_PROPERTY https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonTypeInfo.As.html
- static var

---

The serialization can work with either
- use `JsonTypeInfo.As.PROPERTY`

or
- remove the `JsonTypeId` and make discriminator a (standard JSON property) class variable (not static)

PR #2563, opened 2024-02-20
Remove pass discriminator to children setting

alzimmermsft:
Removes the `pass-discriminator-to-child-deserialization` setting as the discriminator is now always passed to children models.


weidongxu-microsoft:
This is a wide API change? It adds a getter to all of these models.

Also there appear to be a bug with this https://github.com/Azure/autorest.java/issues/2411

weidongxu-microsoft:
And I do not see why "pass to children" is the default.

If user does not use this getter, it does not need to be there. So far, no mgmt user asks for it.

alzimmermsft:
That's good feedback @weidongxu-microsoft.

My thinking is this feature is very confusing and not well used. It'd be better for serialization to always have the discriminator property in the model so it isn't lost, which could happen when we're deserializing an unknown polymorphic type. And I agree, adding a getter everywhere won't be great but this may be nice to just have so our models aren't as mysterious. But if it's going to add a lot of noise, I think we can still remove the current setting and add a much clearer setting `hide-polymorphic-getter` which can default to true.

Generate.ps1
```diff
@@ -139,7 +139,7 @@ $job = @(
     "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/discriminator-flattening.json --namespace=fixtures.discriminatorflattening.clientflatten --modelerfour.flatten-models=false --client-flattened-annotation-target=NONE",
     "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/client-default-value.json --namespace=fixtures.clientdefaultvalue --modelerfour.flatten-models=false",
     "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/discriminator-flattening.json --namespace=fixtures.inheritance.donotpassdiscriminator --stream-style-serialization=false",
-    "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/discriminator-flattening.json --namespace=fixtures.inheritance.passdiscriminator --pass-discriminator-to-child-deserialization=true --stream-style-serialization=false",
+    "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/discriminator-flattening.json --namespace=fixtures.inheritance.passdiscriminator --stream-style-serialization=false",
```
weidongxu-microsoft, line 142: Should we merge the 2, as now they are same?

weidongxu-microsoft (commented):
Yeah, maybe its fine.

I see now the getter is added to the base class, not the terminal class.

The current CI error may relate to the issue I've linked.

Mgmt would generate model test cases like this
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/security/azure-resourcemanager-security/src/test/java/com/azure/resourcemanager/security/generated/ContainerRegistryVulnerabilityPropertiesTests.java

And it seems not work with the Jackson form of this.

weidongxu-microsoft:
Given what's merged to main branch, we now having another bug

tsp https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/model/inheritance/enum-discriminator/main.tsp#L12-L16

But code only return a `String` on `getKind` https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/type/model/inheritance/enumdiscriminator/models/Dog.java#L49-L51

It should return a `DogKind` as in tsp.

Generate.ps1
alzimmermsft: Yeah, I'll get those merged together

alzimmermsft:
> Given what's merged to main branch, we now having another issue (that we need to solve before next typespec-java release)
> 
> tsp https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/model/inheritance/enum-discriminator/main.tsp#L12-L16
> 
> But code only return a `String` on `getKind` https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/type/model/inheritance/enumdiscriminator/models/Dog.java#L49-L51
> 
> It should return a `DogKind` as in tsp.
> 
> I would go figure it out way sometime later as this is typespec. And the cause likely in typescript part. #2571
> 
> It may apply to Swagger as well, since the kind there could also be an enum.

It looks like this does get handled as expected, returning a `DogKind`. It looks to be breaking as the old value was String, but is that due to a change in the tsp or due to code generation changes?

srnagar (approved):
I think it's okay to remove this option and always add the getter for the discriminator property added to the base type of polymorphic models. Having this getter is quite useful for troubleshooting and in general represents the service response more accurately. Btw, .NET  always makes the discriminator visible. So, this will add parity in Java.

While this may cause a lot of classes in mgmt plane to be updated with a getter, it's not a breaking change and is a read-only method.

weidongxu-microsoft:
> It looks like this does get handled as expected, returning a `DogKind`. It looks to be breaking as the old value was String, but is that due to a change in the tsp or due to code generation changes?

It is due to change in codegen https://github.com/Azure/autorest.java/pull/2571 which I did last friday.

It won't break SDK or customer. The "getter" API wasn't there in any released typespec-java, so no one would see the `getKind` and its type.

I just find that the enabling of stream serialization also enable this "getter", so I had to make the type correct (without the getter, type does not matter).

weidongxu-microsoft:
also some error in test code gen
https://github.com/weidongxu-microsoft/autorest.java_ci/actions/runs/8039220851/job/21956004744#step:11:6672

I will have Xiaofei take a look.

weidongxu-microsoft:
I've check the Swagger case, it would also generate the `XXXKind` class, if that is an enum
https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvdmVyc2lvbmluZyI7CskfYXp1cmUtdG9vbHMvyCstxhVjb3JlIjsKCnVzaW5nIFR5cGVTcGVjLlJlc3Q70BVIdHRw0RVWyXLIG0HEWS5Db3JlLlRyYWl0czsKCkBzZXJ2aWNlKHt0aXRsZTogIlRleHQgYXV0aG9y5ACufSkKQOcAvWVkKFPGLsdccykKbmFtZXNwYWNlIMQ4Qcg3OwoKZW51bSDPMCB7CiAgQHVzZURlcGVuZGVuY3ko6wCWyCcudjFfMF9QcmV2aWV3XzIpCiAgdjE6ICIyMDIxLTDEAyIsCn0KCmludGVyZsR9UmVzb3VyY2VPcGVyYXTERAogIGV4dGVuZHPsAPjSKDxOb1JlcGVhdGFibGVSZXF1ZXN0cyAmIE5vQ29uZGnEJGFszhhsaWVudMcTSWQ%2BIHt9CgpAcsdaKCJwcm9qZWN0cyIpCm1vZGVsIFDGEeYA%2FWtlecQHdmlzaWJpbGl0eSgicmVhZCLkAOVpZDogc3TmAUAgIGRlc2NyaXDkAII%2FyhgKICDkAXTLEsR%2BdOQB0MdvTGlzdFJlc3BvbnNlUHJvcGVydGllc3PlAfPmAITFLkNvbnRleHQoxRjHDS7EPOQAinLHQDrFMcgzTG9j5QE1xjTIDuQBVcUtxDkgIHDJczogeyDFEyAgY291bnQ6IGludDMyxRN95ADifeUAveoBxukCP8V2bGlzdCBpc%2FMB3OkBx%2BQArzznAV8s%2FQD5PsRr&e=%40azure-tools%2Ftypespec-autorest&options=%7B%7D

So we are good on parity of TypeSpec and Swagger (mgmt needs this) @XiaofeiCao

PR #2565, opened 2024-02-21
Skip generating clients if there are no operations

srnagar:


javagen/src/main/java/com/azure/autorest/Javagen.java
```diff
@@ -137,121 +137,124 @@ protected NodeTuple representJavaBeanProperty(Object javaBean, Property property
...
         JavaPackage javaPackage = new JavaPackage(this);
-        // Service client
-        if (CoreUtils.isNullOrEmpty(client.getServiceClients())) {
-            javaPackage.addServiceClient(client.getServiceClient().getPackage(),
-                client.getServiceClient().getClassName(), client.getServiceClient());
-        } else {
-            // multi-client from TypeSpec
-            for (ServiceClient serviceClient : client.getServiceClients()) {
-                javaPackage.addServiceClient(serviceClient.getPackage(), serviceClient.getClassName(), serviceClient);
+        if (client.getServiceClient() != null || !CoreUtils.isNullOrEmpty(client.getServiceClients())) {
```
alzimmermsft, line 140: Doesn't `CoreUtils.isNullOrEmpty` handle the nullness check?

alzimmermsft (approved):

javagen/src/main/java/com/azure/autorest/Javagen.java
srnagar: `CoreUtils.isNullOrEmpty()` is only for collections or string. Not for null checking an object.

PR #2567, opened 2024-02-22
tsp, migrate to Azure.Core getUnionAsEnum

weidongxu-microsoft:
Another patch release :-(


typespec-tests/src/main/java/com/cadl/enumservice/models/PriorityModel.java
weidongxu-microsoft, line 1: This class is wrong. Emitter does not support extensible enum of integer/float. But keep it here in test.
https://github.com/Azure/azure-sdk-for-java/issues/38839

typespec-tests/tsp/enum.tsp
```diff
@@ -20,12 +20,22 @@ enum Color {
...
-  Red: "Red",
+  ColorModelBase,
   Blue: "Blue",
   Green: "Green",
 }
 
+union ColorModelBase {
+  string,
+  Red: "Red",
+}
```
weidongxu-microsoft, line 31: `getUnionAsEnum` can handle this correctly.

typespec-tests/src/main/java/com/type/union/EnumsOnlyAsyncClient.java
weidongxu-microsoft, line 1: Change on this test is due to merging of the Union due to getUnionAsEnum
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/union/main.tsp#L83-L89

haolingdong-msft (approved):

Issue #2471, opened 2023-12-18
tsp, support json-merge-patch for model with required properties


haolingdong-msft:
**Case:**

Consider the case of shared model between PUT and Json-Merge-Patch e.g.
```
model SharedModel {
  requiredProp1: string,
  requiredProp2: int32,
  innerModel: InnerModel,
  optionalProp3?: ...
}
```
And how we'd like the class be. Think e.g. what user need to do on the patch, when they only want to modify requiredProp1

Design discussion needed, as current required properties are all set in constructor and do not have setters.

Note: we have one [RFC item](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#rest-response-body-is-resource-schema) saying patch operation's model should not contain required properties. typespec-autorest will create another model named `XXXCreateOrUpdate` to make all the properties optioanl. This is not done by compiler, compiler will not return the model `XXXCreateOrUpdate`.
[...]

PR #2570, opened 2024-02-23
tsp, required model property as optional in json merge patch model

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/2471

- For model used in JSON merge patch, treat all its mutable property as optional. (if property is not mutable, it won't affect patch)
- Add a "requiredForCreate" to client model property, so that we are able to say this property is required when create the resource (but optional when update -- and hence not in ctor).

code https://github.com/Azure/autorest.java/pull/2570/commits/11945da9de526e903fcd3c7f8c52dbc1067f5499


typespec-tests/src/main/java/com/cadl/patch/PatchClient.java
```diff
@@ -48,9 +48,9 @@ public final class PatchClient {
...
      *     name: String (Required)
      *     description: String (Optional)
-     *     map (Required): {
+     *     map (Optional, Required on create): {
      *         String (Required): {
-     *             name: String (Required)
+     *             name: String (Optional, Required on create)
      *             description: String (Optional)
      *         }
      *     }
```
weidongxu-microsoft, line 56: change to javadoc -- optional on patch/update, but required when create

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
```diff
@@ -161,6 +158,22 @@ public int getAge() {
...
+     * 
+     * @param age the age value to set.
+     * @return the Fish object itself.
+     */
+    @Generated
+    public Fish setAge(int age) {
+        this.age = age;
+        this.updatedProperties.add("age");
+        return this;
+    }
```
weidongxu-microsoft, line 175: Now it is via setter (instead of via ctor).

Again Javadoc mention that this is required on create.

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/models/AnalyzeDocumentRequest.java
```diff
@@ -73,6 +73,9 @@ public AnalyzeDocumentRequest setBase64Source(byte[] base64Source) {
         return this;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
```
weidongxu-microsoft, line 81: Most diff is this. Use `@inheritDoc` for consistency to javadoc.

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
weidongxu-microsoft: On the changes of the ctor -> setter, may be easier to see from the changes to test https://github.com/Azure/autorest.java/pull/2570/commits/6f108cffbf4cf24844c6efb735242e69a096ad74

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
```diff
@@ -310,7 +311,11 @@ private static String methodParameterDescriptionOrDefault(ClientMethodParameter
...
     }
 
-    private static String appendOptionalOrRequiredAttribute(boolean isRequired, boolean isRootSchema) {
-        return isRootSchema ? "" : isRequired ? " (Required)" : " (Optional)";
+    private static String appendOptionalOrRequiredAttribute(boolean isRequired, boolean isRequiredForCreate, boolean isRootSchema) {
+        String optionalStr = " (Optional)";
+        if (isRequiredForCreate && !isRequired) {
+            optionalStr = " (Optional, Required on create)";
+        }
+        return isRootSchema ? "" : isRequired ? " (Required)" : optionalStr;
```
alzimmermsft, line 319: Not a big fan of ternary operators in ternary operators as it gets hard to read.

Could we just have an early out of `""` at the beginning if it's the root schema? Then just have an if / else if / else block?

```java
if (isRootSchema) {
    return "";
} else if (isRequired) {
    return " (Required)";
} else if (isRequiredForCreate) {
    return " (Optional, Required on create)";
} else {
    return " (Optional)";
}
```

alzimmermsft (approved):

javagen/src/main/java/com/azure/autorest/template/ClientMethodTemplateBase.java
weidongxu-microsoft: Agree. Nested ternary is hard to read (even with parentheses).

Updated.

haolingdong-msft (approved):

PR #2572, opened 2024-02-23
bug fix, 'partial update' will execute twice when generate from TypeSpec.

haolingdong-msft:
context:
https://dev.azure.com/azure-sdk/public/_build/results?buildId=3514795&view=logs&jobId=b70e5e73-bbb6-5567-0939-8415943fadb9&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=d31d5e2d-d8df-53bc-da94-87e19f5b2ee4

The error is due to the service enabled both customization and partial update. And currently both postprocessor and typespec plugin will execute partial update. When generating from tsp, below things happen:
1. partial update will be executed, which will add the existing file's customized method.
2. postprocessor customization executed which will add the customized method to the generated file.
3. postprocessor partial update will be executed, so in this partial update, there will be duplicate methods.


partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -216,7 +216,7 @@ private static void validateGeneratedClassOrInterface(List<BodyDeclaration<?>> g
         for (BodyDeclaration<?> generatedMember : generatedFileMembers) {
             if (generatedMember.isCallableDeclaration()) {
                 if (methodSignatureSet.contains(generatedMember.asCallableDeclaration().getSignature())) {
-                    throw new RuntimeException("Found duplicate methods in the generated file.");
+                    throw new RuntimeException(String.format("Found duplicate methods in the generated file. Signature: %s", generatedMember.asCallableDeclaration().getSignature()));
```
haolingdong-msft, line 219: Make the error more detailed.

weidongxu-microsoft (approved):
So currently `writeToFiles` does the partial update and format?

haolingdong-msft:
> So currently `writeToFiles` does the partial update and format?

Yes, code for your reference: https://github.com/Azure/autorest.java/blob/285346eb9fd511bcea567376d4e515fe680095a6/postprocessor/src/main/java/com/azure/autorest/postprocessor/Postprocessor.java#L114

alzimmermsft (approved):

PR #2573, opened 2024-02-23
Remove mockito from mock tests

alzimmermsft:


weidongxu-microsoft (approved):
Thanks. I wasn't aware that `MockHttpResponse` is exposed.

PR #2576, opened 2024-02-26
tsp, json-merge-patch, bug fixes

haolingdong-msft:
Code changes:
 1. set `toJsonMergePatch()` as `private` method
 2. switch the order of updatedProperties check and null check
 3. bump minor version.

Link issue: https://github.com/Azure/autorest.java/issues/2471

Remain two issues:
1. about discriminator/hierarchical model. https://github.com/Azure/autorest.java/pull/2576#discussion_r1504012383
2. about deserialization called setter. https://github.com/Azure/autorest.java/pull/2576#issuecomment-1966144039


typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
```diff
@@ -215,18 +215,22 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
...
-            jsonWriter.writeStringField("kind", this.kind);
-        } else if (updatedProperties.contains("kind")) {
-            jsonWriter.writeNullField("kind");
+        if (updatedProperties.contains("kind")) {
+            if (this.kind == null) {
+                jsonWriter.writeNullField("kind");
+            } else {
+                jsonWriter.writeStringField("kind", this.kind);
+            }
         }
```
weidongxu-microsoft, line 226: Please add a UT to verify that `kind` is already set when create the object (should be the current behavior but we won't want regression there).
https://github.com/Azure/autorest.java/blob/18c064e28fabeda10a1db637e4a3096cb9fbf039/typespec-tests/src/main/java/com/cadl/patch/models/Salmon.java#L67-L70

weidongxu-microsoft (commented):
Overall, please have a UT on this case.

E.g. you can initialize a resource via `BinaryData.toObject()` that would set the value but not the `updatedProperties`, and verify the `toJsonMergePatch` only serialize the ones called by the setter.

weidongxu-microsoft:
@haolingdong-msft 

Also please bump the minor version, and do a release after the PR.

Shawn is investigating the stream on OpenAI. With a release, he may help you solve the problems relates to OpenAI.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/models/User.java
```diff
@@ -159,17 +159,21 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
     }
 
     @Generated
-    public JsonWriter toJsonMergePatch(JsonWriter jsonWriter) throws IOException {
+    private JsonWriter toJsonMergePatch(JsonWriter jsonWriter) throws IOException {
```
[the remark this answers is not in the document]
haolingdong-msft: This method is only used by `toJson()`, so I think `private` should be enough. Did I miss/misunderstood anything? What's your thoughts?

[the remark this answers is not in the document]
weidongxu-microsoft: My fault. Yes, private should be good.

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
haolingdong-msft: I'm a bit confused, do you mean change the fish kind from salmon to shark using the same object? This are different class's instances, I don't get how to change. `Fish` won't have a public API called `setKind()` to let customer change as well. Consider below code:
```java
Fish fish = new Salmon().setAge(1);
fish.setColor("pink");
Assertions.assertEquals("salmon", fish.getKind());
JsonMergePatchHelper.getFishAccessor().prepareModelForJsonMergePatch(fish, true);
// patch here
// Could you elaborate how does user change `salmon` to `shark`? 
// I think they will need to call the Shark's ctor, i.e. fish = new Shark(), but this creates a new object , not updating the same object to change fish kind from salmon to shark. So I think line 3 in this code already cover it.
fish.setColor(null);
```

weidongxu-microsoft: No. I mean `setFish` to the the model that refers to the `Fish` class (your test case does not have such model).

You can imagine `Resource` model has a `fish: Fish` property. And user need to set a `salmon` to it. The `kind` need to be there.

weidongxu-microsoft:
Another problem on polymorphic class is that the `fromJson` of the subclass calls the setters to set field on base class.
https://github.com/Azure/autorest.java/blob/5c1f03fd12792048c9d1aa68b162b560313e3311/typespec-tests/src/main/java/com/cadl/patch/models/Salmon.java#L280-L284

I assume it triggers `updatedProperties`.

It proabably not a blocking issue for now (at least it does not affect public API).

Please log it for discussion with Srikanta/Alan.

weidongxu-microsoft (approved):

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
haolingdong-msft: Got it. Thanks.
This case seems to have some issue. e.g.
```java
Resource resource = new Resource();
Fish salmon = new Salmon().setAge(1);
Assertions.assertEquals("salmon", salmon.getKind());
resource.setFish(salmon);
JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);
Fish shark = new Shark().setAge(2);
shark.setColor(null);
resource.setFish(shark);
// here when serialize for resource.fish, it calls shark.toJsonMergePatch(), but kind is not added to shark's updatedProperties, but to fish's updated properties.
String json = BinaryData.fromObject(resource).toString();
Assertions.assertEquals("shark", node.get("fish").get("kind"));
```

I guess we may need to initialize `Shark`'s `updatedProperties` from the parent class `Fish`, maybe from ctor?

weidongxu-microsoft: Maybe we can have `Set<String> updatedProperties = new HashSet<>()` in base class as package private, and the subclass uses it directly, not defining a new one.

Do bring this to discussion as well. There be alternatives (but I don't think base class can init var from subclass in its ctor). -- Do think about alternatives

weidongxu-microsoft:
The case of polymorphic class would not be a blocker to the PR -- let's hope service not using such in PATCH...

But it would be higher priority when service uses it.

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
haolingdong-msft: I will bring it up in tmr's sync up.

typespec-extension/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/typespec-java",
-  "version": "0.13.9",
+  "version": "0.14.1",
```
weidongxu-microsoft, line 3: 0.14.0 ?

Also do a npm install, the package-lock.json also need to be updated.

typespec-extension/changelog.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 0.14.0 (2024-02-29)
```
srnagar, line 3: Why is this 0.14.0 instead of 0.13.11?

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
srnagar: As discussed, let's review the design proposal in next week's sync.

srnagar (approved):

typespec-extension/changelog.md
weidongxu-microsoft: The switch to stream-style-serialization is a huge change. (preview https://github.com/Azure/azure-sdk-for-java/pull/38268/files)
And it may take a week to fully move to this (even when no bug found).

Hence, we may had to release patch on 0.13.x.

weidongxu-microsoft (approved):

PR #2592, opened 2024-02-29
Remove JSON flatten setting

alzimmermsft:
Removes the setting `require-x-ms-flattened-to-flatten` as flattening should always require `x-ms-flattened` to be enabled to produce flattened objects.

### DIFF
--- a/vanilla-tests/src/main/java/fixtures/inheritance/passdiscriminator/models/MetricAlertSingleResourceMultipleMetricCriteria.java
+++ b/vanilla-tests/src/main/java/fixtures/inheritance/passdiscriminator/models/MetricAlertSingleResourceMultipleMetricCriteria.java
@@ -5,7 +5,6 @@
 package fixtures.inheritance.passdiscriminator.models;
 
 import com.azure.core.annotation.Fluent;
-import com.azure.core.annotation.JsonFlatten;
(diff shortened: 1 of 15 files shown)


vanilla-tests/src/main/java/fixtures/inheritance/passdiscriminator/models/MetricAlertSingleResourceMultipleMetricCriteria.java
```diff
@@ -16,13 +15,12 @@
...
     use = JsonTypeInfo.Id.NAME,
-    property = "odata\\.type",
+    property = "odata.type",
     defaultImpl = MetricAlertSingleResourceMultipleMetricCriteria.class,
     visible = true)
 @JsonTypeName("Microsoft.Azure.Monitor.SingleResourceMultipleMetricCriteria")
-@JsonFlatten
 @Fluent
-public class MetricAlertSingleResourceMultipleMetricCriteria extends MetricAlertCriteria {
+public final class MetricAlertSingleResourceMultipleMetricCriteria extends MetricAlertCriteria {
```
weidongxu-microsoft, line 23: Would this affect vanilla that still depends on `JsonFlatten`?

weidongxu-microsoft (approved):

PR #2593, opened 2024-03-01
tsp, partial-update, replace static initializer declaration

weidongxu-microsoft:
partly solve https://github.com/Azure/autorest.java/issues/2468

assumption is this block is always treat as `@Generated` https://github.com/Azure/autorest.java/issues/2468#issuecomment-1897654248, and (since it does not have a name) always corresponding.

code https://github.com/Azure/autorest.java/pull/2593/commits/8417747019b76988f42023216e1b9baed83fab47


partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
```diff
@@ -474,6 +490,11 @@ private static boolean isMembersCorresponding(BodyDeclaration<?> member1, BodyDe
         if (member1.isCallableDeclaration() && member2.isCallableDeclaration()) {
             // compare signature
             return member1.asCallableDeclaration().getSignature().equals(member2.asCallableDeclaration().getSignature());
+        } else if (member1 instanceof InitializerDeclaration && member2 instanceof InitializerDeclaration
+                && ((InitializerDeclaration) member1).isStatic() && ((InitializerDeclaration) member2).isStatic()) {
```
haolingdong-msft, line 494: so assumption is generated static block is already same with existing static block? would add this into comment?

weidongxu-microsoft: No such assumption.

It is "always replace" (same logic, if method on both side is `@Generated`).

haolingdong-msft (approved):

partial-update/src/test/java/com/azure/autorest/partialupdate/util/PartialUpdateHandlerTest.java
```diff
@@ -408,4 +412,64 @@ public void testModuleInfoFileToTestMergeExistingFileAndGeneratedFileToTestIgnor
...
+        assertEquals(1, compilationUnit.getTypes().size());
+        List<InitializerDeclaration> staticInitializerDeclaration = compilationUnit.getTypes().get(0).getMembers().stream()
+                .filter(m -> m instanceof InitializerDeclaration)
+                .map(m -> (InitializerDeclaration) m)
+                .filter(InitializerDeclaration::isStatic)
+                .collect(Collectors.toList());
+        // verify 1 block
+        Assertions.assertEquals(1, staticInitializerDeclaration.size());
+        // verify the block is replaced
+        Assertions.assertTrue(staticInitializerDeclaration.get(0).toString().contains("// token for test"));
```
weidongxu-microsoft, line 432: test for this "replace"

partial-update/src/main/java/com/azure/autorest/partialupdate/util/PartialUpdateHandler.java
haolingdong-msft: Sorry for using the wrong word 'same',  'corresponding' is better. both `@Generated` and method/code block is corresponding, then we will replace.

weidongxu-microsoft: Added the comment to Javadoc of the method as well.

haolingdong-msft (approved):

haolingdong-msft (approved):

PR #2594, opened 2024-03-01
dpg, settings, remove generate-models

weidongxu-microsoft:
Checked that no SDK uses it, in SDK repo md files.


protocol-tests/src/test/java/fixtures/dpgcustomization/DpgCustomizationTests.java
weidongxu-microsoft, line 1: Just make it compile.

We don't want to put effort on DPG from Swagger.

alzimmermsft (approved):

PR #2603, opened 2024-03-05
tsp, patch, skip payload flatten

weidongxu-microsoft:
Do not do payload flatten for json-merge-patch model. The proper patch API impl depends on the setter of that model.


typespec-tests/src/main/java/com/cadl/flatten/FlattenAsyncClient.java
```diff
@@ -132,6 +135,47 @@ public Mono<Response<Void>> sendLongWithResponse(String id, BinaryData request,
...
+     * The update operation.
+     * <p><strong>Request Body Schema</strong></p>
+     * <pre>{@code
+     * {
+     *     patch (Optional, Required on create): {
+     *         title: String (Optional)
+     *         description: String (Optional)
+     *         status: String(NotStarted/InProgress/Completed) (Optional)
+     *     }
+     * }
```
weidongxu-microsoft, line 148: Payload is in this shape ("patch" is just a property in request body), flatten or not.

typespec-tests/src/main/java/com/cadl/flatten/FlattenClient.java
```diff
@@ -225,4 +269,28 @@ public void sendLong(SendLongOptions options) {
...
+     * @throws HttpResponseException thrown if the request is rejected by server.
+     * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
+     * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
+     * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return the response.
+     */
+    @Generated
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public TodoItem update(long id, UpdateRequest request) {
```
haolingdong-msft, line 288: The idea skipping flatten looks good to me. I'm fine with the current naming as well.

haolingdong-msft (approved):

PR #2607, opened 2024-03-06
Generic core update

srnagar:


javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
```diff
@@ -359,7 +359,11 @@ protected void addImports(Set<String> imports, List<ConvenienceMethod> convenien
     }
 
     protected void addGeneratedAnnotation(JavaType typeBlock) {
-        typeBlock.annotation(Annotation.GENERATED.getName());
+        if (JavaSettings.getInstance().isBranded()) {
+            typeBlock.annotation(Annotation.GENERATED.getName());
+        } else {
+            typeBlock.annotation(Annotation.METADATA.getName() + "(generated = true)");
```
haolingdong-msft, line 365: I may be lack of context on this. What is `@MetaData` used for besides setting `generated`?

srnagar: It can be used for any metadata we want to add about a method, class or property. Right now, we are using it to mark methods as generated and classes as either fluent or immutable. In the future, there might be other things added to this annotation.

weidongxu-microsoft: On the chat.

Maybe we could use a full class name here "com.generic.core.annotation.Metadata(generated = true)", and not add the "import com.generic.core.annotation.Metadata"?

The "Metadata" may conflict with existing code https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/resourcemanager/azure-resourcemanager-monitor/src/main/java/com/azure/resourcemanager/monitor/models/DataCollectionRuleMetadata.java#L13

weidongxu-microsoft: Or if it is urgent, you may add Metadata into the reserved word list.
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/util/CodeNamer.java#L65-L81
https://github.com/Azure/autorest.java/blob/main/preprocessor/src/main/java/com/azure/autorest/preprocessor/namer/CodeNamer.java#L63-L79

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/FormRecognizerClientImplBuilder.java
```diff
@@ -47,7 +47,7 @@
 @ServiceClientBuilder(serviceClients = { FormRecognizerClientImpl.class })
 public final class FormRecognizerClientImplBuilder
     implements HttpTrait<FormRecognizerClientImplBuilder>, ConfigurationTrait<FormRecognizerClientImplBuilder>,
-    AzureKeyCredentialTrait<FormRecognizerClientImplBuilder>, EndpointTrait<FormRecognizerClientImplBuilder> {
+    KeyCredentialTrait<FormRecognizerClientImplBuilder>, EndpointTrait<FormRecognizerClientImplBuilder> {
```
weidongxu-microsoft, line 50: nit, I seems not able to find the code change that affect this.

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -458,27 +459,24 @@ protected Map<ServiceClient, com.azure.autorest.extension.base.model.codemodel.C
...
-
         clientBuilder.addBuilderTrait(ClientBuilderTrait.HTTP_TRAIT);
+
         clientBuilder.addBuilderTrait(ClientBuilderTrait.CONFIGURATION_TRAIT);
         if (serviceClient.getSecurityInfo().getSecurityTypes().contains(Scheme.SecuritySchemeType.OAUTH2)) {
             clientBuilder.addBuilderTrait(ClientBuilderTrait.TOKEN_CREDENTIAL_TRAIT);
         }
         if (serviceClient.getSecurityInfo().getSecurityTypes().contains(Scheme.SecuritySchemeType.KEY)) {
-            if (JavaSettings.getInstance().isUseKeyCredential()) {
+            if (JavaSettings.getInstance().isBranded() || JavaSettings.getInstance().isUseKeyCredential()) {
```
weidongxu-microsoft, line 469: Should it be
```
            if (!JavaSettings.getInstance().isBranded() || JavaSettings.getInstance().isUseKeyCredential()) {
````

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
weidongxu-microsoft: Here https://github.com/Azure/autorest.java/pull/2608/commits/a6c34f6dbe4c4442ffd599cc3e614ad34aae3ceb

javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
srnagar: Fixed

javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java
srnagar: Updated code to add the metadata annotation import only for unbranded libraries.

weidongxu-microsoft (approved):

Issue #2596, opened 2024-03-01
Change default generation flavor and flags for generic codegen


lmazuel:
- Emitter should generate "unbranded" code by default
- Add a new flag `flavor` that generates azure code for the value `azure`. This is not case-sensitive. Any other value does nothing, and therefore generates unbranded code.
- Don't remove yet currently supported flag `branded` to help the transition to the new system. If both flags are set, read `flavor` first and ignore `branded` boolean flag.
- If `package-dir` starts with `azure`, auto-set `flavor` to `azure` if `flavor` was not set already. If `package-dir` is anything else, no specific behavior.

weidongxu-microsoft:
We can do this, but as 2024/03/04 it would not be meaningful, as Java don't have a generic core published anywhere.
@srnagar

srnagar:
As discussed, since the generic-core is not published to Maven Central yet, switching the default will result in unresolved dependencies for generic core. Also, the generic core APIs are still work in progress and we should not change the emitter to be generic by default until the APIs are stabilized.

weidongxu-microsoft:
a bit of thought on internal impl

we probably still want to keep the boolean in internal, e.g. `isBranded` or `isAzure` or `isGeneric`. personally I don't want to see 10+ lines of `if ("azure".equalIgnoreCase(settings.getFlavor()) {` scattered around the whole repo.

haolingdong-msft:
I'm going to close this task, and we will create other tasks related with unbranding later, e.g. adopting generic-core, embedding flavor to generated code, make non-azure code as default (after generic-core go public). @lmazuel @srnagar @lirenhe Please let me know if you have any concern on closing this task.
/cc @weidongxu-microsoft

PR #2609, opened 2024-03-07
tsp, support flavor flag from tsp-config

haolingdong-msft:
link https://github.com/Azure/autorest.java/issues/2596


weidongxu-microsoft (commented):
Personally I'd prefer to put `flavor` to JavaSettings, add `getFlavor`, and keep the convenience API `isBranded` in `JavaSettings`.

not a strong opinion though, not visible and can be changed later anyway.

weidongxu-microsoft:
Also update the tspconfig.yaml in our test module to `flavor=azure`.

haolingdong-msft:
> Personally I'd prefer to put `flavor` to JavaSettings, add `getFlavor`, and keep the convenience API `isBranded` in `JavaSettings`.
> 
> not a strong opinion though, not visible and can be changed later anyway.

I did not add `flavor` into it `JavaSettings` just to make the change easy, but I agree with you adding `flavor` to `JavaSettings` would be better. If we want to support embed `flavor`(brand name) into our generated sdk, we will need this in `JavaSettings`. I will update in this pr.

weidongxu-microsoft (approved):

weidongxu-microsoft (commented):
I remember Srikanta say 0.15.0 in meeting.

+changelog

weidongxu-microsoft:
Also make sure you've tested flavor!=azure case (that is what neede by Srikanta).

We may need to have a test setup, but maybe later.

typespec-extension/src/code-model-builder.ts
```diff
@@ -365,7 +365,7 @@ export class CodeModelBuilder {
                 // HTTP Authentication should use "Basic token" or "Bearer token"
                 schemeOrApiKeyPrefix = pascalCase(schemeOrApiKeyPrefix);
 
-                if (!(this.options.branded === false)) {
+                if (this.options["flavor"]?.toLocaleLowerCase() === "azure") {
```
weidongxu-microsoft, line 368: this may need be `!this.options["flavor"] || <existing>`

If no this `flavor`, it is still default to azure, at present.

haolingdong-msft: https://github.com/Azure/autorest.java/blob/a3e61b8ef7544e38785f0b01c480a324d3c646b5/typespec-extension/src/emitter.ts#L63
default is set here, so I did not do the check. I have also tested when flavor is not set, we will generate azure code. Also tested `flavor: xxx` (xxx is not Azure), it generates non-azure.
Let me know if you thought we should do double check.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -291,9 +291,14 @@ public TypeSpecPlugin(EmitterOptions options, boolean sdkIntegration) {
...
         }
 
-        if (options.getBranded() == Boolean.FALSE) {
-            SETTINGS_MAP.put("branded", options.getBranded());
+        if (options.getFlavor() != null) {
+            SETTINGS_MAP.put("flavor", options.getFlavor());
+        }
+        if (options.getPackageDir() != null && options.getPackageDir().toLowerCase().startsWith("azure")) {
+            SETTINGS_MAP.put("flavor", "Azure");
+        }
```
srnagar, line 299: If a `flavor` is explicitly specified, we should not overwrite it based on package directory name. So, `options.getFlavor()` takes precedence and packageDir should be checked only if `options.getFlavor()` is null.

typespec-extension/src/main/java/com/azure/typespec/model/EmitterOptions.java
```diff
@@ -72,6 +72,9 @@ public class EmitterOptions {
     @JsonProperty(value="dev-options")
     private DevOptions devOptions;
 
+    @JsonProperty(value="package-dir")
+    private String packageDir;
```
weidongxu-microsoft, line 76: I don't think we need to put the packageDir into Java part.

I'd prefer the emitter.ts do the config of the default situations, and set the flavor there.
E.g.

```
if (!flavor) {
  if (package-dir.startWith("azure)) {
    flavor = "Azure";
  } else {
    // default, can change later
    flavor = "Azure";
  }
}
```

haolingdong-msft: thanks, updated

typespec-extension/src/emitter.ts
```diff
@@ -118,6 +118,16 @@ export async function $onEmit(context: EmitContext<EmitterOptions>) {
...
+    if (!options["flavor"]) {
+      if (options["package-dir"]?.toLocaleLowerCase().startsWith("azure")) {
+        // Azure package
+        options["flavor"] = "Azure";
+      } else {
+        // default
+        options["flavor"] = "Azure";
+      }
+    }
+
```
weidongxu-microsoft, line 130: Coud we move this block to before
https://github.com/Azure/autorest.java/blob/a20ba150b6c4c4cc57c6de6e5dd0c8b6fdc6b687/typespec-extension/src/emitter.ts#L111-L112

This way, it make sure in code-model-builder, the `options.flavor` is already resolved.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
```diff
@@ -294,9 +294,6 @@ public TypeSpecPlugin(EmitterOptions options, boolean sdkIntegration) {
         if (options.getFlavor() != null) {
             SETTINGS_MAP.put("flavor", options.getFlavor());
         }
-        if (options.getPackageDir() != null && options.getPackageDir().toLowerCase().startsWith("azure")) {
```
weidongxu-microsoft: Please remove this from EmitterOptions class as well.

typespec-extension/src/main/java/com/azure/autorest/TypeSpecPlugin.java
haolingdong-msft: Yeah, it depends on if we set `flavor: Generic` and `package-dir: "azure-ai-documentintelligence"`, what shoud we generate, azure or non-azure? (I know it's unlikely to happen) I was thinking it shall generate `azure` previously, maybe it is wrong understanding. I have updated the logic according to the statement "options.getFlavor() takes precedence" now.

weidongxu-microsoft: I would agree to Srikanta, that explicit `flavor` take precedence.

weidongxu-microsoft (approved):

PR #2612, opened 2024-03-08
tsp, bug fix on grouping

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/pull/2612/commits/196d1705d6b3d6b66f20684a19bb53a834ab5c5e

Error happens when read-only property or constant property in the model get flattened as well.

There could be more bug, e.g. a model (that extends another model) get flattened...


typespec-tests/src/main/java/com/cadl/flatten/implementation/models/SendLongRequest.java
weidongxu-microsoft, line 1: There is bug on this class. It should not be `JsonSerializable`.

XiaofeiCao (approved):

PR #2619, opened 2024-03-12
tsp, refactor multipart/form-data, move the model to typescript

weidongxu-microsoft:
Refactor.

Move the data model to typescript part, so that TCGC SdkMultipartFileType can be mapped to code-model schema in typescript directly.


javagen/src/main/java/com/azure/autorest/mapper/BinaryMapper.java
```diff
@@ -28,6 +30,8 @@ public IType map(BinarySchema binarySchema) {
     if (binarySchema == null) {
       return null;
     }
-    return GenericType.FLUX_BYTE_BUFFER;
+    return JavaSettings.getInstance().isDataPlaneClient()
+            ? ClassType.BINARY_DATA
+            : GenericType.FLUX_BYTE_BUFFER;
```
weidongxu-microsoft, line 35: I didn't know this could work alright?

typespec-extension/src/code-model-builder.ts
```diff
@@ -2510,6 +2513,34 @@ export class CodeModelBuilder {
...
+            property.language.default.description,
+            getFileDetailsSchema(
+              property.language.default.name,
+              schema.language.default.namespace,
+              this.codeModel.schemas,
+              this.binarySchema,
+              this.stringSchema,
+            ),
+          );
+        }
```
weidongxu-microsoft, line 2539: @haolingdong-msft 
TCGC `SdkMultipartFileType` would have this mapping.

haolingdong-msft (approved):

PR #2626, opened 2024-03-15
Remove direct dependencies to Azure SDKs

alzimmermsft:
Removes dependencies on `azure-core`, `generic-core`, and related SDKs to allow for more flexibility in working with non-shipped implementation of those libraries.


javagen/src/main/java/com/azure/autorest/util/TypeUtil.java
weidongxu-microsoft, line 1: @XiaofeiCao please take a look on this part. I remember the check logic could be a bit tricky.

extension-base/src/main/java/com/azure/autorest/extension/base/util/HttpHeaderName.java
weidongxu-microsoft, line 1: I assume we ref this in code mainly to keep the code generated in sync with current HttpHeaderName in azure-core (for name that included there, use var reference; for not included, use `fromString`).

Make it a separate class seems invalidate the intention.

javagen/src/main/java/com/azure/autorest/util/TypeUtil.java
alzimmermsft: I was thinking `GenericType` / `IType` could retain initial information about the type rather than us needing to do this check. Such as tracking `isResponseType` when being created, we know that information at the time of creating the type but it seems we drop is along the way, leading to this logic being needed.

extension-base/src/main/java/com/azure/autorest/extension/base/util/HttpHeaderName.java
alzimmermsft: Agreed this does break some of the reasoning behind the PR, where it changes an explicit reference to an implicit reference. Luckily, `HttpHeaderNames` doesn't commonly add new constants and the way code is generated the lookup is done during initialization of the class and the way `ExpandableStringEnum` is design it will use the same reference as the constant.

javagen/src/main/java/com/azure/autorest/util/TypeUtil.java
XiaofeiCao: The tricky case I can think of  is when `generic-response-type` is set to false. In this case, response type will be ClassType instead of GenericType, so I guess it's fine here.

Though the meaning of this method has changed. Now `isGenericTypeClassSubclassOf(ResponseBase.class, "com.azure.core.http.rest.Response")` will be false, which is quite odd to me.

I'm thinking of renaming this method to `isOneOfTheGivenTypes(GenericType typeToCheck, GenericType... givenTypes)`. This way we can do `isOneOfTheGivenTypes(typeToCheck, GenericType.SyncPoller, GenericType.PollerFlux)`.

fluentgen/src/main/java/com/azure/autorest/fluent/util/FluentUtils.java
```diff
@@ -336,9 +326,10 @@ public static IType getValueTypeFromResponseType(IType clientType) {
         IType bodyType = null;
         if (clientType instanceof GenericType) {
             GenericType type = (GenericType) clientType;
-            if (Response.class.getSimpleName().equals(type.getName())) {
+            if ("Response".equals(type.getName())) {
                 bodyType = type.getTypeArguments()[0];
-            } else if (TypeUtil.isGenericTypeClassSubclassOf(type, Response.class)) {
+            } else if (TypeUtil.isGenericTypeClassSubclassOf(type, "com.azure.core.http.rest.Response",
+                "com.azure.core.http.rest.ResponseBase", "com.azure.core.http.rest.PagedResponse")) {
```
XiaofeiCao, line 332: Not sure if we need `StreamResponse` here.

srnagar:
Do we need to remove these dependencies and duplicate a lot of code that can get outdated if the core libraries change? The generated library still needs released core libraries anyway. Since core ships to Maven Central every month and ships to dev feed almost daily, using the dependency shouldn't be a blocker on the code generator.

weidongxu-microsoft (approved):

alzimmermsft:
> Do we need to remove these dependencies and duplicate a lot of code that can get outdated if the core libraries change? The generated library still needs released core libraries anyway. Since core ships to Maven Central every month and ships to dev feed almost daily, using the dependency shouldn't be a blocker on the code generator.

I agree this ends up duplicating code from `azure-core` but there are some reasons behind that to me.

- This intrinsically ties the code generator to `azure-core`. In recent months though, we've began working and developing `generic-core` which has some drastically different designs. This is pushing the generators into a spot where if these diverge too deeply we'll need to jump through some complex logic to handle that.
- Many of the concepts copied from `azure-core` are tied deeply into Swagger / TypeSpec outputs. To me, it'd be better for the generator to own those types and map them appropriately to `azure-core` / `generic-core` / <insert something in the future here> types.
- Most the copied types don't evolve, and those that do they either happen rarely (HTTP methods) or get handled in a way that doesn't change runtime behaviors of the generated code (calls to `HttpHeaderName.fromString` which will return the same instance as the constant).
- Last, we're reaching a point where we're looking into reducing the shipping cadence of `azure-core`, and related libraries. I don't think shipping cadence one way or another is a big factor for the generator.

All that said, I'll hold this PR in an unmerged state so we can discuss this further.

PR #2630, opened 2024-03-19
Use latest generic core types

srnagar:


javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
```diff
@@ -207,6 +211,21 @@ public static String getTypeReferenceCreation(IType type) {
...
+            }
+            classBlock.privateStaticFinalVariable(String.format("Type %1$s = new ParameterizedType() {"
+                            + "@Override public Type getRawType() { return " + type.getName() + ".class; }"
+                            + "@Override public Type[] getActualTypeArguments() { return new Type[] { " + sb + " }; }"
+                            + "@Override public Type getOwnerType() { return null; } }",
+                    CodeNamer.getEnumMemberName("TypeReference" + type.toJavaPropertyString())));
+            return;
+        }
         classBlock.privateStaticFinalVariable(String.format("TypeReference<%1$s> %2$s = new TypeReference<%1$s>() {}",
             type, CodeNamer.getEnumMemberName("TypeReference" + type.toJavaPropertyString())));
```
weidongxu-microsoft, line 230: Personally prefer a
```
if () {
  return
} else {
  return
}
```
structure.

I assume here for non-branding, we create a `Type` instance inline?

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
```diff
@@ -181,17 +181,21 @@ public static String getLongRunningOperationTypeReferenceExpression(MethodPollin
...
         // or use a singleton instance.
         // Generic types must use a custom instance that supports complex generic parameters.
+        if (!JavaSettings.getInstance().isBranded()) {
+            return (type instanceof ArrayType || type instanceof ClassType || type instanceof EnumType || type instanceof PrimitiveType)
+                    ? type.asNullable() + ".class"
+                    : CodeNamer.getEnumMemberName("TypeReference" + ((GenericType) type).toJavaPropertyString());
+        }
         return (type instanceof ArrayType || type instanceof ClassType || type instanceof EnumType || type instanceof PrimitiveType)
             ? String.format("TypeReference.createInstance(%s.class)", type.asNullable())
             : CodeNamer.getEnumMemberName("TypeReference" + ((GenericType) type).toJavaPropertyString());
```
weidongxu-microsoft, line 201: Same, here. Personally prefer
```java
if () {
  return
} else {
  return
}
```

weidongxu-microsoft:
Is generic-core going to stay?

Do you prefer to have a minimal test for generic flavor (we can use Brian's todo tsp first), to make sure it is not break. (with a minimal one, it should be easy to disable, in the case that either the storage account down, or the emitter and the generic-core not in sync)

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
```diff
@@ -205,15 +205,15 @@ private static String addLogging(LogLevel level, String message) {
...
+                    return String.format("LOGGER.atInfo().log(() -> \"%s\");", message);
                 case WARNING:
-                    return String.format("LOGGER.log(LogLevel.WARNING, () -> \"%s\");", message);
+                    return String.format("LOGGER.atWarning().log(() -> \"%s\");", message);
                 case ERROR:
-                    return String.format("LOGGER.log(LogLevel.ERROR, () -> \"%s\");", message);
+                    return String.format("LOGGER.atError().log(() -> \"%s\");", message);
                 default:
-                    return String.format("LOGGER.log(LogLevel.INFORMATIONAL, () -> \"%s\");", message);
+                    return String.format("LOGGER.atInfo().log(() -> \"%s\");", message);
```
haolingdong-msft, line 216: Will this change the logging output style or just change the way we write logs?

weidongxu-microsoft: This part is unbranded.

Though for branded, we are recommended to output this pattern as well. https://github.com/Azure/autorest.java/issues/2631

srnagar:
> Is generic-core going to stay on that storage account maven?
Do you prefer to have a minimal test for generic flavor (we can use Brian's todo tsp first), to make sure it is not break?
(with a minimal one, it should be easy to disable, in the case that either the storage account down, or the emitter and the generic-core not in sync)

Yes, we should have some tests but let's add it after we rename the package which is happening soon.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ClientBuilderTrait.java
srnagar: > Will this change the logging output style or just change the way we write logs?

Both - but this change is only for unbranded. For branded, we should use this pattern too but we can do that in a separate PR after we review all the current logging calls we generate today.

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
srnagar: For non-branding, we use `ParameterizedType` in `java.lang.reflect` for types with generics. For non-generic types, we can just use the `Model.class` pattern.

weidongxu-microsoft (approved):

Issue #2633, opened 2024-03-21
[MPG] Etag grouping into `MatchConditions` and `RequestConditions`


XiaofeiCao:
DPG has supported it:
https://github.com/Azure/autorest.java/issues/2058

Though for MPG, this will break existing mgmt libs.

Discussion at 21/03:
1. As a first step, we disable Etag grouping for MPG, or possibly with a flag.
2. See if Etag got excessive usage in mgmt tsps. If we gets lots of upcoming new services supporting Etags, we enable this feature for new services.

PR #2638, opened 2024-03-21
Add flag for etag grouping

XiaofeiCao:
https://github.com/Azure/autorest.java/issues/2633

As a first step, add a flag for disabling etag headers grouping.

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -207,8 +207,6 @@ export class CodeModelBuilder {
       this.options["skip-special-headers"].forEach((it) => SPECIAL_HEADER_NAMES.add(it.toLowerCase()));
     }
 
-    this.options["group-etag-headers"] = this.options["group-etag-headers"] ?? true;
-
     this.sdkContext = createSdkContext(context, "@azure-tools/typespec-java");
     const service = listServices(this.program)[0];
     const serviceNamespace = service.type;
@@ -533,6 +531,7 @@ export class CodeModelBuilder {
     for (const client of clients) {
       if (client.arm) {
         this.codeModel.arm = true;
+        this.options["group-etag-headers"] = false;
       }
       const codeModelClient = new CodeModelClient(client.name, this.getDoc(client.type), {
         summary: this.getSummary(client.type),
@@ -640,6 +639,9 @@ export class CodeModelBuilder {
         }
       }
     }
+
+    // postprocess for group-etag-headers
+    this.options["group-etag-headers"] = this.options["group-etag-headers"] ?? true;
 
     return clients;
   }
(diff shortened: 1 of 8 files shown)


typespec-extension/src/code-model-builder.ts
```diff
@@ -781,7 +785,9 @@ export class CodeModelBuilder {
     }
 
     // group ETag header parameters, if exists
-    this.processEtagHeaderParameters(codeModelOperation, op);
+    if (!this.options["disable-etag-grouping"]) {
```
weidongxu-microsoft, line 788: Just check `arm` would be enough for now?

XiaofeiCao: For now it is. Srikanta mentioned that in the future we may want new SDKs enabled and existing ones disabled(avoid breaking). A flag would be helpful in this case. 

Agree a positive one sounds better. I'll go with `group-etag-headers`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -207,6 +207,8 @@ export class CodeModelBuilder {
       this.options["skip-special-headers"].forEach((it) => SPECIAL_HEADER_NAMES.add(it.toLowerCase()));
     }
 
+    this.options["group-etag-headers"] = this.options["group-etag-headers"] ?? true;
```
weidongxu-microsoft, line 210: We do not check `arm`?

typespec-extension/src/emitter.ts
```diff
@@ -42,6 +42,7 @@ export interface EmitterOptions {
   "polling"?: any;
 
   "arm"?: boolean;
```
weidongxu-microsoft, line 44: Not related to this PR.

`arm` is actually not a user settable option, right?
Then it better not be here at all.

If you need set it to Java, just do e.g. `(options as any).arm = codeModel.arm`

typespec-extension/src/code-model-builder.ts
XiaofeiCao: In the long term I think we don't. 
For green fields, before we fix etag headers grouping, we'll have to set this property to false in each tspconfig.yaml. After the fix, we can leave it out for green fields.
For brown fields, we'll always set it false.

Alternatively, as a short-term solution, we could do, if it's ARM and the flag is not set, we set it false. Actually, after a second thought, this one is quite appealing too, as we can selectively enable it for SDKs that really need it.

What's your opinion on this?

weidongxu-microsoft: I am OK either way. Anyway, at present, if tsp has it, arm can only use false, so it just automatically be false, or you need to set it to false.

weidongxu-microsoft (approved):

Issue #2574, opened 2024-02-26
all client, "setKind" cannot be accessed in model of different package


weidongxu-microsoft:
### mgmt (Jackson)

PR https://github.com/Azure/autorest.java/pull/2563 would introduce a getter method on the discriminator property. This going to be shipped in next release.

Please let us know if mgmt is OK with it. Or we want to delay a bit (as this changes API shape of the model)

Also there maybe some issue with the generated test code.
https://github.com/weidongxu-microsoft/autorest.java_ci/actions/runs/8039220851/job/21956004744#step:11:6672

### dpg (azure-core)

https://github.com/Azure/azure-sdk-for-java/pull/39109#discussion_r1515536323

XiaofeiCao:
The root cause is that parent and child class are not in the same package, and child class calls `withKind` which is a package private method. Reason the `withKind` is package private is that parent class is an output-only model, and polymophic discriminator setter method's visibility is set to package private.
https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java#L193-L227

For consumption case, parent class `ReservationRecommendationInner` (inner class) is in `fluent.models` package, while child class `LegacyReservationRecommendation` is in `models` package.

Make `withKind` protected in mgmt could resolve this.

XiaofeiCao:
Meeting notes:

Use [Accessor pattern](https://github.com/Azure/azure-sdk-for-java/wiki/Access-helpers). Set accessor in parent class and use accessor in child class.

A simpler approach is to make setter protected instead of package-private. Though we probably don't want visibility to be bigger than it should be and we'll break existing code when we switch to using accessor pattern.

weidongxu-microsoft:
Accessor pattern would require generate another helper file?

weidongxu-microsoft:
One thing to consider (if we need `setKind`), whether user can add a subclass. If user can, it would also need to be `protected` (as user need to set the "kind").

weidongxu-microsoft:
For Jackson subclass, I can do 

```java
    @JsonTypeId
    @JsonProperty(value = "odata.type", required = true)
    public Odatatype getOdataType() {
        return Odatatype.MICROSOFT_AZURE_MONITOR_SINGLE_RESOURCE_MULTIPLE_METRIC_CRITERIA;
    }
```

To have `getOdataType` behave as type id.

PR #2639, opened 2024-03-25
client, add shadow variable for discriminator property

weidongxu-microsoft:
another fix on https://github.com/Azure/autorest.java/issues/2574

Remove `setKind` method in superclass.
Shadow `kind` variable in subclass.
Override `getKind` method in subclass.

code change https://github.com/Azure/autorest.java/pull/2639/files/e9c1c0dbad94325dacb299fc8f29a1d6c892fd7b..db632bbb53ab4612517996b0298ea42f44824de1

test like this (for both azure-json and Jackson) https://github.com/Azure/autorest.java/pull/2639/commits/e9c1c0dbad94325dacb299fc8f29a1d6c892fd7b

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
+++ b/javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
@@ -136,9 +136,11 @@ public final class ClientModelPropertiesManager {
                 discriminatorProperty = new ClientModelPropertyWithMetadata(model, property, true);
             }
 
-            superPropertyConsumer(property, superRequiredProperties, superConstructorProperties,
-                superReadOnlyProperties, superSetterProperties, settings);
-            hasRequiredProperties |= property.isRequired();
+            if (!property.isPolymorphicDiscriminator()) {
+                superPropertyConsumer(property, superRequiredProperties, superConstructorProperties,
+                    superReadOnlyProperties, superSetterProperties, settings);
+                hasRequiredProperties |= property.isRequired();
+            }
 
             if (property.getNeedsFlatten()) {
                 flattenedProperties.put(property.getName(), new ClientModelPropertyWithMetadata(model, property, true));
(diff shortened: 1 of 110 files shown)


javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
```diff
@@ -136,9 +136,11 @@ public ClientModelPropertiesManager(ClientModel model, JavaSettings settings) {
...
             }
 
-            superPropertyConsumer(property, superRequiredProperties, superConstructorProperties,
-                superReadOnlyProperties, superSetterProperties, settings);
-            hasRequiredProperties |= property.isRequired();
+            if (!property.isPolymorphicDiscriminator()) {
+                superPropertyConsumer(property, superRequiredProperties, superConstructorProperties,
+                    superReadOnlyProperties, superSetterProperties, settings);
+                hasRequiredProperties |= property.isRequired();
+            }
```
weidongxu-microsoft, line 143: Treat polymorphic discriminator property as a property in current model.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -168,23 +168,24 @@ public final void write(ClientModel model, JavaFile javaFile) {
                     ? JavaVisibility.Private
                     : JavaVisibility.Public;
 
-                if (!property.isPolymorphicDiscriminator() || modelDefinesProperty(model, property)) {
```
weidongxu-microsoft, line 171: Here code change is just remove the condition.

typespec-tests/src/main/java/com/cadl/naming/models/BytesData.java
```diff
@@ -17,6 +17,12 @@
  */
 @Immutable
 public final class BytesData extends Data {
+    /*
+     * The @data.kind property.
+     */
+    @Generated
+    private String type = "bytes";
```
weidongxu-microsoft, line 24: My origin attempt is to avoid this `String type` class variable, and directly provide
```java
    @Generated
    @Override
    public String getType() {
        return "bytes";
    }
```
(there is a corner case on behavior difference between the 2, on a case of user directly do `BinaryData("<invalid-bytes-data-json>").toObject(BytesData.class)` with a JSON of unknown type)

But reverted to the shadow variable.
Reason is `fromJson` have code on either `deserializedBytesData.setType(..)` or `deserializedBytesData.type = ..`, and I don't want to disrupt that part of code.

haolingdong-msft (approved):

XiaofeiCao (approved):

typespec-tests/src/main/java/com/cadl/naming/models/BytesData.java
```diff
@@ -30,10 +36,20 @@ public final class BytesData extends Data {
...
     }
 
+    /**
+     * Get the type property: The &#064;data.kind property.
+     * 
+     * @return the type value.
+     */
+    @Generated
+    @Override
+    public String getType() {
```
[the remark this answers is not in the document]
weidongxu-microsoft: https://revapi.org/revapi-java/0.28.1/differences.html#java.method.movedToSuperClass

"Method Moved To Superclass" is "equivalent". So I think this won't be counted as break.

typespec-tests/src/main/java/com/type/model/inheritance/nesteddiscriminator/models/Shark.java
```diff
@@ -52,7 +52,6 @@ public String getSharktype() {
     public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
         jsonWriter.writeStartObject();
         jsonWriter.writeIntField("age", getAge());
-        jsonWriter.writeStringField("kind", getKind());
```
haolingdong-msft, line 55: Just notice serialziation and deserialization part, wonder why this get removed?

PR #2646, opened 2024-03-26
tsp, escape tab in javadoc

weidongxu-microsoft:
Previously this seems done by formatting.


javagen/src/main/java/com/azure/autorest/model/javamodel/JavaJavadocComment.java
```diff
@@ -31,6 +31,8 @@ private static String processText(String value) {
         if (text != null) {
             // escape the "@"
             text = ESCAPE_AT.matcher(text).replaceAll("&#064;");
+            // escape tab
+            text = text.replace("\t", " ");
```
weidongxu-microsoft, line 35: checkstyle does not allow tab in Java code. Replace it with space.

haolingdong-msft (approved):

PR #2652, opened 2024-03-28
Remove setting to disable formatting

alzimmermsft:
Removes the setting that disables code formatting as generated code should always be formatted. This is likely a hold over when code formatting was rapidly changing to prevent too much noise.

Additionally, there is a lot of cleanup of code patterns to reduce `Pattern` usage and `String.format` (when reasonable) to increase performance. And reduction of collecting streams to `List` when an `Iterator` is sufficient.


javagen/src/main/java/com/azure/autorest/mapper/ClientMapper.java
```diff
@@ -489,7 +489,7 @@ private List<XmlSequenceWrapper> parseXmlSequenceWrappers(CodeModel codeModel, J
                 Schema responseBodySchema = SchemaUtil.getLowestCommonParent(operation.getResponses().stream()
                     .map(Response::getSchema)
                     .filter(Objects::nonNull)
-                    .collect(Collectors.toList()));
+                        .iterator());
```
srnagar, line 492: nit: fix formatting. The indentation on the last line seems to be a bit off.

srnagar (approved):

weidongxu-microsoft:
This flag was mainly for skip formatting, when generated code cannot pass formatting (sometimes due to incorrect option, e.g. invalid namespace; sometimes due to bug in codegen).

Issue #2644, opened 2024-03-26
[MPG] versioning for @added mixed with @removed


XiaofeiCao:
```typescript
@added(Versions.v2022_09_02_preview)
@added(Versions.v2023_03_15_preview)
@added(Versions.v2023_06_15_preview)
@added(Versions.v2023_08_15_preview)
@removed(Versions.v2023_10_15)
@added(Versions.v2024_02_02_preview)
hubProfile?: FleetHubProfile;
```

https://github.com/Azure/azure-rest-api-specs/blob/80be470f83f16d693291beb988ee80cd9a0b8023/specification/containerservice/Fleet.Management/fleet.tsp#L50-L58

`TCGCContext` will likely contain a flag to only include operations in latest version: https://github.com/Azure/typespec-azure/issues/129#issuecomment-2011627278

TCGC `getAllOperations` will return operations/models with all the available versions.

Before TCGC supports it and we integrate `getAllOperations`, we could see if we can leverage [existsAtVersion](https://github.com/microsoft/typespec/blob/baac24ac0061910339304ae629fd9317a4e1f03e/packages/versioning/src/versioning.ts#L849) from versioning, using the default(latest) version.

weidongxu-microsoft:
One suggestion is to have a e.g. "api-version" var in class (default to latest).

And codegen (in default mode -- not handling breaking changes caused by non-REST-API-break) would generate API (it may include parameter and model property) that is available in this "api-version".

PR #2653, opened 2024-03-29
all client, support basic versioning for Operation, ModelProperty and Parameter

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2644

code in https://github.com/Azure/autorest.java/pull/2653/commits/01536598dde7316759619016dc3076c486d070cb

### 1. On Operation
```typescript
@removed(ApiVersions.v2022_12_01_preview)
create is ResourceOperations.ResourceCreateOrReplace<Resource>;
```

### 2. On ModelProperty
```typescript
@added(Versions.v2022_06_01_preview)
@removed(Versions.v2023_11_01)
@added(Versions.v2023_12_01_preview)
propertyRemovedInStable: string;
```

### 3. On Parameter
```typescript
model ListQueryParams {
  @removed(ApiVersions.v2022_12_01_preview)
  @query
  expand?: string;
}

op list is ResourceOperations.ResourceList<Resource, ListQueryParametersTrait<ListQueryParams>>;
```


typespec-tests/tsp/arm.tsp
```diff
@@ -139,7 +177,16 @@ interface CustomTemplateResourceInterface
...
     > {
+  @sharedRoute
+  @added(Versions.v2022_09_01_preview)
+  @removed(Versions.v2023_11_01)
   update is ArmResourceUpdateSync<CustomTemplateResource>;
+
+  #suppress "@azure-tools/typespec-azure-core/no-operation-id" "Use operationId to keep same name as deprecated update operation"
+  @sharedRoute
+  @added(Versions.v2023_11_01)
+  @operationId("CustomTemplateResourceInterfaces_Update")
```
XiaofeiCao, line 188: example: https://github.com/Azure/azure-rest-api-specs/blob/d53388b8b7574dfe861a9e6863bd4d137168cbe5/specification/containerservice/Fleet.Management/fleetmember.tsp#L100

This is not supported yet. Will fix in following PR.

typespec-tests/tspconfig.yaml
```diff
@@ -10,5 +10,4 @@ options:
     flavor: "Azure"
     dev-options:
       generate-code-model: true
-      support-versioning: true
```
weidongxu-microsoft, line 13: I think you need to set this flag in script, for resiliency/servicedriven test -- they are for the advanced case to solve SDK API breaks (I am not sure whether tsp cmd can set the flag under a object, if not we can move this one level above).

typespec-extension/src/code-model-builder.ts
```diff
@@ -667,7 +672,28 @@ export class CodeModelBuilder {
       this.trace(`Operation '${operation.name}' is temporary skipped, as it is an overloaded operation`);
       return true;
     }
-    return false;
+    return !this.existsAtCurrentVersion(operation) || false;
```
weidongxu-microsoft, line 675: Seems easier to understand by putting it at top, e.g.

```
if (!this.existsAtCurrentVersion(operation)) {
  return true;
}
```

typespec-tests/tsp/arm.tsp
weidongxu-microsoft: I don't know whether we should honor `@operationId`. Likely not, for greenfield.

typespec-tests/tspconfig.yaml
XiaofeiCao: Yes, it can. Updated.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: fixed

typespec-tests/tspconfig.yaml
weidongxu-microsoft: It may be easier to disable it in tspconfig, and only enable on that particular test. Up to you.

XiaofeiCao: OK. If we are using basic versioning for default mode, this is better.

weidongxu-microsoft: Yeah, this is in dev-options, so it definitely not a default.

typespec-tests/Generate.ps1
```diff
@@ -38,11 +38,18 @@ $generateScript = {
   } elseif ($tspFile -match "resiliency[\\/]srv-driven[\\/]old.tsp") {
     # override namespace for "resiliency/srv-driven/old.tsp" (make it different to that from "main.tsp")
     $tspOptions += " --option ""@azure-tools/typespec-java.namespace=com.resiliency.servicedriven.v1"""
+    # enable advanced versioning for resiliency test
+    $tspOptions += " --option ""@azure-tools/typespec-java.dev-options.support-versioning=false"""
+  } elseif ($tspFile -match "resiliency[\\/]srv-driven[\\/]main.tsp") {
+    # enable advanced versioning for resiliency test
+    $tspOptions += " --option ""@azure-tools/typespec-java.dev-options.support-versioning=false"""
```
weidongxu-microsoft, line 45: =true ?

XiaofeiCao: My bad.. Fixed.

typespec-tests/tspconfig.yaml
XiaofeiCao: Emmm, seems I can't get it to work under an object... Maybe previously I got the test wrong(misconfigured tspconfig.yaml). Let me put it to the root level.

XiaofeiCao: moved, renamed to `advanced-versioning` to distinguish between basic versioning.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2247,7 +2277,8 @@ export class CodeModelBuilder {
       if (
         prop.name === discriminatorPropertyName || // skip the discriminator property
         isNeverType(prop.type) || // skip property of type "never"
-        !isPayloadProperty(this.program, prop)
+        !isPayloadProperty(this.program, prop) ||
+        !this.existsAtCurrentVersion(prop)
```
weidongxu-microsoft, line 2281: @haolingdong-msft for awareness. This may conflict with your Models.

weidongxu-microsoft (approved):

haolingdong-msft:
double confirm my understanding: current support is to generate removed operations, which means even user switch to an old service version which contains a removed operation, they can still not call the newly removed operation?

weidongxu-microsoft:
Yes.

For mgmt, this is same behavior of current Swagger, and there is no "switch to an old service version".

For data-plane, we don't plan to address the "remove" in near future. Service should not "remove" anything that GAed.
For the preview api-versions before API, current guideline is to delete them in typespec (as preview api-version should only live for about 3 months in data-plane).

Issue #2660, opened 2024-04-03
tsp, TypeSpec enum means closed enum


weidongxu-microsoft:
Last task of union as enum. Now tsp `enum` means fixed enum -- Java `enum`.

Remove handling of `@fixed`.

PR #2661, opened 2024-04-03
tsp, enum means closed enum

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2660

blocked by https://github.com/Azure/typespec-azure/issues/584


typespec-extension/src/code-model-builder.ts
```diff
@@ -1707,7 +1706,7 @@ export class CodeModelBuilder {
         return this.processConstantSchemaForLiteral(type, nameHint);
 
       case "Enum":
-        return this.processChoiceSchema(type, this.getName(type), isFixed(this.program, type));
+        return this.processChoiceSchema(type, this.getName(type), true);
```
weidongxu-microsoft, line 1709: @haolingdong-msft 

I think Chenjie said latest TCGC would have this same interpretation, that TypeSpec `enum` is fixed enum.

weidongxu-microsoft:
to https://github.com/Azure/autorest.java/pull/2752

PR #2679, opened 2024-04-09
tsp, unbranded, temporary remove generic-core as explicit dependency in POM

weidongxu-microsoft:
It seems causing some build failure in preview build https://github.com/Azure/autorest.java/runs/23553740795

And we probably don't want to load jar from places not maven.

pick a few from https://github.com/Azure/autorest.java/pull/2626

### DIFF
--- a/javagen/pom.xml
+++ b/javagen/pom.xml
@@ -17,12 +17,12 @@
   <artifactId>azure-autorest-javagen</artifactId>
   <version>1.0.0-beta.1</version>
 
-  <repositories>
-    <repository>
-      <id>clientcore</id>
-      <url>https://clientcore.blob.core.windows.net/artifacts</url>
-    </repository>
-  </repositories>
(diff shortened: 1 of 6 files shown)


javagen/pom.xml
```diff
@@ -17,12 +17,12 @@
...
-      <id>clientcore</id>
-      <url>https://clientcore.blob.core.windows.net/artifacts</url>
-    </repository>
-  </repositories>
+<!--  <repositories>-->
+<!--    <repository>-->
+<!--      <id>clientcore</id>-->
+<!--      <url>https://clientcore.blob.core.windows.net/artifacts</url>-->
+<!--    </repository>-->
+<!--  </repositories>-->
```
srnagar, line 25: Let's remove this completely instead of comment it. This is no longer required.

srnagar (approved):

PR #2681, opened 2024-04-10
tsp, fix naming on anonymous

weidongxu-microsoft:
when anonymous model used in Union

cause https://github.com/joseharriaga/openai-in-typespec

Note that TCGC 0.41.0 has this bug when run on above https://github.com/Azure/typespec-azure/issues/627
0.42.0-dev.6 fixed, but still wait for a stable release


typespec-extension/src/type-utils.ts
```diff
@@ -217,6 +217,11 @@ export function getModelNameForProperty(property: ModelProperty): string {
             (candidateModel === property.model || candidateModel.indexer?.value === property.model)
           ) {
             return getModelNameForProperty(prop);
+          } else if (
+            candidateModel.kind === "Union" &&
```
weidongxu-microsoft, line 221: We had simple naming logic, but previously it does not work for Union.

typespec-tests/tsp/naming.tsp
```diff
@@ -34,6 +34,21 @@ model Response {
...
   @doc("description of status property")
   status: StatusModel;
+
+  anonymous: AnonymouseModel;
+}
+
+model AnonymouseModel {
+  last_error: {
+    code: "server_error" | "rate_limit_exceeded" | "invalid_prompt";
+  } | null;
```
weidongxu-microsoft, line 44: If it is 
```
last_error: {
  code: "server_error" | "rate_limit_exceeded" | "invalid_prompt";
}
```
existing emitter works. But not when it be `| null`.

typespec-tests/tsp/naming.tsp
```diff
@@ -34,6 +34,21 @@ model Response {
   @summary("summary of status property")
   @doc("description of status property")
   status: StatusModel;
+
+  anonymous: AnonymouseModel;
+}
+
+model AnonymouseModel {
+  last_error: {
+    code: "server_error" | "rate_limit_exceeded" | "invalid_prompt";
```
weidongxu-microsoft, line 43: I think .NET currently would name it `AnonymouseModelLastErrorCode`

Wondering what TCGC would name it @haolingdong-msft (try it if you got time)

haolingdong-msft: `model AnonymousModel` -> the model name is a bit strange to me, because anonymous model does not have name. You are referring below part being anonymous, right? I will test what name TCGC would return.
```
last_error: {
  code: "server_error" | "rate_limit_exceeded" | "invalid_prompt";
} | null
```

How about `model ModelContainsAnonymousModel`?

weidongxu-microsoft: It just a test.

Here I am checking what TCGC would do, if this is the case.

weidongxu-microsoft: I renamed it to `RunObject` as the real case.

weidongxu-microsoft: I made a small change.

Now the name would be `RunObjectLastErrorCode`, that should be consistent to .NET and TCGC.

The final solution is still the TCGC model.

haolingdong-msft (approved):

typespec-tests/tsp/naming.tsp
haolingdong-msft: I tried TCGC getAllModels using the definition:
```ts
model RunObject {
  last_error: {
    code: "server_error" | "rate_limit_exceeded" | "invalid_prompt";
  } | null;
}
```

The name looks good:
image

PR #2682, opened 2024-04-10
client, no serialization on GroupSchema

weidongxu-microsoft:
`GroupSchema` is used to group method parameters. It does not correspond to a wire JSON model, and hence should not implement `JsonSerializable`.

code https://github.com/Azure/autorest.java/pull/2682/commits/ada2f3322167e4334af73d60593fc54e44429a5b

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
@@ -13,6 +13,7 @@ import com.azure.autorest.model.clientmodel.ClientModelPropertyAccess;
 import com.azure.autorest.model.clientmodel.ClientModelPropertyReference;
 import com.azure.autorest.model.clientmodel.GenericType;
 import com.azure.autorest.model.clientmodel.IType;
+import com.azure.autorest.model.clientmodel.ImplementationDetails;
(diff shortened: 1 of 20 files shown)


javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -1166,7 +1167,9 @@ private static boolean isGenerateConstantEmptyByteArray(ClientModel model, JavaS
      */
     private static boolean modelRequireSerialization(ClientModel model) {
         // TODO (weidxu): any other case? "binary"?
-        return !ClientModelUtil.isMultipartModel(model);
+        return !ClientModelUtil.isMultipartModel(model)
+                // not GroupSchema
+                && !(model.getImplementationDetails() != null && model.getImplementationDetails().getUsages() != null && model.getImplementationDetails().getUsages().contains(ImplementationDetails.Usage.OPTIONS_GROUP));
```
weidongxu-microsoft, line 1172: currently, 2 condition for a model class to have no serizliation
1. multipart/form-data (and related `##FileDetails` model), as the serialization is via a Helper class.
2. this GroupSchema (here we call it OPTIONS_GROUP, welcome suggestion on naming)

XiaofeiCao (approved):

PR #2684, opened 2024-04-10
Migrate to client core

srnagar:


weidongxu-microsoft (approved):

weidongxu-microsoft:
CI fail due to `typespec-tests/src/main/resources/META-INF/apiview_properties.json`

Maybe you can add it to .gitignore?

typespec-tests/src/main/java/com/cadl/naming/NamingAsyncClient.java
haolingdong-msft, line 1: why do we remove these tests?

weidongxu-microsoft: Seems codegen isssue?

I didn't see any remove in tsp or script.

haolingdong-msft: Yeah, I also checked the `Generate.ps1` script, nothing changed. But the tests are getting removed. Maybe there is some bugs in codegen.

srnagar: Oh this was because the customization failed for these 3 specs. I have regenerated them and pushed an update.

PR #2685, opened 2024-04-10
Surround manual code with Eclipse formatter off in partial update

alzimmermsft:
Surround code that was manually written into a class supporting partial update with the comment tags that disable Eclipse code formatting. This ensures that the code format styles applied to generated code does not modify handwritten code, especially documentation.


weidongxu-microsoft (approved):
@haolingdong-msft for a quick review as well if you have time.

`// @formatter:on` `// @formatter:off` should be from eclipse format. Our job here is just keep this 2 comment intact in partial-update (that involve the java-parser)

haolingdong-msft:
So I guess after we support partial-update, code would be like this.
```
    // @formatter:off
    /**
     * <generated-javadoc-description-start>
     * javadoc
     * <generated-javadoc-description-end>
     *
     * <generated-javadoc-tags-start>
     * @param testProperty javadoc
     * <generated-javadoc-tags-end>
     */
    public void setTestProperty(String testProperty) {
        this.testProperty = testProperty;
    }
    }// @formatter:on
```

weidongxu-microsoft:
> So I guess after we support partial-update on javadoc, code would be like this.
> 
> ```
>     // @formatter:off
>     /**
>      * <generated-javadoc-description-start>
>      * javadoc
>      * <generated-javadoc-description-end>
>      *
>      * <generated-javadoc-tags-start>
>      * @param testProperty javadoc
>      * <generated-javadoc-tags-end>
>      */
>     public void setTestProperty(String testProperty) {
>         this.testProperty = testProperty;
>     }
>     }// @formatter:on
> ```

Yours does not have `@Generated`, so no generated-javadoc

The `@format` here is only about formatting. Whether the code is generated or handwritten, at present formatter would format them all.

PR #2692, opened 2024-04-15
Fix tsp, update cadl-ranch 0.31.8

v-hongli1:
Fixed tsp, update cadl-ranch 0.31.8


weidongxu-microsoft:
your action run? result?

weidongxu-microsoft (approved):

weidongxu-microsoft:
merge it if CI pass

typespec-tests/src/test/java/com/type/property/additionalproperties/ExtendsTests.java
```diff
@@ -65,13 +65,16 @@ public void testExtendsModelArrayClient() {
     public void testExtendsModelClient() {
         ExtendsModelAdditionalProperties body = new ExtendsModelAdditionalProperties();
         Map<String, ModelForRecord> propertyMap = new LinkedHashMap<>();
+        propertyMap.put("knownProp", new ModelForRecord("ok"));
```
weidongxu-microsoft, line 68: "knownProp" should not be set via additionalPorperties, and should not be retrieved from additionalPorperties

If it cannot be set/get via model, disable the test, mark it as a bug.

v-hongli1: Disabled this test case, and `IsModelArrayClientTest#testPullAndGet` and `IsModelClientTest#testPullAndGet` same as this.

typespec-tests/src/test/java/com/type/property/additionalproperties/IsModelClientTest.java
```diff
@@ -3,25 +3,20 @@
...
     public void testPullAndGet() {
-        Map<String, ModelForRecord> properties = new LinkedHashMap<>();
-        properties.put("prop", new ModelForRecord("ok"));
-        client.put(properties);
-
-        properties = client.get();
-        Assertions.assertNotNull(properties);
-        Assertions.assertNotNull(properties.get("prop"));
-        Assertions.assertEquals("ok", properties.get("prop").getState());
+//        Map<String, ModelForRecord> properties = new LinkedHashMap<>();
```
weidongxu-microsoft, line 13: Why comment out?

v-hongli1: Please following as https://github.com/Azure/autorest.java/pull/2692/files#diff-ce230670765308677adff8ade0554d3d5f2935e4ad02368843af75c2150d2875R127

weidongxu-microsoft (approved):

typespec-tests/src/test/java/com/type/property/additionalproperties/IsModelClientTest.java
weidongxu-microsoft: OK. So no test disabled or commented?

haolingdong-msft (approved):

Issue #2564, opened 2024-02-21
dpg, integrate TCGC getAllModels - basic support


haolingdong-msft:
TCGC `getAllModels()` code is already on main. We will need to `getAllModels()` in emitter to replace the logics in `processModels()`.

```[tasklist]
### Tasks
- [x] Check the type mapping from compiler Type to TCGC SDKType
- [x] Replace logics for `intrinsic` types. e.g. `unkonwn`, `void`, `never`, `any`
- [x] Replace logics for `enum` type, pay attention to the case on using union express extensible enum for unbranding
- [x] Replace logics for `union` type
- [x] Replace logics for `Record` type
- [x] Replace logics for `constant` type
- [x] Replace logics for `Array` type, pay attention to `byte` array
- [x] Replace logics for `Scalar` type, e.g. `number`, `decimal`, `string`, `boolean`, `bytes` etc. pay attention to format
- [x] Replace logics for `Model` type, basic cases
- [x] Replace logics for `Model` type, discriminator and inheritance' cases
- [x] Replace logics for `Model` type, additional properties
- [x] Replace logics for `Model` type, multipart form data cases
- [x] Pass all tests cases in typespec-tests/tsp and cadl-ranch
- [ ] https://github.com/Azure/autorest.java/issues/2728
- [ ] https://github.com/Azure/autorest.java/issues/2729
- [ ] https://github.com/Azure/autorest.java/issues/2730
- [ ] https://github.com/Azure/autorest.java/issues/2731
- [ ] https://github.com/Azure/autorest.java/issues/2732
```

Possible gaps between TCGC type and compiler type:
1. `unkown` type in compiler will be converted to `any` type in TCGC `SdkType`.
2. `void` and `never` is not supported in TCGC `SdkType` yet. TCGC issue: https://github.com/Azure/typespec-azure/issues/292
8. `SdkType` does not contain `namespace` information. https://github.com/Azure/typespec-azure/issues/290
9.  `SdkType.description` and `SdkType.detail`: If both `@summary` and `@doc` exist, .`description` will be the content in `@summary` and `detail` will be the content in `@doc`. If only `@doc` exists, then `description` will be the content in `@doc`.
10. TCGC types does not have concept of `@format`, it will map the type with `@format` to an `SdkBuiltInType`. (expected)
11. Optional literal do not have correct generated name. https://github.com/Azure/typespec-azure/issues/553
[...]

PR #2698, opened 2024-04-16
 tsp, TCGC common layer, adopt model types

haolingdong-msft:
API view with diff comparing this PR's generated code for `typespec-tests`  with `main` branch:
https://apiview.dev/Assemblies/Review/e0cf256f814e456bac0f0d751bd4aa9f/d08e9493669445c4bf9916c090262bd5?diffRevisionId=b988ab5926314dc0a7d226750d035115&doc=False&diffOnly=True

Note this: 
Currently TCGC treated `enum` as fixed enum, and this can cause breaking changes in mgmt plane sdk. typespec-azure will fix the tsp definition in 0.42.
image

Fix https://github.com/Azure/autorest.java/issues/2564

Will create issues for below follow up items and create seperate PR to support:
- use TCGC description: https://github.com/Azure/autorest.java/issues/2732
- use TCGC returned name as much as possible: optional literal case, union case: https://github.com/Azure/autorest.java/issues/2731
- remove nullable from type, pay attention to union value nullable case: `A | B | null`: https://github.com/Azure/autorest.java/issues/2730
- use TCGC's access: https://github.com/Azure/autorest.java/issues/2728
- use TCGC's usage: https://github.com/Azure/autorest.java/issues/2729


weidongxu-microsoft:
Chenjie mentioned for the new additionalProperties tests, it would need dev version of TCGC.

Also is there some code we can now delete in code-model-builder tsp? It seems have more lines than before (TCGC supposed to be able to reduce the code).

haolingdong-msft:
I will use latest dev version to regen and test. 
I am also wip on cleaning up the codes, there are codes can be deleted.

weidongxu-microsoft:
There appears to be a bug on [DifferentSpreadStringDerived](https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/java/com/type/property/additionalproperties/models/DifferentSpreadStringDerived.java), which should be a subclass of DifferentSpreadStringRecord, but not so in generated code.
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp#L389-L394

(would take a look on whether the cause is in typescript code or Java code)

If TCGC have change on that, this should be good.

typespec-extension/src/type-utils.ts
```diff
@@ -340,3 +356,27 @@ function getDecoratorScopedValue<T>(
...
+    "uint64",
+  ].includes(kind);
+}
+
+export function isSdkFloatKind(kind: string): boolean {
+  return ["float", "float32", "float64", "decimal", "decimal128"].includes(kind);
+}
+
+export function isSdkStringKind(kind: string): boolean {
+  return ["string", "password", "guid", "url", "uuid", "eTag", "armId", "ipAddress", "azureLocation"].includes(kind);
```
weidongxu-microsoft, line 381: Does this mean every time TCGC add a string type, you need to modify this list? And how do you know they do add one?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1022,9 +1032,10 @@ export class CodeModelBuilder {
         (hasScalarAsBase(param.param.type, "utcDateTime") || hasScalarAsBase(param.param.type, "offsetDateTime"))
       ) {
         // utcDateTime in header maps to rfc7231
-        schema = this.processDateTimeSchema(param.param.type, param.param.name, true);
+        schema = this.processDateTimeSchemaFromSdkType(sdkType as SdkDatetimeType, param.param.name, true);
```
weidongxu-microsoft, line 1035: Why use `as`? What if this is not SdkDatetimeType?

Do not use `as` unless there is no other way. Here a check and fallback should work fine.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1689,275 +1707,250 @@ export class CodeModelBuilder {
...
+      case "string":
+      case "password":
+      case "guid":
+      case "url":
+      case "uuid":
+      case "eTag":
+      case "armId":
+      case "ipAddress":
+      case "azureLocation":
+        return this.processBuiltInFromSdkType(type, nameHint);
```
weidongxu-microsoft, line 1776: Same problem, how do you know when TCGC add a type?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1689,275 +1707,250 @@ export class CodeModelBuilder {
...
+
+      case "utcDateTime":
+      case "offsetDateTime":
+        if ((type as SdkDatetimeType).encode === "unixTimestamp") {
```
weidongxu-microsoft, line 1780: Again, why `as`?

Won't "case" already handle the type?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1689,275 +1707,250 @@ export class CodeModelBuilder {
...
+      case "int16":
+      case "uint16":
+      case "int32":
+      case "uint32":
+      case "int64":
+      case "uint64": {
+        // integer
+        const integerSize = type.kind === "safeint" || type.kind.includes("int64") ? 64 : 32;
+        return this.processIntegerSchemaFromSdkType(type, nameHint, integerSize);
+      }
```
weidongxu-microsoft, line 1822: same issue, and you are duplicating these as type-utils.ts

typespec-extension/src/code-model-builder.ts
```diff
@@ -2347,158 +2209,54 @@ export class CodeModelBuilder {
...
-    }
-    return schema;
-  }
-
-  private processModelProperty(prop: ModelProperty): Property {
-    const schema = this.processSchema(prop, prop.name);
-    let nullable = isNullableType(prop.type);
+  private processModelPropertyFromSdkType(prop: SdkModelPropertyType): Property {
+    // TODO: question: why the schema name is like this? This case is related with literal.tsp
+    const schemaNameHint = pascalCase(getNamePrefixForProperty(prop.__raw as ModelProperty)) + pascalCase(prop.name);
```
weidongxu-microsoft, line 2214: Not sure if you still need this `getNamePrefixForProperty`.

Or is it for the optional Constant case?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: BTW, in processDateTimeSchemaFromSdkType you actually didn't do anything about SdkDatetimeType

typespec-extension/src/code-model-builder.ts
```diff
@@ -1689,275 +1707,250 @@ export class CodeModelBuilder {
...
-      typeof type.members.values().next().value.value === "number"
-        ? isAllValueInteger(Array.from(type.members.values()).map((it) => it.value as number))
-          ? this.integerSchema
-          : this.doubleSchema
+    const namespace = getNamespace(type.__raw as Enum);
+    const valueType = isSdkIntKind(type.valueType.kind)
+      ? this.integerSchema
+      : isSdkFloatKind(type.valueType.kind)
+        ? this.doubleSchema
         : this.stringSchema;
```
weidongxu-microsoft, line 1943: Can we just do `processSchema` here, since TCGC already gives the type of the element?

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yes, I first removed it, but found issue in optional literal case. So add it back.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L1980-L1987
I was refering to our existing code on this logic. 
Good suggestion. Let me try to use `processSchema()` and test.

typespec-extension/src/type-utils.ts
```diff
@@ -340,3 +356,23 @@ function getDecoratorScopedValue<T>(
   }
   return undefined;
 }
+
+export function isSdkIntKind(kind: string): boolean {
```
haolingdong-msft, line 360: TCGC exports two helpers: `isSdkIntKind` and `isSdkFloatKind`, will use theirs after they release.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1696,275 +1715,201 @@ export class CodeModelBuilder {
...
-          return this.processAnySchema(type, nameHint);
-        } else {
-          throw new Error(`Unrecognized intrinsic type: '${type.name}'.`);
-        }
-
-      case "String":
-        return this.processConstantSchemaForLiteral(type, nameHint);
+  private processSchemaFromSdkTypeImpl(type: SdkType, nameHint: string): Schema {
+    if (isSdkBuiltInKind(type.kind)) {
+      return this.processBuiltInFromSdkType(type as SdkBuiltInType, nameHint);
```
haolingdong-msft, line 1724: `as SdkBuiltInType` is required here, otherwise there will be compile error. I checked the TCGC code and confirmed with Chenjie, `SdkBuildInKind` contains and only contains `SdkBuiltInType`'s `kind`, so `as` is safe here. But we both don't understand why TS compiler can't do type narrow down here. Please let me know if you have any idea.

weidongxu-microsoft: TCGC should narrow the type, e.g.

```ts
function isSdkBuiltInKind(type: SdkType): type is SdkBuiltInType {
  ...
}
```

typespec-extension/src/type-utils.ts
haolingdong-msft: updated to use TCGC's helper function now.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yeah, I guess so, but in `else` block, the type is already narrowed down. The screenshot is when I hover over line 1702. So seems the compiler can interprete `if` block's type, but can't do type narrow down. 
image

typespec-extension/src/code-model-builder.ts
```diff
@@ -931,7 +915,9 @@ export class CodeModelBuilder {
       ) {
         const finalResult = useNewPollStrategy ? lroMetadata.finalResult : lroMetadata.finalEnvelopeResult;
         const finalType = this.findResponseBody(finalResult);
-        finalSchema = this.processSchema(finalType, "finalResult");
+        // change to sdk type
```
weidongxu-microsoft, line 918: nit, I don't think you need this comment. Same below.

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: I mean TCGC `isSdkBuiltInKind` method should do the `if` and "narrow down" when it is true. This is feature in TypeScript.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2354,158 +2139,54 @@ export class CodeModelBuilder {
...
-        }
-      }
-    }
-    return schema;
-  }
-
-  private processModelProperty(prop: ModelProperty): Property {
-    const schema = this.processSchema(prop, prop.name);
-    let nullable = isNullableType(prop.type);
+  private processModelPropertyFromSdkType(prop: SdkModelPropertyType): Property {
```
weidongxu-microsoft, line 2142: The process of FileDetails property should be handled when processing model property. (instead of existing code of postprocess when ObjectSchema already created)
https://github.com/Azure/autorest.java/blob/e0c3cc38151f2cce8711a596ce5e91ad9c0764ee/typespec-extension/src/code-model-builder.ts#L2270-L2296

I think TCGC is `isMultipartFileInput` in https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/interfaces.ts#L369)

Please double check the multipart/form-data flow.

typespec-extension/src/type-utils.ts
```diff
@@ -204,7 +220,7 @@ export function getUnionDescription(union: Union, typeNameOptions: TypeNameOptio
 }
 
 export function getNamePrefixForProperty(property: ModelProperty): string {
-  if (property.model) {
+  if (property && property.model) {
```
weidongxu-microsoft, line 223: In which case the property would be undefined?

If it can be undefined, would the method signature be `property: ModelProperty | undefined`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -2354,158 +2139,54 @@ export class CodeModelBuilder {
...
-    if (unionEnum) {
-      // enum
-      return this.processChoiceSchemaForUnion(type, unionEnum, name);
-    }
-
+  private processUnionSchemaFromSdkType(type: SdkUnionType, name: string): Schema {
     // TODO: name from typespec-client-generator-core
-    const namespace = getNamespace(type);
-    const baseName = pascalCase(name) + "Model";
+    const namespace = getNamespace(type.__raw as Union);
```
weidongxu-microsoft, line 2173: If you had to do a `as`, do it once. Not like this in every call of __raw.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1696,275 +1692,200 @@ export class CodeModelBuilder {
...
     return this.anySchema;
   }
 
-  private processStringSchema(type: Scalar, name: string): StringSchema {
+  private processStringSchemaFromSdkType(type: SdkBuiltInType, name: string): StringSchema {
     return this.codeModel.schemas.add(
-      new StringSchema(name, this.getDoc(type), {
-        summary: this.getSummary(type),
+      new StringSchema(name, this.getDoc(type.__raw as Type), {
+        summary: this.getSummary(type.__raw as Type),
```
weidongxu-microsoft, line 1798: Why do you need these `as`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -2354,158 +2139,54 @@ export class CodeModelBuilder {
...
-    }
-    return schema;
-  }
-
-  private processModelProperty(prop: ModelProperty): Property {
-    const schema = this.processSchema(prop, prop.name);
-    let nullable = isNullableType(prop.type);
+  private processModelPropertyFromSdkType(prop: SdkModelPropertyType): Property {
+    // TODO: This case is related with literal.tsp, once TCGC supports giving a name, we can use TCGC generatedName
+    const schemaNameHint = pascalCase(getNamePrefixForProperty(prop.__raw as ModelProperty)) + pascalCase(prop.name);
```
weidongxu-microsoft, line 2144: Same here. `as` only once in func.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: type.__raw is `Type | undefined`. Maybe better change the getDoc() and getSummary()  method's signature to accept `Type | undefined`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1021,17 +1007,20 @@ export class CodeModelBuilder {
...
       // schema
       let schema;
+      const sdkType = getClientType(this.sdkContext, param.param);
       if (
         param.type === "header" &&
         param.param.type.kind === "Scalar" &&
         getEncode(this.program, param.param) === undefined &&
         getEncode(this.program, param.param.type) === undefined &&
-        (hasScalarAsBase(param.param.type, "utcDateTime") || hasScalarAsBase(param.param.type, "offsetDateTime"))
+        (hasScalarAsBase(param.param.type, "utcDateTime") || hasScalarAsBase(param.param.type, "offsetDateTime")) &&
```
XiaofeiCao, line 1016: nit, can this be `sdkType.type`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1523,7 +1516,9 @@ export class CodeModelBuilder {
       for (const response of resp.responses.values()) {
         if (response.headers) {
           for (const [key, header] of Object.entries(response.headers)) {
-            const schema = this.processSchema(header, key);
+            // TODO: change to sdk type
```
XiaofeiCao, line 1519: Is this TODO still needed?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: On the case that you have to do e.g. `__raw as Union`, you may want to do an assert first e.g. `assert __raw.kind === "Union"; Union type = __raw as Union;`.

typespec-extension/src/type-utils.ts
haolingdong-msft: This is for cases for TCGC discriminator property, because TCGC discriminator property can have empty raw type.
And this method is called from `processModelPropertyFromSdkType()`

I think it's better to change signature to be `property: ModelProperty | undefined`.

image

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Thanks,  I already deleted locally, will commit and push.

typespec-extension/src/type-utils.ts
weidongxu-microsoft: Yes.

Also make sure you do the correct `as` on the `__raw`. I guess it had to be `__raw as ModelProperty | undefined`

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yes, it should be removed. This part is related with operation, so my original plan is to remove those TSP related types in processing operation after switch to `SdkPackage` (getAllOperation)

typespec-extension/src/code-model-builder.ts
```diff
@@ -2228,81 +2037,44 @@ export class CodeModelBuilder {
...
+        },
+        description: type.description,
+        nullableValues: false,
+        nullable: false,
+        valueType: type.additionalProperties,
+      };
+      const parentSchema = this.processDictionarySchemaFromSdkType(sdkDictType, "Record");
       objectSchema.parents = new Relations();
       objectSchema.parents.immediate.push(parentSchema);
       pushDistinct(objectSchema.parents.all, parentSchema);
```
weidongxu-microsoft, line 2060: What if `objectSchema.parents` already have value here?

I think it be possible to have 
```ts
model Child extends Parent {
  ...Record<unknown>;
}
```

`Parent` should be processed from `type.baseModel`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2228,81 +2037,44 @@ export class CodeModelBuilder {
...
-    for (const prop of type.properties.values()) {
+    for (const prop of type.properties) {
+      // skip discriminator property
       if (
-        prop.name === discriminatorPropertyName || // skip the discriminator property
-        isNeverType(prop.type) || // skip property of type "never"
-        !isPayloadProperty(this.program, prop)
+        prop.name === type.discriminatorProperty?.name ||
+        prop.name === type.baseModel?.discriminatorProperty?.name ||
+        !isPayloadProperty(this.program, prop.__raw) // why have this check?
```
weidongxu-microsoft, line 2070: Can we use this https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/interfaces.ts#L367 ?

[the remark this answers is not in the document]
weidongxu-microsoft: answer the question. tsp Type could include `@header` `@query` `@path` property.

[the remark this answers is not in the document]
haolingdong-msft: Would like to check if we can remove this check now.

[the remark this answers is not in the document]
weidongxu-microsoft: see https://github.com/Azure/autorest.java/pull/2698#discussion_r1582741965

We should use TCGC info, if available.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2689,13 +2368,13 @@ export class CodeModelBuilder {
     return getWireName(this.sdkContext, target);
   }
 
-  private isReadOnly(target: ModelProperty): boolean {
+  private isReadOnly(target: ModelProperty | undefined): boolean {
```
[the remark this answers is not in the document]
haolingdong-msft: This is a problem brought up before, we checked segment in `isReadOnly`, but TCGC does not handle segment info.
So I keep this method..

https://github.com/Azure/autorest.java/blob/3491b346ae7b6e6cac92417ee59cd1b6dbe76c86/typespec-extension/src/code-model-builder.ts#L2371-L2390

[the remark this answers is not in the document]
weidongxu-microsoft: If this does not affect test or service, we can use theirs.

If something wrong in future, create issue for TCGC.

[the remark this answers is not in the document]
weidongxu-microsoft: Also, TCGC have visibility, use that https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/interfaces.ts#L370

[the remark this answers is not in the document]
haolingdong-msft: I did not modify the two functions (`isSecret`, `isReadOnly`) because there is difference between TCGC's visibility and TSP compiler's. After chatting offline, our agreement is to ignore the difference and modify the two functions using TCGC types. If our our tests can pass, we can keep using TCGC's one.

typespec-extension/src/code-model-builder.ts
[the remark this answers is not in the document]
haolingdong-msft: Yes, I understand we should use TCGC info, rather than `__raw`, I modified the helper functions that used from `processSchema` as much as possible. I did not modify this function because it is also used by `getEffectiveSchemaType`(https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L2309), and `getEffectiveSchemaType` is used when processing operation, I have not started on the operation part yet. So I keep it here. 
I can also write another `isPayloadPropertyFromSdkType(prop: SdkModelPropertyType)` and call the function here, then switch to use this function on other places when working on operation part code (switch to `sdkPackage`).
Please let me know your preference.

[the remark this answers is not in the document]
weidongxu-microsoft: I mean here you do not need to call anything. Just check this `SdkBodyModelPropertyType.discriminator`.
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/interfaces.ts#L367

As in the case of `neverType`, TCGC should have filtered those properties that we previously filter by `!isPayloadProperty`

haolingdong-msft: updated

typespec-extension/src/code-model-builder.ts
haolingdong-msft: good catch, updated

typespec-extension/src/code-model-builder.ts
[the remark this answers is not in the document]
haolingdong-msft: updated

typespec-extension/src/code-model-builder.ts
```diff
@@ -1695,275 +1681,201 @@ export class CodeModelBuilder {
...
+      summary: type.description ? type.description : "",
     });
 
     // cache this now before we accidentally recurse on this type.
     if (!this.schemaCache.has(type)) {
       this.schemaCache.set(type, dictSchema);
     }
 
-    const elementSchema = this.processSchema(type.indexer.value, name);
+    const elementSchema = this.processSchemaFromSdkType(type.valueType, type.valueType.kind);
```
weidongxu-microsoft, line 1854: why `type.valueType.kind` as nameHint?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1695,275 +1681,201 @@ export class CodeModelBuilder {
...
       }),
     );
   }
 
-  private processDictionarySchema(type: RecordModelType, name: string): DictionarySchema {
-    const dictSchema = new DictionarySchema<any>(name, this.getDoc(type), null, {
-      summary: this.getSummary(type),
+  private processDictionarySchemaFromSdkType(type: SdkDictionaryType, name: string): DictionarySchema {
+    const dictSchema = new DictionarySchema<any>(name, type.details ? type.details : "", null, {
+      summary: type.description ? type.description : "",
```
weidongxu-microsoft, line 1846: For summary, if undefined, just set undefined.

haolingdong-msft: Description cannot be undefined, that's why I added the fallback.
image
And I added for summary just for consistensy, but summary can accpet `undefined`. Will update the code.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2228,81 +2037,37 @@ export class CodeModelBuilder {
...
+          kind: "string",
+          encode: "string",
+          nullable: false,
+        },
+        description: type.description,
+        nullableValues: false,
+        nullable: false,
+        valueType: type.additionalProperties,
+      };
+      const parentSchema = this.processDictionarySchemaFromSdkType(sdkDictType, "Record");
```
weidongxu-microsoft, line 2057: see if this works

```suggestion
      const parentSchema = this.processSchemaFromSdkType(sdkDictType, "Record");
```

typespec-extension/src/code-model-builder.ts
```diff
@@ -2228,81 +2037,37 @@ export class CodeModelBuilder {
...
+      objectSchema.parents = objectSchema.parents ? objectSchema.parents : new Relations();
```
weidongxu-microsoft, line 2058: nit
```suggestion
      objectSchema.parents = objectSchema.parents ?? new Relations();
```

typespec-extension/src/code-model-builder.ts
```diff
@@ -2228,81 +2037,37 @@ export class CodeModelBuilder {
...
     // properties
-    for (const prop of type.properties.values()) {
-      if (
-        prop.name === discriminatorPropertyName || // skip the discriminator property
-        isNeverType(prop.type) || // skip property of type "never"
-        !isPayloadProperty(this.program, prop)
-      ) {
-        continue;
+    for (const prop of type.properties) {
+      if (prop.kind === "property" && prop.discriminator === false) {
```
weidongxu-microsoft, line 2066:
```suggestion
      if (prop.kind === "property" && !prop.discriminator) {
```

typespec-extension/src/code-model-builder.ts
```diff
@@ -2228,81 +2037,37 @@ export class CodeModelBuilder {
...
+        objectSchema.addProperty(this.processModelPropertyFromSdkType(prop));
       }
-      objectSchema.addProperty(this.processModelProperty(prop));
     }
 
-    // process all children
-    type.derivedModels?.filter(modelContainsDerivedModel).forEach((it) => this.processSchema(it, this.getName(it)));
```
weidongxu-microsoft, line 2304: The removal of this could cause certain behavior difference. Though I am not yet sure whether the difference be expected.

weidongxu-microsoft (commented):
LGTM

I don't have anything major.
Note that this could cause some behavior change https://github.com/Azure/autorest.java/pull/2698#discussion_r1584308430

Some of it relates to the model in HealthInsights (whether some subclass should be generated, and whether they are public). Please follow the discussion in LLC chat.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: No special reason for this, just provide a name hint and not all `SdkType` has `name`, so I use `kind`. I could change back to use `name`.

typespec-tests/src/main/java/com/cadl/internal/implementation/models/StandAloneUnion.java
haolingdong-msft, line 1: This is removed because TCGC will not return the model if only `@access` is set, but `@usage` is not set.
```ts
@access(Access.internal)
model StandAloneUnion {
  data: url | bytes;
}
```
I can also `@usage` to the test case to make it generated. But we already have this case (with both `@access` and `@usage`), so I did not change the test.
```ts
@access(Access.internal)
@usage(Usage.input | Usage.output)
model StandAloneUnion {
  data: url | bytes;
}
```

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: It basically about the whether a `model B extends A` get generated as public or not, when `B` is not used in any op.

Not a blocker. It likely going to have effect when you do ussage.

weidongxu-microsoft (approved):

weidongxu-microsoft:
You will need to d a minor version bump, before release this change.

typespec-tests/src/main/java/com/cadl/internal/implementation/models/StandAloneUnion.java
iscai-msft: This is by design, I would say update your test to have both access and usage, and you can open an issue about this on tcgc and we can discuss more about how when to return stand alone models

typespec-extension/src/code-model-builder.ts
```diff
@@ -1975,164 +1887,91 @@ export class CodeModelBuilder {
...
-        summary: this.getSummary(type),
+      new ConstantSchema(name, this.getDoc(type.__raw), {
+        summary: this.getSummary(type.__raw),
         valueType: valueType,
         value: new ConstantValue(type.value ?? type.name),
       }),
     );
   }
 
-  private processConstantSchemaForUnionVariant(type: UnionVariant, name: string): ConstantSchema {
```
weidongxu-microsoft, line 2016: Which TCGC type is replacing this?

haolingdong-msft: The logic is to process constant union variant, so in [`processSchemaFromSdkType`](https://github.com/Azure/autorest.java/blob/647a0208d1fb716885b36b3224aa9f574a4dc3ad/typespec-extension/src/code-model-builder.ts#L2190), the sdktype kind is `"constant"`. It will goes into this [logic](https://github.com/Azure/autorest.java/blob/647a0208d1fb716885b36b3224aa9f574a4dc3ad/typespec-extension/src/code-model-builder.ts#L1715).

weidongxu-microsoft: Are you sure?

I thought it may be included in `enumvalue` branch, but unsure.

Basically the difference is
```
model {
  prop: Enum.Value1;
}
```
and whether that `Enum` is `enum`, or `union`.

haolingdong-msft: I thought you are asking `processConstantSchemaForUnionVariant` which processes constant union variant type. So I'm referring to below case, when processing "a", it is constant type. 
```ts
model A {
   p: "a" | "b" | string
}
```


For enum value as property type. Take below test case as example:
```

model Operation {
  name: "Read" | "Write";
  best: true;
  age: 50;
  priority: Priority;
  color: ColorModel;
  unit: Unit;
  priorityValue: Priority.Low;
  colorValue: Color.Green;
  unitValue: Unit.Milligram;
}
```

TCGC response for `Priority.Low` is 
image

weidongxu-microsoft: Yes, I AM asking `processConstantSchemaForUnionVariant`, which is for this `ColorModel.Blue`.

weidongxu-microsoft: The `a` in `"a" | "b" | string` is not constant type either, it would belong to the Enum process.

haolingdong-msft: Yes, you are right, the example is wrong. should be: `"a" | "b" | int32`. And "a" is constant union variant.

ok, I get your point, you are saying `ColorModel` is union. For `colorModelValue: ColorModel.Blue;` field, The response is like below:
image

It is also 'enumvalue' kind. And `ColorModel` is  'enum' kind.

weidongxu-microsoft: `"a" | "b" | int32` won't go here either. It would go to the processUnion.

The "a" in the Union/Enum along is never a constant (union variant or not).

haolingdong-msft:
Our build [passes ](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3777283&view=results) at version "@azure-tools/typespec-client-generator-core": "0.42.0", but [encouter generation error](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3777262&view=results)at version ""@azure-tools/typespec-client-generator-core": "0.42.1".

Seems TCGC has issue on version `0.42.1` which causes the build failure: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=3777204&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=50b9345f-2a7f-5980-cb0f-1a8f9fb09dd0&l=4528

Fixed the TCGC version to `0.42.0` currently.

@tadelesh @iscai-msft

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ManagedServiceIdentity.java
```diff
@@ -37,8 +35,7 @@ public final class ManagedServiceIdentity {
      * The identities assigned to this resource by the user.
      */
     @JsonProperty(value = "userAssignedIdentities")
-    @JsonInclude(value = JsonInclude.Include.NON_NULL, content = JsonInclude.Include.ALWAYS)
-    private Map<String, UserAssignedIdentity> userAssignedIdentities;
+    private UserAssignedIdentities userAssignedIdentities;
```
weidongxu-microsoft, line 38: This is a potential problem for ARM, as Swagger is a `Map`.
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/maps/azure-resourcemanager-maps/src/main/java/com/azure/resourcemanager/maps/models/ManagedServiceIdentity.java#L43-L45

We may reconsider whether `model Model is Record<>` to be a model class or just a Map.

However, I don't know why UserAssignedIdentity is not generated).

XiaofeiCao: For `ManagedServiceIdentity`, I believe no.

XiaofeiCao: Seems we have an extra `SchemaCleanup` logic to clean up unused schemas:
https://github.com/Azure/autorest.java/blob/ca8a1f1f9462c722cfaef60acfa0aa81a13ab1ad/fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/FluentTransformer.java#L60

We have some logic to calculate which schemas to generate/not generate. Though here we have an `ObjectSchema`(UserAssignedIdentities) that has `DictionarySchema`(Record) as its parent, not as one of its properties, which we'll count as inUse(Line 82):
https://github.com/Azure/autorest.java/blob/ca8a1f1f9462c722cfaef60acfa0aa81a13ab1ad/fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/SchemaCleanup.java#L73-L87

We did not count in Dictionary as parent case. Guess we could do either, add this particular case, or use TypeSpec's usage(do not use SchemaCleanup in TypeSpec)?

haolingdong-msft: I may slightly prefer "use usage info and do not use SchemaCleanup in TSP" as this is more straightforward, and in the future, it's easier to debug, but not sure if it will cause other issue in mgmt sdk.

typespec-tests/src/main/java/com/cadl/internal/implementation/models/StandAloneUnion.java
haolingdong-msft: Hi @iscai-msft, https://github.com/Azure/typespec-azure/issues/823 - I've created an issue to discuss about when should standalone models be generated.

haolingdong-msft:
Remaining issues on this PR: 
1. TCGC 0.42.2 released, I  will integrate with the latest version to see if there is any behavior change.
2. Mgmt case error after switching to `0.42`: https://github.com/Azure/autorest.java/pull/2698#discussion_r1596260260, Xiaofe is working on this.

Two pending discussion issues: 
1. Type of model property/enum member when it is defined as constant number: https://github.com/microsoft/typespec/issues/3314
2.  when should standalone models be generated: https://github.com/Azure/typespec-azure/issues/823

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ManagedServiceIdentity.java
haolingdong-msft: "We may reconsider whether model Model is Record<> to be a model class or just a Map. That can be decided later."

For this, we discussed before, the pure Record case, and previously our agreement is to generate the model to align with other language. Because later the model can add other properties, if we use `Map`, there will be breaking changes.

XiaofeiCao:
> 2. Mgmt case error after switching to `0.42`: [ tsp, TCGC common layer, adopt model types #2698 (comment)](https://github.com/Azure/autorest.java/pull/2698#discussion_r1596260260), Xiaofei is working on this.

Offline synced with Haoling. The extra generated `UserAssignedIdentities` class is not related to `0.42`. It's due to that we switched to using TCGC's result, which will generate a model `A` if `model A is Record<>`.

Though, tsp-arm themselves discourage use of `model is Record`: https://azure.github.io/typespec-azure/docs/libraries/azure-resource-manager/rules/no-record#-incorrect-1

@weidongxu-microsoft We probably need a special handling for `UserAssignedIdentity`, or ARM.

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ManagedServiceIdentity.java
weidongxu-microsoft: > Currently we don't use TypeSpec usage in mpg... Instead, we have an extra `SchemaCleanup` logic to clean up unused schemas:
> 
> https://github.com/Azure/autorest.java/blob/ca8a1f1f9462c722cfaef60acfa0aa81a13ab1ad/fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/FluentTransformer.java#L60
> 
> In fluent, we have specific logic to calculate which schemas to generate/not generate. Though here we have an `ObjectSchema`(UserAssignedIdentities) that has `DictionarySchema`(Record) as its parent, not as one of its properties, which we'll count as inUse(Line 82):
> 
> https://github.com/Azure/autorest.java/blob/ca8a1f1f9462c722cfaef60acfa0aa81a13ab1ad/fluentnamer/src/main/java/com/azure/autorest/fluent/transformer/SchemaCleanup.java#L73-L87
> 
> We did not count in Dictionary as parent case. Guess for mgmt TypeSpec, we should use TypeSpec's usage instead of SchemaCleanup.

This one seems a bug in that calculation. Let me know if you can fix it, or me to.

typespec-tests/src/main/java/com/type/union/IntsOnlyAsyncClient.java
```diff
@@ -19,7 +19,7 @@
 import com.type.union.implementation.IntsOnliesImpl;
 import com.type.union.implementation.models.SendRequest3;
 import com.type.union.models.GetResponse6;
-import com.type.union.models.GetResponseProp2;
+import com.type.union.models.GetResponseProp3;
```
weidongxu-microsoft, line 22: any idea why this number changes? Did they change the test case?

typespec-tests/src/main/java/com/cadl/internal/implementation/models/StandAloneUnion.java
weidongxu-microsoft: @haolingdong-msft 
I don't think it worth a discussion... It is just what service can/cannot do in client.tsp.

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ManagedServiceIdentity.java
weidongxu-microsoft: @haolingdong-msft 
I don't think it worth a discussion... Seems just what service can/cannot do in client.tsp. There wasn't strict rule on this from the beginning.

typespec-tests/src/main/java/com/type/union/IntsOnlyAsyncClient.java
haolingdong-msft: Strange, I checked cadl-ranch, seems no change from last 24hours. Checked with Chenjie, the anonymous model logic does not change as well. 
But CI does not have generation diff, which means CI and action generates the same.

weidongxu-microsoft: CI and action should definitely have same result. That is expected that they get same result.

I am just not sure whether this change is due to TCGC 0.42.0 to 0.42.2. (basically, the only change would be from non-TCGC model to TCGC model; after TCGC model, upgrade of TCGC should not affect the number, as long as no change to test case).

weidongxu-microsoft: We can ignore it, if no clue.

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ManagedServiceIdentity.java
XiaofeiCao: I can add this case in SchemaCleanup.

PR #2699, opened 2024-04-17
tsp, fix additionalProperties of type BinaryData

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/pull/2699/commits/7e922eed4c942471ffa0a85ddf7e62bddb95608b (this is not the most efficient runtime code, but this is the code that works...)
test https://github.com/Azure/autorest.java/pull/2699/commits/e25f1b29ffdc18b929c3147a3a9857e77ada5a43

cadl-ranch is adding more complex types for `additionalProperties`...
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/type/property/additional-properties/main.tsp
We currently only support those that are common.

---

bump version as well, plan to release 0.15.10


javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -295,8 +295,14 @@ private static void serializeJsonProperties(JavaBlock methodBlock, ClientModelPr
             methodBlock.ifBlock(additionalPropertiesAccessExpr + " != null", ifAction -> {
                 IType valueType = ((MapType) wireType).getValueType().asNullable();
                 ifAction.line("for (Map.Entry<String, %s> additionalProperty : %s.entrySet()) {", valueType, additionalPropertiesAccessExpr);
-                ifAction.indent(() ->
-                        ifAction.line("jsonWriter.writeUntypedField(additionalProperty.getKey(), additionalProperty.getValue());"));
+                ifAction.indent(() -> {
+                    if (valueType == ClassType.BINARY_DATA) {
+                        // Special handling for BinaryData
+                        ifAction.line("jsonWriter.writeUntypedField(additionalProperty.getKey(), additionalProperty.getValue() == null ? null : additionalProperty.getValue().toObject(Object.class));");
```
haolingdong-msft, line 301: why do we have  `.toObject(Object.class)` here?

haolingdong-msft (approved):

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
weidongxu-microsoft: `writeUntypedField` does not work on `BinaryData`. So it had to be converted to `Object` first.

This is already done in defined properties. Just it missing in `additionalProperties`.

weidongxu-microsoft: Also this is my comment of "this is not the most efficient runtime code, but this is the code that works".

`BinaryData` should already contain serialized code. Here we deserialize it and let the `writeUntypedField` serialize that Object.

If possible the optimal procedure should be a direct "merge" of what in `BinaryData` and the rest of the model, but that may not have support in azure-json.

Issue #2688, opened 2024-04-11
[MPG] models under `xxUpdateProperties` are not generated with `Update` suffix


XiaofeiCao:
DeviceRegistry diff: https://apiview.dev/Assemblies/Review/90da387f53324bba89095f644611247c/e075fa8b740542859e8723c02d2166aa?diffRevisionId=3fce795412de47d78364048e06ee587f#com.azure.resourcemanager.deviceregistry.models.TransportAuthenticationUpdate

Swagger generates additional `xxUpdate` classes, e.g. `TransportAuthenticationUpdate`. 
<img width="668" alt="Screenshot 2024-04-11 at 16 14 28" src="https://github.com/Azure/autorest.java/assets/92354331/e27c7065-8ca2-4c5d-850e-1bf21fb7df8a">

, while tsp doesn't.

weidongxu-microsoft:
seems typespec-autorest make all models in PATCH adding a `Update` suffix (include all model referred there).
I don't think SDK want all such duplications. I guess we are good without them, if this is greenfield.

Also see OperationListResult -> PagedOperation
May check if we can move this class to `implementation.models` (the same processing of "Internal" -- maybe I can check this later)

And some additional setters. Should we enable `output-model-immutable` option?
About options for greenfield, should we disable the name override? https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java#L117-L132

Why this?
![image](https://github.com/Azure/autorest.java/assets/53292327/6b76dd51-077a-4dd9-9361-1d73882ffeb1)

XiaofeiCao:
Seems like a bug in swagger file:

In common model, the property is `int32`:
```typescript
@doc("Percent of the operation that is complete.")
@minValue(0)
@maxValue(100)
percentComplete?: int32;
```
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/arm.foundations.tsp#L171

---

Though in swagger, it's number, and our codegen will treat it as float/decimal:
```json
"percentComplete": {
  "description": "Percent of the operation that is complete.",
  "type": "number",
  "minimum": 0,
  "maximum": 100
},
```
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/common-types/resource-management/v3/types.json#L437-L442

Guess in swagger, it should be `"type": "integer"`?

weidongxu-microsoft:
[...]

PR #2700, opened 2024-04-17
mpg, remove namingOverride and set output-immutable to true

XiaofeiCao:
address https://github.com/Azure/autorest.java/issues/2688#issuecomment-2050841720


typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java
```diff
@@ -111,24 +112,6 @@ protected FluentMapper getFluentMapper() {
...
-        namingOverrides.put("hostName", "hostname");
-        namingOverrides.put("webHook", "webhook");
-        namingOverrides.put("coolDown", "cooldown");
-        namingOverrides.put("resourceregion", "resourceRegion");
-        namingOverrides.put("sTag", "stag");
-        namingOverrides.put("tagname", "tagName");
-        namingOverrides.put("tagvalue", "tagValue");
-
-        return namingOverrides;
     }
```
weidongxu-microsoft, line 115: comment out should be fine

we may need them again for brownfield

XiaofeiCao: updated

weidongxu-microsoft (approved):

PR #2703, opened 2024-04-17
tsp, bug fix on additionalProperties with derived model

weidongxu-microsoft:
2 problem

1. typescript code accidentally overwrite the cache data
2. java code not really sets `additionalProperties` to superclass (from code, this case happens when other properties are all on ctor) https://github.com/Azure/autorest.java/blob/78ed4bf5f631036f81230af6c5f35f73095d0c67/typespec-tests/src/main/java/com/type/property/additionalproperties/models/DifferentSpreadFloatDerived.java#L94-L97
  Should be e.g. `return new DifferentSpreadFloatDerived(name, derivedProp).setAdditionalProperties(additionalProperties);`

- 1 fixed in https://github.com/Azure/autorest.java/pull/2703/commits/a31796aa52bf7eb6137ee250bd6ec843769feebd
- 2 fixed in https://github.com/Azure/autorest.java/pull/2703/commits/3feb16674d9c89b91eeca13ba56189d05573fd3a

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
+++ b/javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
@@ -383,7 +383,7 @@ public class StreamSerializationModelTemplate extends ModelTemplate {
             propertyValueGetter);
         if (fieldSerializationMethod != null) {
             if (isJsonMergePatch && wireType instanceof ClassType && ((ClassType) wireType).isSwaggerType()) {
-                methodBlock.line(propertyValueGetter + ".serializeAsJsonMergePatch(true);");
+                methodBlock.line(String.format("JsonMergePatchHelper.get%1$sAccessor().prepareModelForJsonMergePatch(%2$s, true);", clientType.toString(), propertyValueGetter));
             }
             if (fromSuperType && clientType != wireType && clientType.isNullable()) {
                 // If the property is from a super type and the client type is different from the wire type then a null
@@ -394,7 +394,7 @@ public class StreamSerializationModelTemplate extends ModelTemplate {
                 methodBlock.line(fieldSerializationMethod + ";");
             }
             if (isJsonMergePatch && wireType instanceof ClassType && ((ClassType) wireType).isSwaggerType()) {
-                methodBlock.line(propertyValueGetter + ".serializeAsJsonMergePatch(false);");
+                methodBlock.line(String.format("JsonMergePatchHelper.get%1$sAccessor().prepareModelForJsonMergePatch(%2$s, false);", clientType.toString(), propertyValueGetter));
             }
         } else if (wireType == ClassType.OBJECT) {
             methodBlock.line("jsonWriter.writeUntypedField(\"" + serializedName + "\", " + propertyValueGetter + ");");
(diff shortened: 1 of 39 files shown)


weidongxu-microsoft:
test
```ts
import "@typespec/rest";
import "@azure-tools/typespec-client-generator-core";

using TypeSpec.Http;
using Azure.ClientGenerator.Core;

@service({
  title: "Builtin",
})
namespace Cadl.Builtin;

model DifferentSpreadStringDerived extends DifferentSpreadStringRecord {
  @doc("The index property")
  derivedProp: string;
}

model DifferentSpreadStringRecord {
  @doc("The name property")
  id: float32;

  ...Record<string>;
}

@route("/builtin")
interface BuiltinOp {
  @route("get1")
  get1(): DifferentSpreadStringRecord;

  @route("get2")
  get2(): DifferentSpreadStringDerived;
}
```

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -1290,7 +1290,8 @@ private static void createObjectAndReturn(JavaBlock methodBlock, String modelNam
         if (propertiesManager.hasConstructorArguments()) {
             if (propertiesManager.getSetterPropertiesCount() == 0
                 && propertiesManager.getReadOnlyPropertiesCount() == 0
-                && propertiesManager.getAdditionalProperties() == null) {
+                && propertiesManager.getAdditionalProperties() == null
+                && propertiesManager.getSuperAdditionalPropertiesProperty() == null) {
```
weidongxu-microsoft, line 1294: If the model as `additionalProperties` in superclass, it still need to set it separately.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1923,7 +1923,9 @@ export class CodeModelBuilder {
     });
 
     // cache this now before we accidentally recurse on this type.
-    this.schemaCache.set(type, dictSchema);
+    if (!this.schemaCache.has(type)) {
```
haolingdong-msft, line 1926: why do we add the check? Do we meet any issue?

weidongxu-microsoft: The add to Record is for this issue.

But generally we should not overwrite the cache, anyway. Hence here too.

weidongxu-microsoft: At this line, we call intentionally call `processDictionarySchema` to avoid the cache (as we want to get a new DictonarySchema).
https://github.com/Azure/autorest.java/blob/9af51f7a1ab414f79c267946c1da6e11dfc73d01/typespec-extension/src/code-model-builder.ts#L2240

However, this line still write cache. Hence the problem.

haolingdong-msft: ok, maybe ideally we can let `schemaCache` do this check, instead of having to check every time when we set cache? (I'm ok with having the check here first)

haolingdong-msft (approved):

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: It already handles it https://github.com/Azure/autorest.java/blob/9af51f7a1ab414f79c267946c1da6e11dfc73d01/typespec-extension/src/type-utils.ts#L45-L49

Here just 2 special process.

Issue #2615, opened 2024-03-11
tsp, json-merge-patch, "kind" not set to JSON for polymorphic class


weidongxu-microsoft:
### Background
https://github.com/Azure/autorest.java/pull/2576#discussion_r1504012383
```diff java
Resource resource = new Resource();
Fish salmon = new Salmon().setAge(1);
Assertions.assertEquals("salmon", salmon.getKind());
resource.setFish(salmon);
JsonMergePatchHelper.getResourceAccessor().prepareModelForJsonMergePatch(resource, true);
Fish shark = new Shark().setAge(2);
shark.setColor(null);
resource.setFish(shark);
// here when serialize for resource.fish, it calls shark.toJsonMergePatch(), but kind is not added to shark's updatedProperties, but to fish's updated properties.
+ String json = BinaryData.fromObject(resource).toString();
Assertions.assertEquals("shark", node.get("fish").get("kind"));
```
`Shark` has `updatedProperties` which shadows `Fish`'s `updatedProperties`. And in the ctor of `Shark`, it calls `Fish`'s `setKind()`, which updated `Fish`'s `updatedProperties`. 
In the highlighted line in above code snippet, when serialize for `resource.fish`, it calls `shark.toJsonMergePatch()`, but as `kind` is not added to shark's updatedProperties, but to fish's updated properties. So kind is not serialized. 

test: https://github.com/Azure/autorest.java/tree/main/typespec-tests/src/main/java/com/cadl/patch/models
service using this: [communication jobrouter](https://github.com/Azure/azure-sdk-for-java/blob/8f1c2fb5a54fb547762b76154e6dde2dd7d7ee44/sdk/communication/azure-communication-jobrouter/src/main/java/com/azure/communication/jobrouter/implementation/models/BestWorkerModeInternal.java)

#### Option1:
Make `updatedProperties` from base class as `protected`, and remove override `updatedProperties` from sub class. Thus `this.updatedProperties` in sub class will get base classs' updatedProperties.

**Pros**: Implementation is easy
**Cons**: Once we make `updatedProperties` as `protected`, we can't change back.

#### Option2:
Check both base class and sub class's `updatedProperties` in sub class's `toJsonMergePatch()` logic.

**Pros**:  Will not change the current class's public APIs.
**Cons**: Implementation is a bit complex, we will need to check all the super classes' `updatedProperties`. 

#### Option3:
[...]

PR #2704, opened 2024-04-17
tsp, json-merge-patch, fix issue on patch for discriminator property

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/2615


weidongxu-microsoft (approved):
maybe bump the typespec-java version as well

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -821,6 +821,9 @@ private void addModelConstructor(ClientModel model, JavaVisibility constructorVi
                 if (modelDefinesProperty(model, polymorphicProperty)) {
                     constructor.line("this." + polymorphicProperty.getName() + " = " + discriminatorValue + ";");
                 }
+                if (ClientModelUtil.isJsonMergePatchModel(model)) {
+                    constructor.line(String.format("this.updatedProperties.add(\"%s\");", polymorphicProperty.getName()));
```
srnagar, line 825: nit: avoid using `String.format`. It's performance is very bad. We can just use string concatenation.

```suggestion
                    constructor.line("this.updatedProperties.add(\"" + polymorphicProperty.getName() + "\");");
```

srnagar (approved):

typespec-extension/package.json
weidongxu-microsoft, line 1: run install and update lock file as well.

weidongxu-microsoft:
sorry, forget the changelog. need to update that too.

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
haolingdong-msft: Fixed for the two prs on json merge patch .

PR #2708, opened 2024-04-19
tsp, apiview properties, add ending newline

weidongxu-microsoft:


typespec-extension/src/main/java/com/azure/typespec/Main.java
```diff
@@ -168,7 +168,7 @@ private static void handleDPG(CodeModel codeModel, EmitterOptions emitterOptions
                 }
                 sb.append("    \"").append(key).append("\": \"").append(value).append("\"");
             });
-            sb.append("\n  }\n}");
+            sb.append("\n  }\n}\n");
```
weidongxu-microsoft, line 171: We should use a model and JSON serialization to do this.

Refactor later.

haolingdong-msft (approved):

PR #2712, opened 2024-04-19
tsp, mgmt, sdk integration

weidongxu-microsoft:
The meaning of `sdkIntegration` in ARM is a bit different from that of DPG.

Basically it should be always enabled, if target is the SDK repo.

---

One remaining problem is the tag in readme/changelog/pom.

Swagger readme have tag, but there is no such concept in typespec. Still not sure how we best handle this. Maybe we need an emitter option, e.g. `readme-tag`.

image

### DIFF
--- a/typespec-extension/src/main/java/com/azure/typespec/Main.java
+++ b/typespec-extension/src/main/java/com/azure/typespec/Main.java
@@ -73,14 +73,19 @@ public class Main {
         String outputDir = emitterOptions.getOutputDir();
         Path outputDirPath = Paths.get(outputDir);
         if (Files.exists(outputDirPath)) {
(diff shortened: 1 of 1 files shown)


XiaofeiCao (approved):

XiaofeiCao:
Can we remove this `package-tag` when generating from typespec? 

One blocker that I can think of is the single_service_packages.md, where packages are grouped by package tags.

weidongxu-microsoft:
The remove is simple. But since at least at present, we still have such mapping (as even if Java generate from tsp, it still have a readme for other languages).

Though we may consider adding another line about "api-version", so in future we may switch to api-version based reference.

Issue #2716, opened 2024-04-22
Projected version support Implementation


lmazuel:
Should be as simple as move to TCGC 0.41.4+

weidongxu-microsoft:
Java would need to distinguish a configure of "all" or "latest", depends on emitter option "advanced-versioning"

PR #2717, opened 2024-04-23
all client, upgrade tcgc and support versioning

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2716

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -211,6 +211,11 @@ export class CodeModelBuilder {
     }
 
     this.sdkContext = createSdkContext(context, "@azure-tools/typespec-java");
+    if (this.supportsAdvancedVersioning()) {
+      this.sdkContext.apiVersion = "all";
+    } else {
+      this.sdkContext.apiVersion = "latest";
+    }
     const service = listServices(this.program)[0];
     const serviceNamespace = service.type;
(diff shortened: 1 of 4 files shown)


weidongxu-microsoft (approved):
update lock file

weidongxu-microsoft (commented):
nit, update this as well
https://github.com/Azure/autorest.java/blob/73b8999cd4804565251d90ad55d769085f10b1fb/typespec-tests/package.json#L28

weidongxu-microsoft:
@XiaofeiCao 

They seems to have a `api-version` in emitter.
https://github.com/Azure/typespec-azure/pull/719/files

Please plan a follow-up PR (as this is not always "latest").

typespec-extension/src/code-model-builder.ts
```diff
@@ -211,6 +211,11 @@ export class CodeModelBuilder {
     }
 
     this.sdkContext = createSdkContext(context, "@azure-tools/typespec-java");
+    if (this.supportsAdvancedVersioning()) {
+      this.sdkContext.apiVersion = "all";
+    } else {
+      this.sdkContext.apiVersion = "latest";
+    }
```
weidongxu-microsoft, line 218: Don't do this, if not `this.supportsAdvancedVersioning()`.

Also, please read this `sdkContext.apiVersion` into our own `apiVersion` in class (if it is not "all" or "latest")

Issue #2720, opened 2024-04-24
basic versioning, support specific api-version other than `all` and `latest`


XiaofeiCao:
Follow up issue on https://github.com/Azure/autorest.java/pull/2717#issuecomment-2073769596

1. emitterOptions, `api-version: 2023-11-01-preview`
  https://github.com/Azure/typespec-azure/pull/719/files#diff-500b3c1c9858545cfc35d0be1ebd84bf90746a0bc2116743a03cba712d4351faR237-R238
2. pending discussion on `all` behavior: https://github.com/Azure/typespec-azure/pull/719/files#r1577098644

PR #2722, opened 2024-04-26
bump tcgc and typespec-java, support `api-version`

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2720
1. support `api-version` in tspconfig.yaml
2. bump tcgc to 0.41.6
3. bump typespec-java to 0.15.13 for release

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -181,6 +181,7 @@ import {
 } from "./operation-utils.js";
 import { isArmCommonType } from "./type-utils.js";
 import pkg from "lodash";
+import { log } from "console";
 const { isEqual } = pkg;
 
 export class CodeModelBuilder {
@@ -213,8 +214,8 @@ export class CodeModelBuilder {
     this.sdkContext = createSdkContext(context, "@azure-tools/typespec-java");
     if (this.supportsAdvancedVersioning()) {
       this.sdkContext.apiVersion = "all";
-    } else {
-      this.sdkContext.apiVersion = "latest";
+    } else if (this.options["api-version"]) {
+      this.sdkContext.apiVersion = this.options["api-version"];
     }
     const service = listServices(this.program)[0];
     const serviceNamespace = service.type;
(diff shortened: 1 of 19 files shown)


typespec-extension/src/code-model-builder.ts
```diff
@@ -181,6 +181,7 @@ import {
 } from "./operation-utils.js";
 import { isArmCommonType } from "./type-utils.js";
 import pkg from "lodash";
+import { log } from "console";
```
weidongxu-microsoft, line 184: log not used?

typespec-extension/src/code-model-builder.ts
```diff
@@ -213,8 +214,8 @@ export class CodeModelBuilder {
     this.sdkContext = createSdkContext(context, "@azure-tools/typespec-java");
     if (this.supportsAdvancedVersioning()) {
       this.sdkContext.apiVersion = "all";
-    } else {
-      this.sdkContext.apiVersion = "latest";
+    } else if (this.options["api-version"]) {
+      this.sdkContext.apiVersion = this.options["api-version"];
```
weidongxu-microsoft, line 218: Could it be "latest" or "all" here?

weidongxu-microsoft, line 218: Maybe we don't set this at all.

Instead, we add another `api-version=all` to the cases we now have `advancedVersioning`.

typespec-tests/Generate.ps1
```diff
@@ -43,13 +43,18 @@ $generateScript = {
   } elseif ($tspFile -match "resiliency[\\/]srv-driven[\\/]main.tsp") {
     # enable advanced versioning for resiliency test
     $tspOptions += " --option ""@azure-tools/typespec-java.advanced-versioning=true"""
```
weidongxu-microsoft, line 45: For comment https://github.com/Azure/autorest.java/pull/2722#discussion_r1580502052

We could just add `api-version=all` to this case.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: Yeah, it'll read from emitter-options. Removed.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: No, removed. It's added accidentally by using IDE...

typespec-tests/Generate.ps1
XiaofeiCao: Added.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: Updated.

weidongxu-microsoft (approved):

Issue #2723, opened 2024-04-26
[BUG] when `api-version` is pinned, `ServiceVersion` should only contain api-versions prior to it


XiaofeiCao:
We need to filter out api-versions newer than the pinned one. If not, wrong api-version(latest) will be selected by `ServiceVersion.getLatest()` to call REST API.

And if we generate from stable api-version, the enum should only contain stable api-versions(In case preview versions got removed).

`service-versions` won't work as we've switched to multi client in TypeSpec:
https://github.com/Azure/autorest.java/blob/90517f2278226a7d16de46c0b8ae16bc10b28127/javagen/src/main/java/com/azure/autorest/Javagen.java#L219-L220

PR #2724, opened 2024-04-28
dpg, make `ServiceVersion` only contain api-versions no later than the pinned one

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2723

1. Let `ServiceVersion` contain only versions no later than the pinned api-version.
2. If pinned api-version(or default latest api-version) is stable, make `ServiceVersion` contain only stable versions.
3. Removed previous emitter logic of filtering operations/parameters/model(Properties). TCGC would do it now.

### DIFF
--- a/typespec-extension/src/type-utils.ts
+++ b/typespec-extension/src/type-utils.ts
@@ -52,22 +52,8 @@ export class ProcessingCache<In, Out> {
   }
 }
 
-/**
- * Expose `isStable` and comparison capability of an Api Version.
- */
-export class ComparableApiVersion {
-  version: Version;
-  constructor(version: Version) {
-    this.version = version;
-  }
-
-  isStable(): boolean {
-    return !this.version.value.includes("preview");
-  }
-
-  noLaterThan(anotherVersion: ComparableApiVersion): boolean {
-    return this.version.value <= anotherVersion.version.value;
-  }
+export function isStable(version: Version): boolean {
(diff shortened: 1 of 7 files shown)


typespec-extension/src/type-utils.ts
```diff
@@ -51,6 +52,24 @@ export class ProcessingCache<In, Out> {
...
+  constructor(version: Version) {
+    this.version = version;
+  }
+
+  isStable(): boolean {
+    return !this.version.value.includes("preview");
+  }
+
+  noLaterThan(anotherVersion: ComparableApiVersion): boolean {
+    return this.version.value <= anotherVersion.version.value;
```
weidongxu-microsoft, line 69: which is larger? `2022-02-01` or `2022-02-01-preview`

I don't have idea though.

weidongxu-microsoft (commented):
bump the typespec-java version please

typespec-extension/src/type-utils.ts
XiaofeiCao: Emmm, yes, former should be larger.. I think I should borrow the logic from autocent.js..

XiaofeiCao: Or maybe we should rely on service to do the sort for us. We only filter out those appear after the pinned version, in case of some strange `2022-12-01-preview2` or `2022-12-01-newpreview`.

weidongxu-microsoft (approved):

PR #2733, opened 2024-04-29
Remove eclipse language server

alzimmermsft:
Remove Eclipse language server usage from code customizations. Instead rely completely on JavaParser to handle customizations.

This contains breaking changes where abilities of customizations have been reduced or removed.


customization-base/README.md
```diff
@@ -25,10 +25,7 @@ customizations are supported:
 - [Add an annotation to a class](#add-an-annotation-to-a-class)
 - [Add an annotation to a method](#add-an-annotation-to-a-method)
 - [Remove an annotation from a class](#remove-an-annotation-from-a-class)
-- [Refactor: Rename a class](#refactor-rename-a-class)
 - [Refactor: Rename a method](#refactor-rename-a-method)
-- [Refactor: Generate the getter and setter methods for a property](#refactor-generate-the-getter-and-setter-methods-for-a-property)
-- [Refactor: Rename a property and its corresponding getter and setter methods](#refactor-rename-a-property-and-its-corresponding-getter-and-setter-methods)
```
srnagar, line 31: What are the alternatives for these removed features?

weidongxu-microsoft: Yeah... Usually the power of the language server is on the rename symbol part. I think it will help modify the code that refs the symbol.

java-parser does have a symbol-solver lib. But from the experience (used it 3 years ago to auto rename some API in mgmt lib), it is not very reliable.

customization-base/src/main/java/com/azure/autorest/customization/ClassCustomization.java
```diff
@@ -571,18 +366,21 @@ public ClassCustomization removeAnnotation(String annotation) {
      * @return the current class customization for chaining
      */
     public ClassCustomization renameEnumMember(String enumMemberName, String newName) {
```
weidongxu-microsoft, line 368: I assume it could have problem, if the Enum member is used somewhere.

Though, so far, I cannot think of a case that the Enum member is reference in generated code (except sample/test).

customization-base/src/main/java/com/azure/autorest/customization/ConstructorCustomization.java
```diff
@@ -134,26 +136,35 @@ public ConstructorCustomization replaceBody(String newBody) {
...
+     * <p>
+     * If the new body contains {@code this()} or {@code super()} constructor calls, use
+     * {@link #replaceBody(NodeList, List)} instead.
      *
      * @param newBody New constructor body.
      * @param importsToAdd Any additional imports required by the constructor. These will be custom types or types that
      * are ambiguous on which to use such as {@code List} or the utility class {@code Arrays}.
      * @return A new ConstructorCustomization representing the updated constructor.
      */
     public ConstructorCustomization replaceBody(String newBody, List<String> importsToAdd) {
```
weidongxu-microsoft, line 148: I think previously language server would be able to automatically add some missing imports used in the new body.
I guess java-parser won't.

PR #2739, opened 2024-05-06
tsp, handle the case of optional Constant, but valueType already an enum

weidongxu-microsoft:


typespec-tests/tsp/enum.tsp
```diff
@@ -60,6 +60,7 @@ enum OperationStateValues {
   Failed,
 }
 
+@usage(Usage.input)
```
weidongxu-microsoft, line 63: I change it to `input`, to have the setter generated (so that I can see the `unitValue` now settable.

typespec-tests/tsp/enum.tsp
```diff
@@ -70,7 +71,7 @@ model Operation {
   priorityValue: Priority.Low;
   colorValue: Color.Green;
   colorModelValue: ColorModel.Blue;
-  unitValue: Unit.Milligram;
+  unitValue?: Unit.Milligram;
```
weidongxu-microsoft, line 74: Without this PR, client would generate a `Milligram` enum. To me this does not look OK.

preprocessor/src/main/java/com/azure/autorest/preprocessor/Preprocessor.java
```diff
@@ -152,9 +156,13 @@ public static CodeModel convertOptionalConstantsToEnum(CodeModel codeModel) {
...
-                        SealedChoiceSchema sealedChoiceSchema = convertedChoiceSchemas.computeIfAbsent(constantSchema,
-                            Preprocessor::convertToChoiceSchema);
-                        p.setSchema(sealedChoiceSchema);
+                        if (schemaIsConstantWithChoice.apply(constantSchema)) {
+                            p.setSchema(constantSchema.getValueType());
+                        } else {
+                            SealedChoiceSchema sealedChoiceSchema = convertedChoiceSchemas.computeIfAbsent(constantSchema,
+                                    Preprocessor::convertToChoiceSchema);
+                            p.setSchema(sealedChoiceSchema);
+                        }
```
weidongxu-microsoft, line 165: When `valueType` already an enum (open or closed), use that enum. Do not create another enum.

weidongxu-microsoft: And this way, the API be more consistent to the `required` case (`colorValue` or `colorModelValue`).

weidongxu-microsoft: I think Swagger won't have such case.

XiaofeiCao (approved):

PR #2747, opened 2024-05-08
tsp, bump typespec dependencies to 0.56

weidongxu-microsoft:
need a new cadl-ranch https://github.com/Azure/cadl-ranch/pull/577


typespec-extension/package.json
```diff
@@ -60,24 +61,26 @@
...
-    "@typespec/versioning": "0.55.0"
+    "@typespec/compiler": "0.56.0",
+    "@typespec/http": "0.56.0",
+    "@typespec/openapi": "0.56.0",
+    "@typespec/rest": "0.56.0",
+    "@typespec/versioning": "0.56.0",
+    "@azure-tools/typespec-azure-core": "0.42.0",
+    "@azure-tools/typespec-client-generator-core": "0.42.0",
+    "@azure-tools/typespec-azure-resource-manager": "0.42.0",
+    "@azure-tools/typespec-autorest": "0.42.0"
```
weidongxu-microsoft, line 84: I've also added typespec-azure-resource-manager into dependency.

typespec-tests/tsp/arm.tsp
weidongxu-microsoft, line 1: arm.tsp mainly for change of enum->union, and the rename of some model (old one deprecated).

typespec-tests/tsp/flatten.tsp
weidongxu-microsoft, line 1: tsp on DPG is mostly for the behavior change of http lib 0.56

typespec-tests/src/main/java/com/cadl/flatten/implementation/models/SendLongRequest.java
weidongxu-microsoft, line 1: change in this file may nullify some of our test, but I didn't see a sane way to put it back in tsp.

typespec-tests/tsp/visibility.tsp
```diff
@@ -40,23 +40,31 @@ model WriteDog {
...
-  query(@body dog: ReadDog): Dog;
+  query(@body dog: ReadDog): {
+    @body body: Dog;
+  };
 
   @put
   @route("/roundtrip")
-  roundtrip(@body body: RoundTripModel): RoundTripModel;
+  roundtrip(@body body: RoundTripModel): {
+    @body body: RoundTripModel;
```
haolingdong-msft, line 68: If no `@body` added, it will be interpreted as no response body?

haolingdong-msft (approved):

XiaofeiCao (approved):

typespec-tests/tsp/visibility.tsp
XiaofeiCao: Wonder why some are not affected..
https://github.com/Azure/autorest.java/blob/6df932a061ef99f88eaede77840efa94d221e948/typespec-tests/tsp/enum.tsp#L93-L95

weidongxu-microsoft: model is
```
model RoundTripModel {
  name: string;
  @visibility("create", "update") secretName: string;
}
```

If no explicit body, the `secretName` will not be in response body (as the visibility), and this "response" would be anonymous model.

weidongxu-microsoft: > Wonder why some are not affected..
> 
> https://github.com/Azure/autorest.java/blob/6df932a061ef99f88eaede77840efa94d221e948/typespec-tests/tsp/enum.tsp#L93-L95

I guess `Operation` does not have `@visibility`.

PR #2749, opened 2024-05-10
arm, remove special handling for ARM types

XiaofeiCao:
`typespec-azure-resource-manager` has fixed the naming issue: https://github.com/Azure/typespec-azure/pull/762
Our special handling for `Resource`, `ProxyResource` and `ArmResource` is no longer needed.

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -37,7 +37,6 @@ import {
   getEncode,
   getOverloadedOperation,
   EnumMember,
-  walkPropertiesInherited,
   isVoidType,
   isErrorModel,
 } from "@typespec/compiler";
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -489,10 +489,6 @@ export class CodeModelBuilder {
     // deduplicate model name
     const nameCount = new Map<string, number>();
     const deduplicateName = (schema: Schema) => {
-      // skip models under "Azure.ResourceManager"
-      if (this.isArm() && schema.language.default?.namespace?.startsWith("Azure.ResourceManager")) {
-        return;
-      }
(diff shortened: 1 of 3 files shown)


typespec-extension/src/code-model-builder.ts
```diff
@@ -489,10 +489,6 @@ export class CodeModelBuilder {
     // deduplicate model name
     const nameCount = new Map<string, number>();
     const deduplicateName = (schema: Schema) => {
-      // skip models under "Azure.ResourceManager"
-      if (this.isArm() && schema.language.default?.namespace?.startsWith("Azure.ResourceManager")) {
-        return;
-      }
```
XiaofeiCao, line 495: This deduplication logic is due to same dummy resource schema being added many times here:
https://github.com/Azure/autorest.java/pull/2749/files#diff-14fdfb2e0df46da414f4d04a218f03e4e5f2737d3a9672b9c4e4feb16419445eL2160

weidongxu-microsoft:
ArmResource is deprecated but not removed. Do we still need this till it get removed?

XiaofeiCao:
> ArmResource is deprecated but not removed. Do we still need this till it get removed?

I searched current specs repo, seems (Arm)Resource is only used in template parameters: https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs+%22ArmResource%22&type=code&p=1
I added back one `ArmResource` in our test, seems ok.

I guess service is not allowed to let its models directly extend `ArmResource`, otherwise it'll break JS and Python..

XiaofeiCao:
Offline sync with Chenjie, service will not use `ArmResource` for green field. For brownfields, there exists cases where model directly extends Resource, and will be converted to `Resource` in tsp, still no `ArmResource`. And existing `ArmResource` will be fixed to `Resource`.

weidongxu-microsoft (approved):

Issue #2385, opened 2023-11-07
tsp, investigate whether we can generate new PollingStrategy in generated code


weidongxu-microsoft:
# Context

Some SDK already using [the new PollingStrategy that currently in azure-core-experimental](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/core/azure-core-experimental/src/main/java/com/azure/core/experimental/util/polling), e.g. [HealthInsights](https://github.com/Azure/azure-sdk-for-java/blob/c5d9a28c675fec92183fda969e5f8d5792f6365d/sdk/healthinsights/azure-health-insights-radiologyinsights/src/main/java/com/azure/health/insights/radiologyinsights/implementation/RadiologyInsightsClientImpl.java#L1809-L1822)
(one can see the final result class is not same as polling result class in `SyncPoller<PollOperationDetails, RadiologyInsightsInferenceResult>`)

However, the new LRO design may not be final, and it may be too early to move the new PollingStrategy into azure-core.

One known issue is that some service uses property other than [the designed `result` property](https://github.com/microsoft/api-guidelines/blob/vNext/azure/ConsiderationsForServiceDesign.md#long-running-action-operations) for the final result.
E.g. [DocumentIntelligence uses `analyzeResult`](https://github.com/Azure/azure-rest-api-specs/blob/6fc313ed7a6e926d4409d4ddb435f6532922652f/specification/ai/DocumentIntelligence/models.tsp#L430-L432) in one of their LRO.
We are not sure whether this is a one-off case, or a case that need general support (typespec would not block service dev write this).

Another issue is that there could be a pending different design on LRO with PUT (the PUT to create a status monitor, then poll directly on this status monitor). A new api-version of HealthInsights could be the first -- though at present, this may solved by the legacy PollingStrategy).

# Thought

Generate `PollingStrategy` directly into the SDK, instead of having them in azure-core or azure-core-experimental

For legacy LRO that already exists in azure-core, still use the `PollingStrategy` there.

It would require for codegen to distinguish the type of LRO.

# Implementation consideration

---

Currently we had a few ref to impl packages in azure-core implementation package. They need to be implemented inline, if do it in SDK.

```java
import com.azure.core.implementation.ImplUtils;
import com.azure.core.util.polling.implementation.PollingConstants;
import com.azure.core.util.polling.implementation.PollingUtils;


                PollingUtils.getAbsolutePath(operationLocationHeader.getValue(), endpoint, LOGGER));

[...]

Issue #2756, opened 2024-05-13
tsp, support "LRO action with no related resource pattern"


weidongxu-microsoft:
ref https://github.com/Azure/typespec-azure/issues/787

handle together with https://github.com/Azure/autorest.java/issues/2385

Assume LroMetadata is correct. Java should allow the "result" property as final result, in PUT (previously, and in core-management, this is only for POST).

PR #2757, opened 2024-05-13
tsp, generate PollingStrategy in emitter

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2385

typescript code https://github.com/Azure/autorest.java/pull/2757/commits/721c1bfeacfa9ccda3c19341137a18597fcabe4e
java helper class https://github.com/Azure/autorest.java/pull/2757/commits/13896da03dcc1d38fd96b165a9913f361ad7417a
java code https://github.com/Azure/autorest.java/pull/2757/commits/5b1ecfb66dcdf3bf776277f05bcfacec55eca0f0

The hope is that this PR can also handle case https://github.com/Azure/autorest.java/issues/2756, as long as LroMetadata from typespec-azure-core is correct.


typespec-tests/src/main/java/com/cadl/longrunning/LongRunningClient.java
```diff
@@ -205,11 +210,11 @@ public JobResult getJob(String id) {
      * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
      * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
      * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
-     * @return the {@link SyncPoller} for polling of status details for long running operations.
+     * @return the {@link SyncPoller} for polling of long-running operation.
      */
     @Generated
     @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
-    public SyncPoller<PollOperationDetails, JobResultResult> beginCreateJob(JobData jobData) {
+    public SyncPoller<JobResult, JobResultResult> beginCreateJob(JobData jobData) {
```
weidongxu-microsoft, line 217: Diff here is because a slight change in logic. `JobResult` does not refer `Azure.Core.Foundations.OperationStatus`. So, here it does not use Java core class.

typespec-tests/src/main/java/com/cadl/response/ResponseClient.java
```diff
@@ -398,7 +398,7 @@ public boolean exists() {
      */
     @Generated
     @ServiceMethod(returns = ReturnType.LONG_RUNNING_OPERATION)
-    public SyncPoller<OperationDetails1, OperationDetails1> beginLroInvalidPollResponse(Resource request) {
+    public SyncPoller<OperationDetails1, Resource> beginLroInvalidPollResponse(Resource request) {
```
weidongxu-microsoft, line 401: Diff here is because a slight change in logic. Now if LroMetadata tell us that finalResult need to be from "result" property, we always do that.

javagen/src/main/resources/OperationLocationPollingStrategy.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Creates an instance of the operation resource polling strategy.
+     *
+     * @param pollingStrategyOptions options to configure this polling strategy.
+     * @throws NullPointerException if {@code pollingStrategyOptions} is null.
+     */
+    public OperationLocationPollingStrategy(PollingStrategyOptions pollingStrategyOptions) {
+        this(pollingStrategyOptions, null);
+    }
```
weidongxu-microsoft, line 44: In this mode, it works almost same as azure-core `OperationResourcePollingStrategy`, except that it would not throw if the initial response cannot be de-serialize to `T` (as in PUT pattern, it could be `U`)

haolingdong-msft:
Could we list what polling strategy will we generate in emitter, and what will we use those in azure-core?

weidongxu-microsoft:
> Could we list what polling strategy will we generate in emitter, and what will we use those in azure-core?

Basically, this line (and func) `const useNewPollStrategy = isLroNewPollingStrategy(httpOperation, lroMetadata);`

haolingdong-msft (approved):

PR #2761, opened 2024-05-14
Fix polymorphic XML stream-style serialization

alzimmermsft:
Fixes an issue where XML stream-style serialization was serializing and deserializing the polymorphic property multiple times. This was due to a recent change where polymorphic properties are now retained as formal properties within the `ClientModel` rather than being inferred. Resulting in a polymorphic property each time there was a super type and again for the model itself.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
+++ b/javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
@@ -150,7 +150,9 @@ public final class ClientModelPropertiesManager {
             possibleXmlNameVariableNames.remove(property.getName());
 
             if (property.isXmlAttribute()) {
-                superXmlAttributes.add(property);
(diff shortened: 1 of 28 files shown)


javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
```diff
@@ -150,7 +150,9 @@ public ClientModelPropertiesManager(ClientModel model, JavaSettings settings) {
             possibleXmlNameVariableNames.remove(property.getName());
 
             if (property.isXmlAttribute()) {
-                superXmlAttributes.add(property);
+                if (!property.isPolymorphicDiscriminator()) {
```
srnagar, line 153: Prefer to combine the conditions instead of nested `if`s. 

```java
if(property.isXmlAttribute() && !property.isPolymorphicDiscriminator())
```

srnagar (approved):

PR #2762, opened 2024-05-15
tsp, try remove default on host

weidongxu-microsoft:
not able to handle here https://github.com/Azure/autorest.java/pull/2760

guess we could just remote it (anyway we don't need client default).


haolingdong-msft:
I don't quite get why do we remove client default value?

weidongxu-microsoft:
"next" version deprecated this "default". https://github.com/Azure/autorest.java/pull/2760

Also this wasn't suppose to be "client default".
```ts
    @doc("second-level domain, use httpbin")
    domain?: string = "httpbin",

    @doc("top-level domain, use org")
    tld?: string = "org",
```

It should be "server default".

haolingdong-msft (approved):

weidongxu-microsoft:
I will put the compatibility of "next" for nightly to another PR.

PR #2765, opened 2024-05-16
tsp, make code compatible with compiler/next

weidongxu-microsoft:


typespec-extension/src/type-utils.ts
```diff
@@ -77,7 +77,7 @@ export function isModelReferredInTemplate(template: TemplatedTypeBase, target: M
   return (
     template === target ||
     (template?.templateMapper?.args?.some((it) =>
-      it.kind === "Model" || it.kind === "Union" ? isModelReferredInTemplate(it, target) : false,
+      "kind" in it && (it.kind === "Model" || it.kind === "Union") ? isModelReferredInTemplate(it, target) : false,
```
weidongxu-microsoft, line 80: 0.57 should have `isType`

XiaofeiCao (approved):

PR #2773, opened 2024-05-17
tsp, internal on op won't have effect on its model, if not convenient

weidongxu-microsoft:
Align with TCGC.

And it appears to be reasonable, for API that is convenient=false, its model does not need to be generated -- internal or public (unless specified otherwise).


typespec-tests/main.tsp
weidongxu-microsoft, line 1: mistakenly checked-in, delete

typespec-extension/src/code-model-builder.ts
```diff
@@ -1352,10 +1344,9 @@ export class CodeModelBuilder {
...
     this.trackSchemaUsage(schema, { usage: [SchemaContext.Input] });
 
-    if (op.internalApi) {
-      this.trackSchemaUsage(schema, { usage: [SchemaContext.Internal] });
-    } else if (op.convenienceApi) {
-      this.trackSchemaUsage(schema, { usage: [SchemaContext.Public] });
+    if (op.convenienceApi) {
+      // model/schema does not need to be Public or Internal, if it is not to be used in convenience API
+      this.trackSchemaUsage(schema, { usage: [op.internalApi ? SchemaContext.Internal : SchemaContext.Public] });
     }
```
weidongxu-microsoft, line 1350: Reason here.

haolingdong-msft (approved):

PR #2775, opened 2024-05-17
Fixes for JSON merge patch in polymorphic types

### LINKED ISSUE #2769: Polymorphic JSON merge patch models may cause properties to be included in update when deserializing
While regenerating an SDK to use stream-style serialization I found that polymorphic properties aren't being handled correctly during deserialization if they support JSON merge patch.

 The current design of JSON stream-style deserialization uses the model setter if the property is defined by a super type. JSON merge patch uses the setter to determine if a property should be included in serialization when the model is used in a patch operation. These together result in the parent properties always being included in serialization, and possibly in an incorrect way if the value is `null` and something else sets it in-between deserialization and using the model in a patch operation.

**Code sample**

```java
public class Parent {
  private String property;

  public Parent setProperty(String property) {
    this.property = property;
    // Track property for patch serialization
    return this;
  }
}

public class Child extends Parent {
  public Child setProperty(String property) {
    super.setProperty(property);
    // Track property for patch serialization
    return this;
  }

  public static Child fromJson(JsonReader jsonReader) throws IOException {
    // logic
    if ("property".equals(fieldName) {
      childInstance.setProperty(jsonReader.getString());
    }
    // logic
  }
}
```

**Solutions**

At this time there are two solutions I can think of:

1. The models within the polymorphic hierarchy have internal setters that don't track the property's inclusion into patch serialization.
2. The parent model defines the property as package-private and all children models change deserialization to set the property directly.

My preference is for option #2 as option #1 would require each child model within the hierarchy to add an internal setter as well. Additionally, option #2 can be used universally, even when the model isn't used in JSON merge patch operations and would allow for the removal of calls like:

```java
public String getProperty() {
  return super.getProperty();
}

public Type setProperty(String property) {
  super.setProperty(property);
  return this;
}
```

and allow for direct access to the field:

```java
public String getProperty() {
  return property;
}

public Type setProperty(String property) {
  this.property = property;
  return this;
}
```

### LINKED ISSUE #2770: JsonMergePatchHelper pattern can be simplified
While regenerating an SDK to use stream-style serialization I found that the `JsonMergePatchHelper` pattern used to turn on and off patch serialization doesn't work correctly for polymorphic types.

When the models support JSON merge patch each model will create an access helper and add a boolean flag to determine whether JSON merge patch serialization should be used when calling `toJson`. This isn't necessary and can be a single instance on the super type by having the super type expose the patch serialization status to child types. So instead of:

```java
public class Super {

  static {
    // Configure access helper
  }

  private boolean jsonMergePatch;
  
    if (jsonMergePatch) {
    ...
  }
}

public class Child extends Super {

  static {
    // Configure access helper
  }

  private boolean jsonMergePatch;
  
    if (jsonMergePatch) {
    ...
  }
}
```

it can be simplified to

```java
public class Super {

  static {
    // Configure access helper
  }

  private boolean jsonMergePatch;
  
  boolean isJsonMergePatch() {
    return jsonMergePatch;
  }
  
    if (isJsonMergePatch()) {
    ...
  }
}

public class Child extends Super {
  
    if (isJsonMergePatch()) {
    ...
  }
}
```

And since there is boilerplate code for each accessor this will simplify that to one instance per polymorphic model hierarchy.

Further investigation could be done to see if it is possible for all models within a single package could use the same access helper.

### LINKED ISSUE #2771: JSON merge patch may use BinaryData incorrectly
When using `BinaryData` with JSON merge patch it may be using it incorrectly. The following code is used:

```java
String variableName = convenientParameterName + "InBinaryData";
javaBlock.line(String.format("JsonMergePatchHelper.get%1$sAccessor().prepareModelForJsonMergePatch(%2$s, true);", convenientParameterTypeName, convenientParameterName));
javaBlock.line(String.format("BinaryData %1$s = BinaryData.fromBytes(%2$s.toBytes());", variableName, expression)); // BinaryData.fromObject() will not fire serialization, use toBytes() to fire serialization
javaBlock.line(String.format("JsonMergePatchHelper.get%1$sAccessor().prepareModelForJsonMergePatch(%2$s, false);", convenientParameterTypeName, convenientParameterName));
return variableName;
```

https://github.com/Azure/autorest.java/blob/main/javagen/src/main/java/com/azure/autorest/template/ConvenienceMethodTemplateBase.java#L788

We should be using the `BinaryData.fromObject` directly rather than converting it from `Object` to `byte[]` then landing on `BinaryData`. This may also result in `BinaryData` being used incorrectly later on as it loses type information and may result it being considered binary in JSON (base64 encoded string) rather than a JSON object.

alzimmermsft:
Fixes #2769 
Fixes #2770 
Fixes #2771 

Some fixes when handling stream-style, polymorphic, JSON merge patch models.

- Reduced the number of access helper interfaces created to only the polymorphic parent type. The polymorphic parent will manage whether JSON merge patch serialization will be used and will share this information to child types through an access helper method.
- Change how `BinaryData` is created to just use `BinaryData.fromObject`. To make sure serialization happens before JSON merge patch serialization is disabled `BinaryData.getLength()` is called instead of the previous `BinaryData.fromBytes(BinaryData.fromObject(object).toBytes())` pattern, this will retain information in `BinaryData` that it is based on an object.
- Changed how deserialization works to no longer call the setter methods when the property is defined by a parent model. This was accidentally setting those properties to always be included in JSON merge patch serialization, which could lead to incorrect behavior. Now the parent models expose access helper methods for setting properties which child models will set directly.
- Cleaned up deserialization code when a JSON merge patch model has no parameters in the constructor. Before this was just checking if the model had required properties and required properties were being included in the constructor, but JSON merge patch models don't generate required properties in the constructor if they're optional in the JSON merge patch call.


typespec-tests/src/main/java/com/cadl/patch/models/Salmon.java
```diff
@@ -294,10 +279,10 @@ public static Salmon fromJson(JsonReader jsonReader) throws IOException {
                 }
             }
             Salmon deserializedSalmon = new Salmon();
-            deserializedSalmon.setId(id);
-            deserializedSalmon.setName(name);
-            deserializedSalmon.setAge(age);
-            deserializedSalmon.setColor(color);
+            deserializedSalmon.id = id;
```
alzimmermsft, line 282: Something seems off here as the model doesn't have any constructor parameters but is using the constructor style deserialization with having variables for all the deserialized values. Won't fix in this PR as the behavior hasn't changed but this can get cleaned up.

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/UserAssignedIdentities.java
```diff
@@ -70,7 +69,7 @@ public UserAssignedIdentities withAdditionalProperties(Map<String, UserAssignedI
     @JsonAnySetter
     void withAdditionalProperties(String key, UserAssignedIdentity value) {
         if (additionalProperties == null) {
-            additionalProperties = new HashMap<>();
+            additionalProperties = new LinkedHashMap<>();
```
haolingdong-msft, line 72: Seems CI fails due to `LinkedHashMap` not imported.

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
```diff
@@ -24,43 +24,48 @@ public class Fish implements JsonSerializable<Fish> {
...
     @Generated
-    private String id;
+    String id;
 
     /*
      * The name property.
      */
     @Generated
-    private String name;
+    String name;
```
haolingdong-msft, line 39: Using `package-private` for the properties and remove the setters in the `fromJson` can solve the issue. But this could leave chance for user to inherit this model, would like to hear @srnagar's thoughts as well.

typespec-tests/src/main/java/com/cadl/flatten/implementation/JsonMergePatchHelper.java
```diff
@@ -32,6 +26,12 @@ public static UpdatePatchRequestAccessor getUpdatePatchRequestAccessor() {
         return updatePatchRequestAccessor;
     }
 
+    private static TodoItemPatchAccessor todoItemPatchAccessor;
```
haolingdong-msft, line 29: double confirm: This change is just switch order?

typespec-tests/src/main/java/com/cadl/patch/models/Fish.java
weidongxu-microsoft: Making it package private likely going to cause problem, when class be in different packages...

alzimmermsft: Since it seems we've aimed to make json-merge-patch an implementation detail, even if we use the access helper pattern that'd still be internal only. So, we may be at a point where we either continue with internal only and don't allow customer extension or we need to start adding APIs to the public surface area to support extension.

typespec-tests/src/main/java/com/cadl/flatten/implementation/JsonMergePatchHelper.java
alzimmermsft: Most of this is reordering so the field, interface, and static methods for a type are together instead of the normal layout where all field declarations are together, then interfaces, and finally static methods. I made this change as if you're going to the helper class you are likely caring about a specific access helper and this makes it easier to find all the information for it.

typespec-tests/src/main/java/com/cadl/patch/models/Salmon.java
```diff
@@ -261,47 +246,33 @@ private JsonWriter toJsonMergePatch(JsonWriter jsonWriter) throws IOException {
...
-            deserializedSalmon.setId(id);
-            deserializedSalmon.setName(name);
-            deserializedSalmon.setAge(age);
-            deserializedSalmon.setColor(color);
-            deserializedSalmon.kind = kind;
-            deserializedSalmon.friends = friends;
-            deserializedSalmon.hate = hate;
-            deserializedSalmon.partner = partner;
 
             return deserializedSalmon;
```
weidongxu-microsoft, line 277: I guess I am late.

One possible (simpler) alternative could be call `updatedProperties.clear()` in the end of `fromJson`, and re-add the `kind` to the dict.

weidongxu-microsoft (commented):
@haolingdong-msft 
You are the owner of patch area. We will need you to sign off.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
```diff
@@ -38,76 +42,101 @@ public void write(List<ClientModel> models, JavaFile javaFile) {
...
-            });
-        }
-    }
+        for (ClientModel model : models) {
+            if (!model.getImplementationDetails().isInput()) {
+                // Model is only used as output and doesn't need to support json-merge-patch.
+                continue;
+            }
 
+            if (model.isPolymorphic() && CoreUtils.isNullOrEmpty(model.getDerivedModels())) {
```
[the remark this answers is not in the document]
alzimmermsft: This should work in this scenario and the intermediate polymorphic parent will need accessors for all its new properties so that `level2Terminal` can set those values during deserialization.

I'll update `patch.tsp` to add this scenario so it can be validated.

typespec-tests/src/main/java/com/cadl/patch/models/Salmon.java
alzimmermsft: That could work as well. I'll continue with the access helper pattern though as we may need this elsewhere and this extends the good baseline @haolingdong-msft set up.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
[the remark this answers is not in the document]
weidongxu-microsoft: Yeah, I think current logic is good.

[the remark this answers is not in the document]
weidongxu-microsoft: What's the difference of this line and `model.isPolymorphicParent()`?

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
```diff
@@ -425,55 +433,37 @@ private JsonWriter toJsonMergePatch(JsonWriter jsonWriter) throws IOException {
...
                 } else if ("map".equals(fieldName)) {
-                    map = reader.readMap(reader1 -> InnerModel.fromJson(reader1));
+                    Map<String, InnerModel> map = reader.readMap(reader1 -> InnerModel.fromJson(reader1));
+                    deserializedResource.map = map;
                 } else if ("longValue".equals(fieldName)) {
-                    longValue = reader.getNullable(JsonReader::getLong);
+                    deserializedResource.longValue = reader.getNullable(JsonReader::getLong);
                 } else if ("intValue".equals(fieldName)) {
-                    intValue = reader.getNullable(JsonReader::getInt);
+                    deserializedResource.intValue = reader.getNullable(JsonReader::getInt);
```
haolingdong-msft, line 453: I’m not sure if we want to use the accessor pattern in ‘fromJson’ both for polymorphic json merge patch models and non-polymorphic json merge patch models. I tend to use accessor pattern for both, but not strong opinion.

haolingdong-msft:
The getLength() change  to fire serialization and removing extra code in the accessor look good to me.

javagen/src/main/java/com/azure/autorest/template/JsonMergePatchHelperTemplate.java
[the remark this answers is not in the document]
alzimmermsft: Hmm, doesn't seem like too much now that I look at it again. I got caught up on `model.isPolymorphicParent()` checking both `model.isPolymorphic() && !CoreUtils.isNullOrEmpty(model.getDerivedModels())`, so it ends up being the same in this case. I likely overlooked this as my first design used `!model.isPolymorphicParent()` by itself, which would have been wrong here as non-polymorphic models would have passed the check.

javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
```diff
@@ -227,16 +230,11 @@ public ClientModelPropertiesManager(ClientModel model, JavaSettings settings) {
...
-        boolean readOnlyConstructorProperties = settings.isRequiredFieldsAsConstructorArgs()
-            && settings.isIncludeReadOnlyInConstructorArgs()
-            && (!CoreUtils.isNullOrEmpty(readOnlyProperties) || !CoreUtils.isNullOrEmpty(superReadOnlyProperties));
-
         this.hasRequiredProperties = hasRequiredProperties;
         this.requiredPropertiesCount = requiredProperties.size() + superRequiredProperties.size();
         this.setterPropertiesCount = setterProperties.size() + superSetterProperties.size();
         this.readOnlyPropertiesCount = readOnlyProperties.size() + superReadOnlyProperties.size();
-        this.hasConstructorArguments = requiredConstructorProperties || readOnlyConstructorProperties;
+        this.hasConstructorArguments = hasConstructorArguments;
```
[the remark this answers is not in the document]
alzimmermsft: This is a small logic change which fixed the case in patch models where in the `fromJson` method it created all the local variables when the model could be used directly. This logic was before the "required on create isn't included in the constructor for patch models" was added. Now, stream-style deserialization looks for "are there any constructor parameters" before using the local variables in deserialization.

Here's an example of what this changed: https://github.com/Azure/autorest.java/pull/2775/files#diff-a60642f5b3afc501b9a301c091e0d59a612fcad22726257089ff0dc9b9ae6bfeR199

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
alzimmermsft: For now, I'd say let's go with using the access pattern for polymorphic properties only in JSON merge patch models for now and if needed we can use it everywhere. There's an additional change we'd want to make before enabling this everywhere which is determining if all the generated models are in the same package, if so package-private can be used, otherwise access helpers are needed. I didn't do that for JSON merge patch as:

1. It already was using access helpers to enable JSON merge patch serialization.
2. There aren't many cases of polymorphism + JSON merge patch, so the code noise is limited in scope.

Non-JSON merge patch polymorphism is much more common and I want to hold off on that larger change for a bit. And the non-JSON merge patch polymorphism can use the public setters of the parent models (assuming non-readonly, which won't work at this time anyway) to set the value.

javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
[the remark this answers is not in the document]
haolingdong-msft: I see.

typespec-tests/src/main/java/com/cadl/patch/models/Resource.java
haolingdong-msft: I'm fine with it.

haolingdong-msft (approved):

PR #2781, opened 2024-05-22
Support multi-level polymorphism

alzimmermsft:
Adds support for multi-level polymorphism, only with stream style serialization.


weidongxu-microsoft (approved):

haolingdong-msft:
The changes look good to me. 
Would you update the `PatchTest` to include the multi-level polymorhism case? It can be in this pr or the patch [pr ](https://github.com/Azure/autorest.java/pull/2775#pullrequestreview-2075505204)https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/test/java/com/cadl/patch/PatchClientTest.java#L123

haolingdong-msft (approved):

weidongxu-microsoft:
One thing, is that I never tried multi-level discriminator from Swagger. I cannot say whether m4 gives the same code-model.

Issue #2719, opened 2024-04-23
Feature request: Client builders validate required parameters to build the client


billwert:
Today, `build(Async)Client()` method implementations on builders for service clients do not validate any state. This results in it being possible to build a client that will error when a service call is made. 

Given this:
```
@server("https://example.com/{route}", "My endpoint", {
    route: string,
    otherVar?: string
})
```
I would expect `buildClient()` to throw `InvalidStateException` if I hadn't called `setRoute(...)`.

I'm not entirely sure that optional parameters for services make sense but tsp doesn't complain about it. :)

weidongxu-microsoft:
1st iterate on design https://github.com/Azure/azure-sdk-for-java/pull/39889/files
Would sync with Srikanta next week.

PR #2783, opened 2024-05-24
client, add validateClient to ClientBuilder

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2719

code change mainly in ServiceClientBuilderTemplate.java
https://github.com/Azure/autorest.java/pull/2783/files#diff-011f78c981243301193df6b93ff9aa9373f5486e722793f3ded47bd5425fcbf8

example of generated ClientBuilder
https://github.com/weidongxu-microsoft/autorest.java/blob/client_dpg-validate-builder/typespec-tests/src/main/java/com/resiliency/servicedriven/ResiliencyServiceDrivenClientBuilder.java#L258-L279

---

Would like advice on where we invoke the `validateBuilder()`.
Context: https://github.com/Azure/autorest.java/pull/2783#discussion_r1612764904
Basically, some variable like `credential` is on HttpPipeline, that should not be validated, if user provide their own HttpPipeline. And some variable like `endpoint` is not on HttpPipeline, that can always be validated when creating the client.
However, I don't want to make it more complex by providing a flag or 2 validate method (say `validatePipelineParameter`, `validateNonPipelineParameter`).
^^^
Updated to `validateClient` alone.
If dev need to do similar of `validatePipeline` for credential, they may do such
```java
if (this.pipeline == null) {
  // validate credential
}
```


javagen/src/main/java/com/azure/autorest/template/ServiceClientBuilderTemplate.java
```diff
@@ -516,6 +529,20 @@ protected void addCreateHttpPipelineMethod(JavaSettings settings, JavaClass clas
...
+        addGeneratedAnnotation(classBlock);
+        classBlock.privateMethod("void validateBuilder()", methodBlock -> {
+            methodBlock.line("// This method is invoked from 'createHttpPipeline' when preparing the HTTP pipeline for the new client.");
+            methodBlock.line("// Developer can customize this method, to validate that the necessary conditions are met for the new client.");
+            for (ServiceClientProperty property : properties) {
+                // ServiceVersion is also provided with a defaultValueExpression
+                if (property.isRequired() && property.getDefaultValueExpression() == null) {
+                    methodBlock.line("Objects.requireNonNull(" + property.getName() + ", \"'" + property.getName() + "' cannot be null.\");");
+                }
+            }
```
weidongxu-microsoft, line 542: We can choose not to automatically add the nonNull validation.

javagen/src/main/java/com/azure/autorest/template/TemplateHelper.java
```diff
@@ -41,6 +41,7 @@ public static String getByteCloneExpression(String propertyName) {
     }
 
     public static void createHttpPipelineMethod(JavaSettings settings, String defaultCredentialScopes, SecurityInfo securityInfo, PipelinePolicyDetails pipelinePolicyDetails, JavaBlock function) {
+        function.line("this.validateBuilder();");
```
weidongxu-microsoft, line 44: `validateBuilder()` done in `createHttpPipeline()`.

One reason is that if user provide a `pipeline` instead, validation about "credential" does not need to happen (user could provide an HTTP pipeline that already integrated the credentials into pipeline.

However, the disadvantage here is that the validation on e.g. `endpoint` will also be skipped, if user provides the pipeline.

azure-tests/src/main/java/fixtures/azureparametergrouping/AutoRestParameterGroupingTestServiceBuilder.java
```diff
@@ -241,8 +241,15 @@ public AutoRestParameterGroupingTestService buildClient() {
         return client;
     }
 
+    @Generated
+    private void validateBuilder() {
```
alzimmermsft, line 245: How often do we expect that developers will be customizing this method? I just wonder if it should be generated if there won't be any validations in it.

azure-tests/src/main/java/fixtures/custombaseuri/paging/AutoRestParameterizedHostTestPagingClientBuilder.java
alzimmermsft, line 1: Not really attached to this file but something in all the builders.

We should replace

```java
HttpHeaders headers = new HttpHeaders();
        localClientOptions.getHeaders()
            .forEach(header -> headers.set(HttpHeaderName.fromString(header.getName()), header.getValue()));
        if (headers.getSize() > 0) {
            policies.add(new AddHeadersPolicy(headers));
        }
```

with `CoreUtils.createHttpHeadersFromClientOptions` as it has additional optimizations available due to it being within `azure-core` which can access some module specific things with headers.

javagen/src/main/java/com/azure/autorest/template/TemplateHelper.java
alzimmermsft: I wonder if we should have two methods for validation:

- `validateClient()` which validates anything used to create the client, but not in the pipeline. Ex, endpoint, service version, etc. Or, maybe this is generated into the client constructor.
- `validatePipeline()` which validates anything used to create the pipeline. Ex, credential information, etc.

azure-tests/src/main/java/fixtures/azureparametergrouping/AutoRestParameterGroupingTestServiceBuilder.java
weidongxu-microsoft: From Bill, he mostly wants the validation on `endpoint` (which is usually "required" in ClientBuilder).

The other common case that I can think of, is for the credential. E.g. when SDK accept both KeyCredential and TokenCredential, client likely want to validate that one and only one of them is provided in ClientBuilder.

javagen/src/main/java/com/azure/autorest/template/TemplateHelper.java
weidongxu-microsoft: Emm, yes, that should work.
`validatePipeline()` would be called in `createHttpPipeline()`, `validateClient()` would be called in `createInnerClient()`.

@billwert Let us know your opinion as well.

azure-tests/src/main/java/fixtures/azureparametergrouping/AutoRestParameterGroupingTestServiceBuilder.java
weidongxu-microsoft: https://github.com/Azure/autorest.java/pull/2785

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/dpg/DocumentModelsClientBuilder.java
```diff
@@ -262,8 +262,16 @@ private FormRecognizerClientImpl buildInnerClient() {
         return client;
     }
 
+    @Generated
+    private void validateBuilder() {
+        // This method is invoked from 'createHttpPipeline' when preparing the HTTP pipeline for the new client.
+        // Developer can customize this method, to validate that the necessary conditions are met for the new client.
```
haolingdong-msft, line 268: Currently we will validate required fields not be null?

weidongxu-microsoft: In current (main) code, we won't.

In this PR, see if we can add the default validation code, for required property without default value -- e.g. (typical case) the endpoint.

azure-tests/src/main/java/fixtures/custombaseuri/paging/AutoRestParameterizedHostTestPagingClientBuilder.java
weidongxu-microsoft: done in a separate PR https://github.com/Azure/autorest.java/pull/2785

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/dpg/DocumentModelsClientBuilder.java
haolingdong-msft: I mean this pr. Are we validating all the required property to be non-null, or only `endpoint` this time?

weidongxu-microsoft: To all "required"

javagen/src/main/java/com/azure/autorest/template/TemplateHelper.java
weidongxu-microsoft: I would go `validateClient()` + `validatePipeline()`, if no further feedback.

weidongxu-microsoft: Change to `validateClient()` alone, after discussion with Srikanta.

The thing in `validatePipeline()` maybe done as
```java
if (this.pipeline == null) {
  // validate credential
}
```
in `validateClient()`.

billwert: yeah, I think just having `validateClient` is great. (Sorry I missed this ping!)

azure-tests/src/main/java/fixtures/azureparametergrouping/AutoRestParameterGroupingTestServiceBuilder.java
billwert: I think generating it empty (if that happens) is fine; I assume the compiler is smart enough to elide empty function calls..

haolingdong-msft:
`validateClient()` looks good to me. Just think of below:
1. wonder is there any reason to remove `validatePipeline()`
2. Later, we could collect the use case of validating clients, e.g. validate non-null fields. Provide examples in our docs to recommend user how to write good validation.

haolingdong-msft (approved):

PR #2786, opened 2024-05-27
tsp, visibility to m4 x-ms-mutable

weidongxu-microsoft:
Codegen for fluent uses the x-ms-mutable, as code in create flow and update flow is separate and could be different. Therefore, emitter need to fill it (from `@visibility`).
https://github.com/Azure/autorest/blob/main/docs/extensions/readme.md#x-ms-mutability

However, the change would not affect generated code, as typespec-azure-resource-manager lib already filtered these properties (and separate the properties for update to ##UpdateProperties).
https://github.com/Azure/typespec-azure/blob/typespec-azure%400.42.0/packages/typespec-azure-resource-manager/lib/arm.foundations.tsp#L91-L119

1st and 2nd commit just to see if we can flatten the ##UpdateProperties <-- yes, we can

code https://github.com/Azure/autorest.java/pull/2786/commits/fe751ce11216898029c1b9fdca984b2dbc9c95ac


typespec-tests/tsp/arm.tsp
```diff
@@ -84,6 +84,8 @@ model TopLevelArmResourceProperties {
   userName: string; // this should be renamed to username
   userNames: string; // this isn't a whole word match, shouldn't be renamed to usernames
   accuserName: string; // this isn't a whole word match, shouldn't be renamed to accusername
+
+  @visibility("create", "read")
   startTimeStamp: utcDateTime; // this should be renamed to startTimestamp
```
weidongxu-microsoft, line 89: Test for mutability

code-model
```
                                - language:
                                    default:
                                      name: startTimeStamp
                                      description: ''
                                  protocol: {}
                                  schema: &ref_43
                                    language:
                                      default:
                                        name: >-
                                          TopLevelArmResourcePropertiesStartTimeStamp
                                        description: ''
                                    protocol: {}
                                    type: date-time
                                    format: date-time
                                  serializedName: startTimeStamp
                                  required: true
                                  nullable: false
                                  readOnly: false
                                  extensions:
                                    x-ms-mutability:
                                      - create
                                      - read
```

XiaofeiCao (approved):

Issue #2731, opened 2024-04-29
TSP, TCGC common layer, use TCGC returned name as much as possible


haolingdong-msft:
pay attention to union name, and optional literal name

depends on TCGC issue: https://github.com/Azure/typespec-azure/issues/553

PR #2787, opened 2024-05-27
tsp, use TCGC's name for model property schema and parameter name

haolingdong-msft:
Fix https://github.com/Azure/autorest.java/issues/2731

Code Changes:
1. use TCGC's name for model property schema name, this would apply to optional literal case.
2. use TCGC's parameter name


weidongxu-microsoft (approved):

typespec-tests/src/main/java/com/cadl/literalservice/models/ModelOptionalLiteralOptionalLiteral.java
```diff
@@ -5,35 +5,35 @@
 package com.cadl.literalservice.models;
 
 /**
- * Defines values for ModelOptionalLiteral.
+ * Defines values for ModelOptionalLiteralOptionalLiteral.
  */
-public enum ModelOptionalLiteral {
+public enum ModelOptionalLiteralOptionalLiteral {
```
haolingdong-msft, line 10: Hi @srnagar, we now leverage TCGC to generate the name for optional literal enum. And their logic is to concat the literal value to the enum's name. e.g. the definitiion is `optionalLiteral?: "optionalLiteral"`, so another `OptionalLiteral` is append to the name. Please let me know if this is acceptable or not.

weidongxu-microsoft:
Can we now remove the Model/Scalar related types from `getName`, as we should be using TCGC names from Model?
https://github.com/Azure/autorest.java/blob/894501332ea54568eb091621f17ffe8895fa57d2/typespec-extension/src/code-model-builder.ts#L2246

Also, if above is true, we should be able to remove the "template" related code and "empty name" related code as well.
https://github.com/Azure/autorest.java/blob/894501332ea54568eb091621f17ffe8895fa57d2/typespec-extension/src/code-model-builder.ts#L2274-L2291

haolingdong-msft:
> Can we now remove the Model/Scalar related types from `getName`, as we should be using TCGC names from Model?
> 
> https://github.com/Azure/autorest.java/blob/894501332ea54568eb091621f17ffe8895fa57d2/typespec-extension/src/code-model-builder.ts#L2246
> 
> 
> We probably still need `ModelProperty` and `Operation` as they relates to op.
> Also, if above is true, we should be able to remove the "template" related code and "empty name" related code as well.
> 
> https://github.com/Azure/autorest.java/blob/894501332ea54568eb091621f17ffe8895fa57d2/typespec-extension/src/code-model-builder.ts#L2274-L2291

Thanks for the suggestion. I tried locally, we can remove `Scalar` and `undefined`, but probabyly cannot remove `Model` for now, because there are places that uses `Model` type: 1. https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L1590 2. https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L1341 (body is `Model | ModelProperty`)

weidongxu-microsoft:
#1 is just a doc, and it is already a Model, you can get TCGC and use its name
#2 is the uncompleted migrate, it should use TCGC model name, if Union is Model (you can call getName if it is ModelProperty; just do a if/else)

haolingdong-msft:
Yes, I can do `if/else`, just thinking we can get rid of the whole `getName` function when integrating with `sdkPackage`, so I don't have very strong intention of modifying the logics. But I'm fine of doing this in this pr.

weidongxu-microsoft:
> Yes, I can do `if/else`, just thinking we can get rid of the whole `getName` function when integrating with `sdkPackage`, so I don't have very strong intention of modifying the logics. But I'm fine of doing this in this pr.

If you can finish the work next week, I am good with "no very strong intention". If you cannot finish next month, it is good incentive for me to clean that up.
The point is this `getName` could be vastly simplified, if it does not need to work on Model.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2293,24 +2283,6 @@ export class CodeModelBuilder {
...
 
-    // if no projectedName and friendlyName found, return the name of the target (including special handling for template)
-    if (
-      target.kind === "Model" &&
-      target.templateMapper &&
-      target.templateMapper.args &&
-      target.templateMapper.args.length > 0
-    ) {
-      const tspName = getTypeName(target, this.typeNameOptions);
-      const newName = getNameForTemplate(target);
```
weidongxu-microsoft, line 2304: remove `getNameForTemplate` and related func

PR #2799, opened 2024-05-30
Fixes for multi-level discriminators

alzimmermsft:
Fixes a few bugs for multi-level discriminator structures:

- Types that change the polymorphic discriminator generate getters for all discriminators.
- Enum discriminators now use the right type when there are multiple levels of discriminators.


weidongxu-microsoft (approved):

XiaofeiCao (approved):

XiaofeiCao:
Thanks Alan!

Issue #2800, opened 2024-05-31
tsp, mgmt, bug on LRO POST action in ARM


weidongxu-microsoft:
The response / final result type is wrong.

![image](https://github.com/Azure/autorest.java/assets/53292327/28261c87-a0b4-475d-8b72-47f012795d01)

(source https://github.com/Azure/azure-rest-api-specs/pull/28981)

We need different handling on ARM/DPG

PR #2801, opened 2024-05-31
tsp, fix response of LRO POST action in ARM

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2800

Logic:
- poll/final result of LRO in ARM is the response schema of the activation request (same logic as Swagger)
- poll/final result of LRO in DPG is the `pollResultType`/`finalResultType` from `LongRunningMetadata`; the schema of activation request response is not public and only exists as `BinaryData`

code change in code-model-builder.ts https://github.com/Azure/autorest.java/pull/2801/files#diff-14fdfb2e0df46da414f4d04a218f03e4e5f2737d3a9672b9c4e4feb16419445e

test added to arm.tsp https://github.com/Azure/autorest.java/pull/2801/files#diff-0a599b462ecd977230f46a86db322b6b53c1c0f498b98279cbae7ff58cff13dd


typespec-extension/src/code-model-builder.ts
```diff
@@ -1570,53 +1574,17 @@ export class CodeModelBuilder {
...
-                schema = candidateResponseSchema;
-                const sdkType = getClientType(this.sdkContext, bodyType);
-                const modelName = sdkType.kind === "model" ? sdkType.name : undefined;
-                this.trace(
-                  `Replace TypeSpec model '${modelName}' with '${candidateResponseSchema.language.default.name}'`,
-                );
-              }
-            }
-          } else if (verb === "delete") {
-            // for LRO ResourceDelete, final type will be replaced to "Void" in convenience API, hence do not generate the class
```
weidongxu-microsoft, line 1608: remove as we should now let LroMetadata handle this.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1570,53 +1574,17 @@ export class CodeModelBuilder {
...
-              if (match) {
-                schema = candidateResponseSchema;
-                const sdkType = getClientType(this.sdkContext, bodyType);
-                const modelName = sdkType.kind === "model" ? sdkType.name : undefined;
-                this.trace(
-                  `Replace TypeSpec model '${modelName}' with '${candidateResponseSchema.language.default.name}'`,
-                );
-              }
-            }
-          } else if (verb === "delete") {
```
weidongxu-microsoft, line 1607: remove this would cause some minor Javadoc change on response, e.g.
https://github.com/Azure/autorest.java/pull/2801/files#diff-d5c9a91efac5de9c40e886638c59257e53f04b3fd9670ef7a8464daeb1866351

typespec-extension/src/code-model-builder.ts
```diff
@@ -871,6 +871,10 @@ export class CodeModelBuilder {
   private processLroMetadata(op: CodeModelOperation, httpOperation: HttpOperation): LongRunningMetadata {
     const operation = httpOperation.operation;
 
+    // ARM uses the response of activation request as poll/final result.
+    // Therefore these schema will not be public.
+    const trackConvenienceApi: boolean = Boolean(op.convenienceApi) && !this.isArm();
```
haolingdong-msft, line 876: So ARM will not use `getLroMetadata` to calculate poll/final result? Will `getLroMetadata` enhance later to support ARM as well?

weidongxu-microsoft: ARM still uses `getLroMetadata`.

I may need to double check whether in all ARM cases (CRUD+action), the response of activation request is always same as the poll/result response of that `getLroMetadata`. If it is true, we only need to modify the DPG path.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1570,53 +1574,17 @@ export class CodeModelBuilder {
...
-          } else if (verb === "delete") {
-            // for LRO ResourceDelete, final type will be replaced to "Void" in convenience API, hence do not generate the class
+        if (candidateResponseSchema) {
+          // The operation is LRO
+          if (trackConvenienceApi && !this.isArm()) {
+            // DPG uses the LroMetadata as poll/final result, not the response of activation request
+            // Therefore this schema will not be public.
             trackConvenienceApi = false;
           }
         }
```
XiaofeiCao, line 1584: Not sure I understand the logic here. Why do we check whether `candidateResponseSchema`(which should be `lroMetadata.pollResultType`) exists before we set convenience usage for activation response type?

weidongxu-microsoft: After the removal of the code, `candidateResponseSchema` basically same as `isLro`

XiaofeiCao: Got it. Seems it can be simplified to `if (candidateResponseSchema && !this.isArm()){...}`.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: ok, I see.

XiaofeiCao (approved):

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: I've changed the signature to `longRunning: boolean`

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Appears I can unify the ARM/DPG (to DPG code flow), without affect existing ARM cases.

Hope this can stay correct while we add new ARM cadl-ranch cases.

haolingdong-msft (approved):

PR #2807, opened 2024-06-07
mgmt, verify operationId works in example

weidongxu-microsoft:


XiaofeiCao (approved):
Do we want to support `@OperationId` in the future?

weidongxu-microsoft:
We have no plan to support `@operationId` at any other places, in SDK emitter.

But the matching of example and operationId is done at openapi level, so for this we had to read it.

PR #2809, opened 2024-06-10
Migrate all tests to JUnit 5

### LINKED ISSUE #1253: Migrate unit tests for AutoRest Java from JUnit 4 to Junit 5

alzimmermsft:
Fixes #1253 

Migrate tests to JUnit 5, as part of this migration it was realized many tests weren't running due to them using JUnit 4 configurations but the vintage JUnit engine wasn't included so they weren't being ran. As part of migrating to JUnit 5 they began running again and a few errors in code generation were found.

- stream-style serialization wasn't removing escape `\.` in JSON flattening scenarios.
- stream-style serialization had an incorrect interpretation of lists in XML where serialization would generate in an XML wrapped format even if `wrapped: true` wasn't present. Now if there is a list in XML without `wrapped: true` it will consume the values using the XML element name as OpenAPI 2.0 specifies.
- stream-style serialization wasn't try-catch'ing URL creation in XML serialization. XmlSerializable only checks XMLStreamException and not IOException which URL creation throws.

And some cases of code generation disabled stream-style serialization as they were testing Jackson annotation cases with JSON flattening.

Some tests were disabled due to either incorrect behavior with Java object types and disallowed null values (this won't throw as boxed typed are allowed to be null and there is no logic generated to prevent this) and direct usage of enum types in serialization as stream-style falls back to Jackson for non-JsonSerializable and non-XmlSerializable scenarios and enums are handled as their name and not toString value by default.


weidongxu-microsoft (approved):

vanilla-tests/src/test/java/fixtures/bodyfile/FilesTests.java
```diff
@@ -40,7 +40,8 @@ public void getFile() throws Exception {
         assertArrayEquals(expected, actual);
     }
 
-    @Ignore("This is causing OOM error with upgraded version of core https://github.com/Azure/autorest.java/issues/980")
+    @Disabled(
+        "This is causing OOM error with upgraded version of core https://github.com/Azure/autorest.java/issues/980")
```
srnagar, line 44: This issue seems to be resolved. We should enable this test again.

vanilla-tests/src/test/java/fixtures/lro/LROsTests.java
```diff
@@ -4,48 +4,47 @@
...
-//    }
+    //
+    //    @Test
+    //    public void beginPatch200SucceededIgnoreHeaders() throws Exception {
+    //        Product product = new Product();
+    //        product.setLocation("West US");
+    //
+    //        SyncPoller<Product, Product> poller = client.getLROs().beginPatch200SucceededIgnoreHeaders(product);
+    //        assertEquals("Succeeded", poller.getFinalResult().getProvisioningState());
+    //    }
```
srnagar, line 44: Delete these commented lines.

vanilla-tests/src/test/java/fixtures/lro/LROsTests.java
```diff
@@ -93,19 +92,19 @@ public void beginPut200SucceededNoState() throws Exception {
...
+    //    @Test
+    //    public void beginPut202Retry200() throws Exception {
+    //        Product product = new Product();
+    //        product.setLocation("West US");
+    //
+    //        SyncPoller<Product, Product> poller = client.getLROs().beginPut202Retry200(product);
+    //        Product actual = poller.getFinalResult();
+    //        assertEquals("100", actual.getId());
+    //        assertEquals("foo", actual.getName());
+    //    }
```
srnagar, line 104: Delete

androidgen/pom.xml
```diff
@@ -37,13 +37,6 @@
       <groupId>com.azure</groupId>
       <artifactId>azure-core</artifactId>
     </dependency>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <version>4.13.2</version>
-      <scope>test</scope>
-    </dependency>
```
srnagar, line 45: This is a much needed upgrade from JUnit 4 to JUnit 5. Thanks for making these changes!

srnagar (approved):

PR #2811, opened 2024-06-12
client, graalvm, error model alone need reflect, if stream-style-serialization

weidongxu-microsoft:
We don't need reflection for model/enum classes, if using azure-json.

But we still need reflection for error/exception classes, as azure-core initialize them.

Tested on compute, which have error model as `ApiError` and `InnerError`


haolingdong-msft:
No test is impacted?

weidongxu-microsoft:
> No test is impacted?

We only generate these (by default) for mgmt, and currently ths only mgmt tsp test has `stream-style-serialization=false`.

We should see them if there is new mgmt tsp test (i.e. Xiaofei's unmerged one in cadl-ranch).

XiaofeiCao:
I added one when enabling stream-style for mgmt:
https://github.com/Azure/autorest.java/blob/main/typespec-tests/tsp/arm-stream-style-serialization.tsp

Seems no reflect config is generated. 
https://github.com/Azure/autorest.java/blob/main/typespec-tests/src/main/resources/META-INF/native-image/com.azure.resourcemanager/azure-resourcemanager-armstreamstyleserialization-generated/reflect-config.json

Though I don't have error types there.

weidongxu-microsoft:
Yeah, for most common cases, generated SDK directly use ManagementError from azure-core-manangment, so it won't generate error model.

haolingdong-msft (approved):

PR #2816, opened 2024-06-14
tsp, bump dependency

weidongxu-microsoft:
Main change
1. request body of multipart <- I will do a follow-up PR, but probably need to wait for TCGC update first (as we uses their model now)
2. nullable type change from TCGC


typespec-extension/src/code-model-builder.ts
```diff
@@ -1018,7 +1020,7 @@ export class CodeModelBuilder {
     } else {
       // schema
       let schema;
-      const sdkType = getClientType(this.sdkContext, param.param);
+      const sdkType = getNonNullSdkType(getClientType(this.sdkContext, param.param));
```
weidongxu-microsoft, line 1023: For parameter, the `nullable` is from `param.param.type` (the raw type).

I didn't change the logic.

typespec-extension/src/code-model-builder.ts
```diff
@@ -789,9 +791,9 @@ export class CodeModelBuilder {
     this.addAcceptHeaderParameter(codeModelOperation, op.responses);
     // body
     if (op.parameters.body) {
-      if (op.parameters.body.parameter) {
-        if (!isVoidType(op.parameters.body.parameter.type)) {
-          this.processParameterBody(codeModelOperation, op, op.parameters.body.parameter);
+      if (op.parameters.body.property) {
+        if (!isVoidType(op.parameters.body.property.type)) {
+          this.processParameterBody(codeModelOperation, op, op.parameters.body.property);
```
weidongxu-microsoft, line 796: `.parameter` is deprecated, changed to `property`

typespec-tests/src/main/java/com/cadl/flatten/implementation/models/SendLongRequest.java
```diff
@@ -71,7 +71,7 @@ public final class SendLongRequest implements JsonSerializable<SendLongRequest>
      * The status of the todo item
      */
     @Generated
-    private final SendLongRequestStatus status;
+    private final TodoItemStatus status;
```
weidongxu-microsoft, line 74: Unsure about this rename.

typespec-tests/src/main/java/com/cadl/literalservice/models/Model.java
```diff
@@ -27,7 +27,7 @@ public final class Model implements JsonSerializable<Model> {
      * The optionalLiteral property.
      */
     @Generated
-    private ModelOptionalLiteralOptionalLiteral optionalLiteral;
+    private ModelOptionalLiteral optionalLiteral;
```
weidongxu-microsoft, line 30: expected from naming change of constant

typespec-tests/src/main/java/com/type/union/IntsOnlyAsyncClient.java
```diff
@@ -120,7 +120,7 @@ public Mono<GetResponse6> get() {
      */
     @Generated
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public Mono<Void> send(GetResponseProp3 prop) {
+    public Mono<Void> send(GetResponseProp2 prop) {
```
weidongxu-microsoft, line 123: Again :-(

XiaofeiCao (approved):

haolingdong-msft (approved):

typespec-tests/src/main/java/com/type/union/IntsOnlyAsyncClient.java
haolingdong-msft: /cc @tadelesh Seems the anonymous model name's suffix number changes again.

typespec-tests/src/main/java/com/cadl/flatten/implementation/models/SendLongRequest.java
haolingdong-msft: Yeah, seems due to TCGC change. But this name makes sense to me. 
BTW, I think it's from [`TodoItem`](https://github.com/Azure/autorest.java/blob/main/typespec-tests/tsp/flatten.tsp#L69) model, instead of [`TodoItemPatch`](https://github.com/Azure/autorest.java/blob/main/typespec-tests/tsp/flatten.tsp#L93)

typespec-tests/src/main/java/com/cadl/flatten/implementation/FlattenClientImpl.java
```diff
@@ -257,7 +257,7 @@ Response<Void> uploadFileSync(@HostParam("endpoint") String endpoint, @PathParam
         @UnexpectedResponseExceptionType(value = ResourceModifiedException.class, code = { 409 })
         @UnexpectedResponseExceptionType(HttpResponseException.class)
         Mono<Response<Void>> uploadTodo(@HostParam("endpoint") String endpoint,
-            @HeaderParam("content-type") String contentType, @HeaderParam("accept") String accept,
+            @HeaderParam("Content-Type") String contentType, @HeaderParam("accept") String accept,
```
haolingdong-msft, line 260: Is this due to TCGC change? Guessing the lower-case or upper-case do not matter here?

typespec-tests/tsp/nested-enum-discriminator.tsp
```diff
@@ -66,183 +66,183 @@ model GoblinShark extends Shark {
 @scenario
 @route("/model")
 @scenarioDoc("""
-Generate and receive polymorphic model in multiple levels inheritance with 2 discriminators.
```
haolingdong-msft, line 69: Is there any change to this file? Or just format?

weidongxu-microsoft: Just format.

Probably due to update of `@typespec/prettier-plugin-typespec`

typespec-tests/src/main/java/com/cadl/flatten/implementation/FlattenClientImpl.java
weidongxu-microsoft: More probably dependency lib change like http lib

tsp is
https://github.com/Azure/autorest.java/blob/9c3d8ed0575d107f2e2519033ba67625ecdade85/typespec-tests/tsp/flatten.tsp#L141

If written full it should be `@header("content-type") contentType: "multipart/form-data",`

My guess is that without the explicit header value in `@header`, http lib converts property name `contentType` to header name `Content-Type`.

PR #2818, opened 2024-06-17
mgmt, ExtensionResource as ResourceType

XiaofeiCao:
scvmm: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/scvmm/ScVmm.Management/VirtualMachineInstance.tsp#L14-L15

`ExtensionResource` is a special kind of `ProxyResource`. We don't want to generate an extra `ExtensionResource` class for it.

code: https://github.com/Azure/autorest.java/pull/2818/commits/b32965d204fe44daca398477b5580e357503a618


fluentnamer/src/main/java/com/azure/autorest/fluent/model/ResourceTypeName.java
```diff
@@ -19,6 +20,7 @@ public class ResourceTypeName {
     public static final String SUB_RESOURCE_AUTO_GENERATED = SUB_RESOURCE + AUTO_GENERATED;
     public static final String PROXY_RESOURCE_AUTO_GENERATED = PROXY_RESOURCE + AUTO_GENERATED;
     public static final String TRACKED_RESOURCE_AUTO_GENERATED = TRACKED_RESOURCE + AUTO_GENERATED;
+    public static final String EXTENSION_RESOURCE_AUTO_GENERATED = EXTENSION_RESOURCE + AUTO_GENERATED;
```
weidongxu-microsoft, line 23: We probably don't need it.

m4 for Swagger do this AutoGenerated. Here I guess we won't see this.

weidongxu-microsoft (approved):

fluentnamer/src/main/java/com/azure/autorest/fluent/model/ResourceTypeName.java
XiaofeiCao: Yeah, removed.

Issue #2671, opened 2024-04-05
Method signature (alias / spread) Implementation


lmazuel:

weidongxu-microsoft:
WIP https://github.com/Azure/autorest.java/pull/2821

need TCGC release (so that it avoid `getEffectiveType`)

weidongxu-microsoft:
The update of commit ID to that after `BodyParameter` is done via PR https://github.com/Azure/azure-sdk-for-java/pull/40992

There is some question about how best to customize the name of `body` parameter to e.g. keep backward-compatible (that applies to most languages). https://github.com/Azure/azure-rest-api-specs/issues/29696

weidongxu-microsoft:
0.18.0 released

- merge to SDK repo https://github.com/Azure/azure-sdk-for-java/pull/41080

PR #2821, opened 2024-06-18
tsp, spread implicit request body

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2671

version 0.18.0

Need a new cadl-ranch release
- https://github.com/Azure/cadl-ranch/pull/603

Also need a new TCGC release (reason is that TCGC would call `getEffectiveType` on model, which could map an anonymous model to a named model -- this [in some cases](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/parameters/body-optionality/main.tsp#L43-L64) causes problem)
- https://github.com/Azure/typespec-azure/pull/886


typespec-extension/src/code-model-builder.ts
```diff
@@ -1341,7 +1346,7 @@ export class CodeModelBuilder {
       schema = this.processSchemaFromSdkType(sdkType, body.name);
     }
 
-    const isAnonymousModel = sdkType.kind === "model" && sdkType.isGeneratedName === true;
+    const isAnonymousModel = sdkType.kind === "model" && body.kind === "Model" && !this.isArm();
```
haolingdong-msft, line 1349: why do we remove the condition `sdkType.isGeneratedName === true` for checking anonymous model?
"

weidongxu-microsoft: Because this is no longer about anonymous model.

The feature ask is to treat `op(...Model)` the same, regardless of `alias Model` or `model Model`

typespec-tests/src/main/java/com/client/naming/ClientModelClient.java
weidongxu-microsoft, line 1: This change is due to TCGC 0.43.2

But seems it is correct
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/client/naming/main.tsp#L145-L147

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Rename the var to `bodyParameterFlatten`. Also added comment to make the logic understandable.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1371,7 +1375,7 @@ export class CodeModelBuilder {
       this.trackSchemaUsage(schema, { serializationFormats: [KnownMediaType.Multipart] });
     }
 
-    if (schema instanceof ObjectSchema && isAnonymousModel) {
+    if (schema instanceof ObjectSchema && bodyParameterFlatten) {
       // anonymous model
```
haolingdong-msft, line 1379: Please modify this comment as well

typespec-tests/src/main/java/com/parameters/bodyoptionality/BodyOptionalityClient.java
```diff
@@ -116,9 +118,11 @@ public void requiredExplicit(BodyModel body) {
      */
     @Generated
     @ServiceMethod(returns = ReturnType.SINGLE)
-    public void requiredImplicit(BodyModel bodyModel) {
+    public void requiredImplicit(String name) {
```
weidongxu-microsoft, line 121: This is expected after behavior change.

Test is "requiredImplicit", so I didn't add `@body` to this case in cadl-ranch.

typespec-tests/src/main/java/com/client/naming/ClientModelClient.java
```diff
@@ -15,25 +15,25 @@
...
 import com.client.naming.models.ClientModel;
 import com.client.naming.models.JavaModel;
 
 /**
  * Initializes a new instance of the synchronous NamingClient type.
  */
 @ServiceClient(builder = NamingClientBuilder.class)
-public final class ModelClient {
+public final class ClientModelClient {
     @Generated
```
haolingdong-msft, line 27: curious why the client name changed？

weidongxu-microsoft: https://github.com/Azure/autorest.java/pull/2821#discussion_r1670368435

looks like tcgc bug fix

typespec-tests/src/main/java/com/parameters/spread/implementation/models/SpreadAsRequestBodyRequest1.java
```diff
@@ -0,0 +1,83 @@
...
+import com.azure.json.JsonSerializable;
+import com.azure.json.JsonToken;
+import com.azure.json.JsonWriter;
+import java.io.IOException;
+
+/**
+ * The SpreadAsRequestBodyRequest1 model.
+ */
+@Immutable
+public final class SpreadAsRequestBodyRequest1 implements JsonSerializable<SpreadAsRequestBodyRequest1> {
```
haolingdong-msft, line 19: Seems this model looks similar as [`SpreadAsRequestBodyRequest1`](https://github.com/Azure/autorest.java/blob/0cb0d48d8fa24ce8f7766cd7eea6d00b6fde04d8/typespec-tests/src/main/java/com/parameters/spread/implementation/models/SpreadAsRequestBodyRequest.java)?

weidongxu-microsoft: It is same. But they are from different operation (one in alias og, another in model og), so they are different model (even the content is same).

They probably also different entity in typespec.

haolingdong-msft (approved):

PR #2836, opened 2024-06-25
tsp, set request mediaTypes from body.contentTypes

weidongxu-microsoft:
code https://github.com/Azure/autorest.java/pull/2836/commits/706a4234663348a01fb75c95344c2b72cf0ea8e3

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -135,7 +135,6 @@ import { EmitterOptions } from "./emitter.js";
 import { createPollOperationDetailsSchema, getFileDetailsSchema } from "./external-schemas.js";
 import { ClientContext } from "./models.js";
 import {
-  CONTENT_TYPE_KEY,
   ORIGIN_API_VERSION,
   SPECIAL_HEADER_NAMES,
   cloneOperationParameter,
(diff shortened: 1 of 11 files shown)


typespec-extension/src/code-model-builder.ts
```diff
@@ -1327,6 +1317,9 @@ export class CodeModelBuilder {
   }
 
   private processParameterBody(op: CodeModelOperation, httpOperation: HttpOperation, body: ModelProperty | Model) {
+    // set contentTypes to mediaTypes
+    op.requests![0].protocol.http!.mediaTypes = httpOperation.parameters.body!.contentTypes;
```
weidongxu-microsoft, line 1321: This `body.contentTypes` was added to http lib a few weeks ago (part of multipart PR).

Given this, we should no longer need to use the `content-type` header.

typespec-tests/src/main/java/com/cadl/optional/OptionalAsyncClient.java
```diff
@@ -55,6 +55,8 @@ public final class OptionalAsyncClient {
      * <table border="1">
      * <caption>Header Parameters</caption>
      * <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
+     * <tr><td>Content-Type</td><td>String</td><td>No</td><td>The content type. Allowed values:
+     * "application/json".</td></tr>
```
weidongxu-microsoft, line 59: This appears to be actually a bug fix.

The request body is optional, hence we actually ask user to set the `content-type` header, even this is just 1 content-type.
```java
                if (MethodUtil.getContentTypeCount(operation.getRequests()) == 1 && !isBodyParamRequired) {
                    Parameter contentTypeParameter = MethodUtil.createContentTypeParameter(request, operation);
                    allParameters.add(Mappers.getProxyParameterMapper().map(contentTypeParameter));
                }
```

haolingdong-msft (approved):

PR #2837, opened 2024-06-25
mgmt, update readme template, avoid AZURE_CLIENT_SECRET

weidongxu-microsoft:
we also need to update https://aka.ms/azsdk/java/mgmt related docs.

### DIFF
--- a/fluentgen/src/main/resources/Readme.txt
+++ b/fluentgen/src/main/resources/Readme.txt
@@ -45,15 +45,11 @@ Azure Management Libraries require a `TokenCredential` implementation for authen
 
 ### Authentication
 
-By default, Microsoft Entra ID token authentication depends on correct configuration of the following environment variables.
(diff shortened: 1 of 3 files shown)


fluentgen/src/main/resources/Readme.txt
```diff
@@ -45,15 +45,11 @@ Azure Management Libraries require a `TokenCredential` implementation for authen
 
 ### Authentication
 
-By default, Microsoft Entra ID token authentication depends on correct configuration of the following environment variables.
+Microsoft Entra ID token authentication relies on the [credential class](azure_identity_credentials) from [Azure Identity][azure_identity] package.
```
weidongxu-microsoft, line 48: It's safer to direct user to azure-identity doc..

XiaofeiCao: `[credential class][azure_identity_credentials]` instead of `[credential class](azure_identity_credentials)`?

XiaofeiCao (approved):

PR #2840, opened 2024-06-26
client, include file content on format error

weidongxu-microsoft:
logging for diagnose the cause of such format failure

### DIFF
--- a/postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
+++ b/postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
@@ -4,6 +4,7 @@
 package com.azure.autorest.postprocessor.implementation;
 
 import com.azure.autorest.extension.base.plugin.NewPlugin;
+import com.azure.autorest.extension.base.plugin.PluginLogger;
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.ImportDeclaration;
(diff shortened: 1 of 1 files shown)


postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -44,6 +47,7 @@ public final class CodeFormatterUtil {
      * @throws Exception If code formatting fails.
      */
     public static void formatCode(Map<String, String> files, NewPlugin plugin) throws Exception {
+        AtomicReference<Logger> loggerReference = new AtomicReference<>();
```
XiaofeiCao, line 50: Why do we use `AtomicReference` here?

postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
```diff
@@ -52,6 +56,12 @@ public static void formatCode(Map<String, String> files, NewPlugin plugin) throw
 
                 plugin.writeFile(fileEntry.getKey(), file, null);
             } catch (Exception e) {
+                // print file content
+                Logger logger = loggerReference.updateAndGet(logger1 ->
+                    logger1 == null ? new PluginLogger(plugin, CodeFormatterUtil.class) : logger1);
```
haolingdong-msft, line 61: I like this change! friendly for deubbing on formatting failure

postprocessor/src/main/java/com/azure/autorest/postprocessor/implementation/CodeFormatterUtil.java
weidongxu-microsoft: We have `files.entrySet().parallelStream().forEach` below. Here is just a trick to try set the Logger var only once (and no set at all if none used).

haolingdong-msft (approved):

XiaofeiCao (approved):

PR #2844, opened 2024-07-01
tsp, bump cadl-ranch

weidongxu-microsoft:
The `@body` for spread

Pan's PR of namespace change

PR of some float number change


typespec-tests/src/test/java/com/azure/resourcemanager/models/resources/generated/NestedProxyResourceInnerTests.java
XiaofeiCao, line 1: Generate.ps1 may need update as well to avoid test gen for arm.
https://github.com/Azure/autorest.java/blob/827c35454caae5e71a7954467fff1f2088dea5f3/typespec-tests/Generate.ps1#L48

weidongxu-microsoft: I see, this be the reason I see some additional generated tests.

XiaofeiCao (approved):

typespec-tests/src/main/java/com/_specs_/azure/core/lro/rpc/RpcClient.java
```diff
@@ -68,7 +68,7 @@ public final class RpcClient {
      * }
      * }</pre>
      * 
-     * @param generationOptions Options for the generation.
```
haolingdong-msft, line 71: curious why would the description change to default?

weidongxu-microsoft: Good question.

The "after" is due to this https://github.com/Azure/cadl-ranch/pull/603/files#r1660519786

The "before" seems from here
https://github.com/weidongxu-microsoft/cadl-ranch/blob/d8ca94bef39fbdfb188ca4288ec77858e2584d33/packages/cadl-ranch-specs/http/azure/core/lro/rpc/main.tsp#L29-L33
But I am not sure about the process that the doc on model end up to the doc on the parameter...

haolingdong-msft: I see. I guess it's because we get description from body type. https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L817

I will pay attention when migrating to `sdkpackage`, to use `bodyParam.description`, instead of `bodyParam.type.description`

haolingdong-msft (approved):

typespec-tests/src/main/java/com/_specs_/azure/core/lro/rpc/RpcClient.java
weidongxu-microsoft: Ah, yes. On implicitly body, the model itself is treated as body.

Issue #2611, opened 2024-03-08
[BUG] tsp sample generation, constructor error when required value is not present in example file


XiaofeiCao:
missing required example value for constructor will result in generated code like `new LongRequest( , , )`.
issue PR: https://github.com/Azure/autorest.java/pull/2612

weidongxu-microsoft:
Candidate:

1. `null` for non-primitive, 0 etc. for priminite
2. log error but keep current behavior
3. throw

XiaofeiCao:
Missing required value seems invalid example to me. We should either throw or skip this file.

XiaofeiCao:
I'll do throw first, see if we need to catch and skip.

PR #2849, opened 2024-07-08
tsp, TCGC getHttpOperationWithCache

weidongxu-microsoft:
Chenjie asks emitter to use this `getHttpOperationWithCache` (I guess the lots of diff in generated code is the reason -- calling `getHttpOperation` twice causes problem on anonymous models)
https://github.com/Azure/autorest.java/pull/2849/commits/be2195beac81348c0d4849ae27843293e4171c7d

Also fix an issue of not fallback to "body" parameter name, in grouping case.
https://github.com/Azure/autorest.java/pull/2849/commits/dff4f57d9036dadb430a8ee615119656af0aba76

The issue that emitter generate sample that cannot compile (if there is missing required values) be  https://github.com/Azure/autorest.java/issues/2611

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/example/ClientMethodExampleWriter.java
+++ b/javagen/src/main/java/com/azure/autorest/template/example/ClientMethodExampleWriter.java
@@ -3,6 +3,7 @@
 
 package com.azure.autorest.template.example;
 
+import com.azure.autorest.extension.base.model.codemodel.RequestParameterLocation;
 import com.azure.autorest.model.clientmodel.ClassType;
 import com.azure.autorest.model.clientmodel.ClientMethod;
 import com.azure.autorest.model.clientmodel.ClientMethodParameter;
@@ -42,6 +43,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class ClientMethodExampleWriter {
@@ -284,7 +286,7 @@ public class ClientMethodExampleWriter {
                         // output parameter's name is the "escaped reserved client method parameter name" of the real parameter's serialized name
                         // since flattened parameter is always in body, we can deal with that explicitly
                         ClientMethodParameter outputParameter = detail.getOutParameter();
-                        Map<String, Object> flattenedParameterValue = getFlattenedBodyParameterExampleValue(proxyMethodExample, outputParameter.getName());
+                        Map<String, Object> flattenedParameterValue = getFlattenedBodyParameterExampleValue(proxyMethodExample, outputParameter);
                         if (flattenedParameterValue != null) {
                             exampleValue.putAll(flattenedParameterValue);
                         }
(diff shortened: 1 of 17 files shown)


typespec-tests/tsp/examples/2022-06-01-preview/FlattenOp_SendLong.json
```diff
@@ -4,7 +4,7 @@
   "parameters": {
     "name":  "myRequiredId",
     "filter": "name=myName",
-    "request": {
+    "sendLongRequest": {
```
weidongxu-microsoft, line 7: TCGC parameter name would be `sendLongRequest`.

Also, the fix within this PR makes it work if this is called `body`.

javagen/src/main/java/com/azure/autorest/template/example/ClientMethodExampleWriter.java
```diff
@@ -329,18 +331,25 @@ private ExampleNode parseNodeFromParameter(ClientMethod convenienceMethod, Proxy
...
                 .orElse(null);
-        if (parameterValue == null) {
-            return null;
+        ProxyMethodExample.ParameterValue parameterValue = getParameterValue.apply(clientMethodParameterName);
+
+        if (parameterValue == null && clientMethodParameter.getRequestParameterLocation() == RequestParameterLocation.BODY && !"body".equalsIgnoreCase(clientMethodParameterName)) {
+            // fallback, "body" is commonly used in example JSON for request body
+            clientMethodParameterName = "body";
+            parameterValue = getParameterValue.apply(clientMethodParameterName);
         }
```
weidongxu-microsoft, line 350: Fallback to `body` if this is BODY request parameter.

weidongxu-microsoft:
@haolingdong-msft This PR should solve the `request` problem with examples.

javagen/src/main/java/com/azure/autorest/template/example/ClientMethodExampleWriter.java
weidongxu-microsoft, line 350: This part is new.

typespec-tests/tsp/examples/2022-06-01-preview/FlattenOp_SendLong.json
haolingdong-msft: Yeah, the fix should solve the example issue, but I'm not sure if service know they should write in this way? `request` seems a common name, like `responses`, while `sendLongRequest` is a special name for each operation.

weidongxu-microsoft: So far (at least in mgmt), the `body` is actually the common name (and that's the reason we had code to fallback to `body`).

XiaofeiCao (approved):

typespec-tests/tsp/examples/2022-06-01-preview/FlattenOp_SendLong.json
XiaofeiCao: Should we use `body` in the example here? Guess in the future, no service will use TCGC given name for example body name...

weidongxu-microsoft: OK, let me change it to "body".

PS: so far, the observation of what JSON example typically does is from Swagger. I'd assume things may change it source is TypeSpec (as it is less clear to service about what the parameter name be, if deduced from tsp across templates and anonymous models).

haolingdong-msft (approved):

typespec-tests/tsp/examples/2022-06-01-preview/FlattenOp_SendLong.json
haolingdong-msft: Yeah, `body` is much better

PR #2853, opened 2024-07-10
Update default settings and add new setting

alzimmermsft:
Updates the following setting defaults:

`--no-custom-headers` is now defaulted to true, meaning all new code generation will add overloads without `ResponseBase` for REST APIs that have typed headers. These overloads are meant to be a performance improvement as they skip converting the `HttpHeaders` to the typed header object, which is commonly ignored in handwritten APIs.

`--generic-response-type` is now defaulted to true, meaning all new code generation will no longer generate named subtypes of `ResponseBase`, ex `NameResponse extends ResponseBase<TypedHeaders, BodyType>`. This reduces the amount of reflective access that is required by `azure-core` by using `ResponseBase` directly which `azure-core` can instantiate directly rather than needing to access the named subtype with reflection.

The following setting was added:

`--disable-typed-headers-methods`, this setting is defaulted to `false`. When set to `true`, this will skip generating any `ResponseBase` and named `ResponseBase` REST API methods in the interface client. This reduces the number of overloads that exist in the interface. Typed headers will still be generated but require being instantiated with handwritten code, making it even clearer when there will be a performance hit for using typed headers.

An additional miscellaneous change was added. Now when code generation is completing, or is halting due to an error, a `shutdown` call will be sent to the Eclipse Language Server if it was being used. The aim of this change is to have a more graceful shutdown process, hopefully reducing how often the Eclipse Language Server gets stuck in a permanent bad state, requiring it to be deleted and re-downloaded.


weidongxu-microsoft (approved):

weidongxu-microsoft:
@XiaofeiCao after this PR, please check fluentgen for MGMT settinga.

XiaofeiCao:
> @XiaofeiCao after this PR, please check fluentgen for MGMT settings.

Sure. I'm curious what the generation error is for fluent. I fixed some in https://github.com/Azure/autorest.java/pull/2541.

weidongxu-microsoft:
> > @XiaofeiCao after this PR, please check fluentgen for MGMT settings.
> 
> Sure. I'm curious what the generation error is for fluent. I fixed some in #2541.

I guess we would have `no-custom-headers=false` as we probably don't want to have `##NoCustomHeadersWithResponse` APIs.

Not sure about `generic-response-type`. We would still need it to be false for libs that already generates the wrapper class. Not sure about new libs.

alzimmermsft:
> > > @XiaofeiCao after this PR, please check fluentgen for MGMT settings.
> > 
> > 
> > Sure. I'm curious what the generation error is for fluent. I fixed some in #2541.
> 
> I guess we would have `no-custom-headers=false` as we probably don't want to have `##NoCustomHeadersWithResponse` APIs.
> 
> Not sure about `generic-response-type`. We would still need it to be false for libs that already generates the wrapper class. Not sure about new libs.

I left both settings in instead of making them hard coded for these reasons, making them enabled may be breaking for some generated code. In the long run for the `ResponseBase` scenarios, this would be recreating the previously generated class manually and just mapping `ResponseBase` to the subtype that was previously generated.

PR #2855, opened 2024-07-11
tsp, convenience api skip param of constant value

weidongxu-microsoft:


typespec-tests/tsp/flatten.tsp
```diff
@@ -101,7 +101,12 @@ model TodoItemPatch {
 interface FlattenOp {
   @route("send")
   @post
-  send(@query id: string, ...Request, ...ApiVersionParameter): OkResponse;
+  send(
+    @query id: string,
+    @query constantQueryParam: "constant",
```
weidongxu-microsoft, line 106: Add a case of constant param. Verify it is set in Impl.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1123,7 +1123,7 @@ export class CodeModelBuilder {
         },
         language: {
           default: {
-            serializedName: param.name,
+            serializedName: this.getSerializedName(param.param),
```
weidongxu-microsoft, line 1126: After TCGC, we should not use this.

typespec-tests/src/main/java/com/cadl/flatten/implementation/FlattenClientImpl.java
```diff
@@ -302,8 +304,9 @@ Response<Void> uploadTodoSync(@HostParam("endpoint") String endpoint,
     @ServiceMethod(returns = ReturnType.SINGLE)
     public Mono<Response<Void>> sendWithResponseAsync(String id, BinaryData sendRequest,
         RequestOptions requestOptions) {
+        final String constantQueryParam = "constant";
```
weidongxu-microsoft, line 307: constantQueryParam = constant

typespec-extension/src/code-model-builder.ts
```diff
@@ -1409,7 +1409,8 @@ export class CodeModelBuilder {
             // parameter
             if (
               existParameter.implementation === ImplementationLocation.Method &&
-              (existParameter.origin?.startsWith("modelerfour:synthesized/") ?? true)
+              (existParameter.origin?.startsWith("modelerfour:synthesized/") ?? true) &&
+              !(existParameter.schema instanceof ConstantSchema)
             ) {
```
weidongxu-microsoft, line 1414: Bug was here.

Now we just don't add the param to convenience API if it is constant.

XiaofeiCao (approved):

typespec-extension/src/code-model-builder.ts
haolingdong-msft: I will try to rewrite this part to use TCGC's `(SdkQueryParameter | SdkPathParameter | SdkHeaderParameter)` types and use the `serializedName` on the type.

haolingdong-msft (approved):

Issue #2779, opened 2024-05-22
[MPG] Empty models in tsp, what should we generate


XiaofeiCao:
tsp: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/devopsinfrastructure/Microsoft.DevOpsInfrastructure/main.tsp#L314-L315

converted swagger: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/devopsinfrastructure/resource-manager/Microsoft.DevOpsInfrastructure/preview/2024-04-04-preview/devopsinfrastructure.json#L1511-L1514

TCGC issue: https://github.com/Azure/typespec-azure/issues/846
copied from the issue: 
| Swagger | m4 | TypeSpec | TCGC |
| --- | --- | --- | --- |
| `{"type": "object"}` | `anyObject` | `{}` | anonymous model with no properties |
| `{"type": "object"}` | `anyObject` | `model EmptyModel {}` | named model with no properties |
| `{}` | `any` | `unknown` | `any` |

Considerations:
We probably need to check what code-model would it be, if it is
1. {} from tsp
2. model Model {} from tsp
Is it only anonymous model vs named model? Do we want to generate them both as empty class (DPG would do this)?
 
And can we distinguish this "empty model" from typespec, from e.g. "type": "object" from Swagger?

XiaofeiCao:
> Is it only anonymous model vs named model?

From code-model point of view, it is. The generated code-model is the same. Though TCGC's result contains `isGeneratedName` for SdkModelType:
<img width="198" alt="Screenshot 2024-07-11 at 18 31 49" src="https://github.com/Azure/autorest.java/assets/92354331/0579f377-0902-4448-a313-4f8055106c0b">

> And can we distinguish this "empty model" from typespec, from e.g. "type": "object" from Swagger?

I'm not sure if TCGC has a way to represent `anyObject` in M4. If it has, we can modify our old M4 logic to detect `anyObject`(instead of treating it as `AnySchema`). If not, guess we can't.

XiaofeiCao:
Discussion here: https://github.com/Azure/typespec-azure/issues/846#issuecomment-2224310829
> 1. Empty model in TypeSpec means model without properties, not `anyObject`. 
> 2. For brownfields, by default we convert `anyObject` as TypeSpec (anonymous)empty model. If service really wants `anyObject`, TypeSpec's equivalence is `Record<unknown>`.

[...]

PR #2859, opened 2024-07-12
tsp arm, generate empty class for emptymodel

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2779

code: https://github.com/Azure/autorest.java/commit/be4de91becac78516a85f6f87eed441a999b60de
test: https://github.com/Azure/autorest.java/commit/bfa99d93fcc24a5f3feac555423ab57ad14aaa2b

changes to `update` are expected.

fixed `package-info` in https://github.com/Azure/autorest.java/pull/2859/commits/e59d78744300114f9985d5d235982fdc8c9f0329


typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ChildExtensionResource.java
```diff
@@ -130,7 +130,7 @@ interface WithProperties {
     /**
      * The template for ChildExtensionResource update.
      */
-    interface Update extends UpdateStages.WithProperties {
+    interface Update {
```
XiaofeiCao, line 133: I need to check why this change.

typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java
```diff
@@ -31,7 +32,7 @@ public class TypeSpecFluentPlugin extends FluentGen {
 
     public TypeSpecFluentPlugin(EmitterOptions emitterOptions, boolean sdkIntegration) {
         super(new TypeSpecPlugin.MockConnection(), "dummy", "dummy");
-
+        SchemaUtilAdapter.adaptForTypeSpec();
```
weidongxu-microsoft, line 35: Emm... Given all these, would subclass make more sense (I assume we need subclass for ObjectMapper and ModelMapper)?

XiaofeiCao: Yeah, I had the same thought before. 
It seemed a bit overkill to have 2 subclasses only to override two identical methods `isPlainObject()`.
But yeah, they are cleaner and less hacky..

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ChildExtensionResourceInterfaces.java
```diff
@@ -40,37 +40,6 @@ Response<ChildExtensionResource> getWithResponse(String resourceUri, String topL
...
-     * @param properties The resource properties to be updated.
-     * @throws IllegalArgumentException thrown if parameters fail the validation.
-     * @throws com.azure.core.management.exception.ManagementException thrown if the request is rejected by server.
-     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
-     * @return extensionResource of Top Level Arm Resource.
-     */
-    ChildExtensionResource update(String resourceUri, String topLevelArmResourceName, String childExtensionResourceName,
-        Object properties);
-
     /**
```
XiaofeiCao, line 43: This is expected. As we now have `bodyClientModel` (instead of Object):
https://github.com/Azure/autorest.java/blob/9b59bd39524bd50b79d04458b464d5d4a241680b/fluentgen/src/main/java/com/azure/autorest/fluent/mapper/ResourceParser.java#L235-L247

Thus, the update methods should be in resource model's API, not resource collection's.

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ChildExtensionResource.java
XiaofeiCao: Finally find the cause:
https://github.com/Azure/autorest.java/blob/9b59bd39524bd50b79d04458b464d5d4a241680b/fluentgen/src/main/java/com/azure/autorest/fluent/mapper/ResourceParser.java#L228-L233

We didn't find `PATCH` operation, and fallback to `PUT`, hence the `properties` be present and of type `ChildExtensionResourceInner`.
Reason we didn't find `PATCH` is because its request body is of type `Object`, not ClientModel. We only count request body of type ClientModel be resource `create` or `update` methods. I doubt there are such cases in Swagger brownfield since I can't imagine `PATCH` request body be {"type": "object"}..

In TypeSpec however, the generated [`ResourceUpdateModel`(xxUpdate)](https://github.com/Azure/typespec-azure/blob/f8ee0803166022ec66cc982222d1090aab10cacb/packages/typespec-azure-resource-manager/lib/foundations/arm.foundations.tsp#L114) will [omit](https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/src/private.decorators.ts#L36-L42) `properties` property if it's empty, making `xxUpdate` has no updatable properties.

typespec-tests/src/main/java/com/cadl/armresourceprovider/models/ChildExtensionResourceInterfaces.java
XiaofeiCao: Turns out the cause be a little different. 
Since the found method is actually `PUT`, we do have `resourceUpdate`. The `update` methods in the resource collection are not considered resource update methods, hence not be removed. 

The new generated code should be right.

typespec-extension/src/main/java/com/azure/autorest/fluent/TypeSpecFluentPlugin.java
XiaofeiCao: Updated. Though in fact, we need to subclass 5 classes: ObjectMapper, ModelPropertyMapper, ModelMapper, FluentObjectMapper and FluentModelMapper..

Alternative is something like the combination: we rely on MapperFactory to return different `PlainObjectChecker` implementations. 
When needing the `PlainObjectChecker`, we do something like `Mappers.getPlainObjectChecker().check()`.

typespec-extension/src/main/java/com/azure/typespec/mapper/TypeSpecModelMapper.java
weidongxu-microsoft, line 1: I am fine it is there.

But I thought DPG already be this logic?

XiaofeiCao: Right.. I'll try removing the DPG mappers.

XiaofeiCao: updated

javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
```diff
@@ -24,12 +24,9 @@
 import java.util.Locale;
 import java.util.stream.Collectors;
 
-public class ModelPropertyMapper implements IMapper<Property, ClientModelProperty> {
+public class ModelPropertyMapper implements IMapper<Property, ClientModelProperty>, NeedsPlainObjectCheck {
     private static final ModelPropertyMapper INSTANCE = new ModelPropertyMapper();
 
-    private ModelPropertyMapper() {
```
weidongxu-microsoft, line 30: protected?

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/mapper/ModelPropertyMapper.java
XiaofeiCao: fixed

Issue #2665, opened 2024-04-04
Implement TCGC's SdkPackage types


iscai-msft:
Language emitters should work to move entirely to TCGC's new ecosystem. The entry point for the ecosystem is accessing .sdkPackage on the SdkContext returned from tcgc

https://azure.github.io/typespec-azure/docs/howtos/DataPlane%20Generation%20-%20DPG/07tcgcTypes
```[tasklist]
### Tasks
- [x] Replace logics for process operation parameters
- [x] Replace logics for server parameters
- [x] Replace logics for process operation
- [x] Replace logics for process clients
- [x] Replace logics for process response
- [x] Replace logics for process LRO
- [x] Replace logics for process body parameters
- [x] Replace logics for process examples
- [x] Replace logics for process auth
- [x] Replace logics for the util functions, e.g. addAcceptedHeader, processEtagHeader etc.
- [x] Replace logics for versioning
- [ ] Integrate with access and usage
- [ ] Pass cadl-ranch and typespec-tests
- [x] integrate with new SdkClientType design
- [x] integrate with new multipart design
- [x] integrate with uri template
```

TCGC dependencies:
```[tasklist]
### TCGC dependency issues
- [ ] https://github.com/Azure/typespec-azure/issues/954
- [ ] https://github.com/Azure/typespec-azure/issues/917
- [ ] https://github.com/Azure/typespec-azure/issues/916
- [ ] https://github.com/Azure/typespec-azure/issues/963
- [ ] https://github.com/Azure/typespec-azure/issues/965
- [ ] https://github.com/Azure/typespec-azure/issues/980
- [ ] https://github.com/Azure/typespec-azure/issues/217
- [ ] https://github.com/Azure/typespec-azure/issues/1088
- [ ] https://github.com/Azure/typespec-azure/issues/1060
- [ ] https://github.com/Azure/typespec-azure/issues/1164
- [ ] https://github.com/Azure/typespec-azure/issues/1259
- [ ] https://github.com/Azure/typespec-azure/issues/1153
- [ ] https://github.com/microsoft/typespec/issues/4132
```

```[tasklist]
### TCGC dependency issues - continue.
- [ ] https://github.com/Azure/typespec-azure/issues/965
```

haolingdong-msft:
Hi @iscai-msft @lmazuel, Java may need to start this work at June as we are working on remaining items related with TCGC model types in May. Please let me know if there is any concern.
[...]

PR #2861, opened 2024-07-15
tsp, tcgc, adopt sdkpackage

haolingdong-msft:
Link https://github.com/Azure/autorest.java/issues/2665

WIP, have not cleaned up code in emitter yet, will let you know once it's done. But would like to create PR to track generated code diff. I'm checking the diffs generated impl files. Especially on `accept` and `content-type` headers, as well as descriptions.

Latest API view and diff summary: https://github.com/Azure/autorest.java/issues/2665#issuecomment-2227958175

-------- update on 8/7----------------
[api view](https://apiview.dev/Assemblies/Review/e0cf256f814e456bac0f0d751bd4aa9f/e8c51901bc82471792b67e3b19771caf?diffRevisionId=d2d206fc3c5d44db9135fbf261aa5392&diffOnly=true) and diff summary: https://github.com/Azure/autorest.java/issues/2665#issuecomment-2272925419

I have also done a round of code clean up. If you have time, please feel free to take a look in the meanwhile I'm fixing the two diffs.
--------update on 8/9--------------------
Note that after adopting new `SdkclientType` design, those previously constant endpoint would be overridable, so there are diffs generated: 955620e82c61c85a05ce0fc0f0ed0324c9975f43
---------update on 8/12---------------------
new api-view: https://apiview.dev/Assemblies/Review/e0cf256f814e456bac0f0d751bd4aa9f/e0f4e0c3e4704de89332b6229b14ad64?diffRevisionId=fbdae726b3a04e839587766478ca88eb&diffOnly=true

Diff summary in the issue: https://github.com/Azure/autorest.java/issues/2665#issuecomment-2283231163

-----------8/13 update------
latest api view: https://apiview.dev/Assemblies/Review/e0cf256f814e456bac0f0d751bd4aa9f/0e83c8742b5942379d330f0e8f40b667?diffRevisionId=af2d36f4bf4243a39d2cab13ace75d25&diffOnly=true

public API diff: 
 - allow override endpoint for constant tsp endpoint definition
 - parameter order, I will sync to sdk repo to see if there is existing sdk impacted, if not, I will keep the parameter order change.
 
image

impl diff: 
 - `allowReserved` on path not supoprted by TCGC.

--------------------update on 8/19----------------------
API view: 
https://apiview.dev/Assemblies/Review/e0cf256f814e456bac0f0d751bd4aa9f/b6c1bc44f2a64d90bd19587aab003913?diffRevisionId=1eed99f40d2d46f0a66d7184e26c5d86&diffOnly=true

API view diff
1. endpoint always overridable - expected
2. parameter order change for spread parameter case. - will regen on sdk repo to see the impact
Reason: There is limitation on `parameters` list of `SdkServiceMethod` that it cannot map to corresponding `SdkHttpOperation`'s `parameters/bodyType` which causes we can't use the `SdkServiceMethod.parameters`. TCGC is considering adding a mapping for us to map SdkServiceMethod's parameters to `SdkHttpOperation`'s `parameters` and `bodyType`. But before it is supported, we are now using `SdkServiceMethod.SdkHttpOperation`.
For spread parameters, we will process the `parameters` and `bodyType` under `SdkHttpOperation`. So the order is first `query/header/path` parameters, then `body` parameter.

Notable changes comparing to last version:
- update spread part code: https://github.com/Azure/autorest.java/pull/2861#issuecomment-2295660875
- add test case for HttpStatusCodeRange: b7ff02c70f5ab694cf858c9304725d1e6bce56ea
- support `allowReserved` in path: a423afaf64feb57783a12857305f84f7f3bae973


typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/usage/implementation/ModelInOperationsImpl.java
```diff
@@ -65,7 +65,7 @@ public interface ModelInOperationsService {
         @UnexpectedResponseExceptionType(value = ResourceNotFoundException.class, code = { 404 })
         @UnexpectedResponseExceptionType(value = ResourceModifiedException.class, code = { 409 })
         @UnexpectedResponseExceptionType(HttpResponseException.class)
-        Mono<Response<Void>> inputToInputOutput(@HeaderParam("accept") String accept,
+        Mono<Response<Void>> inputToInputOutput(@HeaderParam("Content-Type") String contentType,
```
haolingdong-msft, line 68: This is because I removed the [logic](https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L1141) to add accept header.  And content-type header is added by TCGC. 
@weidongxu-microsoft I'm not quite sure on why do we add accept header to this case, the response is `void`, do you think we need to add accept header in this case?

typespec-tests/src/main/java/com/_specs_/azure/core/basic/BasicAsyncClient.java
```diff
@@ -541,10 +541,10 @@ public PagedFlux<User> list(Integer top, Integer skip, List<String> orderBy, Str
             }
         }
         PagedFlux<BinaryData> pagedFluxResponse = list(requestOptions);
-        return PagedFlux.create(() -> (continuationTokenParam, pageSizeParam) -> {
-            Flux<PagedResponse<BinaryData>> flux = (continuationTokenParam == null)
```
haolingdong-msft, line 545: I will check where the name change come from, but I think it will not have functional impact.

typespec-tests/src/main/java/com/_specs_/azure/clientgenerator/core/usage/implementation/ModelInOperationsImpl.java
weidongxu-microsoft: Yeah, we don't need the `accept`, as the response is 204.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/BasicAsyncClient.java
weidongxu-microsoft: This should be from a recent PR of Alan.

typespec-tests/src/main/java/com/_specs_/azure/core/basic/implementation/TwoModelsAsPageItemsImpl.java
```diff
@@ -369,6 +369,9 @@ public PagedIterable<BinaryData> listSecondItem(RequestOptions requestOptions) {
     }
 
     /**
+     * Two operations with two different page item types should be successfully generated. Should generate model for
```
haolingdong-msft, line 372: This comes from operation's doc.
https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/core/basic/main.tsp#L317

typespec-tests/src/main/java/com/_specs_/azure/core/model/ModelAsyncClient.java
```diff
@@ -55,7 +55,7 @@ public final class ModelAsyncClient {
      * @throws ClientAuthenticationException thrown if the request is rejected by server on status code 401.
      * @throws ResourceNotFoundException thrown if the request is rejected by server on status code 404.
      * @throws ResourceModifiedException thrown if the request is rejected by server on status code 409.
-     * @return an embedding vector along with {@link Response} on successful completion of {@link Mono}.
+     * @return the response body along with {@link Response} on successful completion of {@link Mono}.
```
haolingdong-msft, line 58: Response description related changes depends on this issue: https://github.com/Azure/typespec-azure/issues/1088. Response description should be updated once we updated the TCGC version. I will check together after upgrading.

typespec-tests/src/main/java/com/cadl/builtin/implementation/BuiltinClientImpl.java
```diff
@@ -16,12 +16,12 @@
  */
 public final class BuiltinClientImpl {
     /**
-     * Server parameter.
+     * Service host.
```
haolingdong-msft, line 19: `endpoint` parameter desription change due to TCGC use this description, and I just hornor TCGC's desc.

typespec-tests/src/main/java/com/cadl/specialheaders/EtagHeadersOptionalBodyAsyncClient.java
```diff
@@ -55,8 +55,6 @@ public final class EtagHeadersOptionalBodyAsyncClient {
      * <table border="1">
      * <caption>Header Parameters</caption>
      * <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
-     * <tr><td>Content-Type</td><td>String</td><td>No</td><td>The content type. Allowed values:
-     * "application/json".</td></tr>
```
haolingdong-msft, line 59: This is deleted because TCGC now will add `content-type` as `application/json` by default. 
We previously always added content type if request body is optional, to avoid not adding `content-type` header when fire the call.
I think this is acceptable because it does not affect the runtime functionality. @weidongxu-microsoft  @srnagar what do you think?

typespec-tests/src/main/java/com/cadl/union/implementation/UnionClientImpl.java
```diff
@@ -88,7 +89,7 @@ public UnionFlattenOpsImpl getUnionFlattenOps() {
     /**
      * Initializes an instance of UnionClient client.
      * 
-     * @param endpoint
+     * @param endpoint Union.
```
haolingdong-msft, line 92: mark, the union added looks strange, I will check.

typespec-tests/tsp/internal.tsp
```diff
@@ -94,7 +94,7 @@ model ProtocolInternalModel {
 interface InternalOp {
   // test ApiRequest with Access.public
   @access(Access.public, "python")
-  @access(Access.internal, "client")
+  @access(Access.internal, "java")
```
haolingdong-msft, line 97: update the test case because TCGC does not have "client" scope now.

typespec-tests/src/main/java/com/_specs_/azure/core/model/ModelAsyncClient.java
haolingdong-msft: The response description is not updated after using `sdkResponse.description` as we already get description from raw type previously, so the result are the same. I'm checking where these diffs come from.

haolingdong-msft: ok, I get the root cause of the response description difference now.

In codegen, in [createSimpleSyncClientMethods](https://github.com/Azure/autorest.java/blob/8ee365815f616d708a129b92610e8e20f24e8b38/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java#L811), it calls [createSimpleSyncRestResponseReturnValue](https://github.com/Azure/autorest.java/blob/8ee365815f616d708a129b92610e8e20f24e8b38/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java#L1094) to get the return value type, and it has special handling in [processing description](https://github.com/Azure/autorest.java/blob/8ee365815f616d708a129b92610e8e20f24e8b38/javagen/src/main/java/com/azure/autorest/mapper/ClientMethodMapper.java#L1589). It will try to get the description from operation, and put it to the return type description.

But after adoption sdkpackage, I just use `sdkMethod.details ?? ""` to get the description. Since the details is empty(the description is in `summary`), so codegen cannot build the return type description from operation. refer [code](https://github.com/Azure/autorest.java/blob/e2fd61bbaf89359c4cf273de6d1f49de4c3926cc/typespec-extension/src/code-model-builder.ts#L939-L947)

@weidongxu-microsoft , please let me know if this diff is acceptable.

If it's not acceptable, I can think of two solutions:
1. handling in emitter: change `sdkMethod.details ?? ""` to `sdkMethod.details ? sdkMethod.details : sdkMethod.description` when getting operation's description.
2. handling from codegen: use `summary` as well do build the description for return type.

weidongxu-microsoft: I think (2) be better. A merge of "summary" and "description" in Java part, similar to existing logic on model/property/parameter.

typespec-tests/src/main/java/com/cadl/specialheaders/EtagHeadersOptionalBodyAsyncClient.java
weidongxu-microsoft: And the Impl now also sets the `content-type` header explicitly, right?

Here I think the case is that the request body is optional, and hence the Javadoc here reminds user to set the header in `requestOption` when they set the request body there.

typespec-tests/src/main/java/com/cadl/union/implementation/UnionClientImpl.java
weidongxu-microsoft: Maybe OK... `@param endpoint` without a description seems not right.

typespec-tests/src/main/java/com/cadl/specialheaders/EtagHeadersOptionalBodyAsyncClient.java
haolingdong-msft: Yes, `impl` explicity set the `content-type` header: https://github.com/Azure/autorest.java/blob/7a5479dd75c522a9b02dad7d4ab072d112548b82/typespec-tests/src/main/java/com/cadl/specialheaders/implementation/EtagHeadersOptionalBodiesImpl.java#L150

haolingdong-msft: remove the content-type added by TCGC in emitter for optional body case.

typespec-tests/src/main/java/com/_specs_/azure/core/model/ModelAsyncClient.java
haolingdong-msft: Fixed in ab74bad4c423c0d37ce3ac929d05413769394b1d

typespec-tests/src/main/java/com/cadl/specialheaders/EtagHeadersOptionalBodyAsyncClient.java
haolingdong-msft: Fixed in 9e5146e7e931973a8c1c1d78a9b7bf4c909db772, and the generated file does not have diff now.

typespec-tests/src/main/java/com/parameters/bodyoptionality/OptionalExplicitAsyncClient.java
```diff
@@ -40,14 +40,6 @@ public final class OptionalExplicitAsyncClient {
 
     /**
      * The set operation.
-     * <p><strong>Header Parameters</strong></p>
-     * <table border="1">
-     * <caption>Header Parameters</caption>
-     * <tr><th>Name</th><th>Type</th><th>Required</th><th>Description</th></tr>
-     * <tr><td>Content-Type</td><td>String</td><td>No</td><td>The content type. Allowed values:
```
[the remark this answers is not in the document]
haolingdong-msft: Fixed and the generated file does not have diff now.

typespec-extension/src/code-model-builder.ts
```diff
@@ -548,121 +515,196 @@ export class CodeModelBuilder {
...
         if (!this.sdkContext.apiVersion || ["all", "latest"].includes(this.sdkContext.apiVersion)) {
-          this.apiVersion = getDefaultApiVersion(this.sdkContext, client.service);
+          this.apiVersion = versions[versions.length - 1];
         } else {
-          this.apiVersion = versioning.getVersions().find((it: Version) => it.value === this.sdkContext.apiVersion);
+          this.apiVersion = versions.find((it: string) => it === this.sdkContext.apiVersion);
           if (!this.apiVersion) {
             throw new Error("Unrecognized api-version: " + this.sdkContext.apiVersion);
           }
         }
```
weidongxu-microsoft, line 530: Should we always take `versions[versions.length - 1]` ? As now TCGC handles the version.

@XiaofeiCao for opinion as well.

typespec-extension/src/code-model-builder.ts
```diff
@@ -673,17 +715,17 @@ export class CodeModelBuilder {
    * @param versions api-versions to filter
    * @returns filtered api-versions
    */
-  private getFilteredApiVersions(
-    pinnedApiVersion: Version | undefined,
-    versions: Version[],
+  private getFilteredApiVersionsFromString(
```
weidongxu-microsoft, line 718: Similar, if you do not have another function of `getFilteredApiVersionsFrom<Type>`, just use simpler name.

The type is on the parameter.

typespec-extension/src/code-model-builder.ts
```diff
@@ -548,121 +515,196 @@ export class CodeModelBuilder {
...
-        const apiVersions = client.apiVersions;
-        if (apiVersions) {
-          client.serviceVersion = getServiceVersion(client);
+    }
+    return operationGroups;
+  }
+
+  private listServiceMethodsUnderClient(
+    client: SdkClientType<SdkHttpOperation>,
+    includeNestedServiceMethods: boolean,
```
weidongxu-microsoft, line 687: This `includeNestedServiceMethods` param seems always `false`? Any case you would pass `true`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -548,121 +515,196 @@ export class CodeModelBuilder {
...
+        methods.push(method);
+      }
     }
+    return methods;
+  }
 
-    return clients;
+  private removeClientSuffix(clientName: string): string {
+    return clientName.endsWith("Client") ? clientName.slice(0, -6) : clientName;
   }
```
weidongxu-microsoft, line 708: Move these to util class.
It does not need to be a class method.

Also `-6` is magic number. Just do a `const clientSuffix = "Client";` then `slice(0, -clientSuffix.length)`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2178,6 +2180,14 @@ export class CodeModelBuilder {
     );
   }
 
+  private processBinarySchemaFromSdkType(type: SdkBuiltInType): BinarySchema {
```
weidongxu-microsoft, line 2183: Delete `processBinarySchema` if no longer needed?

typespec-extension/src/code-model-builder.ts
```diff
@@ -2382,6 +2392,15 @@ export class CodeModelBuilder {
     }
   }
 
+  private getConvenienceApiNameFromServiceMethod(sdkMethod: SdkServiceMethod<SdkHttpOperation>): string | undefined {
+    // check @convenienceMethod
+    if (sdkMethod.__raw && shouldGenerateConvenient(this.sdkContext, sdkMethod.__raw)) {
```
weidongxu-microsoft, line 2397: Do we still need to do __raw here?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1529,114 +1488,157 @@ export class CodeModelBuilder {
     }
   }
 
+  private addParameterOrBodyToCodeModelRequest(
```
weidongxu-microsoft, line 1491: `addParameterOrBodyPropertyTo...`?

I assume it be body property, not body.

typespec-extension/src/code-model-builder.ts
```diff
@@ -279,67 +276,38 @@ export class CodeModelBuilder {
     return this.codeModel;
   }
 
-  private processHost(server: HttpServer | undefined): Parameter[] {
+  private processHostParametersFromSdkType(sdkPathParameters: SdkPathParameter[]): Parameter[] {
```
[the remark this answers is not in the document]
haolingdong-msft: I just put the suffix to make the name more obvious. I can do one-time suffix removal.

typespec-extension/src/code-model-builder.ts
XiaofeiCao: Guess we can. For pinned scenario, seems TCGC now will filter out all api-versions that's larger than the pinned one. So the latest one will always be the pinned.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: No, we don't pass in `true` for now, just have the flag in case we need to do so. I can delete it.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: TCGC also horners the sequence of the versions, if there is no pinned api version, the last one should be pinned API version. https://github.com/Azure/typespec-azure/pull/1045#discussion_r1678426224

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Good catch. There is a but previously and I used `raw`, https://github.com/Azure/typespec-azure/issues/980, now it is fixed, I will switch to use TCGC to check `@convenientAPI`.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yes, it's body property, will update.

typespec-extension/src/code-model-builder.ts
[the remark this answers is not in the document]
haolingdong-msft: Also I found typescript does not support a traditional OO language function overloading, which means if you define two functions with the same name but difference parameters and write implementation for each of them. There would be compilation error.
image

If you want to do overload in typescript, you need to write code like below- declare the functions and have one implementation.
```ts
function add(a: number, b: number): number;
function add(a: string, b: string): number;

function add(a: unknown, b: unknown): number {
  if (typeof a === "string" && typeof b === "string") {
    return Number(a) + Number(b);
  } else if (typeof a === "number" && typeof b === "number") {
    return a + b;
  }
  return 0;
}
```
references: https://www.telerik.com/blogs/how-implement-function-overloading-typescript
https://stackoverflow.com/questions/13212625/typescript-function-overloading

That's why I name the functions differently to avoid ambiguity and be more obvious.
But I can do a one-time suffix removal if you all prefer. Please let me know your opinion. Thanks!

typespec-extension/src/code-model-builder.ts
haolingdong-msft: See the reply here: https://github.com/Azure/autorest.java/pull/2861#discussion_r1713462201

typespec-extension/src/code-model-builder.ts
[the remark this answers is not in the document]
weidongxu-microsoft: Yeah, I know about the ts overload.

What I mean is there is no `processHost` in your ts file (correct me if wrong). So overload does not take effect (if there were, it is history that not concern the PR).

We'd generally prefer shorter name, if that already get things clear.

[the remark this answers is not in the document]
haolingdong-msft: Sure, I will do a one-time suffix removal.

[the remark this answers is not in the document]
haolingdong-msft: Removed the suffix for process clients/operation and related functions. Did not touch the process schema yet as some helper functions are still using raw type.

typespec-extension/src/operation-utils.ts
```diff
@@ -41,33 +42,33 @@ export function isKnownContentType(contentTypes: string[]): boolean {
...
 }
 
-function operationIsContentType(op: HttpOperation, contentType: string): boolean {
-  for (const param of op.parameters.parameters) {
-    if (param.type === "header" && param.name.toLowerCase() === CONTENT_TYPE_KEY) {
-      if (param.param.type.kind === "String" && param.param.type.value === contentType) {
+function sdkHttpOperationIsContentType(op: SdkHttpOperation, contentType: string): boolean {
+  for (const param of op.parameters) {
+    if (param.kind === "header" && param.serializedName.toLowerCase() === CONTENT_TYPE_KEY) {
+      if (param.type.kind === "constant" && param.type.value === contentType) {
```
weidongxu-microsoft, line 56: Same ask on naming of these. If you do not need to change the function name, keep it same.

typespec-extension/src/code-model-builder.ts
```diff
@@ -246,7 +237,9 @@ export class CodeModelBuilder {
   }
 
   public async build(): Promise<CodeModel> {
-    this.sdkContext = await createSdkContext(this.emitterContext, "@azure-tools/typespec-java");
+    this.sdkContext = await createSdkContext(this.emitterContext, "@azure-tools/typespec-java", {
+      versioning: { previewStringRegex: /$/ },
```
weidongxu-microsoft, line 241: Does it need to depends on the `options["service-version-exclude-preview"]`?

haolingdong-msft: Yes, the current logic is to get all the api-versions and do the filtering by ourselves. This is the logics to get `options["service-version-exclude-preview"]` and filter the api-versions: https://github.com/Azure/autorest.java/blob/c007e8225c52dfd4c53446fd7ea0785ddeee8d74/typespec-extension/src/code-model-builder.ts#L518-L521

typespec-extension/src/code-model-builder.ts
```diff
@@ -677,34 +670,22 @@ export class CodeModelBuilder {
...
-   * `@armProviderNamespace` currently will add a default server if not defined globally:
-   * https://github.com/Azure/typespec-azure/blob/8b8d7c05f168d9305a09691c4fedcb88f4a57652/packages/typespec-azure-resource-manager/src/namespace.ts#L121-L128
-   * TODO: if the synthesized server has the right hostParameter, we can use that instead
-   *
-   * @param server returned by getServers
-   * @returns whether it's synthesized by `@armProviderNamespace`
-   */
-  private isArmSynthesizedServer(server: HttpServer): boolean {
-    return this.isArm() && (!server.parameters || server.parameters.size == 0);
+      .filter((version) => !excludePreview || pinnedApiVersion.includes("preview") || !version.includes("preview"));
```
weidongxu-microsoft, line 682: We are not using TCGC for this filering?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Should we migrate to use TCGC on this?

typespec-extension/src/code-model-builder.ts
haolingdong-msft: We are using TCGC returned api-versions and apply our emitter options to do filtering.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: I'm not sure. I may tend to keep the logics for now as using the regex to keep preview is not an ideal way. I will discuss with Chenjie and Crystal on how TCGC support filtering api versions, I will update the code after TCGC design finalized. I can create an issue to track.

typespec-extension/src/code-model-builder.ts
```diff
@@ -818,67 +795,71 @@ export class CodeModelBuilder {
...
-    // body
-    if (op.parameters.body) {
-      if (op.parameters.body.property) {
-        if (!isVoidType(op.parameters.body.property.type)) {
-          this.processParameterBody(codeModelOperation, op, op.parameters.body.property);
+    // path/query/header parameters
+    for (const param of httpOperation.parameters) {
+      // if it's paged operation with request body, skip content-type header added by TCGC, as next link call should not have content type header
+      if ((sdkMethod.kind === "paging" || sdkMethod.kind === "lropaging") && httpOperation.bodyParam) {
+        if (param.serializedName.toLocaleLowerCase() === "content-type") {
```
weidongxu-microsoft, line 799: check param location?

operation can do a `@query("content-type) contentType`

typespec-extension/src/code-model-builder.ts
```diff
@@ -818,67 +795,71 @@ export class CodeModelBuilder {
...
+  private processRouteForPaged(
+    op: CodeModelOperation,
+    responses: Map<number | HttpStatusCodeRange, SdkHttpResponse>,
+    sdkMethod: SdkMethod<SdkHttpOperation>,
+  ) {
+    if (sdkMethod.kind === "paging" || sdkMethod.kind === "lropaging") {
+      for (const [_, response] of responses) {
+        const bodyType = response.type;
+        if (bodyType && bodyType.kind === "model") {
+          const pagedResult = sdkMethod.__raw_paged_metadata;
```
weidongxu-microsoft, line 859: Why do we need to use raw? Does TCGC miss this information? I assume emitter must know which property is for value, which is for nextLink.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1003,10 +983,10 @@ export class CodeModelBuilder {
...
-    for (const resp of responses) {
-      if (resp.responses && resp.responses.length > 0 && resp.responses[0].headers) {
-        for (const [_, header] of Object.entries(resp.responses[0].headers)) {
-          if (isPollingLocation(this.program, header)) {
+    for (const [_, response] of responses) {
+      if (response.headers) {
+        for (const header of response.headers) {
+          if (isPollingLocation(this.program, header.__raw)) {
             op.extensions = op.extensions ?? {};
             op.extensions["x-ms-long-running-operation"] = true;
```
weidongxu-microsoft, line 989: I thnk we can drop this logic.

weidongxu-microsoft:
Do we have code about the "endpoint override"? https://github.com/Azure/typespec-azure/pull/1155/files#diff-c5102ce4d47a7a165842f82ead490df78ae2f05d05b6957cf893001086dcdf58R1001-L985

Especially, how do we handle this `SdkUnionType<SdkEndpointType>`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1540,114 +1433,157 @@ export class CodeModelBuilder {
...
+
   private findResponseBody(bodyType: Type): Type {
     // find a type that possibly without http metadata like @statusCode
     return this.getEffectiveSchemaType(bodyType);
   }
 
-  private processResponse(op: CodeModelOperation, resp: HttpOperationResponse, longRunning: boolean) {
+  private processResponse(
+    op: CodeModelOperation,
+    statusCode: number | HttpStatusCodeRange | "*",
```
weidongxu-microsoft, line 1487: Do we have existing test case for `HttpStatusCodeRange` (probably local)?

typespec-extension/src/code-model-builder.ts
```diff
@@ -2384,10 +2284,10 @@ export class CodeModelBuilder {
     }
   }
 
-  private getConvenienceApiName(op: Operation): string | undefined {
-    // check @convenienceMethod
-    if (shouldGenerateConvenient(this.sdkContext, op)) {
-      return this.getName(op);
+  private getConvenienceApiNameFromServiceMethod(sdkMethod: SdkServiceMethod<SdkHttpOperation>): string | undefined {
```
weidongxu-microsoft, line 2284: Use previous (shorter) name if possible.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1019,76 +999,67 @@ export class CodeModelBuilder {
...
-        const parameter = param.type === "query" ? this.apiVersionParameter : this.apiVersionParameterInPath;
+        const parameter = param.kind === "query" ? this.apiVersionParameter : this.apiVersionParameterInPath;
         op.addParameter(parameter);
         clientContext.addGlobalParameter(parameter);
       }
-    } else if (this.isSubscriptionId(param)) {
+    } else if (param.kind === "path" && param.onClient && this.isSubscriptionId(param)) {
       const parameter = this.subscriptionIdParameter(param);
       op.addParameter(parameter);
       clientContext.addGlobalParameter(parameter);
```
weidongxu-microsoft, line 1027: You need to add logic to handle `param.onClient`. (part of it need to be merged with `subscriptionId`)

If TCGC gives `param.onClient`, it could be `clientContext.addGlobalParameter(parameter)`.
```
if (param.onClient) {
  clientContext.addGlobalParameter(parameter);
}
```

This applies to all parameters. So this logic probably near the end of this function.

This is not easy to test right now. It may depend on https://github.com/Azure/typespec-azure/issues/914

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: OK

haolingdong-msft:
> Do we have code about the "endpoint override"? https://github.com/Azure/typespec-azure/pull/1155/files#diff-c5102ce4d47a7a165842f82ead490df78ae2f05d05b6957cf893001086dcdf58R1001-L985
> 
> Especially, how do we handle this `SdkUnionType<SdkEndpointType>`?

Yes，this part is to support the new client endpoint design: https://github.com/Azure/autorest.java/blob/f5626c1f260c6d758f7ec080d7b6d385a497b17b/typespec-extension/src/code-model-builder.ts#L532-L557

typespec-extension/src/code-model-builder.ts
```diff
@@ -564,108 +517,148 @@ export class CodeModelBuilder {
...
+            if (initializationProperty.type.values.length <= 2) {
+              // only get the path parameters from the endpoint whose serverUrl is not {"endpoint"}
+              for (const endpointType of initializationProperty.type.values) {
+                if (endpointType.kind === "endpoint" && endpointType.serverUrl !== "{endpoint}") {
+                  sdkPathParameters = endpointType.templateArguments;
+                  baseUri = endpointType.serverUrl;
+                }
+              }
+            } else {
+              throw new Error("multiple server url defined for one client is not supported yet");
```
weidongxu-microsoft, line 545: Let us be more consistent on error message. Make it a whole sentence, start with uppercase letter and end with `.`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -564,108 +517,148 @@ export class CodeModelBuilder {
...
+            }
+          } else if (initializationProperty.type.kind === "endpoint") {
+            sdkPathParameters = initializationProperty.type.templateArguments;
+            baseUri = initializationProperty.type.serverUrl;
+          } else {
+            throw new Error("unexpected endpoint parameter type");
```
weidongxu-microsoft, line 551: Is it possible to reach this `else`? Given data is `type: SdkEndpointType | SdkUnionType<SdkEndpointType>;`?

weidongxu-microsoft:
> Yes，this part is to support the new client endpoint design:
> 
> https://github.com/Azure/autorest.java/blob/f5626c1f260c6d758f7ec080d7b6d385a497b17b/typespec-extension/src/code-model-builder.ts#L532-L557

Got it.

Create a tracking issue (if not having one)?

typespec-extension/src/code-model-builder.ts
haolingdong-msft: issue created： https://github.com/Azure/autorest.java/issues/2913

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Does `onClient` only applies to header/query/path params? Or it also applies to body parameters? From the issue seems it also applies to body parameters, but the design is not finalized yet.

For appying to header/query/path params case, I am a bit unusre about the requirement and use case (besides subscription id and api-version). I tried to add logics in this commit. Please correct me if I understand wrongly.

typespec-extension/src/code-model-builder.ts
```diff
@@ -564,108 +517,146 @@ export class CodeModelBuilder {
...
-            operationGroupName = groupPath[0];
-          }
-          codeModelGroup = new OperationGroup(operationGroupName);
-          for (const operation of operations) {
-            if (!this.needToSkipProcessingOperation(operation, clientContext)) {
-              codeModelGroup.addOperation(this.processOperation(operationGroupName, operation, clientContext));
+        if (serviceMethods.length > 0) {
+          codeModelGroup = new OperationGroup(subClient.name);
+          for (const serviceMethod of serviceMethods) {
+            if (!this.needToSkipProcessingOperation(serviceMethod.__raw, clientContext)) {
```
XiaofeiCao, line 585: Can we get rid of this raw?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: I don't think it could be in body.

Use case be that e.g. a header (or path segment) that going to be same for the whole client.

E.g. `@header region` that going to be same e.g. "WESTUS" / "EASTUS" for all API calls in the client.

weidongxu-microsoft: Could you double check whether `op.addParameter(parameter);` need to be called in both case (for double confirm, do we call `addParameter` on current api-version and subscriptionId?

Other part of the commit LGTM.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: `needToSkipProcessingOperation` will check if the operation is overloaded or not by checking if the operation has `@overload` decorator, and TCGC does not have a flag indicating the operation is overload. So I use raw. I synced with Chenjie before and he said the overload method support is a big feature and needs many design discussions. I can confirm with him on whether there is plan and issue created, if no issue, I can help to create.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Thanks for catching. We call `op.addParameter()` currently for subscription id and api-version. https://github.com/Azure/autorest.java/blob/main/typespec-extension/src/code-model-builder.ts#L1048

We may need to call this in both cases. I will update code and test.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2516,19 +2421,13 @@ export class CodeModelBuilder {
...
 
-  private isSubscriptionId(param: HttpOperationParameter): boolean {
-    return (
-      "subscriptionId".toLocaleLowerCase() === param?.name?.toLocaleLowerCase() &&
-      param.param &&
-      isArmCommonType(param.param) &&
-      isPathParam(this.program, param.param)
-    );
+  private isSubscriptionId(param: SdkPathParameter): boolean {
+    return "subscriptionId".toLocaleLowerCase() === param.name.toLocaleLowerCase();
```
XiaofeiCao, line 2425: Maybe `param.serializedName`?

haolingdong-msft: updated

typespec-extension/src/code-model-builder.ts
haolingdong-msft: https://github.com/Azure/autorest.java/issues/2920 - created an issue for Java.
depends on TCGC issue: https://github.com/Azure/typespec-azure/issues/1391

haolingdong-msft:
https://github.com/Azure/autorest.java/blob/e8a8d3b3b2bdd3309274434eaa947b416f4a87c1/typespec-extension/src/code-model-builder.ts#L1461-L1468
This part is for integrating with the bug fix for flattened parameter. TCGC will put the body parameters on the flattened model to `SdkHttpOperation.bodyParam` and other parameters to ``HttpServiceOperation.parameters`, while TSP compiler will put all the flattened properties to "Operation.parameters". I believe the above logic should be enough after adopting TCGC and regeneration looks good.

Take our local test [tsp/flatten.tsp](https://github.com/Azure/autorest.java/blob/main/typespec-tests/tsp/flatten.tsp#L102-L110) as example:
raw `Operation`:
image

TCGC `SdkHttpOperation`:
image
image

Please let me know if you think above logics are not enough.

typespec-extension/src/code-model-builder.ts
```diff
@@ -829,67 +801,75 @@ export class CodeModelBuilder {
...
-          this.processParameterBody(codeModelOperation, op, op.parameters.body.property);
+    // path/query/header parameters
+    for (const param of httpOperation.parameters) {
+      // if it's paged operation with request body, skip content-type header added by TCGC, as next link call should not have content type header
+      if (
+        (sdkMethod.kind === "paging" || sdkMethod.kind === "lropaging") &&
+        httpOperation.bodyParam &&
+        param.kind === "header"
+      ) {
+        if (param.serializedName.toLocaleLowerCase() === "content-type") {
```
weidongxu-microsoft, line 812: there is a CONTENT_TYPE_KEY defined somewhere

haolingdong-msft: Got it, I saw it in operation-utils.ts, but the `CONTENT_TYPE_KEY` is not exported, I will export the constant and use it in code-model-builder.ts

weidongxu-microsoft:
> Please let me know if you think above logics are not enough.

For `SdkHttpOperation.params`, for non-spread, it would contain param of type `path`/`query`/`header` (could it contain `body`?)
And `SdkHttpOperation.bodyParam` is `body`.

For spread, it would contain param of type `path`/`query`/`header`/`property`.
And `SdkHttpOperation.bodyParam` is `body` (though now this body won't be on method signature).

Is this correct understanding?

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Now TCGC has the info, I have updated the code.

haolingdong-msft:
> > Please let me know if you think above logics are not enough.
> 
> For `SdkHttpOperation.params`, for non-spread (where request body exists), it would contain param of type `path`/`query`/`header` (could it contain `body`?) And `SdkHttpOperation.bodyParam` is `body`.
> 
> For spread, it would contain param of type `path`/`query`/`header`/`property`. And `SdkHttpOperation.bodyParam` is `body` (though now this body won't be on method signature).
> 
> Is this correct understanding?

I don't think so.

In TCGC, for both spread and non-spread, `SdkHttpOperation.parameters` would only contain `query/header/path` parameters. `SdkHttpOperation.bodyParam` is `body`, whose `type` is model which has `properties`, but for spread, current models are marked as anonymous, and for non-spread, it would be anonymous only when user defines it as anonymous.

PS: There is a TCGC pr to change the model used in spread as non-anomymous if spread happens on model being explicitly defined. https://github.com/Azure/typespec-azure/pull/1388 So later is't not all spread models are anonymous.

The spread part is complex, we can have a call offline and I can present my screen to show the debugging code.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Removed and tested.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Seems not, I added one in our local, see tsp and generated code in this commit b7ff02c70f5ab694cf858c9304725d1e6bce56ea 
will consider added to cadl-ranch as well.

weidongxu-microsoft:
Then what's this line for?

```
    if (opParameter.kind !== "property") {
```

> In TCGC, for both spread and non-spread, `SdkHttpOperation.parameters` would only contain `query/header/path` parameters.

haolingdong-msft:
> Then what's this line for?
> 
> ```
>     if (opParameter.kind !== "property") {
> ```
> 
> > In TCGC, for both spread and non-spread, `SdkHttpOperation.parameters` would only contain `query/header/path` parameters.
it is body property, tcgc name it as "property" type

weidongxu-microsoft:
> > > In TCGC, for both spread and non-spread, `SdkHttpOperation.parameters` would only contain `query/header/path` parameters.
> 
> it is property of the body model. Because body parameter and query/header/path parameters are processed seperately, so I guess there is no need to check existing parameter for body parameter. Please correct me if I understand wrongly.
> 
> https://github.com/Azure/autorest.java/blob/00264171ea00641383e5f3d7c900ac097762f850/typespec-extension/src/code-model-builder.ts#L1360-L1373

Yes, I think it is not needed in your code. And that is the reason I am verifying whether TCGC gives you the "property" type parameter.

Also, from this code, the order of parameters of "spread" op is not what TCGC gives you. It is by your code, that non-property parameter is processed, then the property of the `bodyParam` is processed.

The order in TCGC on this would be in its `SdkServiceMethod`, and I assume you didn't use that type.

I likely going to refactor and use `SdkServiceMethod` later (probably during option bag impl). But please check the case that order changes, that your order here is same as order in `SdkServiceMethod`. I would want to avoid the case that we change it today, then had to change it again after migrating to `SdkServiceMethod`.

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Could you confirm with Chenjie that this `__raw_paged_metadata` is the final decision by TCGC?

Should it be `SdkPagingServiceMethodOptions`
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/interfaces.ts#L610-L614
and `resultPath`?
https://github.com/Azure/typespec-azure/blob/main/packages/typespec-client-generator-core/src/interfaces.ts#L534

I am fine we don't switch to this. Just want to make sure that we know what should be used.

haolingdong-msft:
> > > > In TCGC, for both spread and non-spread, `SdkHttpOperation.parameters` would only contain `query/header/path` parameters.
> > 
> > 
> > it is property of the body model. Because body parameter and query/header/path parameters are processed seperately, so I guess there is no need to check existing parameter for body parameter. Please correct me if I understand wrongly.
> > https://github.com/Azure/autorest.java/blob/00264171ea00641383e5f3d7c900ac097762f850/typespec-extension/src/code-model-builder.ts#L1360-L1373
> 
> Yes, I think it is not needed in your code. And that is the reason I am verifying whether TCGC gives you the "property" type parameter
> 
> Also, from this code, the order of parameters of "spread" op is not what TCGC gives you. It is by your code, that non-property parameter is processed, then the property of the `bodyParam` is processed.
> 
> The order in TCGC on this would be in its `SdkServiceMethod`, and I assume you didn't use that type.
> 
> I likely going to refactor and use `SdkServiceMethod` later (probably during option bag impl). But please check that your order here is same as order in `SdkServiceMethod` (just take a look on param in it, for the cadl-ranch case that we see the change in apivew -- `spreadParameterWithInnerAlias` I think). I would want to avoid the case that we change it today, then had to change it again after migrating to `SdkServiceMethod`.

No, it is not same as `SdkServiceMethod`'s parameters, as `SdkServiceMethod` returns body and non-body parameters together in TSP compiler's order. But as there is limitation on parameters list of `SdkServiceMethod` that it cannot map to corresponding operation parameter , we can't use the parameters list. So for spread parameters, we will process the `parameters` and `bodyType` under `SdkHttpOperation`. I'm afraid the order would change later if we use `SdkServiceMethod`.

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Why we no longer use `isStable` func? It seems to be a good abstraction.

BTW, we should actually check `.toLowerCase().includes("-preview")`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -681,34 +672,22 @@ export class CodeModelBuilder {
...
-
-  private needToSkipProcessingOperation(operation: Operation, clientContext: ClientContext): boolean {
+  private needToSkipProcessingOperation(operation: Operation | undefined, clientContext: ClientContext): boolean {
     // don't generate protocol and convenience method for overloaded operations
     // issue link: https://github.com/Azure/autorest.java/issues/1958#issuecomment-1562558219 we will support generate overload methods for non-union type in future (TODO issue: https://github.com/Azure/autorest.java/issues/2160)
-    if (getOverloadedOperation(this.program, operation)) {
+    if (operation && getOverloadedOperation(this.program, operation)) {
       this.trace(`Operation '${operation.name}' is temporary skipped, as it is an overloaded operation`);
       return true;
     }
```
weidongxu-microsoft, line 693: Given the func, it seems to say "do not skip this operation, if it is undefined"?

typespec-extension/src/code-model-builder.ts
```diff
@@ -736,73 +715,66 @@ export class CodeModelBuilder {
...
+    const operationId = groupName ? `${groupName}_${operationName}` : `${operationName}`;
     const operationGroup = this.codeModel.getOperationGroup(groupName);
-    const operationName = this.getName(operation);
-    const opId = groupName ? `${groupName}_${operationName}` : `${operationName}`;
 
-    const operationExamples = this.getOperationExample(op);
+    let operationExamples = undefined;
+    if (sdkMethod.operation && sdkMethod.operation.__raw) {
+      operationExamples = this.getOperationExample(sdkMethod);
+    }
```
weidongxu-microsoft, line 739: Why would the condition include `sdkMethod.operation.__raw`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -736,73 +715,66 @@ export class CodeModelBuilder {
...
+    (codeModelOperation as CrossLanguageDefinition).crossLanguageDefinitionId = sdkMethod.crossLanguageDefintionId;
+    codeModelOperation.internalApi = sdkMethod.access === "internal";
 
-    const convenienceApiName = this.getConvenienceApiName(operation);
+    const convenienceApiName = this.getConvenienceApiName(sdkMethod);
     let generateConvenienceApi: boolean = Boolean(convenienceApiName);
-    let generateProtocolApi: boolean = shouldGenerateProtocol(this.sdkContext, operation);
+    let generateProtocolApi: boolean = sdkMethod.__raw
+      ? shouldGenerateProtocol(this.sdkContext, sdkMethod.__raw)
+      : true;
```
weidongxu-microsoft, line 756: net, it should be more natural to put the `sdkMethod.__raw` check into `shouldGenerateProtocol`.

typespec-extension/src/code-model-builder.ts
```diff
@@ -829,93 +801,101 @@ export class CodeModelBuilder {
...
-            }
-          }
-
-          this.processParameterBody(codeModelOperation, op, bodyType);
+      }
+      // if the request body is optional, skip content-type header added by TCGC
+      if (httpOperation.bodyParam && httpOperation.bodyParam.optional) {
+        if (param.serializedName.toLocaleLowerCase() === CONTENT_TYPE_KEY) {
+          continue;
         }
```
weidongxu-microsoft, line 820: nit, here you remove the header, instead of what you ask in issue "make content-type optional if body is optional".

I am OK you do not change this, but you need to add a `// TODO` to modify this after TCGC decision.

typespec-extension/src/code-model-builder.ts
```diff
@@ -829,93 +801,101 @@ export class CodeModelBuilder {
...
       }
+      this.processParameter(codeModelOperation, param, clientContext);
+    }
+
+    // body
+    if (httpOperation.bodyParam && httpOperation.__raw && sdkMethod.__raw && httpOperation.bodyParam.type.__raw) {
+      this.processParameterBody(codeModelOperation, httpOperation.__raw, httpOperation, httpOperation.bodyParam);
     }
```
weidongxu-microsoft, line 828: Similar to above. Why it checks so many `__raw`, while in the func call it only pass a single `__raw`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1004,102 +984,79 @@ export class CodeModelBuilder {
...
-      } else {
-        schema = this.processSchemaFromSdkType(sdkType, param.param.name);
-      }
+      const sdkType = getNonNullSdkType(param.type);
+      const schema = this.processSchemaFromSdkType(sdkType, param.name);
 
       let extensions: { [id: string]: any } | undefined = undefined;
-      // skip-url-encoding
-      if (param.type === "path") {
+      // TODO haoling: skip-url-encoding, pending on TCGC issue https://github.com/Azure/typespec-azure/issues/1318
```
weidongxu-microsoft, line 1040: This TODO is handled?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1109,68 +1066,68 @@ export class CodeModelBuilder {
...
+              style = SerializationStyle.Form;
+              explode = true;
+              break;
+
+            default:
+              if (format) {
+                this.logWarning(`Unrecognized query parameter format: '${format}'.`);
+              }
+              break;
           }
```
weidongxu-microsoft, line 1099: I think Chenjie has explained all format. https://github.com/Azure/typespec-azure/pull/1363#discussion_r1716327246

Please handle them, e.g. simple and form

weidongxu-microsoft (commented):
Mostly LGTM

- please check the "if" branch, seems to see some extra `__raw` check that seems not used in the code block
- there is some new `collectionFormat` from TCGC https://github.com/Azure/autorest.java/pull/2861#discussion_r1722755533

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yes, it is handled, I will remove TODO

typespec-extension/src/code-model-builder.ts
haolingdong-msft: We need to get namespace, but TCGC does not return namespace for now. https://github.com/Azure/autorest.java/blob/33a3cef4b461fe24982d5c49669dc542b7aa55c2/typespec-extension/src/code-model-builder.ts#L1379

typespec-extension/src/code-model-builder.ts
haolingdong-msft: issue created: https://github.com/Azure/autorest.java/issues/2930

typespec-extension/src/code-model-builder.ts
haolingdong-msft: right, better to add check to return true for `operation === undefined`, I will update code.

typespec-extension/src/code-model-builder.ts
```diff
@@ -566,108 +519,146 @@ export class CodeModelBuilder {
...
-      }
-      const hostParameters = this.processHost(servers?.length === 1 ? servers[0] : undefined);
-      codeModelClient.addGlobalParameters(hostParameters);
+      let hostParameters: Parameter[] = [];
+      client.initialization.properties.forEach((initializationProperty) => {
+        if (initializationProperty.kind === "endpoint") {
+          let sdkPathParameters: SdkPathParameter[] = [];
+          if (initializationProperty.type.kind === "union") {
+            if (initializationProperty.type.values.length <= 2) {
+              // only get the path parameters from the endpoint whose serverUrl is not {"endpoint"}
```
srnagar, line 539: nit: this may not always be path parameters in the template. It could also be domain params like `https://{region}.monitor.azure.com`

typespec-extension/src/code-model-builder.ts
```diff
@@ -566,108 +519,146 @@ export class CodeModelBuilder {
...
-        baseUri = servers[0].url;
-      }
-      const hostParameters = this.processHost(servers?.length === 1 ? servers[0] : undefined);
-      codeModelClient.addGlobalParameters(hostParameters);
+      let hostParameters: Parameter[] = [];
+      client.initialization.properties.forEach((initializationProperty) => {
+        if (initializationProperty.kind === "endpoint") {
+          let sdkPathParameters: SdkPathParameter[] = [];
+          if (initializationProperty.type.kind === "union") {
+            if (initializationProperty.type.values.length <= 2) {
```
srnagar, line 538: When would we have a union with values length < 2?

typespec-extension/src/code-model-builder.ts
srnagar: Do we have an issue to track this in TCGC? If not, can we please create one?

typespec-extension/src/code-model-builder.ts
srnagar: Yeah, we should use `isStable` to not duplicate the logic of checking for `-preview` string and also should be case insensitive.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1104,68 +1059,70 @@ export class CodeModelBuilder {
...
+            case "tsv":
+              style = SerializationStyle.TabDelimited;
+              break;
+
+            case "pipes":
+              style = SerializationStyle.PipeDelimited;
+              break;
+
+            case "multi":
+            case "form":
```
srnagar, line 1084: Do we have `multi` and `form` now when it previously was just `multi`?

typespec-extension/src/code-model-builder.ts
```diff
@@ -1104,68 +1059,70 @@ export class CodeModelBuilder {
...
+              style = SerializationStyle.Form;
+              explode = true;
+              break;
+
+            default:
+              if (format) {
```
srnagar, line 1090: What is the behavior is format is undefined?

typespec-extension/src/code-model-builder.ts
```diff
@@ -191,7 +182,7 @@ export class CodeModelBuilder {
   readonly typeUnionRefCache = new Map<Type, Union | null | undefined>(); // Union means it ref a Union type, null means it does not ref any Union, undefined means type visited but not completed
 
   // current apiVersion name to generate code
-  private apiVersion: Version | undefined;
+  private apiVersion: string | undefined;
```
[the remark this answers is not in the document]
haolingdong-msft: The reason is TCGC returns api-version as `string`. So we can only get `string`, but it should be enough for now.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: I used `<=2` becauase we can't handle multiple endpoint case which is `>2`, and `>2`'s opposite is `<=2`.  But I agree it is not possible in real case to have only one union value.  I can change it to to `===2`.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yes, TCGC adds "form" and "simple" format. So we add it accordingly. https://github.com/Azure/typespec-azure/pull/1363#discussion_r1716327246

typespec-extension/src/code-model-builder.ts
haolingdong-msft: TCGC name it as `SdkPathParameter`, so I use `path parameters`, I will change to `only get the SdkPathParameters from the endpoint whose serverUrl is not {"endpoint"}`, it could be more clear.

typespec-extension/changelog.md
```diff
@@ -1,10 +1,12 @@
 # Release History
 
-## 0.19.3 (2024-08-21)
+## 0.20.0 (2024-08-23)
 
 Compatible with compiler 0.59.
 
-## 0.19.2 (2024-08-21)
+- Adopt TCGC `sdkPackage`.
```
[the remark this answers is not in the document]
weidongxu-microsoft: BTW, the merge is wrong. Please keep the 0.19.3 item.

weidongxu-microsoft (approved):

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Does TCGC uses a union, and do we currently handled all cases? If yes, we may remove this `default` case.

`default` was there for http lib, to my memory the reason is theirs is a string.

haolingdong-msft:
```
export type CollectionFormat = "multi" | "csv" | "ssv" | "tsv" | "pipes" | "simple" | "form";

export interface SdkQueryParameter extends SdkModelPropertyTypeBase {
    kind: "query";
    collectionFormat?: CollectionFormat;
    serializedName: string;
    correspondingMethodParams: SdkModelPropertyType[];
    explode: boolean;
}
```

TCGC definition is above, we should have handled all the cases, so I can remove default.

typespec-extension/src/code-model-builder.ts
haolingdong-msft: Yes, issue created https://github.com/Azure/typespec-azure/issues/1409

PR #2862, opened 2024-07-15
Call into overload with context instead of duplicating logic

alzimmermsft:
Adds additional logic into `ClientMethodTemplate` for asynchronous `Response` methods to call into the overload taking `Context` when the current method being generated doesn't have a `Context` parameter. This reduces the amount of code generated as each of these calls would generate all validation and constant logic, which in service clients with many API calls can reduce out 100s of lines of code, making it easier to read and reason with generated client methods.


weidongxu-microsoft (approved):
LGTM

I've merged your prior PR, that having some conflict with files in this PR.

PR #2867, opened 2024-07-17
Fix possible NPE in ProxyMethodExample

alzimmermsft:
Fixes a `NullPointerException` issue if the value of the header is null. Also removes usage of `Optional` for nullness and null checking.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
+++ b/javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
@@ -10,7 +10,6 @@ import com.azure.core.http.HttpHeaders;
 import com.azure.core.util.CoreUtils;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
 import org.slf4j.Logger;
 
 import java.net.MalformedURLException;
@@ -23,8 +22,6 @@ import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
```diff
@@ -241,32 +232,29 @@ public String getRelativeOriginalFileName() {
                 URL url = new URI(originalFileName).toURL();
                 switch (url.getProtocol()) {
                     case "http":
-                    case "https":
-                    {
-                        String[] segments = url.getPath().split(QUOTED_SLASH);
+                    case "https": {
+                        String[] segments = url.getPath().split(SLASH);
```
weidongxu-microsoft, line 236: Any reason we change this?

I think the argument to `split()` is a regex, and it would be safer to do a `Pattern.quote` if we are split on literal `/`?
Performance concern?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
```diff
@@ -77,13 +73,9 @@ public Object getUnescapedQueryValue() {
         @Override
         public String toString() {
             try {
-                return "ParameterValue{" +
-                        "objectValue=" + PRETTY_PRINTER.writeValueAsString(objectValue) +
-                        '}';
+                return "ParameterValue{" + "objectValue=" + NORMAL_PRINTER.writeValueAsString(objectValue) + '}';
```
weidongxu-microsoft, line 76: I assume the change of PRETTY_PRINTER to NORMAL_PRINTER here is just because of this `toString` is only for logging purpose and we don't want to waste CPU here?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
```diff
@@ -110,9 +102,8 @@ public Response(int statusCode, Object response) {
                 Map<String, Object> responseMap = (Map<String, Object>) response;
                 if (responseMap.containsKey("headers") && responseMap.get("headers") instanceof Map) {
                     Map<String, Object> headersMap = (Map<String, Object>) responseMap.get("headers");
-                    headersMap.forEach((header, value) -> {
-                        httpHeaders.add(HttpHeaderName.fromString(header), value.toString());
-                    });
+                    headersMap.forEach(
+                        (header, value) -> httpHeaders.add(HttpHeaderName.fromString(header), String.valueOf(value)));
```
weidongxu-microsoft, line 106: So the only logic change is here?

javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
alzimmermsft: That and eventually the hope is to replace with azure-json instead of Jackson to go full circle on stream-style.

javagen/src/main/java/com/azure/autorest/model/clientmodel/ProxyMethodExample.java
alzimmermsft: Yeah, this is the meaningful logic change

weidongxu-microsoft (approved):

PR #2868, opened 2024-07-18
tsp, update typespec-tests/package.json

haolingdong-msft:


XiaofeiCao (approved):

typespec-tests/package.json
```diff
@@ -18,16 +18,16 @@
...
-    "@azure-tools/typespec-azure-resource-manager": ">=0.43.0 <1.0.0",
-    "@azure-tools/typespec-autorest": ">=0.43.0 <1.0.0"
+    "@typespec/compiler": ">=0.58.0 <1.0.0",
+    "@typespec/http": ">=0.58.0 <1.0.0",
+    "@typespec/rest": ">=0.58.0 <1.0.0",
+    "@typespec/versioning": ">=0.58.0 <1.0.0",
+    "@typespec/openapi": ">=0.58.0 <1.0.0",
+    "@typespec/xml": ">=0.58.0 <1.0.0",
+    "@azure-tools/typespec-azure-core": ">=0.44.0 <1.0.0",
+    "@azure-tools/typespec-client-generator-core": "0.44.1",
```
weidongxu-microsoft, line 28: Do we want to use `>=0.44.1 <1.0.0`

I kind of remember we fix the version because some newer TCGC lib don't work.

But now latest TCGC should work. Though there is some possibility that latest TCGC have breaking changes.

haolingdong-msft: Yes, we can have try and pin back if later version not work

weidongxu-microsoft (approved):

PR #2870, opened 2024-07-19
tsp, bug fix multiple content-type + json-patch+json

weidongxu-microsoft:
source is appconfiguration
https://github.com/Azure/azure-rest-api-specs/blob/c65e8cbbb18fc9d79165f12291d38d91eb340f0f/specification/appconfiguration/AppConfiguration/routes.tsp#L323-L336

which generates a `PutKeyValueRequestContentType contentType`, where it should be `String contentType`, for WithResponse API.


typespec-tests/tsp/multiple-content-types.tsp
```diff
@@ -42,7 +42,11 @@ interface MultipleContentTypesOnRequest {
   @post
   uploadBytesWithSingleBodyTypeForMultiContentTypes(
     @body data: bytes,
-    @header contentType: "application/octet-stream" | "image/jpeg" | "image/png",
+    @header contentType:
+      | "application/octet-stream"
+      | "image/jpeg"
+      | "image/png"
+      | "application/json-patch+json",
```
weidongxu-microsoft, line 49: this `application/json-patch+json` triggers the bug.

usual case is good.

javagen/src/main/java/com/azure/autorest/mapper/CustomClientParameterMapper.java
```diff
@@ -28,11 +28,14 @@ public static CustomClientParameterMapper getInstance() {
 
     @Override
     public ClientMethodParameter map(Parameter parameter) {
+        return map(parameter, false);
+    }
+
+    public ClientMethodParameter map(Parameter parameter, boolean isProtocolMethod) {
```
weidongxu-microsoft, line 34: I added a `isProtocolMethod` boolean and the logic here.

@haolingdong-msft 
I am not sure about the background that we add this `CustomClientParameterMapper`. Is it purely for json-merge-patch? Can we merge it with `ClientParameterMapper`, or extract the shared logic (I assume at least half of the logic here is duplicated)?

typespec-tests/src/main/java/com/cadl/multicontenttypes/implementation/MultipleContentTypesOnRequestsImpl.java
```diff
@@ -176,9 +175,8 @@ public Mono<Response<Void>> uploadBytesWithSingleBodyTypeForMultiContentTypesWit
      * BinaryData
      * }</pre>
      * 
-     * @param contentType The contentType parameter. Allowed values: "application/octet-stream", "image/jpeg",
-     * "image/png".
-     * @param data The data parameter.
+     * @param contentType The contentType parameter.
```
weidongxu-microsoft, line 178: This change may also due to the difference of logic in `CustomClientParameterMapper`

typespec-tests/tsp/multiple-content-types.tsp
haolingdong-msft: So the CI failure is due to ‘json-patch+json’ content type, not due to defining multiple content types as enum? I’m a bit confused why defining ‘json-patch+json’ will cause the contentType in method use enum rather than string.

weidongxu-microsoft: https://github.com/Azure/autorest.java/pull/2870#discussion_r1683874251

javagen/src/main/java/com/azure/autorest/mapper/CustomClientParameterMapper.java
weidongxu-microsoft: I can merge `CustomClientParameterMapper` with `ClientParameterMapper`, as there seems to be only one place of code diff.

But there is `CustomProxyParameterMapper` as well... Guess I would wait the background of these classes first.

typespec-tests/tsp/multiple-content-types.tsp
haolingdong-msft: ok, I see.

haolingdong-msft (approved):

XiaofeiCao (approved):

PR #2874, opened 2024-07-22
Replace Jackson usage with azure-json

alzimmermsft:
Replaces the usage of Jackson with `azure-json`.


customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageClient.java
```diff
@@ -215,14 +209,7 @@ public void close() {
...
     private static <T> T sendRequest(Connection connection, String method, Object param, Type responseType) {
-        try {
-            String response = OBJECT_MAPPER.writeValueAsString(connection.requestWithSerializedObject(
-                OBJECT_MAPPER.constructType(JsonNode.class), method, GSON.toJson(param)));
-
-            return GSON.fromJson(response, responseType);
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
-        }
+        return GSON.fromJson(connection.requestWithSerializedObject(method, GSON.toJson(param)), responseType);
```
weidongxu-microsoft, line 212: Are we replacing it wit gson?

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
```diff
@@ -81,9 +79,12 @@ public class TemplateUtil {
      * @return the JSON string
      */
     public static String prettyPrintToJson(Object jsonObject) {
```
weidongxu-microsoft, line 81: I guess this is no longer "pretty print".

As long as SDK CI won't report anything, I am fine with the formatting of json file.

weidongxu-microsoft (approved):

XiaofeiCao (approved):

customization-base/README.md
```diff
@@ -181,14 +181,14 @@ with customization
...
     ClassCustomization foo = models.getClass("Foo");
-    foo.addAnnotation("JsonClassDescription(\"Foo class\")");
+    foo.addAnnotation("Deprecated");
 }
 ```
 
 will generate
 
 ```java readme-sample-add-class-annotation-result
-@JsonClassDescription("Foo class")
```
srnagar, line 191: Why is `@JsonClassDescription` replaced with `@Deprecated`?

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageClient.java
srnagar: Why do we want to use GSON here?

extension-base/src/main/java/com/azure/autorest/extension/base/jsonrpc/CallerResponse.java
```diff
@@ -1,71 +0,0 @@
...
-
-import java.util.concurrent.CompletableFuture;
-import java.util.function.Function;
-
-/**
- * Represents a response from a remote procedure call.
- *
- * @param <T> The type of the result of the remote procedure call.
- */
-public class CallerResponse<T> extends CompletableFuture<T> {
```
srnagar, line 16: Why is this type removed? Was it never used?

fluentgen/pom.xml
```diff
@@ -56,6 +56,18 @@
...
+      <artifactId>jackson-databind</artifactId>
+      <version>2.13.5</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+      <version>2.13.5</version>
+      <scope>test</scope>
+    </dependency>
```
srnagar, line 70: Why do we add jackson dependency here? Should we use azure-json for tests as well and not require this dependency?

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageClient.java
alzimmermsft: The types used by the Eclipse dependencies use GSON

customization-base/README.md
alzimmermsft: This was to get Jackson removed as a dependency, I can revert this by using Jackson as a test dependency. But with our overall shift on limiting the usage of Jackson, I thought to remove using it as an example as well.

fluentgen/pom.xml
alzimmermsft: Needed as fluentgen has tests that use Azure SDK types that still depend on Jackson.

javagen/src/main/java/com/azure/autorest/util/TemplateUtil.java
alzimmermsft: I haven't run across issues with this being removed. Unless this is something I need to regenerate an actual Azure SDK with to verify.

customization-base/src/main/java/com/azure/autorest/customization/implementation/ls/EclipseLanguageClient.java
srnagar: It doesn't have to be done in this PR. Please file an issue and resolve this comment.

srnagar (approved):

PR #2875, opened 2024-07-24
tsp, move createSdkContext to async func

weidongxu-microsoft:
I am doing this refactor as anticipation of TCGC examples break, which would make `createSdkContext` async, and async cannot happen in constructor.

Then when TCGC really break, we only need to add a `await` on the func.


typespec-extension/src/code-model-builder.ts
```diff
@@ -206,23 +212,23 @@ export class CodeModelBuilder {
...
+    // java namespace
     const javaNamespace = getJavaNamespace(this.namespace);
 
     const namespace1 = this.namespace;
     this.typeNameOptions = {
       // shorten type names by removing TypeSpec and service namespace
       namespaceFilter(ns) {
         const name = getNamespaceFullName(ns);
-        return name !== "Cadl" && name !== namespace1;
+        return name !== "TypeSpec" && name !== namespace1;
```
weidongxu-microsoft, line 231: This should only be used by logging.

XiaofeiCao (approved):

Issue #2826, opened 2024-06-19
mgmt azure-json, stream-style issue for ManagementError subclass


XiaofeiCao:
Failed fluent runtime deserialization test:
https://github.com/Azure/autorest.java/blob/6627e64eea35b25163422971b07bb3d372d46b52/fluent-tests/src/test/java/com/azure/mgmttest/RuntimeTests.java#L125

Generated code:
```java
public final class DefaultErrorResponseError extends ManagementError {
  private String innererror;
    /**
     * {@inheritDoc}
     */
    @Override
    public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
        jsonWriter.writeStartObject();
        return jsonWriter.writeEndObject();
    }

    /**
     * Reads an instance of DefaultErrorResponseError from the JsonReader.
     * 
     * @param jsonReader The JsonReader being read.
     * @return An instance of DefaultErrorResponseError if the JsonReader was pointing to an instance of it, or null if
     * it was pointing to JSON null.
     * @throws IOException If an error occurs while reading the DefaultErrorResponseError.
     */
    public static DefaultErrorResponseError fromJson(JsonReader jsonReader) throws IOException {
        return jsonReader.readObject(reader -> {
            DefaultErrorResponseError deserializedDefaultErrorResponseError = new DefaultErrorResponseError();
            while (reader.nextToken() != JsonToken.END_OBJECT) {
                String fieldName = reader.getFieldName();
                reader.nextToken();

                if ("innererror".equals(fieldName)) {
                    deserializedDefaultErrorResponseError.innererror = reader.getString();
                } else {
                    reader.skipChildren();
                }
            }

            return deserializedDefaultErrorResponseError;
        });
    }
}
```
`fromJson` is wrong, as it doesn't deserialize properties of parent class `ManagementError`.

Swagger: 
https://github.com/Azure/azure-rest-api-specs/blob/d9f06f3de6cb00796a91b86b622dcf50340952a2/specification/web/resource-manager/Microsoft.Web/stable/2023-12-01/CommonDefinitions.json#L1186

```[tasklist]
### Affected packages
- [ ] azure-resourcemanager-compute
- [ ] azure-resourcemanager-appservice
- [ ] azure-resourcemanager-keyvault
- [ ] azure-resourcemanager-network
```

XiaofeiCao:
### Proposal

#### Simple case(without polymophic, or polymophic terminal class)

```diff
[...]

PR #2876, opened 2024-07-24
azure-json, fix deserialization bug on ManagementError subclasses

XiaofeiCao:
- fix https://github.com/Azure/autorest.java/issues/2826
- code
  - introduce Error Client model, now properties in `ManagementError` will properly get shadowed in child classes 
    https://github.com/Azure/autorest.java/commit/1050009be648ceb8539d706906fae3ea5459344f
  - deserialization adaption(special deserialization handling for `ManagementError` subclass: https://github.com/Azure/autorest.java/commit/e927f629b71663d0af514134e79b44243e1c3ce2
  - A minor deserialization bug fix on shadow properties:
     https://github.com/Azure/autorest.java/commit/7aa99363cb9db567cec2b21c446db1a8545d7ae0
    When properties are shadowed in child class, deserialization in `fromJson` will be something like:
    ```java
    if ("details".equals(fieldName)) { // field from parent property
        List<ManagementError> details = reader.readArray(reader1 -> Error.fromJson(reader1));
        deserializedError.details = details;
    } else if ("details.equals(fieldName)) { // field from child property
        List<Error> details = reader.readArray(reader1 -> Error.fromJson(reader1));
        deserializedError.details = details;
    }
    ```
    After the fix, we will only deserialize those with the same serialized name in child classes.


typespec-tests/src/main/java/com/cadl/armstreamstyleserialization/models/Error.java
```diff
@@ -0,0 +1,190 @@
...
+                    return fromJson0(bufferedReader);
+                } else {
+                    bufferedReader.skipChildren();
+                }
+            }
+            return fromJson0(bufferedReader.reset());
+        });
+    }
+
+    private static Error fromJson0(JsonReader reader) throws IOException {
```
weidongxu-microsoft, line 163: `fromJson0` :-)

Not picky about private name. core-management is `readManagementError`

weidongxu-microsoft (approved):
Please loop in Alan for review, when PR ready.

typespec-tests/src/main/java/com/cadl/armstreamstyleserialization/models/Error.java
XiaofeiCao: I used `fromJson0` for generalization. Though `readManagementError` seems good enough too. Updated.

fluentgen/src/main/java/com/azure/autorest/fluent/model/arm/ErrorClientModel.java
```diff
@@ -0,0 +1,99 @@
...
+                .name("info")
+                .serializedName("info")
+                .description("The additional info.")
+                .readOnly(true)
+                .wireType(ClassType.OBJECT)
+                .clientType(ClassType.OBJECT)
+                .build()
+        )).build();
+
+    public static Optional<ClientModel> getErrorClientModel(String modelName) {
```
alzimmermsft, line 90: Seeing how this return is used, not a big fan of the `return Optional.empty()` only to check `optional.isPresent()` then `optional.get()` as this is equivalent to `return null` and checking `value != null`.

haolingdong-msft (approved):

fluentgen/src/main/java/com/azure/autorest/fluent/model/arm/ErrorClientModel.java
XiaofeiCao: Thanks Alan. 
Seems like I can chain the logic together with `ResourceClientModel.getResourceClientModel` like this:
```java
clientModel = ResourceClientModel.getResourceClientModel(name)
    .or(() -> ErrorClientModel.getErrorClientModel(name))
    .orElse(null);
```

Updated.

PR #2879, opened 2024-07-26
azure-json mgmt, fix pageable lro

XiaofeiCao:
Switch `TypeReference` from that of Jackson to that of azure-json.
`getType` in Jackson is now `getJavaType` in azure-json.
generated code is like:
```java
private PollerFlux<PollResult<Map<String, List<PirCommunityGalleryResource>>>, Map<String, List<PirCommunityGalleryResource>>>
    beginListAsync() {
    Mono<Response<Flux<ByteBuffer>>> mono = listWithResponseAsync();
    return this.client
        .<Map<String, List<PirCommunityGalleryResource>>, Map<String, List<PirCommunityGalleryResource>>>getLroResult(
            mono, this.client.getHttpPipeline(),
            new TypeReference<Map<String, List<PirCommunityGalleryResource>>>() {
            }.getJavaType(), new TypeReference<Map<String, List<PirCommunityGalleryResource>>>() {
            }.getJavaType(), this.client.getContext());
}
```

### DIFF
--- a/fluent-tests/src/test/java/com/azure/mgmttest/LiteCompilationTests.java
+++ b/fluent-tests/src/test/java/com/azure/mgmttest/LiteCompilationTests.java
@@ -40,6 +40,7 @@ import com.azure.mgmtlitetest.storage.models.StorageAccountExpand;
 import com.azure.mgmtlitetest.storage.models.StorageAccountListKeysResult;
 import com.azure.mgmtlitetest.storage.models.StorageAccountRegenerateKeyParameters;
 import com.azure.mgmtlitetest.storage.models.StorageAccounts;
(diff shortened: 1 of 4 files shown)


haolingdong-msft:
Is there any generated test code diff?

weidongxu-microsoft (approved):

XiaofeiCao:
> Is there any generated test code diff? Seems I did not see generated code diff.

Generated code is in `fluent-tests` and it'll not be checked in. If fluent ci passes without compilation error, guess it's fine. 
I could add a class initialization to ensure that.

haolingdong-msft (approved):

Issue #2669, opened 2024-04-05
Multipart Form Data Implementation


lmazuel:

weidongxu-microsoft:
July: support "required" filename

weidongxu-microsoft:
All cases that used by existing service and cadl-ranch test is supported.

We had basic agreement on advanced feature like specifying content-type and additional headers for non-File HttpPart (filename and content-type is good on File HttpPart). The implementation is not started due to lack of real service ask.

PR #2881, opened 2024-07-29
tsp, basic case for multipart HttpPart

weidongxu-microsoft:
link https://github.com/Azure/autorest.java/issues/2669


typespec-extension/src/common/client.ts
```diff
@@ -50,9 +48,6 @@ export class ServiceVersion extends Metadata {
   }
 }
 
-export class ObjectScheme extends ObjectSchema {
+export interface CrossLanguageDefinition {
   crossLanguageDefinitionId?: string;
-  constructor(name: string, description: string, objectInitializer?: DeepPartial<ObjectSchema>) {
-    super(name, description, objectInitializer);
-  }
```
weidongxu-microsoft, line 57: This part of change is not related to multipart.

I was just surprised to see we had `ObjectScheme` and `ObjectSchema`.

I had plan to refactor the representation of `crossLanguageDefinitionId` in code model, but not having time though.

typespec-tests/tsp/multipart.tsp
```diff
@@ -35,3 +35,25 @@ op upload(
...
+
+@route("/uploadHttpPart/images/{name}")
+@post
+op uploadHttpPart(
+  @path name: string,
+  @query compress?: boolean,
+  @header contentType: "multipart/form-data",
+  @multipartBody body: {
+    fileData1: HttpPart<Inherit2File>;
+    fileData2: HttpPart<File>;
```
weidongxu-microsoft, line 56: 2 test cases.

1st for inherit and override of default property in `File`.
2nd for just using the `File` from http lib.

typespec-extension/src/external-schemas.ts
```diff
@@ -108,30 +109,73 @@ export function createPollOperationDetailsSchema(schemas: Schemas, stringSchema:
...
+    // property.type is File, use name and properties from property.type for the File schema
+    /*
+    Current logic:
+    - Class name suffix "FileDetails"
+    - No class hierachy for File
+    - File has 3 properties: "content", "filename", "contentType" (Note that it is "contents" in TypeSpec)
+    - No adjustment on "content" property, it is always BinaryData and required
+    - Allow constant type for "filename" and "contentType" (to be discussed for other types e.g. enum)
+    - Allow required for "filename" and "contentType"
+     */
```
weidongxu-microsoft, line 152: I summarize current logic here. We would have more discussion in Scrum.

weidongxu-microsoft: Though I have certain logic (above) to handle some complex case, but the logic is not finalized, the goal of this PR is to allow service write basic `@multipartBody` and get good code generated, e.g.

```ts
  @multipartBody body: {
    file: HttpPart<File>
  }
```

and still generate reasonable code on complex cases.

weidongxu-microsoft: The rationale for
- Class name suffix "FileDetails"
is allow dev write `HttpPart<File>` and generate `FileDetails`

If we use the model name directly, user would need to do a `@clientName(File, "FileDetails", "java")`

typespec-extension/src/common/client.ts
haolingdong-msft: Previously `ObjectScheme` is just used for adding `crossLanguageDefinitionId` ?

weidongxu-microsoft: That's my guess too.

For now, I changed to `(schema as CrossLanguageDefinition).crossLanguageDefinitionId = ...`

Future plan is to move this to `schema.language.default.crossLanguageDefinitionId` but that would involve Java changes.

typespec-extension/src/external-schemas.ts
```diff
@@ -108,53 +115,164 @@ export function createPollOperationDetailsSchema(schemas: Schemas, stringSchema:
...
+      default: {
+        namespace: namespace,
+      },
+      java: {
+        namespace: getJavaNamespace(namespace),
+      },
+    },
+    serializationFormats: [KnownMediaType.Multipart],
+  });
+  fileDetailsSchema.serializationFormats;
```
XiaofeiCao, line 141: What does this line do?

typespec-extension/src/external-schemas.ts
```diff
@@ -108,53 +115,164 @@ export function createPollOperationDetailsSchema(schemas: Schemas, stringSchema:
...
+  schemas.add(fileDetailsSchema);
+  fileDetailsMap.set(schemaName, fileDetailsSchema);
```
XiaofeiCao, line 143: Is this necessary? Seems all other places will set fileDetailsMap after calling this function.

weidongxu-microsoft: I think it is here
https://github.com/Azure/autorest.java/blob/14572ec8399685c072ac4366f889ede9ed0310f8/typespec-extension/src/external-schemas.ts#L224

If FileDetails of same name in the map, it won't get to this `createFileDetailsSchema` again.

typespec-extension/src/external-schemas.ts
weidongxu-microsoft: Yeah, this is a mistake.

typespec-extension/src/external-schemas.ts
XiaofeiCao: Yes, and we seem to have put into the map outside create function:
e.g. 
https://github.com/Azure/autorest.java/blob/14572ec8399685c072ac4366f889ede9ed0310f8/typespec-extension/src/external-schemas.ts#L227-L228

We could remove either one. Having them both seems excessive to me.

weidongxu-microsoft: Ah, I see. I thought `fileDetailsMap` is local var and won't be in the func. It is not.

typespec-extension/src/code-model-builder.ts
```diff
@@ -2229,35 +2223,39 @@ export class CodeModelBuilder {
 
   private processMultipartFormDataFilePropertySchemaFromSdkType(
```
haolingdong-msft, line 2224: So this function is not only for `FileProperty`? Do we want to rename the function to `processMultipartFormDataSchemaFromSdkType`?

weidongxu-microsoft: Currently it is only for File property
https://github.com/Azure/autorest.java/blob/6f322419cb50c6a342855ab70617bce96365e6cf/typespec-extension/src/code-model-builder.ts#L2104-L2106

haolingdong-msft: I'm a bit confused. Then why `getFileDetailsSchema` function can have 'model' type, I thought files are all bytes.

https://github.com/Azure/autorest.java/blob/6f322419cb50c6a342855ab70617bce96365e6cf/typespec-extension/src/external-schemas.ts#L209

weidongxu-microsoft: `HttpPart<File>`

that `File` is model

haolingdong-msft: I see, Thanks

typespec-extension/src/external-schemas.ts
```diff
@@ -108,53 +115,161 @@ export function createPollOperationDetailsSchema(schemas: Schemas, stringSchema:
...
+      // crossLanguageDefinitionId
+      (fileDetailsSchema as CrossLanguageDefinition).crossLanguageDefinitionId =
+        property.type.crossLanguageDefinitionId;
+
+      let contentTypeProperty;
+      let filenameProperty;
+
+      // find "filename" and "contentType" property in current model and its base models
+      let type: SdkModelType | undefined = property.type;
+      while (type !== undefined) {
```
haolingdong-msft, line 244: I tend to let TCGC return the file related info, e.g. filename, headers(content-type or other headers), and we just use it. I'm fine if we handle it by ourselves currently, we can discuss later.

haolingdong-msft (approved):
Overall looks good to me.

typespec-extension/src/external-schemas.ts
weidongxu-microsoft: 1. it could be in base class like `File`
2. it could be in multiple base class, if there is multiple `FileA extends FileB`, `FileB extends FileC`

Yeah, we may (or may not) be able to use the SdkModelProperty directly. E.g. I don't know what the SDK should generate, if typespec write
```
model FileA extends File {
  filename: "name1" | "name2"
}
```

Good news is that lint appears to error if adding any other property (other than this 3).

PR #2882, opened 2024-07-30
tsp, migrate tcgc getHttpOperationExamples

weidongxu-microsoft:


typespec-extension/src/operation-utils.ts
weidongxu-microsoft, line 1: All change in this file is deleting code that no longer needed.

typespec-extension/src/code-model-builder.ts
```diff
@@ -706,22 +704,19 @@ export class CodeModelBuilder {
...
+        operationExample["x-ms-original-file"] = pathToFileURL(example.filePath).toString();
+        operationExamples[operationExample.title ?? operationExample.operationId ?? operation.operation.name] = operationExample;
       }
-      operationExample = this.operationExamples.get(operation);
+      return operationExamples;
+    } else {
+      return undefined;
     }
-    return operationExample;
   }
```
weidongxu-microsoft, line 720: Most logic is same during the change.

One change is to make it directly return a Record that can be plugged in "x-ms-examples"

XiaofeiCao (approved):

typespec-extension/src/code-model-builder.ts
```diff
@@ -706,22 +704,20 @@ export class CodeModelBuilder {
...
-      if (operation.projectionSource?.kind === "Operation") {
-        operation = operation.projectionSource;
+  private getOperationExample(operation: HttpOperation): Record<string, any> | undefined {
+    const httpOperationExamples = getHttpOperationExamples(this.sdkContext, operation);
+    if (httpOperationExamples && httpOperationExamples.length > 0) {
+      const operationExamples: Record<string, any> = {};
+      for (const example of httpOperationExamples) {
+        const operationExample = example.rawExample;
+        operationExample["x-ms-original-file"] = pathToFileURL(example.filePath).toString();
+        operationExamples[operationExample.title ?? operationExample.operationId ?? operation.operation.name] =
```
XiaofeiCao, line 714: maybe the `operation.operation.name` can be removed, as TCGC will ensure either `title` or `operationId`:
https://github.com/Azure/typespec-azure/pull/1076/files#diff-e186052723c0e3dfe654f3a7fd0f36bd8126c371ad017fdee5a8643ba3d8da10R87

weidongxu-microsoft: The rawExample is type `any`, so we don't have any constraint on type of this behavior.

Though I am actually relying on their logic on duplicate title.
https://github.com/Azure/typespec-azure/pull/1076/files#diff-d276532a6e7d2da177119679a7dfdc92a59980b7bbb70000f99ad0e4d50f3183R180

PR #2885, opened 2024-07-31
tsp, multipart bug fix on HttpPart<File>[]

weidongxu-microsoft:
The case would be covered in future cadl-ranch https://github.com/Azure/cadl-ranch/pull/613

### DIFF
--- a/typespec-extension/src/external-schemas.ts
+++ b/typespec-extension/src/external-schemas.ts
@@ -206,7 +206,13 @@ export function getFileDetailsSchema(
   stringSchema: StringSchema,
   processSchemaFunc: (type: SdkType) => Schema,
 ): ObjectSchema {
+  let fileSdkType: SdkModelType | undefined;
   if (property.type.kind === "model") {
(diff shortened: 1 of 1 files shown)


typespec-extension/src/external-schemas.ts
```diff
@@ -206,7 +206,13 @@ export function getFileDetailsSchema(
   stringSchema: StringSchema,
   processSchemaFunc: (type: SdkType) => Schema,
 ): ObjectSchema {
+  let fileSdkType: SdkModelType | undefined;
   if (property.type.kind === "model") {
+    fileSdkType = property.type;
+  } else if (property.type.kind === "array" && property.type.valueType.kind === "model") {
+    fileSdkType = property.type.valueType;
+  }
```
weidongxu-microsoft, line 214: Bug was due to the revert back to Array<File> from TCGC property.

So here we need to either take the File from property, or the File inside the Array.

haolingdong-msft (approved):

PR #2891, opened 2024-08-05
azure-json, fix incorrect initialization for flatten+required+outputonly property

XiaofeiCao:
https://github.com/Azure/autorest.java/blob/d7bc07ad8d0d7afd8fdeb498d52a556ad0ac501f/typespec-tests/src/main/java/com/cadl/armstreamstyleserialization/fluent/models/OutputOnlyModelInner.java#L39

When the property is flatten+required and the model is output-only, the initialization is impossible since the output-only model doesn't have public ctor and will result in compilation error

code in: https://github.com/Azure/autorest.java/pull/2891/commits/dc5555fc338d9d8eb9baa25057642831b3b6b4dc

another minor fix for improper @Override on private `innerProperties()`. the bug appears due to flattened property become effectively read-only through its model becoming immutable: https://github.com/Azure/autorest.java/pull/2891/commits/310233413f34ecd3aa963c27810c001c076bff75


vanilla-tests/src/main/java/fixtures/requiredfieldsascotrargstransformation/models/TransformationAsParentRequiredFields.java
```diff
@@ -225,7 +210,7 @@ public static TransformationAsParentRequiredFields fromJson(JsonReader jsonReade
                 TransformationAsParentRequiredFields deserializedTransformationAsParentRequiredFields
                     = new TransformationAsParentRequiredFields(rfc1123Required, nameRequired, urlBase64EncodedRequired,
                         unixTimeLongRequired, unixTimeDateTimeRequired, rfc1123RequiredChild);
-                deserializedTransformationAsParentRequiredFields.rfc1123NonRequired = rfc1123NonRequired;
+                deserializedTransformationAsParentRequiredFields.setRfc1123NonRequired(rfc1123NonRequired);
```
XiaofeiCao, line 213: Expected fix for non-readonly properties in Immutable parent model.

weidongxu-microsoft (approved):

vanilla-tests/src/main/java/fixtures/streamstyleserializationimmutableoutput/models/DotSalmon.java
```diff
@@ -118,7 +103,7 @@ public static DotSalmon fromJson(JsonReader jsonReader) throws IOException {
                 reader.nextToken();
 
                 if ("species".equals(fieldName)) {
-                    deserializedDotSalmon.species = reader.getString();
+                    deserializedDotSalmon.setSpecies(reader.getString());
```
XiaofeiCao, line 106: Expected fix: 
```json
"DotFish": {
  "type": "object",
  "discriminator": "fish.type",
  "required": ["fish.type"],
  "properties": {
    "fish.type": {
      "type": "string"
    },
    "species": {
      "type": "string"
    }
  }
  },
  "DotSalmon": {
  "allOf": [
    {
      "$ref": "#/definitions/DotFish"
    }
  ],
  "type": "object",
  "properties": {
    "location": {
      "type": "string"
    },
    "iswild": {
      "type": "boolean"
    }
  }
 }
```

typespec-tests/src/main/java/com/cadl/armstreamstyleserialization/fluent/models/OutputOnlyModelInner.java
```diff
@@ -0,0 +1,215 @@
...
+                .log(new IllegalArgumentException("Missing required property name in model OutputOnlyModelInner"));
+        }
+        if (innerProperties() == null) {
+            throw LOGGER.atError()
+                .log(new IllegalArgumentException(
+                    "Missing required property innerProperties in model OutputOnlyModelInner"));
+        } else {
+            innerProperties().validate();
+        }
+    }
```
XiaofeiCao, line 142: Seems we'll not validate read-only required properties: https://github.com/Azure/autorest.java/blob/6fabb93051072929ea2641bcae63a3276fa7fd50/javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java#L1101

Here `id` is read-only and required, and we don't check it in `validate()`.
So the scenario where we `validate()` locally will only limit to flatten properties in parent class.

weidongxu-microsoft: `validate()` is for sending the request.

If property is `readonly`, it won't be in the request at all.

XiaofeiCao: As offline discussed, we'll validate all properties locally.

haolingdong-msft:
Trying to understand the problem encountered when we shadowed the properties in sub class?

XiaofeiCao:
> Trying to understand the problem encountered when we shadowed the properties in sub class?

At first, we thought there's an unnecessary shadowing for output-only model subclass. https://github.com/Azure/azure-sdk-for-java/pull/41161/files#diff-21fe9ae3fc0e7c1c656f277adadb7dfbcd9321cf3b7fd84b644cc869df3c024d
Turned out it's by design, as output-only model doesn't have public setters. 
Though there're still a few bugs in output-only model and `validate()` method. I'll fix the output-only + flatten case here and the other in another PR.

haolingdong-msft (approved):

Generate.ps1
```diff
@@ -148,7 +148,7 @@ $job = @(
     "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/xml-tag-with-attribute-and-value.json --namespace=fixtures.complexstreamstylexmlserialization",
     "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/security-info.json --namespace=fixtures.securityinfo --use-key-credential",
     "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/special-header.json --namespace=fixtures.specialheader",
-    "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/required-fields-as-ctor-args-transformation.json --namespace=fixtures.requiredfieldsascotrargstransformation --required-fields-as-ctor-args=true --output-model-immutable --null-byte-array-maps-to-empty-array",
+    "$VANILLA_ARGUMENTS --input-file=vanilla-tests/swagger/required-fields-as-ctor-args-transformation.json --namespace=fixtures.requiredfieldsascotrargstransformation --required-fields-as-ctor-args=true --output-model-immutable --null-byte-array-maps-to-empty-array --modelerfour.flatten-models=false --client-flattened-annotation-target=NONE ",
```
weidongxu-microsoft, line 151: Why we change the config for vanilla test?

XiaofeiCao: `modelerfour.flatten-models` defaults to `true`, and m4 will generate an anonymous model for the flattened property with `clientFlatten` false. 

The initialization only happens for `--modelerfour.flatten-models=false`. This test didn't have flatten case before, and can also be used to test `required flatten + required-fields-as-ctor-args` case.

XiaofeiCao: test here: https://github.com/Azure/autorest.java/pull/2891/files#diff-1e1169627c38fd2e98bf825ac7ba63135a7e8a262f2f281379b61a885d81221f

XiaofeiCao: Oh, It's meant for flatten issue found in mgmt. 

I didn't notice there is difference between flatten in mgmt and dpg... If the config isn't valid here, I can revert this test.

weidongxu-microsoft: Yeah, flatten in vanilla would be `modelerfour.flatten-models=true` + `client-flattened-annotation-target=TYPE`

XiaofeiCao: OK. Flatten in dpg seems disabled. Then this case seems not valid. I'll remove it.

weidongxu-microsoft (approved):

PR #2894, opened 2024-08-07
tsp, multipart model naming

weidongxu-microsoft:
Do not modify the name of the File model, in multipart. Exception is `TypeSpec.Http.File`, which is renamed to `FileDetails`.

code https://github.com/Azure/autorest.java/pull/2894/commits/633075bcae35dd8c8325d6b0cdfffb1f703e4990

Most changes is due to cadl-ranch version bump (they've moved some files).


typespec-extension/src/code-model-builder.ts
weidongxu-microsoft, line 1: Change to this file is bug fix.

We should only process the schema of the property once.

typespec-extension/src/external-schemas.ts
```diff
@@ -115,7 +115,13 @@ export function createPollOperationDetailsSchema(schemas: Schemas, stringSchema:
 
 const fileDetailsMap: Map<string, ObjectSchema> = new Map();
 
-function getFileSchemaName(baseName: string) {
+function getFileSchemaName(baseName: string, modelNamespace?: string): string {
+  // "modelNamespace" exists only if "baseName" is from a TypeSpec Model
+  // If the TypeSpec Model exists and is not TypeSpec.Http.File, directly use its name
+  if (modelNamespace && !(modelNamespace === "TypeSpec.Http" && baseName === "File")) {
```
weidongxu-microsoft, line 121: This is changed to use `crossLanguageDefinitionId !== "TypeSpec.Http.File"`

XiaofeiCao (approved):

haolingdong-msft (approved):

Issue #2890, opened 2024-08-02
azure-json, bug in validate() when required properties are duplicated in child class


XiaofeiCao:
## Bug

validate() should call local properties validate(), not super.validate()

https://github.com/Azure/azure-sdk-for-java/pull/41410/files#diff-c94d4ac25fa0574751b403d057581d42019912a532ef016f7cb37d7b3288ac4dR138-R140

current:
```java
public void validate() {
    super.validate();
}
```

expected:
```java
public void validate() {
    if (innerProperties() == null) {
        throw LOGGER.atError()
            .log(new IllegalArgumentException(
                "Missing required property innerProperties in model RedisFirewallRuleInner"));
    } else {
        innerProperties().validate();
    }
}
```

## Proposal
1.  Don't call `super.validate()`, validate all properties in child classes' `validate()` method

Issue #2742, opened 2024-05-06
mgmt azure-json, flatten property in parent class will cause issue in child class in a different package


XiaofeiCao:
* part of https://github.com/Azure/azure-sdk-for-java/issues/38333
* In Child class, `toJson` can't access parent class's flattened property getter, as the getter is private(no matter whether they are in the same package)
* In Child class, `fromJson` can't access parent class's flattened property setter, as it's package private and parent class is in a different package

Parent class:
```java
// child class in different package
public class PirCommunityGalleryResource implements JsonSerializable<PirCommunityGalleryResource> {
    private CommunityGalleryIdentifier innerIdentifier;

    private CommunityGalleryIdentifier innerIdentifier() {
        return this.innerIdentifier;
    }

    PirCommunityGalleryResource withInnerIdentifier(CommunityGalleryIdentifier innerIdentifier) {
        this.innerIdentifier = innerIdentifier;
        return this;
    }

    public String uniqueId() {
        return this.innerIdentifier() == null ? null : this.innerIdentifier().uniqueId();
    }

    public PirCommunityGalleryResource withUniqueId(String uniqueId) {
        if (this.innerIdentifier() == null) {
            this.innerIdentifier = new CommunityGalleryIdentifier();
        }
        this.innerIdentifier().withUniqueId(uniqueId);
        return this;
    }
}
```

Child class:
```java
// child class in different package
public final class CommunityGalleryInner extends PirCommunityGalleryResource {

    @Override
    public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
        jsonWriter.writeStartObject();
        // `innerIdentifier()` can't be accessed, as it is private
        jsonWriter.writeJsonField("identifier", innerIdentifier());
        return jsonWriter.writeEndObject();
    }

    public static CommunityGalleryInner fromJson(JsonReader jsonReader) throws IOException {
        return jsonReader.readObject(reader -> {
            CommunityGalleryInner deserializedCommunityGalleryInner = new CommunityGalleryInner();
            while (reader.nextToken() != JsonToken.END_OBJECT) {
                String fieldName = reader.getFieldName();
                reader.nextToken();

                if ("identifier".equals(fieldName)) {
                    // `withInnerIdentifier()` can't be accessed, as parent class is in a different package and the method is package private
                    deserializedCommunityGalleryInner.withInnerIdentifier(CommunityGalleryIdentifier.fromJson(reader));
                } else {
                    reader.skipChildren();
                }
            }

[...]

PR #2895, opened 2024-08-07
azure-json, fix validate() on shadow properties

XiaofeiCao:
- fix https://github.com/Azure/autorest.java/issues/2890
- flattened properties will be shadowed in child class due to https://github.com/Azure/autorest.java/issues/2742. child class will also override setters and getters, in order to manipulate shadowed flattened properties instead of their parent variants
   turns out that its `validate()` method should be updated accordingly as well, otherwise, when doing `super.validate()` in `validate()`, it'll validate uninitialized flattened properties in parent class
  as discussed, for azure-json, we'll always do local validate(), `super.validate()` will not be called
- code: https://github.com/Azure/autorest.java/pull/2895/commits/3f7c2a6172803f7970210481ac28be19128c86e9
- added test: https://github.com/Azure/autorest.java/commit/5e4bce2b0f245652272da01f3ef7e6de2184f8a7
- CI passed for redis: https://github.com/Azure/azure-sdk-for-java/pull/41410


typespec-tests/src/main/java/com/cadl/armstreamstyleserialization/fluent/models/SalmonInner.java
```diff
@@ -230,14 +254,35 @@ public String pattenAnotherPropertiesPatten() {
...
+        this.innerAnotherProperties().withRequiredString(requiredString);
+        return this;
+    }
+
     /**
      * Validates the instance.
      * 
      * @throws IllegalArgumentException thrown if the instance is not valid.
      */
-    @Override
```
weidongxu-microsoft, line 238: Any reason we remove the `@Override`?

XiaofeiCao: Right, it was not intended.. This is controlled by a flag which also controls whether to call `super.validate()`. 

Let me add it back.

weidongxu-microsoft (approved):

PR #2903, opened 2024-08-09
tsp, bump compiler to 0.59

weidongxu-microsoft:


typespec-extension/src/code-model-builder.ts
```diff
@@ -1097,33 +1097,35 @@ export class CodeModelBuilder {
...
-              break;
-
-            default:
-              if (queryParamOptions?.format) {
-                this.logWarning(`Unrecognized query parameter format: '${queryParamOptions?.format}'.`);
-              }
-              break;
+          // eslint-disable-next-line deprecation/deprecation
+          const queryParamFormat = queryParamOptions?.format;
+          if (queryParamFormat) {
```
weidongxu-microsoft, line 1102: I had to keep this, as cadl-ranch uses this `@format`

I didn't do `// TODO` as this should be replaced by TCGC SdkPackage.

typespec-tests/tsp/enum.tsp
weidongxu-microsoft, line 1: Change to this file is to move to the "recommended" way to define csv/multi, in 0.59.0

eng/sdk/sync_sdk.py
```diff
@@ -17,8 +17,7 @@
 
 skip_artifacts: List[str] = [
     'azure-ai-anomalydetector',         # deprecated
-    'azure-ai-vision-imageanalysis',    # temporary disabled for modification on Javadoc
-    'azure-analytics-defender-easm'     # https://github.com/Azure/azure-sdk-for-java/issues/39862
```
weidongxu-microsoft, line 21: I've refreshed easm, so we no longer need to skip it.

haolingdong-msft: would like to confirm the current sync sdk process before release, we need to merge the pr first and use the pipeline to sync sdk, set the 'devpackage' to true, right? Or is there a CI can be configured in the PR, that we can check before merge the pr?
image

haolingdong-msft (approved):

eng/sdk/sync_sdk.py
weidongxu-microsoft: If one has doubt that the release could have large impact, it is recommended to run a devpackage first.

In normal situations, just release and sync.

E.g. when I add support to `multipartBody` or subnamespace, I know no existing SDK uses this. And I won't run the devpackage.

For this bump, I would likely go release as well. (there could be problem when sync, as the `query(format)` is deprecated, and without bump commit ID it could fail the sync. But this is not a emitter problem.

Issue #2896, opened 2024-08-07
Support AllowReserved in Path Implementation


lmazuel:

weidongxu-microsoft:
reopen

Find our current implementation of x-ms-skip-url-encoding is not same as the URI template [Reserved Expansion](https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.3).

When input is `/foo/bar/a b`, ours appears to keep it same, while URI template would give `/foo/bar/a%20b`

weidongxu-microsoft:
Currently use same impl of `x-ms-skip-url-encoding`.

If we need to improve. it likely be same for both.

PR #2904, opened 2024-08-09
tsp, path allowReserved

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2896

### DIFF
--- a/eng/sdk/sync_sdk.py
+++ b/eng/sdk/sync_sdk.py
@@ -77,7 +77,7 @@ def update_emitter(package_json_path: str, use_dev_package: bool):
             logging.error('Failed to locate the dev package.')
 
     logging.info('Update emitter-package-lock.json')
-    subprocess.check_call(['tsp-client', '--generate-lock-file'], cwd=sdk_root)
+    subprocess.check_call(['tsp-client', 'generate-lock-file'], cwd=sdk_root)
 
 
 def get_generated_folder_from_artifact(module_path: str, artifact: str, type: str) -> str:
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -1131,16 +1131,16 @@ export class CodeModelBuilder {
             }
           }
         } else if (param.type === "header") {
-          switch (param.format) {
-            case "csv":
-              style = SerializationStyle.Simple;
-              break;
+          if (param.format) {
+            switch (param.format) {
+              case "csv":
+                style = SerializationStyle.Simple;
+                break;
 
-            default:
-              if (param.format) {
+              default:
                 this.logWarning(`Unrecognized header parameter format: '${param.format}'.`);
-              }
-              break;
+                break;
+            }
           }
         }
       }
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -118,6 +118,7 @@ import {
   getHeaderFieldName,
   getHeaderFieldOptions,
   getPathParamName,
+  getPathParamOptions,
   getQueryParamName,
   getQueryParamOptions,
   getServers,
@@ -306,6 +307,7 @@ export class CodeModelBuilder {
               },
             },
             extensions: {
+              // TODO: deprecate this logic of string/url for x-ms-skip-url-encoding
               "x-ms-skip-url-encoding": schema instanceof UriSchema,
             },
             // // make the logic same as TCGC, which takes the server-side default of host as client-side default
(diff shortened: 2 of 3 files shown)


typespec-tests/tsp/server.tsp
```diff
@@ -59,8 +59,8 @@ namespace Cadl.ContosoServer {
     name: "Contoso.Sub.ContosoClient",
     service: Cadl.ContosoServer,
   })
-  @route("/contoso/{group}")
+  @route("/contoso/")
   interface ServerOp {
-    get(@path group: url): OkResponse | NoContentResponse;
+    get(@path(#{ allowReserved: true }) group: string): OkResponse | NoContentResponse;
   }
```
weidongxu-microsoft, line 65: 0.59.0 requires either way
- `/contoso/{+group}` without `allowReserved`
- `/contoso/` with `allowReserved`. compiler or http lib would append the `group` to path

typespec-extension/src/code-model-builder.ts
```diff
@@ -1072,14 +1074,23 @@ export class CodeModelBuilder {
...
+      // TODO: deprecate this logic of string/url for x-ms-skip-url-encoding
       if (
         (param.type === "query" || param.type === "path") &&
         param.param.type.kind === "Scalar" &&
         schema instanceof UriSchema
       ) {
-        extensions = { "x-ms-skip-url-encoding": true };
+        extensions = extensions ?? {};
+        extensions["x-ms-skip-url-encoding"] = true;
       }
```
weidongxu-microsoft, line 1094: I didn't deprecate the old logic of `uri for skip encoding, string for not skip encoding`.

Plan to check later.

haolingdong-msft: Yeah, TCGC should handle it and returns correct collection format. I am integrating with latest TCGC. I will pay attention to the test case.

haolingdong-msft (approved):

typespec-tests/tsp/server.tsp
haolingdong-msft: Do we want to add another test case using URI template? (`/contoso/{+group}`)? Or cadl-ranch already has case to cover it?

eng/sdk/sync_sdk.py
```diff
@@ -77,7 +77,7 @@ def update_emitter(package_json_path: str, use_dev_package: bool):
             logging.error('Failed to locate the dev package.')
 
     logging.info('Update emitter-package-lock.json')
-    subprocess.check_call(['tsp-client', '--generate-lock-file'], cwd=sdk_root)
+    subprocess.check_call(['tsp-client', 'generate-lock-file'], cwd=sdk_root)
```
weidongxu-microsoft, line 80:
```
$ tsp-client --help

Usage: tsp-client <COMMAND> [OPTIONS]

Use one of the supported commands to get started generating clients from a
TypeSpec project. This tool will default to using your current working directory
to generate clients in and will use it to look for relevant configuration files.
To specify a different directory, use the -o or --output-dir option.

Commands:
  init                         Initialize the SDK project folder from a
                               tspconfig.yaml
  sync                         Sync TypeSpec project specified in
                               tsp-location.yaml
  generate                     Generate from a TypeSpec project
  update                       Sync and generate from a TypeSpec project
  convert                      Convert a swagger specification to TypeSpec
  generate-lock-file           Generate a lock file under the eng/ directory
                               from an existing emitter-package.json
  sort-swagger <swagger-file>  Sort a swagger file to be the same content order
                               with TypeSpec generated swagger

Options:
      --version     Show version number                                [boolean]
  -d, --debug       Enable debug logging                               [boolean]
  -o, --output-dir  Specify an alternate output directory for the generated
                    files.                               [string] [default: "."]
  -y, --no-prompt   Skip any interactive prompts.                      [boolean]
      --help        Show help                                          [boolean]
```

typespec-tests/tsp/server.tsp
weidongxu-microsoft: Timothee is adding a cadl-ranch test.

My local test, that `allowReserved` does not work if route written as `@route("/contoso/{+group}")`. Could be the http lib just don't want to support this (say it require we use URI template directly), or it could be a bug.

weidongxu-microsoft: Ok, it need to be 

```
  @route("/contoso/{+group}")
  interface ServerOp {
    get(group: string): OkResponse | NoContentResponse;
  }
```

XiaofeiCao (approved):

typespec-extension/src/code-model-builder.ts
```diff
@@ -401,14 +399,12 @@ export class CodeModelBuilder {
...
-  private isInternal(context: SdkContext, operation: Operation): boolean {
+  private isInternal(operation: Operation): boolean {
     const access = getAccess(operation);
     if (access) {
       return access === "internal";
     } else {
-      // TODO: deprecate "internal"
-      // eslint-disable-next-line deprecation/deprecation
-      return isInternal(context, operation);
+      return false;
```
weidongxu-microsoft, line 407: It is being long enough. I guess we can remove this now.

typespec-tests/tsp/server.tsp
weidongxu-microsoft: > Timothee is adding a cadl-ranch test.
> 
> My local test, that `allowReserved` does not work if route written as `@route("/contoso/{+group}")`. Could be the http lib just don't want to support this (say it require we use URI template directly), or it could be a bug.
> 
> Same effect on playground (Swagger having `x-ms-skip-url-encoding` or not).

This appears to be a bug in http lib.

The `getPathParamOptions` gives `undefined`. But `param.allowReserved` is `true`. Will report a bug there.

Therefore I switched to `param.allowReserved`.

PR #2905, opened 2024-08-09
fix, mgmt shadow properties from Resource/ProxyResource

XiaofeiCao:
The shadow property is in the middle of the inheritance hierarchy. In this case, "ResourceWithWritableName", since it has parent class `ProxyResource`.

Previously, we only check if property is shadowed in the leaf subclass. 
Now, we check if property is shadowed in the inheritance hierarchy, and use the one that's closest to the leaf.

code: 9149f75adc80f837b97267c7eebdc791ebdc917d

generated code diffs are reordering of the shadowed properties(a side effect of the fix logic). Now properties from subclass will come before the ones from parent class. 

```json
"ResourceWithWritableName": {
  "description": "ARM resource.",
  "type": "object",
  "properties": {
    "id": {
      "description": "Resource ID.",
      "type": "string",
      "readOnly": true
    },
    "name": {
      "description": "Resource name.",
      "type": "string"
    },
    "type": {
      "description": "Resource type.",
      "type": "string",
      "readOnly": true
    }
  },
  "x-ms-azure-resource": true
},
"ProxyResourceWithWritableName": {
  "description": "ARM proxy resource.",
  "type": "object",
  "allOf": [
    {
      "$ref": "#/definitions/ResourceWithWritableName"
    }
  ],
  "properties": {}
},
"FirewallRule": {
  "description": "A server firewall rule.",
  "type": "object",
  "allOf": [
    {
      "$ref": "#/definitions/ProxyResourceWithWritableName"
    }
  ],
  "properties": {
    "properties": {
      "$ref": "#/definitions/ServerFirewallRuleProperties",
      "description": "Resource properties.",
      "x-ms-client-flatten": true
    }
  }
}
```

### DIFF
--- a/fluent-tests/src/test/java/com/azure/mgmttest/CompilationTests.java
+++ b/fluent-tests/src/test/java/com/azure/mgmttest/CompilationTests.java
@@ -31,6 +31,7 @@ import com.azure.mgmttest.resources.fluent.DeploymentsClient;
 import com.azure.mgmttest.resources.fluent.models.DeploymentExtendedInner;
 import com.azure.mgmttest.resources.fluent.models.ResourceGroupInner;
 import com.azure.mgmttest.resources.models.IdentityUserAssignedIdentities;
+import com.azure.mgmttest.resourcewithwritablename.fluent.models.FirewallRuleInner;
 import com.azure.mgmttest.storage.fluent.StorageAccountsClient;
 import com.azure.mgmttest.storage.fluent.models.StorageAccountInner;
 import com.azure.mgmttest.trafficmanager.fluent.models.EndpointInner;
@@ -169,5 +170,9 @@ public class CompilationTests {
 
     public void testSchemaCleanup() {
         UserAssignedIdentity userAssignedIdentity = new UserAssignedIdentity();
+    }
+
+    public void testResourceWithWritableName() {
+        FirewallRuleInner firewallRuleInner = new FirewallRuleInner();
     }
 }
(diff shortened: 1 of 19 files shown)


fluent-tests/src/test/java/com/azure/mgmttest/CompilationTests.java
```diff
@@ -170,4 +171,8 @@ public void testPolymophicSubClass() {
     public void testSchemaCleanup() {
         UserAssignedIdentity userAssignedIdentity = new UserAssignedIdentity();
     }
+
+    public void testResourceWithWritableName() {
+        FirewallRuleInner firewallRuleInner = new FirewallRuleInner();
```
weidongxu-microsoft, line 176: What would happen of the class, if we did not have this fix?

XiaofeiCao: diff after fix here: https://github.com/Azure/azure-sdk-for-java/pull/41476/commits/806b4e05d097f5c22f27deb047d64dc02c6e758c#diff-a6e36e0731c287838e46e86be1e06840866bec2766bbe888a0de3deac129ee80

1. `name` will be improperly shadowed in `FirewallRuleInner`, also getter improperly generated for it. 
  This is due to in current `getFieldPropeties`, we use `name` from `ProxyResource`, not the shadowed one in `ResourceWithWritableName`.  Now we start search from subclass to parent, to choose the one closest to the leaf.
  codegen fix [here](https://github.com/Azure/autorest.java/pull/2905/files#diff-9f20042b5cf02a7bf58023ee835bca8b87fd603b2304e10d68242142b3d61120R250-R251). 
2. setter won't be generated for the writable `name`. 
  Cause is the same.
3. fromJson would set `name` twice, once for `name` from `ProxyResource`, once for `name` from `ResourceWithWritableName`. 
  codegen fix [here](https://github.com/Azure/autorest.java/pull/2905/files#diff-9f20042b5cf02a7bf58023ee835bca8b87fd603b2304e10d68242142b3d61120R972-R1000)

weidongxu-microsoft (approved):

Issue #2863, opened 2024-07-16
Encode int as string Implementation


lmazuel:

weidongxu-microsoft:
Are service just able to cast every type to every other type?
 
If this comes from Batch (on the "contentLength" in a model that their backend wrongly send a string, but they want to expose a int to customer), they should customize the code for that (they've done that in Jasva, and it is there we notify service on that bug).
From my understanding, this request is not a generic feature. It is just for fixing for a backend bug (that cannot be fixed in backend, maybe due to backward-compatibility issue).
 
I assume for such sporadic incident (only happen on 1 API in 1 service, and that even a bug on backend), customization would be better. We'd talk about feature when we see a 2nd one.

weidongxu-microsoft:
Code merged. Next release 0.19.2 (or 0.20.0) would have the feature.

PR #2908, opened 2024-08-12
tsp, support encode int as string

weidongxu-microsoft:
fix https://github.com/Azure/autorest.java/issues/2863

code is pretty ugly... also it won't work for Jackson.
currently I only did this for int64, which the service needs.

I did local test.

cadl-ranch at https://github.com/Azure/cadl-ranch/pull/656


haolingdong-msft:
Could you pasted the local generated test code?

weidongxu-microsoft:
> Could you paste the local generated test code?

image

haolingdong-msft:
> > Could you paste the local generated test code?
> 
> image

I mean what does the generated code look like. I suppose all the changes are in impl?

weidongxu-microsoft:
```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Code generated by Microsoft (R) TypeSpec Code Generator.

package com.encode.asstring.models;

import com.azure.core.annotation.Generated;
import com.azure.json.JsonReader;
import com.azure.json.JsonSerializable;
import com.azure.json.JsonToken;
import com.azure.json.JsonWriter;
import java.io.IOException;
import java.util.Objects;

/**
 * The IntegerAsStringProperty model.
 */
@Fluent
public final class IntegerAsStringProperty implements JsonSerializable<IntegerAsStringProperty> {
    /*
     * The value property.
     */
    @Generated
    private final long value;

    /*
     * The optionalValue property.
     */
    @Generated
    private Long optionalValue;

    /**
     * Creates an instance of IntegerAsStringProperty class.
     * 
     * @param value the value value to set.
     */
    @Generated
    public IntegerAsStringProperty(long value) {
        this.value = value;
    }

    /**
     * Get the value property: The value property.
     * 
     * @return the value value.
     */
    @Generated
    public long getValue() {
        return this.value;
    }

    /**
     * Get the optionalValue property: The optionalValue property.
     * 
     * @return the optionalValue value.
     */
    @Generated
    public Long getOptionalValue() {
        return this.optionalValue;
    }

    /**
     * Set the optionalValue property: The optionalValue property.
     * 
     * @param optionalValue the optionalValue value to set.
     * @return the IntegerAsStringProperty object itself.
     */
    @Generated
    public IntegerAsStringProperty setOptionalValue(Long optionalValue) {
        this.optionalValue = optionalValue;
        return this;
    }

    /**
     * {@inheritDoc}
     */
    @Generated
    @Override
        jsonWriter.writeStringField("value", Objects.toString(this.value, null));
        jsonWriter.writeStringField("optionalValue", Objects.toString(this.optionalValue, null));
        return jsonWriter.writeEndObject();
    }

    /**
     * Reads an instance of IntegerAsStringProperty from the JsonReader.
     * 
     * @param jsonReader The JsonReader being read.
     * @return An instance of IntegerAsStringProperty if the JsonReader was pointing to an instance of it, or null if it
     * was pointing to JSON null.
     * @throws IllegalStateException If the deserialized JSON object was missing any required properties.
     * @throws IOException If an error occurs while reading the IntegerAsStringProperty.
     */
    @Generated
    public static IntegerAsStringProperty fromJson(JsonReader jsonReader) throws IOException {
        return jsonReader.readObject(reader -> {
            long value = Long.parseLong("0");
            Long optionalValue = null;
                reader.nextToken();

                if ("value".equals(fieldName)) {
                    value = reader.getNullable(nonNullReader -> Long.parseLong(nonNullReader.getString()));
                } else if ("optionalValue".equals(fieldName)) {
                    optionalValue = reader.getNullable(nonNullReader -> Long.parseLong(nonNullReader.getString()));
                } else {
                    reader.skipChildren();
                }
            }
            IntegerAsStringProperty deserializedIntegerAsStringProperty = new IntegerAsStringProperty(value);
            deserializedIntegerAsStringProperty.optionalValue = optionalValue;

            return deserializedIntegerAsStringProperty;
        });
    }
}
```

Yeah, all impl. User only see a `long` or `Long`. That's all.

haolingdong-msft (approved):

XiaofeiCao (approved):
nit, since it doesn't support Jackson, should we limit the scope to sss only? Or do we want to support it in the future?

weidongxu-microsoft:
> nit, since it doesn't support Jackson, should we limit the scope to sss only? Or do we want to support it in the future?

My thought is that we'd probably going to deprecate Jackson eventually. Therefore, my feeling is to avoid putting much effort on this branching, unless necessary.

Let us see what other reviewer thought.

PS: If we need support Jackson, it would need to have different wireType as `String` (here I didn't do a different wireType, just make serialization/de-serialization code different -- that is the reason it only work with azure-json).

PR #2909, opened 2024-08-12
Change deserialization and merge patch to bitwise

alzimmermsft:
Updates JSON merge patch and `fromJson` code to use bitwise tracking instead of a `Set<String>` and a `boolean` per tracking instance.

### DIFF
--- a/azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/models/DocumentModelComposeOperationDetails.java
+++ b/azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/models/DocumentModelComposeOperationDetails.java
@@ -36,36 +36,6 @@ public final class DocumentModelComposeOperationDetails extends OperationDetails
     }
 
     /**
-     * Get the kind property: Type of operation.
-     * 
-     * @return the kind value.
-     */
-    @Override
-    public String getKind() {
-        return this.kind;
-    }
-
-    /**
-     * Get the result property: Operation result upon success.
-     * 
-     * @return the result value.
-     */
-    public DocumentModelDetails getResult() {
-        return this.result;
-    }
-
-    /**
-     * Set the result property: Operation result upon success.
-     * 
-     * @param result the result value to set.
-     * @return the DocumentModelComposeOperationDetails object itself.
-     */
-    public DocumentModelComposeOperationDetails setResult(DocumentModelDetails result) {
-        this.result = result;
-        return this;
-    }
-
-    /**
      * {@inheritDoc}
      */
     @Override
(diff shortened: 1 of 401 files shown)


weidongxu-microsoft (commented):
@haolingdong-msft could you take review on this, as it is about json-merge-patch?

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/models/DocumentModelComposeOperationDetails.java
```diff
@@ -146,6 +116,36 @@ public DocumentModelComposeOperationDetails setError(Error error) {
...
+     * @return the result value.
+     */
+    public DocumentModelDetails getResult() {
+        return this.result;
+    }
+
+    /**
+     * Set the result property: Operation result upon success.
+     * 
+     * @param result the result value to set.
```
haolingdong-msft, line 141: Is it just changing the order?

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
```diff
@@ -1339,7 +1346,7 @@ private void addJsonMergePatchRelatedPropertyAndAccessors(JavaClass classBlock,
         classBlock.javadocComment(comment ->
             comment.description("Stores updated model property, the value is property name, not serialized name"));
         addGeneratedAnnotation(classBlock);
-        classBlock.privateFinalMemberVariable("Set<String> updatedProperties = new HashSet<>()");
+        classBlock.privateMemberVariable("long updatedProperties = 0L");
```
haolingdong-msft, line 1349: Once concern from me is would bitwise make it harder to debug?

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/models/DocumentModelComposeOperationDetails.java
alzimmermsft: It is just a change in order but it's a good time to have a discussion on how we want to order overridden methods. Before this change, the methods defined the type came first then overridden methods, but in serialization code the properties defined by super types came first. It'd be nice to have a standard system for both serialization and getters/setters, such as properties and setters/getters defined by super types always come first (or last).

javagen/src/main/java/com/azure/autorest/template/ModelTemplate.java
alzimmermsft: It would make debugging moderately more difficult; it'd be a change from checking `updatedProperties.contains(string)` to doing a bitwise operation that matches the logic used by the setter. In both cases you'll still need to look at the setter or serialization code to see what's going on.

weidongxu-microsoft: Emm.. what if the model has more than 64 properties?

azure-dataplane-tests/src/main/java/com/azure/ai/formrecognizer/documentanalysis/implementation/models/DocumentModelComposeOperationDetails.java
haolingdong-msft: To me, as long as the two orders (methods order and properties processing order during serialziation) are consistent, and both override methods and new methods have their own section, that would be fine.
I also asked copilot. It says override methods from super class comes first. So I would vote for this - override methods come first. :)
image

PR #2911, opened 2024-08-13
tsp, bug fix, flatten body property that having same serialized name with path parameter

weidongxu-microsoft:
bug reported by openai

code change https://github.com/Azure/autorest.java/pull/2911/commits/5efc992ccdd4a0dccc4364293f9c5e286ed8a969

SDK repo change seems not due to this PR https://github.com/Azure/azure-sdk-for-java/pull/41497

### DIFF
--- a/typespec-extension/src/code-model-builder.ts
+++ b/typespec-extension/src/code-model-builder.ts
@@ -119,7 +119,10 @@ import {
   getQueryParamName,
   getServers,
   getStatusCodeDescription,
+  isBody,
+  isBodyRoot,
   isHeader,
+  isMultipartBodyProperty,
   isPathParam,
   isQueryParam,
 } from "@typespec/http";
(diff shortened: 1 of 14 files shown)


typespec-tests/tsp/flatten.tsp
```diff
@@ -33,6 +33,7 @@ model ApiVersionParameter {
 
 alias Request = {
   @visibility("read") @key name: string;
+  endpoint: string;
```
weidongxu-microsoft, line 36: Without the fix, this `endpoint` would not show up on method signature.

Cause is that it has same serializedName with `endpoint` on host
https://github.com/Azure/autorest.java/blob/7998c94e70361d50277866e3e59b842815e5fcc0/typespec-tests/tsp/flatten.tsp#L12-L18

typespec-extension/src/code-model-builder.ts
```diff
@@ -1440,7 +1443,24 @@ export class CodeModelBuilder {
...
+              const sameSerializedName = it.language.default.serializedName === serializedName;
+              if (paramLocation === ParameterLocation.Body) {
+                // body, same paramLocation, as there could only be 1 body in operation
+                return sameParamLocation;
+              } else {
+                // path, query, header, require same serializedName and same paramLocation
+                return sameParamLocation && sameSerializedName;
+              }
+            });
+          }
```
weidongxu-microsoft, line 1463: We should be migrating to TCGC operation/method on this.

To be done after TCGC package migration.

Also, the "options bag" feature would also touch this part.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1440,7 +1443,24 @@ export class CodeModelBuilder {
 
         for (const [_, opParameter] of parameters.properties) {
           const serializedName = this.getSerializedName(opParameter);
-          const existParameter = op.parameters.find((it) => it.language.default.serializedName === serializedName);
+          const paramLocation = this.getParameterLocation(opParameter);
+          let existParameter: Parameter | undefined;
+          if (paramLocation === "BodyProperty") {
```
haolingdong-msft, line 1448: use `paramLocation === ParameterLocation.Body`?

weidongxu-microsoft: Here is intended to be `BodyProperty`.

The `opParameter` here is from the operation (when spread, the parameter there be the "body property", not the "body" itself).

haolingdong-msft: ok, I see.

haolingdong-msft (approved):

XiaofeiCao (approved):

PR #2917, opened 2024-08-15
tsp, bump tcgc

weidongxu-microsoft:
Also handle the behavior change of example.filePath
code https://github.com/Azure/autorest.java/pull/2917/files#diff-14fdfb2e0df46da414f4d04a218f03e4e5f2737d3a9672b9c4e4feb16419445e


typespec-extension/src/code-model-builder.ts
```diff
@@ -727,7 +728,13 @@ export class CodeModelBuilder {
       const operationExamples: Record<string, any> = {};
       for (const example of httpOperationExamples) {
         const operationExample = example.rawExample;
-        operationExample["x-ms-original-file"] = pathToFileURL(example.filePath).toString();
+
+        // resolve absolute path of the example file
+        const examplesDir = (this.options as any)["examples-dir"] ?? (this.options as any)["examples-directory"];
```
weidongxu-microsoft, line 733: I am being lazy here.

I don't want to define `examples-dir` or `examples-directory` in emitter.ts, as they are supposed to be handled by TCGC.
But here I still need to concat it with the `example.filePath` from TCGC. Else we loss part of the path, that cannot be re-constructed afterwards.

weidongxu-microsoft:
there is some problem in multipart, but I am not sure where is this introduced https://github.com/Azure/azure-sdk-for-java/pull/41533/files#diff-b9cb1f70f320d20a7f6db797215c3e3c2c2472dc4bece4f50d325434333b0574

need more invesitgation

haolingdong-msft:
> there is some problem in multipart, but I am not sure where is this introduced (I don't think I've changed emitter code about this) https://github.com/Azure/azure-sdk-for-java/pull/41533/files#diff-b9cb1f70f320d20a7f6db797215c3e3c2c2472dc4bece4f50d325434333b0574
> 
> need more invesitgation

Our local tests and cadl-ranch tests are not impacted?

haolingdong-msft (approved):

XiaofeiCao (approved):

weidongxu-microsoft:
> Our local tests and cadl-ranch tests are not impacted?

Yes, they are not impacted. This is a puzzle to me...

weidongxu-microsoft:
> Yes, they are not impacted. This is a puzzle to me...

It is from a bug exists in TCGC for some time, but only appears after some other change in 0.45.3

Condition:
1. more than 1 client in client.tsp
2. multipart not in 1st client

Reason is that after 1st client, TCGC process unused models, and the model used in multipart op is handled as if it is not a multipart model (because it is in 2nd client that not yet processed)

haolingdong-msft:
Do we want to revert this TCGC version upgrade until the multipart issue is fixed?

weidongxu-microsoft:
Yuchao or Chenjie will fix it (depends on "how to fix"). They will handle that.

weidongxu-microsoft:
Probably not, .3 involves change on `examples-directory` to `examples-dir`, `allowPreserved` etc.

If we revert to .2, there is actually not much different from our previously released typespec-java.

So we likely don't need to revert, unless there is some urgent need for a new release (bug fix etc.).

haolingdong-msft:
Sure, if the issue can be fixed soon, it would be fine to wait for their next version.

PR #2918, opened 2024-08-15
Support for shared toJson code in polymorphic types

alzimmermsft:
Adds support for polymorphic to share `toJson` code when all models are contained in the same package. When all models of the polymorphic structure are in the same package, the following code changes are made:

- Types that don't define the polymorphic discriminator property no longer generate a field for the value.
- The type that defines the polymorphic property defines a non-final, package-private field for the value.
- Super types generate a package-private `toJsonShared` that handles serializing all properties defined by the model.
- Subtypes no longer override the getter for the polymorphic discriminator.

There is also one more check performed before generating the `toJsonShared` method, which is the model doesn't have JSON merge patch support. In the future this will be supported, as JSON merge patch JSON serialization can also share logic between types, but it is much more complex and the number of models using this functionality is much smaller, as polymorphism and JSON merge patch alone are less commonly used and together it is even less common.


typespec-tests/src/main/java/com/azure/resourcemanager/models/resources/fluent/models/TopLevelTrackedResourceInner.java
```diff
@@ -85,7 +85,6 @@ public SystemData systemData() {
      * 
      * @return the type value.
      */
-    @Override
     public String type() {
```
XiaofeiCao: Is this change expected? They are getters of the shadowed read-only properties from parent class.

typespec-tests/src/main/java/com/cadl/armstreamstyleserialization/fluent/models/SalmonInner.java
```diff
@@ -75,6 +75,7 @@ public String kind() {
      * 
      * @return the friends value.
      */
+    @Override
```
XiaofeiCao: Seems not correct. These properties are defined in `SalmonInner` itself:
https://github.com/Azure/autorest.java/blob/bf0facc6325a2b5fcfffddd816a7c98328aab395/typespec-tests/tsp/arm-stream-style-serialization.tsp#L52-L57

typespec-tests/src/main/java/com/azure/resourcemanager/models/resources/fluent/models/TopLevelTrackedResourceInner.java
XiaofeiCao: Oh, I saw it's added back in later commits!

vanilla-tests/src/main/java/fixtures/streamstyleserialization/models/Salmon.java
```diff
@@ -132,13 +118,14 @@ public void validate() {
...
-        jsonWriter.writeFloatField("length", getLength());
-        jsonWriter.writeStringField("species", getSpecies());
-        jsonWriter.writeArrayField("siblings", getSiblings(), (writer, element) -> writer.writeJson(element));
-        jsonWriter.writeStringField("fishtype", this.fishtype);
+        toJsonShared(jsonWriter);
+        return jsonWriter.writeEndObject();
+    }
+
+    void toJsonShared(JsonWriter jsonWriter) throws IOException {
+        super.toJsonShared(jsonWriter);
```
haolingdong-msft, line 126: Is this only for the case that super class and sub class in the same package? Can super class and sub class be in different packages? Then `super.toJsonShared` could not be called.

XiaofeiCao (approved):

PR #2919, opened 2024-08-16
Allow sharing fromJson code

alzimmermsft:
Add support for generating `fromJsonShared` methods, similar to what was done with `toJsonShared`, when polymorphic models meet the following requirements:

- The models doesn't have any constructor parameters. Sharing deserialization logic isn't helpful when the model has constructor parameters as we'd need to create container types and pass those to the shared logic. In the future more work could be done here to allow this for all polymorphic models, such as making fields non-final internally and having a private no-args constructor.
- All models in the polymorphic structure are in the same package.

When those requirements are met, polymorphic parent types will generate a `fromJsonShared` method for all the properties it defines. Child models will call into that rather than writing the deserialization logic themselves. `fromJsonShared` will return a boolean indicating whether a field was consumed during the call, if not the caller will either skip the unknown property or do further checks in the child type for the property. Intermediate parents will still call into their parent for shared properties as well.

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
+++ b/javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
@@ -57,8 +57,6 @@ public final class ClientModelPropertiesManager {
     private final LinkedHashMap<String, ClientModelProperty> readOnlyProperties;
     private final ClientModelProperty additionalProperties;
     private final ClientModelPropertyWithMetadata discriminatorProperty;
-    private final boolean allPolymorphicModelsInSamePackage;
-    private final boolean polymorphicDiscriminatorDefinedByModel;
     private final String expectedDiscriminator;
     private final JsonFlattenedPropertiesTree jsonFlattenedPropertiesTree;
     private final boolean allFlattenedPropertiesFromParent;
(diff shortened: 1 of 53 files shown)


javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
```diff
@@ -57,8 +57,11 @@ public final class ClientModelPropertiesManager {
     private final LinkedHashMap<String, ClientModelProperty> readOnlyProperties;
     private final ClientModelProperty additionalProperties;
     private final ClientModelPropertyWithMetadata discriminatorProperty;
+    private final boolean allPolymorphicModelsInSamePackage;
```
XiaofeiCao, line 60: Is this just the result of `PolymorphicDiscriminatorHandler.isAllPolymorphicModelsInSamePackage(model)`? 
It's seems weird to me as field of `ClientModelPropertiesManager`, can we put it in `ClientModel`?

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
```diff
@@ -1002,37 +1009,50 @@ private static void writeManagementErrorAdaption(JavaBlock methodBlock, ClientMo
...
-                .ifBlock("\"error\".equals(" + fieldNameVariableName + ")", ifAction -> {
-                    ifAction.line("return " + READ_MANAGEMENT_ERROR_METHOD_NAME + "(bufferedReader);");
-                }).elseBlock(elseAction -> {
-                    elseAction.line("bufferedReader.skipChildren();");
-                });
-        });
+        addReaderWhileLoop("bufferedReader", methodBlock, true, fieldNameVariableName, false, whileBlock ->
+            methodBlock.ifBlock("\"error\".equals(" + fieldNameVariableName + ")",
+                    ifAction -> ifAction.line("return " + READ_MANAGEMENT_ERROR_METHOD_NAME + "(bufferedReader);"))
+                .elseBlock(elseAction -> elseAction.line("bufferedReader.skipChildren();")));
```
XiaofeiCao, line 1015: Only format change?

javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
alzimmermsft: I should add some documentation around this. It doesn't work well being in `ClientModel`, as when the `ClientModel` is initialized the parents or children in the structure may not exist yet. That said, I could make this something that is set on access later on.

javagen/src/main/java/com/azure/autorest/template/StreamSerializationModelTemplate.java
alzimmermsft: Yes, this was just a formatting change

XiaofeiCao (approved):
LGTM.

Haven't dived into flatten code, though the generated code looks good.

PR #2921, opened 2024-08-19
sample/test, default value for missing required properties

XiaofeiCao:
fix https://github.com/Azure/autorest.java/issues/2611

### DIFF
--- a/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
+++ b/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
@@ -201,7 +201,7 @@ public class ModelExampleWriter {
                     if (node.getClientType() instanceof PrimitiveType) {
                         return node.getClientType().defaultValueExpression();
                     } else {
-                        literalValue = "null";
+                        return "null";
                     }
                 }
 
--- a/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
+++ b/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
@@ -296,7 +296,7 @@ public class ModelExampleWriter {
                         ctorPosition.put(properties.get(i), i);
                     }
 
-                    List<String> initAtCtors = new ArrayList<>(Collections.nCopies(properties.size(), ""));
+                    List<String> initAtCtors = new ArrayList<>(Collections.nCopies(properties.size(), null));
                     List<String> initAtSetters = new ArrayList<>();
                     for (ExampleNode childNode : node.getChildNodes()) {
                         ModelProperty modelProperty = clientModelNode.getClientModelProperties().get(childNode);
@@ -309,7 +309,7 @@ public class ModelExampleWriter {
                     }
                     for (int i = 0; i < properties.size(); ++i) {
                         String ctorParameterValue = initAtCtors.get(i);
-                        if ("".equals(ctorParameterValue)) { // not present, due to missing required property
+                        if (ctorParameterValue == null) { // not present, due to missing required property
                             if (properties.get(i).getClientType() instanceof PrimitiveType) {
                                 initAtCtors.set(i, properties.get(i).getClientType().defaultValueExpression());
                             } else {
--- a/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
+++ b/javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
@@ -196,7 +196,16 @@ public class ModelExampleWriter {
                     helperFeatures.add(ExampleHelperFeature.ThrowsIOException); // MalformedURLException from URL ctor
                 }
 
-                return node.getClientType().defaultValueExpression(((LiteralNode) node).getLiteralsValue());
+                String literalValue = ((LiteralNode) node).getLiteralsValue();
+                if (literalValue == null) {
+                    if (node.getClientType() instanceof PrimitiveType) {
+                        return node.getClientType().defaultValueExpression();
+                    } else {
+                        literalValue = "null";
+                    }
+                }
+
+                return node.getClientType().defaultValueExpression(literalValue);
             } else if (node instanceof ObjectNode) {
                 IType simpleType = null;
                 if (node.getObjectValue() instanceof Integer) {
@@ -296,6 +305,16 @@ public class ModelExampleWriter {
                         } else {
                             // .setProperty(...)
                             initAtSetters.add(String.format(".%1$s(%2$s)", modelProperty.getSetterName(), this.accept(childNode)));
+                        }
+                    }
+                    for (int i = 0; i < properties.size(); ++i) {
+                        String ctorParameterValue = initAtCtors.get(i);
+                        if ("".equals(ctorParameterValue)) { // not present, due to missing required property
+                            if (properties.get(i).getClientType() instanceof PrimitiveType) {
+                                initAtCtors.set(i, properties.get(i).getClientType().defaultValueExpression());
+                            } else {
+                                initAtCtors.set(i, "null");
+                            }
                         }
                     }
                     // model constructor
--- a/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
+++ b/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
@@ -179,7 +179,11 @@ public class ModelExampleUtil {
                 node = defaultNode(type, wireType, objectValue);
             }
         } else if (objectValue == null) {
-            node = null;
+            if (ClassType.CONTEXT.equals(type)) {
+                node = new LiteralNode(ClassType.CONTEXT, "").setLiteralsValue("");
+            } else {
+                node = new LiteralNode(type, null);
+            }
         } else {
             node = defaultNode(type, wireType, objectValue);
         }
--- a/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
+++ b/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
@@ -178,12 +178,6 @@ public class ModelExampleUtil {
                 // e.g. do not throw exception, use defaultValueExpression
                 node = defaultNode(type, wireType, objectValue);
             }
-        } else if (objectValue == null) {
-            if (ClassType.CONTEXT.equals(type)) {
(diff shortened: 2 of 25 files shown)


typespec-tests/src/test/java/com/cadl/flatten/generated/FlattenOpSendLongTests.java
```diff
@@ -17,7 +17,7 @@ public final class FlattenOpSendLongTests extends FlattenClientTestBase {
     public void testFlattenOpSendLongTests() {
         // method invocation
         flattenClient.sendLong(
-            new SendLongOptions("myRequiredId", "myRequiredInput", 11, "title", SendLongRequestStatus.NOT_STARTED)
+            new SendLongOptions("myRequiredId", "myInput", 0, null, "title", SendLongRequestStatus.NOT_STARTED)
```
XiaofeiCao, line 20: 0 is absence of required primitive.
null is absence of required model.

typespec-tests/src/test/java/com/cadl/model/generated/ModelOpPutNestedTests.java
```diff
@@ -17,7 +17,7 @@ public final class ModelOpPutNestedTests extends ModelClientTestBase {
     @Disabled
     public void testModelOpPutNestedTests() {
         // method invocation
-        NestedModel response = modelClient.putNested(null);
+        NestedModel response = modelClient.putNested(new NestedModel(new NestedModel1(new NestedModel2("123"))));
```
XiaofeiCao, line 20: This is due to a fix [here](https://github.com/Azure/autorest.java/pull/2921/files#diff-ad84edf0aad661e6e721088dc8c894fa8099fc3ce43a39226a22ea79971ea65d)

typespec-tests/tsp/examples/2022-06-01-preview/FlattenOp_SendLong.json
```diff
@@ -8,8 +8,7 @@
       "user": {
         "user": "myOptionalUser"
       },
-      "input": "myRequiredInput",
-      "dataInt": 11,
+      "input": "myInput",
```
XiaofeiCao, line 11: `myRequiredInput` -> `myInput` is not intended.. Just don't want to regen the code again..

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
```diff
@@ -179,7 +179,11 @@ public static ExampleNode parseNode(IType type, IType wireType, Object objectVal
                 node = defaultNode(type, wireType, objectValue);
             }
         } else if (objectValue == null) {
-            node = null;
+            if (ClassType.CONTEXT.equals(type)) {
+                node = new LiteralNode(ClassType.CONTEXT, "").setLiteralsValue("");
```
weidongxu-microsoft, line 183: Why we need this? I think we already had special handling for `ClassType.CONTEXT`.

javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
```diff
@@ -298,6 +307,16 @@ public String accept(ExampleNode node) {
...
+                    for (int i = 0; i < properties.size(); ++i) {
+                        String ctorParameterValue = initAtCtors.get(i);
+                        if ("".equals(ctorParameterValue)) { // not present, due to missing required property
+                            if (properties.get(i).getClientType() instanceof PrimitiveType) {
+                                initAtCtors.set(i, properties.get(i).getClientType().defaultValueExpression());
+                            } else {
+                                initAtCtors.set(i, "null");
+                            }
+                        }
+                    }
```
weidongxu-microsoft, line 319: Should this logic be on line 304, when setting the ctor?
https://github.com/Azure/autorest.java/pull/2921/files#diff-5ad12a8cf60cd3edf947aae6e321e17758b17fedb00713c2cdbc29a1e1087bf9R304

setting it, then getting it and compare to `""` is strange.

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
XiaofeiCao: Got it. It should already be handled in `parseNodeFromParameter` before entering `parseNode`.

javagen/src/main/java/com/azure/autorest/template/example/ModelExampleWriter.java
XiaofeiCao: missing node won't reach L304, this is the cause of the bug.. `""` is set during initialization on L299

Maybe we could reverse the iteration, iterating model properties instead of child nodes. This would guarantee each property gets processed.

weidongxu-microsoft: Can we init it as `null`? That at least would make the missing of value apparent.

XiaofeiCao: sure

weidongxu-microsoft (approved):

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
```diff
@@ -178,8 +178,6 @@ public static ExampleNode parseNode(IType type, IType wireType, Object objectVal
                 // e.g. do not throw exception, use defaultValueExpression
                 node = defaultNode(type, wireType, objectValue);
             }
-        } else if (objectValue == null) {
-            node = null;
         } else {
             node = defaultNode(type, wireType, objectValue);
```
weidongxu-microsoft, line 182: Is it expected that we merge this 2 case into 1?

typespec-tests/src/test/java/com/cadl/flatten/generated/FlattenOpSendLongTests.java
haolingdong-msft: Do we have document somewhere to list different type's default value?

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
XiaofeiCao: Good catch. I should've brought this up. 

May I ask what does null mean for ExampleNode previously? I'm guessing it means absence of node, but I wasn't able to find any null checks for returned `parseNode`. Closest I can find is that everywhere we call `parseNode`, we ensure the passed in `objectValue` is not null, e.g. https://github.com/Azure/autorest.java/blob/9b59bd39524bd50b79d04458b464d5d4a241680b/fluentgen/src/main/java/com/azure/autorest/fluent/mapper/ExampleParser.java#L479-L490
, and this behaves identical to `defaultNode`. 

In this sense, I kinda feel like null is an unexpected return value. Would like your opinion on this.

typespec-tests/src/test/java/com/cadl/flatten/generated/FlattenOpSendLongTests.java
XiaofeiCao: I wasn't able to find any. 

In Javadoc, for general case, it's in `IType`:https://github.com/Azure/autorest.java/blob/403ffcb32add0d2b9457bb86494ca68ae15605c5/javagen/src/main/java/com/azure/autorest/model/clientmodel/IType.java#L81-L90

For other scenarios, you'll need to check each sub type's definition, e.g. https://github.com/Azure/autorest.java/blob/403ffcb32add0d2b9457bb86494ca68ae15605c5/javagen/src/main/java/com/azure/autorest/model/clientmodel/ClassType.java#L197

weidongxu-microsoft: here https://github.com/Azure/autorest.java/blob/403ffcb32add0d2b9457bb86494ca68ae15605c5/javagen/src/main/java/com/azure/autorest/model/clientmodel/PrimitiveType.java#L36

javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java
XiaofeiCao: I feel like letting downstream decide what to do with node with null value, instead of a null node.

weidongxu-microsoft: For some (commonly seen) cases, if the value is null, this is already filtered out, e.g. for class model.

```
                for (Object childObjectValue : elements) {
                    ExampleNode childNode = parseNode(elementType, childObjectValue);
                    node.getChildNodes().add(childNode);
                }
```

for parameter, it seems also already handled directly
```
        if (exampleValue == null) {
            if (ClassType.CONTEXT.equals(methodParameter.getClientMethodParameter().getClientType())) {
                node = new LiteralNode(ClassType.CONTEXT, "").setLiteralsValue("");
            } else {
                node = new LiteralNode(methodParameter.getClientMethodParameter().getClientType(), null);
            }
        } else {
```

I guess one remaining place would be a `null` value in JSON array...

weidongxu-microsoft: > I feel like letting downstream decide what to do with node with null value, instead of a null node. Added comment on this.

Yeah, make sense, as long as we've handled the known types, and what's left be Literal (basically types about primitive or datetime etc.).

XiaofeiCao: Thanks! 
```
                for (Object childObjectValue : elements) {
                    ExampleNode childNode = parseNode(elementType, childObjectValue);
                    node.getChildNodes().add(childNode);
                }
```
This seems to be the JSON array case? But yeah, we filtered out null for class model: 
https://github.com/Azure/autorest.java/blob/9b59bd39524bd50b79d04458b464d5d4a241680b/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java#L112

And
https://github.com/Azure/autorest.java/blob/9b59bd39524bd50b79d04458b464d5d4a241680b/javagen/src/main/java/com/azure/autorest/util/ModelExampleUtil.java#L146-L148

Then I prefer to use a default node with null value instead of a null node. 
1. We need to deal with every node, either property or parameter. For missing value, we have to determine its default value based on its client type.
2. We don't have to do null checks for node when we use it.
3. For List case, we don't have to pay extra attention about null value in example array.

PR #2927, opened 2024-08-20
bug fix, use package-private setter if not usingFromJsonShared

XiaofeiCao:
compilation error: https://github.com/Azure/azure-sdk-for-java/blob/970eed98c5b0054ab82b04beefbc24a3f695fb68/sdk/easm/azure-analytics-defender-easm/src/main/java/com/azure/analytics/defender/easm/models/AsAssetResource.java#L129

cause: 
Child model has a required property, thus it doesn't use fromJsonShared. We should use package-private setters in fromJson, not through property reference.


XiaofeiCao:
Should already be fixed by https://github.com/Azure/autorest.java/pull/2928

PR #2928, opened 2024-08-20
Simplify StreamSerializationModelTemplate

alzimmermsft:
Cleans up a lot of code in `StreamSerializationModelTemplate` by using an instance class to generate code. This removes the need to pass many parameters all over the place.

There is also a fix to generating fields in `ModelTemplate` by having the overload of `getFieldProperties` check if `fromJsonShared` can be used to determine whether the subclass should shade fields of a super class. The only time shading shouldn't happen with polymorphic types is when `fromJsonShared` can be used as those properties will be deserialized by the super type's `fromJsonShared` method, meaning the subtype doesn't need to shade.


weidongxu-microsoft:
test https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4067144&view=results
https://github.com/Azure/azure-sdk-for-java/pull/41571

vanilla-tests/src/main/java/fixtures/bodycomplex/models/Goblinshark.java
```diff
@@ -41,7 +49,15 @@ public final class Goblinshark extends Shark {
...
         super(length, birthday);
-        this.fishtype = "goblin";
+    }
+
+    /**
+     * Get the fishtype property: The fishtype property.
+     * 
+     * @return the fishtype value.
+     */
+    public String getFishtype() {
```
XiaofeiCao, line 59: There should be an @Override? Seems it overrides parent discriminator getter.

haolingdong-msft:
Hi @alzimmermsft, I merged your pr with main branch in a feature branch, and tried to regenerate for TypeSpec SDKs based on the branch. I found API view diff like below, `@override` is removed from `getKind()` in subclass. I suppose after setting `share-jsonserializable-code` to `false`, there shoud be no diff. It seems a bug. Could u please check? 
image

PR: https://github.com/Azure/azure-sdk-for-java/pull/41617
diff example: https://github.com/Azure/azure-sdk-for-java/pull/41617/files#diff-90b30f05da823f3aaaf96caa3a6234d229c84b3e93c10224183857a36a9c3b02L63

Thanks.

vanilla-tests/src/main/java/fixtures/bodycomplex/models/Goblinshark.java
alzimmermsft: This is fixed now

weidongxu-microsoft (approved):

PR #2934, opened 2024-08-26
tsp, handle spread behavior change from TCGC

weidongxu-microsoft:
nightly dev passes

diff https://dev.azure.com/azure-sdk/public/_build/results?buildId=4083808&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=8d0a7aee-8090-5653-72a7-7c81ab87cde7


typespec-extension/src/code-model-builder.ts
```diff
@@ -1316,15 +1319,20 @@ export class CodeModelBuilder {
       this.trackSchemaUsage(schema, { serializationFormats: [KnownMediaType.Multipart] });
     }
 
-    // Implicit body parameter would have usage flag: UsageFlags.Spread, for this case we need to do body parameter flatten
-    const bodyParameterFlatten = sdkType.kind === "model" && sdkType.usage & UsageFlags.Spread && !this.isArm();
+    // Explicit body parameter @body or @bodyRoot would result to the existance of rawHttpOperation.parameters.body.property
+    // Implicit body parameter would result to rawHttpOperation.parameters.body.property be undefined
+    // see https://typespec.io/docs/libraries/http/cheat-sheet#data-types
+    const bodyParameterFlatten =
+      sdkType.kind === "model" && !rawHttpOperation.parameters.body?.property && !this.isArm();
```
weidongxu-microsoft, line 1326: flatten or not should not be based on the model, it should be based on the client method (which is yet to be included in this migration).

typespec-extension/src/code-model-builder.ts
```diff
@@ -1302,119 +1305,138 @@ export class CodeModelBuilder {
...
+        if (!parameter.language.default.name) {
+          // name the parameter for documentation
+          parameter.language.default.name = "request";
+        }
 
-      if (operationIsJsonMergePatch(sdkHttpOperation)) {
-        // skip model flatten, if "application/merge-patch+json"
-        schema.language.default.name = pascalCase(op.language.default.name) + "PatchRequest";
-        return;
-      }
```
weidongxu-microsoft, line 1338: This change does not relate to the issue.

But we should use name from TCGC in almost all cases.

typespec-extension/src/code-model-builder.ts
```diff
@@ -1302,119 +1305,138 @@ export class CodeModelBuilder {
...
-
-      if (!parameter.language.default.name) {
-        // name the parameter for documentation
-        parameter.language.default.name = "request";
-      }
+    if (op.convenienceApi) {
+      // Explicit body parameter @body or @bodyRoot would result to the existance of rawHttpOperation.parameters.body.property
+      // Implicit body parameter would result to rawHttpOperation.parameters.body.property be undefined
+      // see https://typespec.io/docs/libraries/http/cheat-sheet#data-types
+      const bodyParameterFlatten =
```
haolingdong-msft, line 1331: Use Chenjie's recommended check for is body spread? https://teams.microsoft.com/l/message/19:meeting_MWQ2ODAwOTgtMmI4Mi00MWY2LTkxZjgtNzE0N2FjOGJhNDVi@thread.v2/1724754309829?context=%7B%22contextType%22%3A%22chat%22%7D

typespec-extension/src/common/schemas/usage.ts
```diff
@@ -20,7 +20,7 @@ export enum SchemaContext {
   Paged = "paged",
 
   /** Schema as anonymous model. This usage does not propagate. */
```
haolingdong-msft, line 22: update the comment?

typespec-tests/src/main/java/com/type/union/implementation/models/SendRequest9.java
```diff
@@ -67,13 +67,13 @@ public JsonWriter toJson(JsonWriter jsonWriter) throws IOException {
...
     public static SendRequest9 fromJson(JsonReader jsonReader) throws IOException {
         return jsonReader.readObject(reader -> {
-            MixedTypesCases prop = null;
+            GetResponseProp4 prop = null;
             while (reader.nextToken() != JsonToken.END_OBJECT) {
                 String fieldName = reader.getFieldName();
                 reader.nextToken();
 
                 if ("prop".equals(fieldName)) {
-                    prop = MixedTypesCases.fromJson(reader);
```
haolingdong-msft, line 76: ok, order changes again..

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: Both is temporary solution.

haolingdong-msft: May I know why is Chenjie's one temp solution? Is there any futhur plan for checking if a body needs spread? 
And also using Chenjie's check could reduce our usage on raw type.

typespec-tests/src/main/java/com/type/union/implementation/models/SendRequest9.java
weidongxu-microsoft: Mentioned here. The reason is you are renaming the model where TCGC already generate the name.
https://github.com/Azure/autorest.java/pull/2934#discussion_r1732119327

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: I think I told multiple times, that we should use `SdkServiceMethod` for method signature, not `SdkServiceOperation`. And grouping/flattening is the method signature.

haolingdong-msft: I just would like to know do we have furthur plan on how to check if a body needs spread when using `SdkServiceOperation`. 
Yeah, we should adopt `SdkServiceMethod` and I agree if we use `SdkServiceMethod`, we could check grouping/spread there. But I'm not sure when will we adopt `SdkServiceMethod`, if it is not short term, personally I feel it is better to follow the guidance when using `SdkServiceOperation`.

typespec-tests/src/main/java/com/type/union/implementation/models/SendRequest9.java
haolingdong-msft: ok, I thought the [comment](https://github.com/Azure/autorest.java/pull/2934#discussion_r1732119327) is corresponding to this diff: https://github.com/Azure/autorest.java/pull/2934/files#diff-069b8a8df9585df2f250fe704a3f33872d32991db1db3959d3b943bc9fba2042L181
So the order changes are also due to we renamed?

typespec-extension/src/code-model-builder.ts
weidongxu-microsoft: We DON'T check body flatten on `SdkServiceOperation`.

I WILL DO THAT WHEN I HAVE TIME, if you don't do that.

haolingdong-msft: It's not to check directly on `SdkServiceOperation`, it checks propertiese on  `SdkServiceOperation` like `SdkServiceOperation.bodyParam`. There is limitation if we want to use `SdkServiceMethod`, see this [comment](https://github.com/Azure/autorest.java/pull/2861#issuecomment-2297882441) in my previous pr. That's why I said I don't know if we can use `SdkServiceMethod` soon. (If we can use `SdkMethod`, I will use  it in the begining. It's not about who do it.. I can do it as well when I get time. )
I will check if there is existing TCGC issue on this, previously I just talked with Chenjie offline, and he said getting correct parameters mapping will be supported. If no issue, I will create one.

typespec-extension/src/code-model-builder.ts
```diff
@@ -463,11 +460,17 @@ export class CodeModelBuilder {
...
-          schemaUsage.push(SchemaContext.Internal);
+      // Internal on PublicSpread, but Public takes precedence
+      if (schemaUsage?.includes(SchemaContext.PublicSpread)) {
+        // remove PublicSpread as it served its purpose
+        schemaUsage.splice(schemaUsage.indexOf(SchemaContext.PublicSpread), 1);
+
+        // Public would override PublicSpread, hence do nothing if this schema is Public
+        if (!schemaUsage?.includes(SchemaContext.Public)) {
+          const index = schemaUsage.indexOf(SchemaContext.Internal);
+          if (index < 0) {
```
haolingdong-msft, line 471: I'm a bit confused, What does `index < 0` mean? `schemaUsage` include `SchemaContext.Internal` or not include `SchemaContext.Internal` ? Can we use `schemaUsage?.includes(SchemaContext.Internal)` instead as it is more straightforward?

weidongxu-microsoft: Yes, it is for `includes`. Changed.

haolingdong-msft (approved):

weidongxu-microsoft:
I will merge the PR after Haoling done the release and sync of 0.20.0

haolingdong-msft:
publishing to npm, I will notify here once I triggered sync sdk pipeline on main so that you can merge the pr.

haolingdong-msft:
Hi @weidongxu-microsoft, I have triggered sync sdk pipeline on main branch.

PR #2937, opened 2024-08-28
typespec submodule

weidongxu-microsoft:
Goal: remove most of the TypeScript and Java files from autorest.java, while still have CI pass.

## Generator Code

- removed code in typespec-extension (but kept package.json, so it can still pack to typespec-java)
- removed preprocessor, postprocessor, fluentnamer
- reduced code to Main.java in javagen, fluentgen

### Swagger

In short, only 2 `Main.java` remains.

`@autorest/java` is built from javagen and fluentgen. The lib itself depends on "http-client-generator-core" and "http-client-generator-mgmt".

(PS: in future, we likely going to move Swagger/Azure specific code out of microsoft/typesepc repo and put to autorest.java repo -- especially code only for mgmt)

### TypeSpec

On `typespec-extension`, TypeScript code will be copied from http-client-java/emitter, JAR will be copied from http-client-java/generator, before pack as typespec-java. See [Build-TypeSpec.ps1](https://github.com/Azure/autorest.java/pull/2937#discussion_r1744899421)

## Unit test

Unit test is kept. Plan to move some to typespec repo.

## E2E test

Swagger related test is not touched, and will be regen and run, as before.

Code in typespec-test is not removed in this PR.
Currently they are duplicate of what's in microsoft/typespec repo, but in future they can be different. (as code here would focus on flavor=azure)

## Other

Secondary goal: able to release `@autorest/java` package. If necessary also typespec-java package.

Didn't really publish, but it should work: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4102946&view=results
sample sync PR https://github.com/Azure/azure-sdk-for-java/pull/41704


Build-TypeSpec.ps1
weidongxu-microsoft, line 1: Updated this `Build-TypeSpec.ps1` script to build the lib (TypeScript and Java) in core.

It also made the patch.

core.diff
weidongxu-microsoft, line 1: The patch to update core to use "azure-autorest-customization" lib for customization code.

Generate-TypeSpec.ps1
```diff
@@ -56,10 +36,10 @@ Push-Location ./typespec-tests
 
 try {
   Write-Host "Generating code ('Generate.ps1' in './typespec-tests')"
-  invokeExpressionAndCaptureOutput("./Generate.ps1 -Parallelization $Parallelization")
+  pwsh ./Generate.ps1 -Parallelization $Parallelization
 
-  Write-Host "Checking format of generated code ('npm run check-format')"
-  invokeExpressionAndCaptureOutput("npm run check-format")
+#   Write-Host "Checking format of generated code ('npm run check-format')"
+#   invokeExpressionAndCaptureOutput("npm run check-format")
```
weidongxu-microsoft, line 42: Skip check-format as there are slight difference of format between repo.

eng/pipelines/ci-typespec-java.yaml
```diff
@@ -38,6 +38,9 @@ jobs:
       vmImage: "MMSUbuntu20.04"
 
     steps:
+      - checkout: self
```
haolingdong-msft, line 41: Is the release process for typespec-java same as before?

weidongxu-microsoft: I hope so. (I think it is covered in this PR, but cannot be really sure till I trigger one)

I will test upon next release (probably on a branch that at first disables the publish).

.gitignore
```diff
@@ -104,3 +104,7 @@ typespec-tests/package-lock.json
 
 # ApiView properties files
 /typespec-tests/src/main/resources/META-INF/*apiview_properties.json
+
+# typespec-extension
+/typespec-extension/src
+/typespec-extension/generator
```
haolingdong-msft, line 110: what is `/typespec-extension/generator` for?

weidongxu-microsoft: These are the places we copy JAR and TypeScript code into typespec-extension (from core) -- so that typespec-extension can still pack to typespec-java

Since they are copied, we don't want to commit them.

typespec-extension/package.json
```diff
@@ -42,7 +42,7 @@
     "lib/*.tsp",
     "dist/**",
     "!dist/test/**",
-    "target/emitter.jar"
+    "generator/http-client-generator/target/emitter.jar"
```
XiaofeiCao, line 45: Is it possible to reference outer folder? e.g. "../core/packages/http-client-generator/target/emitter.jar"?

weidongxu-microsoft: Appears no.

```
  "files": [
    "lib/*.tsp",
    "dist/**",
    "!dist/test/**",
    "../emitter.jar"
  ],
```

And I cannot find the jar in tgz.

Though not an expert on npm pack...

XiaofeiCao (approved):

haolingdong-msft (approved):
overall looks good to me, we can have a try first.

eng/pipelines/ci.yaml
```diff
@@ -32,6 +32,15 @@ jobs:
...
 
     steps:
+      - checkout: self
+        submodules: true
+
+      - script: |
+          git checkout .
+          git apply ../core.diff --ignore-whitespace
+        displayName: 'Patch core'
+        workingDirectory: ./core
```
weidongxu-microsoft, line 42: I've put this patch to `Generate-TypeSpec.ps1` (actually this does the Build too).

But I didn't add this to `Generate.ps1` as it does not contain the Build step (it is in follow step in yaml)
```
      - task: Maven@3
        displayName: 'Build and Package'
        inputs:
          mavenPomFile: pom.xml
          goals: 'clean install'
          options: '$(DefaultOptions) -P local -T 1C'
          mavenOptions: '$(MemoryOptions) $(LoggingOptions)'
          javaHomeOption: 'JDKVersion'
          jdkVersionOption: $(JavaVersion)
          jdkArchitectureOption: 'x64'
          publishJUnitResults: false
```

Let me know what's the preference. (usually this patch only need to be done once after submodule init).

PR #2941, opened 2024-09-02
tsp, pin cadl-ranch-expect

weidongxu-microsoft:


XiaofeiCao (approved):

typespec-extension/src/code-model-builder.ts
```diff
@@ -2159,14 +2159,14 @@ export class CodeModelBuilder {
       case "Model":
         if (isArrayModelType(this.program, type)) {
           ++option.depth;
-          if (option.depth == 1) {
+          if (option.depth === 1) {
```
weidongxu-microsoft, line 2162: These are the changes in http-client-java (check there would flag the `==`)

PR #2943, opened 2024-09-06
Fix case where property is defined in two different ways

alzimmermsft:
Fixes an issue where the polymorphic structure defines a property once as a non-polymorphic property and once as a polymorphic property, which resulted in it being serialized and deserialized twice when using stream-style-serialization.


weidongxu-microsoft (commented):
@XiaofeiCao Could you take a look (that this at least won't affect other common scenarios)?

@alzimmermsft If this is merged, please also sync it to microsoft/typespec repo https://github.com/microsoft/typespec/tree/main/packages/http-client-java. We are likely going to release next version of typespec-java from Java code in that repo.

javagen/src/main/java/com/azure/autorest/implementation/ClientModelPropertiesManager.java
```diff
@@ -234,6 +234,24 @@ public ClientModelPropertiesManager(ClientModel model, JavaSettings settings) {
             }
         }
 
+        // Temporary fix to a larger problem where the discriminator property is defined by a parent model, but not as
+        // a discriminator. This results in the discriminator property being serialized and deserialized twice as it
+        // shows up once as a regular property and once as a discriminator property. This will remove the regular
+        // property from the super properties and indicate that the discriminator came from a parent model.
+        if (discriminatorProperty != null) {
```
XiaofeiCao, line 241: Does this `discriminatorProperty` always mean it's from current(child) class?

XiaofeiCao (approved):
There could be an edge case where different inheritance hierarchy share different discriminators:
```typespec
model modelA {
  disA: string,
  prop1: string
}

@discriminator("disB")
model modelB extends modelA {
  prop2: string
}

@discriminator("disA")
model modelC extends modelA {
  prop3: string
}

model modelD extends modelB {
  disB: "D"
}

model modelE extends modelC {
  disA: "E"
}
```

Could you help see if it works? From the code, I think it works, just in case.

PR #2953, opened 2024-09-23
fix non-string expandable enum serialization, mgmt test case

XiaofeiCao:
- added test case for mgmt Swagger
- depends on https://github.com/microsoft/typespec/pull/4492


fluent-tests/src/test/java/com/azure/mgmttest/CompilationTests.java
```diff
@@ -175,4 +179,9 @@ public void testSchemaCleanup() {
     public void testResourceWithWritableName() {
         FirewallRuleInner firewallRuleInner = new FirewallRuleInner();
     }
+
+    public void testNonStringExpandableEnum() throws IOException {
+        ScheduledQueryRuleProperties properties = new ScheduledQueryRuleProperties();
+        String jsonString = properties.toJsonString();
```
weidongxu-microsoft, line 185: Why does the model have `toJsonString` method? I didn't recall we had any model with this method?

XiaofeiCao: This is provided by JsonSerializable's default implementation: 
https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/serialization/azure-json/src/main/java/com/azure/json/JsonSerializable.java#L78-L84

weidongxu-microsoft: I see. 1.2.0?

weidongxu-microsoft (approved):

fluent-tests/src/test/java/com/azure/mgmttest/CompilationTests.java
XiaofeiCao: Yes. https://github.com/Azure/azure-sdk-for-java/pull/41253/files#diff-0197788260b7e501fdb9e98100746846527a9a78b1c06dfa6327ec39e93ed9beR78

PR #2954, opened 2024-09-24
eng, bump core jar

weidongxu-microsoft:


vanilla-tests/pom.xml
```diff
@@ -42,14 +42,14 @@
     <dependency>
       <groupId>com.azure</groupId>
       <artifactId>azure-core-test</artifactId>
-      <version>1.26.2</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.azure</groupId>
-      <artifactId>azure-core-http-netty</artifactId>
```
weidongxu-microsoft, line 50: netty was removed because azure-core-test now contains all http clients.

vanilla-tests/pom.xml
```diff
@@ -42,14 +42,14 @@
...
-      <version>1.15.3</version>
+      <version>1.27.0-beta.1</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>com.azure</groupId>
+          <artifactId>azure-core-http-jdk-httpclient</artifactId>
```
weidongxu-microsoft, line 50: excludes azure-core-http-jdk-httpclient as it does not support Java 11

XiaofeiCao (approved):

PR #2959, opened 2024-09-30
eng, bump core, prepare 0.21.2

weidongxu-microsoft:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@autorest/java",
-  "version": "4.1.40",
+  "version": "4.1.39",
```
weidongxu-microsoft, line 3: This was a mistake in prior PR. I seems bumped it twice without release 4.1.39

XiaofeiCao (approved):

PR #2961, opened 2024-10-11
tsp, prepare release for upgrading tsp and related libs to 0.61

haolingdong-msft:


weidongxu-microsoft:
"update core" should be the 1st step

haolingdong-msft:
> "update core" should be the 1st step
> 
> the "sync" is meaningless when your core is not the correct version

Thanks. Yeah, I realized it, so I run sync_versions and syncTests script again after update core. But I committed the sync files first and then core. Sorry for the confusion.

weidongxu-microsoft:
sync_version would update into the core. So if you run it, you should run it before core check-in...
But hope it does not matter...

haolingdong-msft:
> sync_version would update into the core. So if you run it, you should run it before core check-in... But hope it does not matter... And this should not be the cause of your diff.

My previous diff is caused by I did not regen for vanilla-tests.

vanilla-tests/src/test/java/fixtures/nonstringenum/NonStringEnumTests.java
```diff
@@ -27,6 +27,7 @@ public void getInt() {
         assertEquals(IntEnum.FOUR_HUNDRED_TWENTY_NINE, actual);
     }
 
+    @Disabled("wire type changed for Swagger extensible non-string enum")
```
haolingdong-msft, line 30: Actually seems all tests in this class are disabled. Shall we delete this test class?

weidongxu-microsoft:
> > sync_version would update into the core. So if you run it, you should run it before core check-in... But hope it does not matter... And this should not be the cause of your diff.
> 
> My previous diff is caused by I did not regen for vanilla-tests.

I know. What I mean is the jar dependency in both repo best to be same. If different there could be unexpected situation that could waste dev effort to investigate.

vanilla-tests/src/test/java/fixtures/nonstringenum/NonStringEnumTests.java
weidongxu-microsoft: Leave it. It is autorest

haolingdong-msft:
> > > sync_version would update into the core. So if you run it, you should run it before core check-in... But hope it does not matter... And this should not be the cause of your diff.
> > 
> > 
> > My previous diff is caused by I did not regen for vanilla-tests.
> 
> I know. What I mean is the jar dependency in both repo best to be same. If different there could be unexpected situation that could waste dev effort to investigate.

ok, I see.

weidongxu-microsoft (approved):
unblock

XiaofeiCao (approved):

PR #2976, opened 2024-11-07
eng, report error, but do not fail during regen SDK

weidongxu-microsoft:
Problem:
1. some lib fails, but we are not able to bump the commit ID because problem on specs repo (e.g. tspconfig removed from https://github.com/Azure/azure-rest-api-specs/tree/main/specification/ai/HealthInsights/HealthInsights.TrialMatcher)
2. typespec source change frequently require us to bump commit ID; it would be easier to see what's fails, and do a batch update

Dev need to check the sync SDK CI to see if there is failure.

At present, known failure is
- azure-health-insights-clinicalmatching

### DIFF
--- a/eng/sdk/sync_sdk.py
+++ b/eng/sdk/sync_sdk.py
@@ -144,7 +144,7 @@ def update_sdks():
     subprocess.check_call(cmd, cwd=sdk_root)
 
     if failed_artifact:
-        logging.error(f'Failed modules f{failed_artifact}')
+        logging.error(f'Failed modules {failed_artifact}')
 
 
 def main():
--- a/eng/sdk/sync_sdk.py
+++ b/eng/sdk/sync_sdk.py
@@ -17,7 +17,9 @@ sdk_root: str
 
 skip_artifacts: List[str] = [
     'azure-ai-anomalydetector',         # deprecated
-    'azure-ai-vision-imageanalysis'     # temporary disabled for modification on Javadoc
+    'azure-ai-vision-imageanalysis',    # temporary disabled for modification on Javadoc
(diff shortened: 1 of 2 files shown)


eng/sdk/sync_sdk.py
```diff
@@ -137,6 +143,9 @@ def update_sdks():
     cmd = ['git', 'add', '.']
     subprocess.check_call(cmd, cwd=sdk_root)
 
+    if failed_artifact:
+        logging.error(f'Failed modules {failed_artifact}')
```
weidongxu-microsoft, line 147: Though, generally we actually want a failure. This can be done here by re-throwing an error.

eng/sdk/sync_sdk.py
```diff
@@ -17,6 +17,7 @@
 
 skip_artifacts: List[str] = [
     'azure-ai-anomalydetector',         # deprecated
+    'azure-health-insights-cancerprofiling', # deprecated
```
weidongxu-microsoft, line 20: Service told me this lib get deprecated too.

eng/sdk/sync_sdk.py
```diff
@@ -89,6 +90,7 @@ def get_generated_folder_from_artifact(module_path: str, artifact: str, type: st
 
 
 def update_sdks():
+    failed_artifact = []
```
XiaofeiCao, line 93: nit, failed_artifacts

XiaofeiCao (approved):

eng/sdk/sync_sdk.py
weidongxu-microsoft: thanks. renamed to `failed_modules`

PR #2985, opened 2024-11-15
mgmt, add mock test for CAE

XiaofeiCao:
- fix https://github.com/Azure/azure-sdk-for-java/issues/42935
- Per @archerzz 's ask, add e2e test for SDK CAE behavior. Personally, I think [HttpPipeline test](https://github.com/Azure/azure-sdk-for-java/blob/4cf3d9f1c9a1e5c8d3696079940cc94fc65545fe/sdk/core/azure-core/src/test/java/com/azure/core/http/policy/BearerTokenAuthenticationPolicyTests.java#L44) in azure-core should suffice, though it doesn't hurt to have an e2e test for generated SDK.

### DIFF
--- a/typespec-tests/src/test/java/tsptest/armresourceprovider/ArmTests.java
+++ b/typespec-tests/src/test/java/tsptest/armresourceprovider/ArmTests.java
@@ -1,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package tsptest.armresourceprovider;
+
+import com.azure.core.credential.AccessToken;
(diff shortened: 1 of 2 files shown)


weidongxu-microsoft:
I think mock test in azure-core should be enough. I am fine if you want to add one more here.

XiaofeiCao:
> I think mock test in azure-core should be enough. I am fine if you want to add one more here.

Yeah, I agree. 
Let's make this test to showcase that Java SDK is CAE-ready with latest 1.54.0 azure-core.

weidongxu-microsoft (approved):

PR #2987, opened 2024-11-21
release azure autorest customization 1.0.0 beta.9

weidongxu-microsoft:
There seems some problem in customization test. It appears to loading the azure-autorest-customization lib from Maven, instead of getting it from local installed lib (which we are testing). Not sure why though.

### DIFF
--- a/javagen/src/main/resources/readme/pom.xml
+++ b/javagen/src/main/resources/readme/pom.xml
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>com.azure.tools</groupId>
       <artifactId>azure-autorest-customization</artifactId>
-      <version>1.0.0-beta.8</version>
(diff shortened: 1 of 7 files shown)


javagen/src/main/resources/readme/pom.xml
```diff
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>com.azure.tools</groupId>
       <artifactId>azure-autorest-customization</artifactId>
-      <version>1.0.0-beta.8</version>
+      <version>1.0.0-beta.9</version>
```
weidongxu-microsoft, line 21: Not sure whether we are required to release a new version of typespec-java as well.

haolingdong-msft (approved):

PR #2994, opened 2024-12-09
Prepare autorest 4.1.43/typespec 0.26.1

XiaofeiCao:
no issue in sync-sdk: https://github.com/Azure/azure-sdk-for-java/pull/43293


weidongxu-microsoft (approved):
you probably need to bump tcgc and/or cadl-ranch (I guess their versions are bumped in typespec repo, since last sync).

PR #2999, opened 2024-12-17
eng, update core, default flavor should be azure now

weidongxu-microsoft:
fix https://github.com/microsoft/typespec/issues/5380

I am adding more details about commonly used emitter options in readme

As expected, the enabling of TCGC diagnostic to compiler, result in 1 failure on azure-ai-inference. Guess we ignore it for now, as we already contacted the owner about their potential error in typespec.


typespec-extension/readme.md
weidongxu-microsoft, line 1: Major thing in this PR, is that I am adding more detailed documentation on emitter options here. We may move it elsewhere later.

typespec-extension/readme.md
```diff
@@ -36,23 +36,87 @@ emit:
...
     api-version: "2023-11-01"
 ```
 
+### SDK Related Configuration in Emitter options `tspconfig.yaml`
+
+#### `api-version`
+
+By default, emitter generates code from the latest api-version in TypeSpec.
+
+There is cases that service wishes to generate code from a previous api-version. In such case, set the value to that api-version.
```
XiaofeiCao, line 52:
```suggestion
In cases where a service needs to generate code from a previous api-version, set the value to that specific api-version.
```
--- from copilot

typespec-extension/readme.md
```diff
@@ -36,23 +36,87 @@ emit:
...
+
+Default value is `true`. Emitter generates code samples under `generated` package.
+
+For data-plane SDK, the generated samples are for reference. The motivation is to show how samples can be written. It is expected that developers write correct and concise samples outside of `generated` package.
+
+If there is customization of the generated SDK, the generated samples may not compile after customization. One can delete the `generated` package, set the value to `false` to disable the generated samples.
+
+#### `generate-tests`
+
+Default value is `true`. Emitter generates tests under `generated` package. The motivation is to show how samples can be written.
```
XiaofeiCao, line 80:
```suggestion
Default value is `true`. Emitter generates tests under `generated` package.
```

weidongxu-microsoft: thanks. probably a paste triggered on the wrong position...

XiaofeiCao (approved):

PR #3016, opened 2025-01-21
eclipse language server, handle directory corruption

XiaofeiCao:
- Met this in autorest repo too. 
image

- Copied from TypeSpec: https://github.com/microsoft/typespec/pull/5307


weidongxu-microsoft (approved):
Does this require a release of azure-autorest-customization lib?

Just checking, we may not need to release now, even if it requires.

XiaofeiCao:
> Does this require a release of azure-autorest-customization lib?
> 
> Just checking, we may not need to release now, even if it requires.

Probably not? 
In autorest.java, we used maven-shade-plugin which I assume would pack all dependencies into released package.
In typespec-java/http-client-java, seems we don't depend on it.

weidongxu-microsoft:
> Probably not? In autorest.java, we used maven-shade-plugin which I assume would pack all dependent modules(including customization-base) into released package. In typespec-java/http-client-java, seems we don't depend on azure-autorest-customization lib.

Yeah, if http-client-java fixed after this change, it won't need the azure lib.

azure-dataplane-tests/src/main/java/com/azure/containers/containerregistry/implementation/ContainerRegistriesImpl.java
```diff
@@ -1104,6 +1142,27 @@ public PagedFlux<String> getRepositoriesAsync(String last, Integer n, Context co
...
+     * @param last Query parameter for the last item in previous query. Result set will include values lexically after
+     * last.
+     * @param n query parameter for max number of items.
+     * @throws IllegalArgumentException thrown if parameters fail the validation.
+     * @throws AcrErrorsException thrown if the request is rejected by server.
+     * @throws RuntimeException all other wrapped checked exceptions if the request fails to be sent.
+     * @return list of repositories along with {@link PagedResponse} on successful completion of {@link Mono}.
+     */
+    @ServiceMethod(returns = ReturnType.SINGLE)
+    public Mono<PagedResponse<String>> getRepositoriesNoCustomHeadersSinglePageAsync(String last, Integer n) {
```
XiaofeiCao, line 1157: This is due to previous code change here: https://github.com/microsoft/typespec/pull/5665/files#diff-d354f2c590be72c845717b155ef58df3371ccda76aa853abff7af464079b9c65R786-R787

specs definition defines `sync-methods: none`: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/containerregistry/azure-containers-containerregistry/swagger/autorest.md?plain=1#L39

Now we are adding pageable with full parameters + context regardless of `sync-methods`.
After this change, we seem to align with create/deleteXXAsync methods, which already have full parameters + context variant:
https://github.com/Azure/autorest.java/blob/8ee365815f616d708a129b92610e8e20f24e8b38/azure-dataplane-tests/src/main/java/com/azure/containers/containerregistry/implementation/ContainerRegistriesImpl.java#L844-L845

@weidongxu-microsoft Do you know if this is expected?

weidongxu-microsoft: Likely not expected. I guess regression is here https://github.com/microsoft/typespec/pull/5665/commits/5b0b44641d1b84614c46548832f75c931f64a3d9 (I would use "not sync=none" here, instead of removing the if)

I would do a fix soon.

PR #3023, opened 2025-02-08
mgmt, add properties file

XiaofeiCao:
- sync from typespec repo: https://github.com/microsoft/typespec/pull/5914

### DIFF
--- a/typespec-tests/src/main/resources/azure-resourcemanager-armresourceprovider-generated.properties
+++ b/typespec-tests/src/main/resources/azure-resourcemanager-armresourceprovider-generated.properties
@@ -1,0 +1,1 @@
+version=${project.version}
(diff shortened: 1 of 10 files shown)


typespec-tests/src/main/resources/azure-resourcemanager-armresourceprovider-generated.properties
weidongxu-microsoft, line 1: Why all mgmt ones has `-generated`?

XiaofeiCao: It's added when package name doesn't start with `com.azure.resourcemanager`. 
https://github.com/microsoft/typespec/blob/08384aaf73215343cb75aec26f02a22740730077/packages/http-client-java/generator/http-client-generator-mgmt/src/main/java/com/microsoft/typespec/http/client/generator/mgmt/util/FluentUtils.java#L147

We are using `azure.resourcemanager` in tests.

weidongxu-microsoft: Got it.

weidongxu-microsoft (approved):

PR #3044, opened 2025-03-17
sync mgmt sync-stack

XiaofeiCao:
[CI failure](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4695735&view=logs&j=a782880c-df6e-5f49-eb14-3a6fe2aebdb9&t=98722542-7ad9-5e5a-5a2b-3bbc1f21c363&l=51085) for fluent-test is strange:
> [ERROR]   SmbSettingTests.testDeserialize:17 » IllegalState java.lang.IllegalAccessException: access to public member failed: com.azure.mgmtlitetest.storage.models.SmbSetting.fromJson[Ljava.lang.Object;@7f53ddf8/invokeStatic, from public Lookup

Source code is suggesting module issue:
```java
if (Modifier.isPublic(mods))
    return "access to public member failed";  // (how?, module not readable?)
```
Though I couldn't find any.

Can only find one similar issue: https://issues.apache.org/jira/browse/CASSANDRA-18658
I'm not sure what does their solution do...

Strangely, it reported there's `module-info.java` file which should have been deleted in the previous script...
https://dev.azure.com/azure-sdk/public/_build/results?buildId=4700233&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=738c417a-4a48-5620-b571-4f9e77e0bfd2

`Wait-Job` will not terminate the job or throw exception if the timeout is reached. It'll simply abort the wait and the job will still be running in background. See [example](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/wait-job?view=powershell-7.5#example-6-wait-for-a-period-then-allow-job-to-continue-in-background).


weidongxu-microsoft (approved):

weidongxu-microsoft (commented):
We can start with some service that we had live test, e.g. datafactory with sync-stack.

If these works, we can e.g. switch the default of TypeSpec to sync-stack, then Swagger on Lite.

typespec-tests/src/main/java/tsptest/armresourceprovider/fluent/models/ChildExtensionResourceInner.java
weidongxu-microsoft, line 1: Any reason why this changed from Jackson to sss?

weidongxu-microsoft: Later we may consider merge the 2 tsp into 1 (or have 2 but let them focus on different area), as prior reason of separation not longer valid.

weidongxu-microsoft:
Not sure why CI fails on this PR.

My other PR is good https://github.com/microsoft/typespec/pull/6750 https://github.com/Azure/autorest.java/pull/3054

XiaofeiCao:
> Not sure why CI fails on this PR.
> 
> My other PR is good #3052 #3054

My local test run fails too.
After I specified java8 for fluent-test module, my local test passed. Let me see if it works for CI...

weidongxu-microsoft:
My guess is "module-info.java" too.

XiaofeiCao:
> My guess is "module-info.java" too.

Yes, it's the module-info.java file causing the issue. 

Root cause turns out to be PowerShell's `Wait-Job`: https://github.com/Azure/autorest.java/blob/1a2e2fe01f14b1a4f1986f185e7f59b4b1d3e623/fluent-tests/Initialize-Tests.ps1#L163

It will not terminate the job or throw exception if the timeout is reached. It'll simply abort the wait and the job will still be running in background. See [example](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/wait-job?view=powershell-7.5#example-6-wait-for-a-period-then-allow-job-to-continue-in-background).
Even if we delete the module-info.java in https://github.com/Azure/autorest.java/blob/1a2e2fe01f14b1a4f1986f185e7f59b4b1d3e623/fluent-tests/Initialize-Tests.ps1#L166-L169
, unfinished jobs will keep running and generating new module-info.java...

I'll increase the timeout for now, and see if we need to check the jobs state afterwards.

PR #3055, opened 2025-03-28
verify parameter transformation cleanup

anuchandy:


weidongxu-microsoft:
There is something in the middle (sync-stack) that Xiaofei is working on https://github.com/Azure/autorest.java/pull/3044, hence the diff

I think there is no problem on the clean up.

PR #3059, opened 2025-04-02
mgmt, sync PR to replace Flux<ByteBuffer> with BinaryData

XiaofeiCao:
- typespec PR: https://github.com/microsoft/typespec/pull/6825
- added test for fluent/vanilla


azure-dataplane-tests/src/main/java/com/azure/containers/containerregistry/implementation/ContainerRegistryBlobsImpl.java
```diff
@@ -85,7 +84,7 @@ Mono<ResponseBase<ContainerRegistryBlobsGetBlobHeaders, BinaryData>> getBlob(@Ho
         @Get("/v2/{name}/blobs/{digest}")
         @ExpectedResponses({ 200, 307 })
         @UnexpectedResponseExceptionType(HttpResponseException.class)
-        Mono<Response<BinaryData>> getBlobNoCustomHeaders(@HostParam("url") String url, @PathParam("name") String name,
+        Mono<StreamResponse> getBlobNoCustomHeaders(@HostParam("url") String url, @PathParam("name") String name,
```
XiaofeiCao, line 87: This is unexpected. I need to investigate why.

XiaofeiCao: Cause here: 
```java
return isByteStream(bodyType) ? mono(ClassType.STREAM_RESPONSE) : mono(GenericType.Response(bodyType));
```
```java
private static boolean isByteStream(IType type) {
    return (type == ClassType.INPUT_STREAM) || (type == ClassType.BINARY_DATA);
}
```

For non-DPG, previously we would [map](https://github.com/microsoft/typespec/blob/main/packages/http-client-java/generator/http-client-generator-core/src/main/java/com/microsoft/typespec/http/client/generator/core/mapper/BinaryMapper.java#L33) BinarySchema to FluxBB:
```java
return JavaSettings.getInstance().isDataPlaneClient() ? ClassType.BINARY_DATA : GenericType.FLUX_BYTE_BUFFER;
```

`isByteStream()` check would return false.

Now that we always map BinarySchema to BinaryData, `isByteStream()` check becomes true, hence StreamResponse.

We should be able to replace `StreamResponse` with `Response<BinaryData` now.

weidongxu-microsoft: I see.

BTW, your change to `BinaryData` would also change the behavior in vanilla too, right?

XiaofeiCao: Yes, it's for removing FluxBB across the board(not only mgmt).
I'll mention this when PR's ready.

weidongxu-microsoft (approved):

PR #3106, opened 2025-05-06
run ci with core v2

anuchandy:


weidongxu-microsoft:
there is a minor issue with the `core.diff`. I've made another try at https://github.com/Azure/autorest.java/pull/3107 (WIP)

PR #3123, opened 2025-05-30
tsp, fix Generate.ps1 script

weidongxu-microsoft:
I am checking nightly build error.

1st step let me sync the tests and fix the script.


typespec-tests/Generate.ps1
```diff
@@ -73,6 +73,8 @@ $generateScript = {
     $tspOptions += " --option ""@azure-tools/typespec-java.service-version-exclude-preview=true"""
     # enable sync-stack
     $tspOptions += " --option ""@azure-tools/typespec-java.enable-sync-stack=true"""
+    # rename model
+    $tspOptions += " --option ""@azure-tools/typespec-java.rename-model=TopLevelArmResourceListResult:ResourceListResult,CustomTemplateResourcePropertiesAnonymousEmptyModel:AnonymousEmptyModel"""
```
haolingdong-msft, line 77: wonder why do we need to rename the model?

weidongxu-microsoft: It is for test.

The `rename-model` was in Swagger codegen. Here we also put to TypeSpec emitter as last resort. TypeSpec code cannot rename anonymous model, or templated model.

weidongxu-microsoft: The directly cause is that the name generated for anonymous model is too long, and SDK repo has a limit of path length 222

haolingdong-msft (approved):

PR #3140, opened 2025-06-16
eng, sync image.yml

weidongxu-microsoft:
use what's in SDK repo latet PR https://github.com/Azure/azure-sdk-for-java/blob/main/eng/pipelines/templates/variables/image.yml
(this file is in PR and to be merged)


weidongxu-microsoft:
use https://github.com/Azure/autorest.java/pull/3143

PR #3207, opened 2025-10-17
eng, update jar deps

weidongxu-microsoft:


pom.xml
```diff
@@ -178,7 +178,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-javadoc-plugin</artifactId>
-          <version>3.11.3</version>
```
weidongxu-microsoft: There is a bug in 3.11.3 that be fixed in 3.12.0

PR #3211, opened 2025-10-22
eng, disable test on extension resource

weidongxu-microsoft:


typespec-tests/src/test/java/azure/resourcemanager/resources/ArmResourceTest.java
```diff
@@ -66,6 +67,7 @@ public class ArmResourceTest {
     private ExtensionsResource extensionResource;
     private List<ExtensionsResource> extensionResources;
 
+    @Disabled
```
weidongxu-microsoft: to pass nightly build

change to the test is merged but not released

XiaofeiCao: Yeah, this should do.

PR #3268, opened 2026-01-16
eng, add openai-typespec lib, prepare release 0.38.4

weidongxu-microsoft:
foundry is using this lib for API/models from native openai https://github.com/Azure/azure-rest-api-specs/blob/feature/foundry/package.json#L4


typespec-extension/package.json
```diff
@@ -62,6 +62,7 @@
   },
   "dependencies": {
     "@autorest/codemodel": "~4.20.1",
+    "@azure-tools/openai-typespec": "^1.6.1",
```
weidongxu-microsoft: In short-term, we still add it to emitter dependencies, to be synced to SDK repo.

eng/sdk/sync_sdk.py
```diff
@@ -20,8 +20,6 @@
 skip_artifacts: List[str] = [
     "azure-ai-anomalydetector",  # deprecated
     # expect failure on below
-    "azure-ai-projects",  # wait for service dev update tsp-location.yaml
-    "azure-ai-agents-persistent", # tspconfig does not have java
```
weidongxu-microsoft: They likely would still fail, if run.

But in SDK repo they renamed tsp-location.yaml to _tsp-location.yaml, so it won't run.

eng/pipelines/post-publish-sdk.yaml
```diff
@@ -22,7 +22,7 @@ resources:
 jobs:
 - job: Generate_SDK
 
-  timeoutInMinutes: 60
+  timeoutInMinutes: 120
```
weidongxu-microsoft: The running time seems not very stable. I see range from 40min to more than 1hour (where the job times out).

Let's see if increase here would work.

PR #3323, opened 2026-04-15
eng, use CFS for maven and npm

weidongxu-microsoft:
Also added a contributing.md to typespec-extension, to auth the npm

However, I seem not need to auth the maven


eng/pipelines/templates/stages/1es-redirect.yml
```diff
@@ -21,6 +21,7 @@ extends:
   parameters:
     settings:
       skipBuildTagsForGitHubPullRequests: true
+      networkIsolationPolicy: Permissive,CFSClean,CFSClean2
```
raych1: This override is not necessary with CFS enabled.

eng/pipelines/post-publish-sdk.yaml
raych1, line 70: Just confirm if this step would use the 'npmrc' with azure artifacts feed.

eng/pipelines/templates/stages/1es-redirect.yml
weidongxu-microsoft: removed

eng/pipelines/post-publish-sdk.yaml
weidongxu-microsoft: Thanks. I've updated the step to copy and auth global `.npmrc` (under $HOME) -- similar to the step for Maven.

I think this way, any access to npm would be limited to CFS.

Run of the pipeline is good (this pipeline does not use 1ES, so it won't report anything though).

weidongxu-microsoft:
nightly dev fails due to missing `@azure/identity 4.13.1` in azure registry. Will do workaround in other PR.

raych1 (approved):

PR #3341, opened 2026-05-21
eng, release npm first

weidongxu-microsoft:
Release npm first, before create GitHub release and devop feed.

So that if npm release fails, we can just run it again.


eng/sdk/patches/0001-revert-azure-ai-speech-transcription.patch
weidongxu-microsoft, line 1: transcription did customization to change return type (from `int` to `Duration`).
Hence the generated tests do not work.

Their dev manually modified the generated code. I've notified them to disable the `generate-tests` in tspconfig.

XiaofeiCao (approved):

PR #3348, opened 2026-06-12
Switch PAT to GitHubApps

chidozieononiwu:
This pull request updates the `eng/pipelines/publish-autorest-java.yaml` pipeline to improve security and reliability when publishing releases. The main changes involve updating GitHub authentication and adding a login step.

Authentication and security improvements:

* Added a step to log in to GitHub using the shared template `login-to-github.yml` to ensure authenticated operations before release steps.
* Changed the environment variable used for the GitHub token from `azuresdk-github-pat` to `GH_TOKEN` for consistency and improved secret management.


weidongxu-microsoft (approved):
The pipeline itself is disabled (due to CFS). autorest itself is also deprecated. So we likely won't use this pipeline anymore.

Let me know if we'd better just delete it? I can do that.

